PR #430, opened 2022-01-13
make `HashOutTarget`'s `elements` public

Sladuca:
`CircuitBuilder`'s hashing gadgets currently take a `Vec<Target>` and return a `HashOutTarget` but there's no way to turn the `HashOutTarget` back into a `Vec<Target>`.

I made `elements` public so the user can do `hash_out.elements.to_vec()`. I also made `from_vec` and `from_partial` public for the same reason.


dlubarov (approved):
Thanks for fixing that 👍

PR #433, opened 2022-01-15
Merkle tree optimizations

nbgl:
Getting around the overhead of Rayon's parallel iterators when building Merkle Trees. These changes speed up Merkle Tree benchmarks by 5% for width 2 ** 15 and 8% for width 2 ** 13 (both on Poseidon) on my machine. Speeds up proofs by 2%. (In general, there's more benefit for smaller trees and for machines with more cores.)

### DIFF
--- a/plonky2/src/hash/merkle_tree.rs
+++ b/plonky2/src/hash/merkle_tree.rs
@@ -2,6 +2,7 @@ use std::mem::MaybeUninit;
 use std::slice;
 
 use plonky2_util::log2_strict;
+use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 
 use crate::hash::hash_types::RichField;
--- a/plonky2/src/hash/merkle_tree.rs
+++ b/plonky2/src/hash/merkle_tree.rs
@@ -1,4 +1,7 @@
-use rayon::prelude::*;
+use std::mem::MaybeUninit;
+use std::slice;
+
+use plonky2_util::log2_strict;
 use serde::{Deserialize, Serialize};
 
 use crate::hash::hash_types::RichField;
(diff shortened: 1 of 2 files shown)


dlubarov:
Hm I think this is probably worthwhile if we're pretty sure it's the fastest approach, I just want to try a few other things first, since once we merge this more complex version it will become a bit harder to experiment with other methods. I made #438 with a couple trivial optimizations. After that I'll rebase your branch if you don't mind, and try a couple other things.

I suspect this might be a really significant improvement if we had large trees with small leaves, just in practice we have a large leaf layer (135 elements per leaf in the largest) while the rest of the tree is pretty compact (2^15 * 64 bits for recursion).

plonky2/src/hash/merkle_tree.rs
```diff
@@ -27,36 +30,108 @@ pub struct MerkleTree<F: RichField, H: Hasher<F>> {
...
+            digests_chunks.zip(cap_refs).zip(leaves_chunks)
+        {
+            // We have `1 << cap_height` sub-trees, one for each entry in `cap`. They are totally
+            // independent, so we schedule one task for each. `digests_buf` and `leaves` are split
+            // into `1 << cap_height` slices, one for each sub-tree.
+            s.spawn(|_| {
+                subtree_cap.write(fill_subtree::<F, H>(subtree_digests, subtree_leaves));
+            });
+        }
+    });
```
dlubarov, line 110: Maybe consider using parallel iterators -

```suggestion
    let digests_chunks = digests_buf.par_chunks_exact_mut(subtree_digests_len);
    let leaves_chunks = leaves.par_chunks_exact(subtree_leaves_len);
    assert_eq!(digests_chunks.len(), cap_buf.len());
    assert_eq!(digests_chunks.len(), leaves_chunks.len());
    digests_chunks.zip(cap_buf).zip(leaves_chunks).for_each(
        |((subtree_digests, subtree_cap), subtree_leaves)| {
            // We have `1 << cap_height` sub-trees, one for each entry in `cap`. They are totally
            // independent, so we schedule one task for each. `digests_buf` and `leaves` are split
            // into `1 << cap_height` slices, one for each sub-tree.
            subtree_cap.write(fill_subtree::<F, H>(subtree_digests, subtree_leaves));
        },
    );
```

It's a bit shorter and the docs seem to encourage it for performance reasons, though that probably doesn't matter since we only spawn a few (`cap.len()`) tasks.

plonky2/src/hash/merkle_tree.rs
```diff
@@ -27,36 +30,108 @@ pub struct MerkleTree<F: RichField, H: Hasher<F>> {
...
+        // Layout is: left recursive output || left child digest
+        //             || right child digest || right recursive output.
+        // Split `digests_buf` into the two recursive outputs (slices) and two child digests
+        // (references).
+        let (left_digests_buf, right_digests_buf) = digests_buf.split_at_mut(digests_buf.len() / 2);
+        let (left_digest_mem, left_digests_buf) = left_digests_buf.split_last_mut().unwrap();
+        let (right_digest_mem, right_digests_buf) = right_digests_buf.split_first_mut().unwrap();
+        // Split `leaves` between both children.
+        let (left_leaves, right_leaves) = leaves.split_at(leaves.len() / 2);
+        let (left_digest, right_digest) = rayon::join(
```
dlubarov, line 76: I thought this might be a bit much overhead, spawning a task for each hash. It's easy to adjust the granularity with e.g.

```rust
        let mut fill_left = || fill_subtree::<F, H>(left_digests_buf, left_leaves);
        let mut fill_right = || fill_subtree::<F, H>(right_digests_buf, right_leaves);
        let (left_digest, right_digest) = if leaves.len() > 16 {
            rayon::join(fill_left, fill_right)
        } else {
            (fill_left(), fill_right())
        };
```

But I couldn't really detect a change in performance. Might be worth trying on another machine but I guess the overhead is small.

dlubarov (approved):
LGTM

plonky2/src/hash/merkle_tree.rs
nbgl: [The docs](https://docs.rs/rayon/latest/rayon/fn.join.html) claim that `rayon::join` is actually more clever than blindly spinning up tasks. In summary, the calling worker thread executes the first task, but the second task is marked as “available for stealing” if another thread is looking for work in the meantime. If it doesn’t get stolen (as we would expect the vast majority of the time), then the calling thread proceeds to execute the second task.

You’re still right though that this adds _some_ overhead—just the atomic writes involved would not be cheap—but Poseidon (at ~2K per hash and many hashes per leaf) would dwarf that.

plonky2/src/hash/merkle_tree.rs
nbgl: Good idea! I’ll give it a go. I don’t think it matters performance-wise, but it would definitely be more elegant.

nbgl: Oops I didn't notice your suggested change… That looks much cleaner!

PR #434, opened 2022-01-15
Optimize + test log2 functions

nbgl:
These changes halve the ASM size for `log2_strict` and `log2_ceil`. I also added some tests.

### DIFF
--- a/plonky2/src/util/mod.rs
+++ b/plonky2/src/util/mod.rs
@@ -58,60 +58,9 @@ pub(crate) fn reverse_bits(n: usize, num_bits: usize) -> usize {
 
 #[cfg(test)]
 mod tests {
-    use plonky2_util::{log2_ceil, log2_strict, reverse_index_bits, reverse_index_bits_in_place};
+    use plonky2_util::{reverse_index_bits, reverse_index_bits_in_place};
 
     use crate::util::reverse_bits;
-
-    #[test]
-    fn test_log2_strict() {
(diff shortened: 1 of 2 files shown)


plonky2/src/util/mod.rs
```diff
@@ -58,10 +58,61 @@ pub(crate) fn reverse_bits(n: usize, num_bits: usize) -> usize {
 
 #[cfg(test)]
 mod tests {
-    use plonky2_util::{reverse_index_bits, reverse_index_bits_in_place};
+    use plonky2_util::{log2_ceil, log2_strict, reverse_index_bits, reverse_index_bits_in_place};
 
     use crate::util::reverse_bits;
 
+    #[test]
```
dlubarov, line 65: nit: I think it would be idiomatic to have the tests in the util crate. (I realize I didn't follow that with `reverse_index_bits`... just didn't notice it as I was moving things around.)

dlubarov (approved):
👍

plonky2/src/util/mod.rs
nbgl: Ah! I thought that was weird, but I figured there must be a technical reason for that! I'll move them all to `util`

nbgl: Actually, to move the `reverse_index_bits` tests, I'd have to move `reverse_bits` as well, and that touches more code than I'm comfortable changing in this PR

PR #436, opened 2022-01-17
Fix degree of polynomial used in LDT

wborgeaud:
In our FRI-based polynomial commitment, we commit to polynomials of degree `< n` and use a low-degree test to check that `(f - f(z)) / (X - z)` has degree `< n` (where `f` is a linear combination of the original polynomials). 
However, this quotient polynomial has degree `deg(f)-1`. So by starting with polynomials of degree `n`, the LDT would pass. This could potentially be exploited.

This PR fixes this by multiplying the quotient by `X`, so that the LDT actually checks that the quotient has degree `< n-1`.

### DIFF
--- a/plonky2/src/fri/oracle.rs
+++ b/plonky2/src/fri/oracle.rs
@@ -154,6 +154,12 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
             alpha.shift_poly(&mut final_poly);
             final_poly += quotient;
         }
+        final_poly.trim();
+        let mut final_poly_coeffs = final_poly.coeffs;
+        final_poly_coeffs.insert(0, F::Extension::ZERO);
+        final_poly = PolynomialCoeffs {
+            coeffs: final_poly_coeffs,
+        };
 
         let lde_final_poly = final_poly.lde(fri_params.config.rate_bits);
         let lde_final_values = timed!(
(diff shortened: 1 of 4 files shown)


plonky2/src/fri/oracle.rs
```diff
@@ -154,6 +154,10 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
             alpha.shift_poly(&mut final_poly);
             final_poly += quotient;
         }
+        final_poly.trim();
+        // Multiply the final polynomial by `X`, so that `final_poly` has the maximum degree for
```
dlubarov, line 158: nit: let's copy this to the two verifier changes also, otherwise I think it's hard to see why what the mul is for

plonky2/src/fri/oracle.rs
```diff
@@ -154,6 +154,10 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
             alpha.shift_poly(&mut final_poly);
             final_poly += quotient;
         }
+        final_poly.trim();
```
dlubarov, line 157: Seems like it's possible this would result in a non-power-of-two size, if there were multiple `0`s at the end. How about using `rotate_right(1)` (and maybe asserting that a `0` was rotated)? To ensure the size doesn't change

dlubarov (approved):
LGTM otherwise, good catch

plonky2/src/fri/oracle.rs
wborgeaud: Hmm good point.
Now that I look at it, the 
```rust
quotient.padded(quotient.degree_plus_one().next_power_of_two())
```
in `compute_quotient` seems a bit weird, especially now that we only divide by linear factors. Maybe we could remove this padding, remove the `final_poly.trim()` here, and keep the `final_poly.coeffs.insert(0,0)` to multiply by `X`.
What do you think?

dlubarov: Sounds good to me 👍

PR #438, opened 2022-01-18
Have hash functions take references to avoid cloning

dlubarov:
And other tweaks to `MerkleTree::new`

### DIFF
--- a/plonky2/src/hash/hashing.rs
+++ b/plonky2/src/hash/hashing.rs
@@ -18,7 +18,7 @@ pub fn hash_or_noop<F: RichField, P: PlonkyPermutation<F>>(inputs: Vec<F>) -> Ha
     if inputs.len() <= 4 {
         HashOut::from_partial(inputs)
     } else {
-        hash_n_to_hash::<F, P>(inputs, false)
+        hash_n_to_hash::<F, P>(&inputs, false)
     }
 }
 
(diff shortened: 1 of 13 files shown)


plonky2/src/hash/hashing.rs
```diff
@@ -101,16 +101,18 @@ pub trait PlonkyPermutation<F: RichField> {
...
-        while (inputs.len() + 1) % SPONGE_WIDTH != 0 {
-            inputs.push(F::ONE);
+        let mut padded_inputs = inputs.to_vec();
+        padded_inputs.push(F::ZERO);
+        while (padded_inputs.len() + 1) % SPONGE_WIDTH != 0 {
+            padded_inputs.push(F::ONE);
         }
-        inputs.push(F::ZERO);
+        padded_inputs.push(F::ZERO);
+        return hash_n_to_m::<F, P>(&padded_inputs, num_outputs, false);
```
nbgl, line 115: I think we can avoid this cloning too. We process the inputs in one loop as usual, and have a second loop to deal with the padding. But Merkle trees don't use padding, so maybe it doesn't matter.

nbgl (approved):
Looks good!

plonky2/src/hash/hashing.rs
dlubarov: (Discussed offline, decided to hold off for now since we don't use padding in perf sensitive code)

PR #439, opened 2022-01-18
Make all FFTs in-place

nbgl:
Perform FFTs without cloning vectors.

While it is ready for review, this PR should not be merged yet. It turns out that `reverse_index_bits_in_place` is significantly slower than its out-of-place counterpart. This disparity can and should be fixed, but until that happens, merging this PR would actually make proofs slower.

### DIFF
--- a/field/src/fft.rs
+++ b/field/src/fft.rs
@@ -1,7 +1,7 @@
 use std::cmp::{max, min};
 use std::option::Option;
 
-use plonky2_util::{log2_strict, reverse_index_bits};
+use plonky2_util::{log2_strict, reverse_index_bits_in_place};
 use unroll::unroll_for_loops;
 
 use crate::field_types::Field;
@@ -34,10 +34,10 @@ pub fn fft_root_table<F: Field>(n: usize) -> FftRootTable<F> {
 
 #[inline]
 fn fft_dispatch<F: Field>(
-    input: &[F],
+    input: &mut [F],
     zero_factor: Option<usize>,
     root_table: Option<&FftRootTable<F>>,
-) -> Vec<F> {
+) {
     let computed_root_table = if root_table.is_some() {
         None
     } else {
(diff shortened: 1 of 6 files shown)


nbgl:
Having worked out how to make bit reverses fast, I'd like to merge this PR.

field/src/polynomial/mod.rs
```diff
@@ -64,7 +64,7 @@ impl<F: Field> PolynomialValues<F> {
     }
 
     pub fn degree_plus_one(&self) -> usize {
-        self.ifft().degree_plus_one()
+        self.clone().ifft().degree_plus_one()
```
dlubarov, line 67: This might add a non-negligible cost, since we `trim` polynomials in a couple places which calls this. Though I think we could remove the `trim`s since we always know what degree a poly in our argument should have (unless it happens to have extra 0s by chance). So probably not worth worrying about. I think @wborgeaud was going to remove one of the `trim` calls

field/src/fft.rs
```diff
@@ -200,13 +199,13 @@ pub(crate) fn fft_classic<F: Field>(input: &[F], r: usize, root_table: &FftRootT
...
-        fft_classic_simd::<F>(&mut values[..], r, lg_n, root_table);
+        fft_classic_simd::<F>(values, r, lg_n, root_table);
     } else {
-        fft_classic_simd::<<F as Packable>::Packing>(&mut values[..], r, lg_n, root_table);
+        fft_classic_simd::<<F as Packable>::Packing>(values, r, lg_n, root_table);
     }
-    values
 }
 
+// JNTODO
```
dlubarov, line 208: If this is meant to stay could you add a note?

dlubarov (approved):
👍

field/src/fft.rs
nbgl: Oops. This was just a bookmark. I'll delete it!

field/src/polynomial/mod.rs
nbgl: The buffer was cloned before this change, so this PR doesn't introduce any _new_ inefficiencies. I'll make a new issue to look into the `trim`s.

PR #440, opened 2022-01-19
Remove `compute_quotient` and update division tests

wborgeaud:
Remove `PolynomialBatch::compute_quotient()` and replace it with simple calls to `PolynomialCoeffs::divide_by_linear()` since we don't divide by higher degree polynomials anymore.
Also update the division tests to reflect that.


dlubarov (approved):
LGTM

PR #442, opened 2022-01-20
Jakub does in 200 lines what can be done in 5 (bit-order reversal optimizations)

nbgl:
Implements a cache-friendly bit-order reversal algorithm. (description of said algorithm to go here…)

Benchmark changes (arrays of `u64`s):
| length   | change |
| -------- | ------ |
|      256 |   -19% |
|    65536 |   -56% |
| 16777216 |   -75% |

### DIFF
--- a/util/src/lib.rs
+++ b/util/src/lib.rs
@@ -7,6 +7,7 @@
 
 use std::arch::asm;
 use std::hint::unreachable_unchecked;
+use std::mem::{size_of, swap};
 
 pub fn bits_u64(n: u64) -> usize {
     (64 - n.leading_zeros()) as usize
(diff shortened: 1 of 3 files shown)


nbgl:
The algorithm (also explained in the code):

Treat `arr` as a sqrt(`arr.len()`) by sqrt(`arr.len()`) row-major matrix. (Assume for now that `arr.len()` is a square number.) To perform bit-order reversal we:
 1. Bit-reverse the order of the rows. (They are contiguous in memory, so this is
    basically a series of large `memcpy`s.)
 2. Transpose the matrix.
 3. Bit-reverse the order of the rows.

This is equivalent to, for every index 0 ≤ `i` < `arr.len()`:
 1. bit-reversing `i[lb(arr.len()) / 2..lb(arr.len())]`,
 2. swapping `i[0..lb(arr.len()) / 2]` and `i[lb(arr.len()) / 2..lb(arr.len())]`,
 3. bit-reversing `i[lb(arr.len()) / 2..lb(arr.len())]`.

If `arr.len()` is not a square number, then the above procedure requires
only slight modification, to ensure that we don't change the indices' middle bits.

For big enough inputs, we read the array three times, regardless of cache size or cacheline length.

dlubarov:
@wborgeaud would you mind reviewing this one? I'm pretty backlogged, lots of open PRs on the mir-core side

Sounds like a cool technique though!

util/src/lib.rs
```diff
@@ -80,57 +86,123 @@ fn reverse_index_bits_large<T: Copy>(arr: &[T], n_power: usize) -> Vec<T> {
...
-/* Both functions below are semantically equivalent to:
-        for src in 0..n {
-            let dst = reverse_bits(src, n_power);
-            if src < dst {
-                arr.swap(src, dst);
-            }
+/// Bit-reverse the order of elements in `arr`.
+/// SAFETY: ensure that `arr.len() == 1 << lb_n`.
+#[cfg(target_arch = "aarch64")]
+unsafe fn reverse_index_bits_in_place_small<T>(arr: &mut [T], lb_n: usize) {
```
wborgeaud, line 126: This function panics when `lb_n=0` in debug mode, due to a shr overflow. It works in release mode though, so maybe we could just add a `debug_assert`.

util/src/lib.rs
```diff
@@ -26,6 +31,7 @@ pub fn log2_ceil(n: usize) -> usize {
 pub fn log2_strict(n: usize) -> usize {
     let res = n.trailing_zeros();
     assert!(n.wrapping_shr(res) == 1, "Not a power of two: {}", n);
+    assume(n == 1 << res);
```
wborgeaud, line 34: It's not so clear what this does, maybe add a comment.

wborgeaud (approved):
Looks good to me!

util/src/lib.rs
nbgl: Ah, that tells the optimizer about the semantics of `log2_strict`. This way it knows that it can replace `n` with `1 << res` and vice versa. I'll add a comment!

PR #445, opened 2022-01-24
Replace `proof_to_proof_target`

dlubarov:
With a `add_virtual_proof_with_pis` method that uses the inner circuit data, but does not require constructing a proof.

Note that this doesn't support IVC type patterns (where a single circuit can be used for any proof in a PCD graph) yet. For that, I think we can add a variant of `add_virtual_proof_with_pis` that takes several parameters like FRI arities, but does not involve `CommonCircuitData` (since no circuit has been build yet). It might also be best to avoid large objects like `FriParams`, and pass just the data we need.

Then there will be some nontrivial work to do recursion with "estimated" parameters (degree, arities, etc), check if the estimates were correct, and try again if not.

### DIFF
--- a/plonky2/src/plonk/circuit_data.rs
+++ b/plonky2/src/plonk/circuit_data.rs
@@ -234,6 +234,8 @@ pub struct CommonCircuitData<
 
     pub(crate) num_virtual_targets: usize,
 
+    pub(crate) num_public_inputs: usize,
+
     /// The `{k_i}` valued used in `S_ID_i` in Plonk's permutation argument.
     pub(crate) k_is: Vec<F>,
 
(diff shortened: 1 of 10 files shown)


dlubarov:
cc @Sladuca

plonky2/src/plonk/circuit_data.rs
```diff
@@ -234,6 +234,8 @@ pub struct CommonCircuitData<
 
     pub(crate) num_virtual_targets: usize,
 
+    pub(crate) num_public_inputs: usize,
```
wborgeaud, line 237: I think we could add a check
```rust
assert_eq!(proof_with_pis.public_inputs.len(), common_data.num_public_inputs);
```
in `crate::verifer::verify` and `CircuitBuilder::verify_proof_with_pis`, now that this is part of the common data.

plonky2/src/plonk/recursive_verifier.rs
```diff
@@ -123,137 +142,75 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+            openings: self.add_opening_set(common_data),
+            opening_proof: self.add_virtual_fri_proof(num_leaves_per_oracle, fri_params),
+        };
+
+        let public_inputs = self.add_virtual_targets(common_data.num_public_inputs);
+        ProofWithPublicInputsTarget {
+            proof,
+            public_inputs,
+        }
+    }
```
wborgeaud, line 174: nit: Maybe we could add a `pub fn add_virtual_proof()` that this function would call. This would match the pattern in `verify_proof{_with_pis}()`.

wborgeaud (approved):
Looks good to me!

plonky2/src/plonk/circuit_data.rs
dlubarov: Done

plonky2/src/plonk/recursive_verifier.rs
dlubarov: Done

PR #447, opened 2022-01-24
Make `set_proof_target` publicly accessible

### LINKED ISSUE #431: Need public API for creating proof targets and populating them using proof objects
As @Sladuca pointed out,

> How do I construct recursive circuit outside the `plonky2` crate? I see in the test suite for `recursive_verifier.rs` that `recursive_proof` uses  `proof_to_proof_target` and `set_proof_target`, which turns a proof into a target and adds the proof as an input in a `PartialWitness`. Is there a reason these two functions are only in the test suite? 
>
> `proof_to_proof_target` seems to me like it would fit as a  `CircuitBuilder` method (it takes a `&mut CircuitBuilder`) and `set_proof_target` seems similar to other `PartialWitness` methods.
>
> Also, does one need to build a new recursive circuit for each proof, or is it possible to build one recursive 'outer' circuit and feed it different proofs for the same 'inner' circuit as input? 

I'm not sure about the `proof_to_proof_target` API though; let's think about some alternatives.

dlubarov:
Same code as before, except I broke it into a few functions and renamed a couple things.

Resolves #431.

### DIFF
--- a/plonky2/src/iop/witness.rs
+++ b/plonky2/src/iop/witness.rs
@@ -201,7 +201,12 @@ pub trait Witness<F: Field> {
         );
         self.set_cap_target(&proof_target.quotient_polys_cap, &proof.quotient_polys_cap);
 
-        for (&t, &x) in proof_target.openings.wires.iter().zip_eq(&proof.openings.wires) {
+        for (&t, &x) in proof_target
+            .openings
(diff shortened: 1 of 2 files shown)


wborgeaud (approved):
👍

PR #449, opened 2022-01-27
Matrix type

nbgl:
A proposal for a simple matrix type that supports the following operations:
- Construction in a few ways
  - From iterator of iterators
  - From flat vector
  - Using closure that initializes a matrix of `MaybeUninit`
- Indexing by row (each row is a slice)
- Mutable and non-mutable iteration (the iterators yield slices, one per row)
- Transposition
- Cloning

### DIFF
--- a/util/src/matrix.rs
+++ b/util/src/matrix.rs
@@ -5,10 +5,13 @@ use std::ops::{Index, IndexMut};
 use std::ptr;
 use std::slice;
 
+/// Calculates the number of elements in a matrix with the given height and width, panicking on
+/// overflow.
 fn size_checked(height: usize, width: usize) -> usize {
     height.checked_mul(width).expect("memory overflow")
 }
 
+/// Returns an uninitialized `Vec` of lenth `len`.
 fn uninit_vec<T>(len: usize) -> Vec<MaybeUninit<T>> {
     let mut res: Vec<MaybeUninit<T>> = Vec::with_capacity(len);
     unsafe {
@@ -20,6 +23,7 @@ fn uninit_vec<T>(len: usize) -> Vec<MaybeUninit<T>> {
     res
 }
 
+/// A non-resizeable matrix.
 pub struct Matrix<T> {
     data: *mut T,
     height: usize,
@@ -28,6 +32,20 @@ pub struct Matrix<T> {
 }
 
 impl<T> Matrix<T> {
+    /// Creates a `Matrix<T>` directly from the raw components of another matrix. The matrix takes
+    /// ownership of `data` and becomes responsible for deallocating it.
+    ///
+    /// # Safety
+    ///
+    /// This is unsafe because a number of invariants are assumed and not checked:
+    /// * `height * width * size_of::<T>()` must not overflow `usize`.
+    /// * `data` must be correctly aligned.
+    /// * `data` must point to the start of an allocation (i.e., returned by `std::alloc::alloc` or
+    ///   similar), and that allocation must hold exactly `height * width` instances of `T`.
+    /// * All `height * width` elements pointed to by `data` must be correctly initialized.
+    ///
+    /// Note that these invariants hold when `data` was allocated by a `Vec<T>` whose length and
+    /// capacity both equal `height * width` (and the latter does not overflow).
     pub unsafe fn from_raw(data: *mut T, height: usize, width: usize) -> Self {
         Matrix {
             data,
@@ -37,6 +55,9 @@ impl<T> Matrix<T> {
         }
     }
 
+    /// Creates a `Matrix<T>` from a vector. The vector is interpreted as a storing the matrix in a
+    /// row-major order. In other words, `v` is split into contiguous chunks, with each chunk
+    /// treated as a row of the matrix.
     pub fn from_flat_vec(height: usize, width: usize, v: Vec<T>) -> Self {
         // NB: We ask for the height _and_ the width to handle edge cases when `v.len() == 0`.
 
@@ -51,9 +72,13 @@ impl<T> Matrix<T> {
         // because to correctly deallocate memory we need to know its allocated size.
         let buf = v.into_boxed_slice();
 
+        // SAFETY: The pointer comes from a vector with `len()` and `capacity()` both equal to
+        // `height * width`. We can assume ownership of the memory since `Box::into_raw`
+        // relinguishes it.
         unsafe { Self::from_raw(Box::into_raw(buf).cast(), height, width) }
     }
 
+    /// Creates an uninitialized matrix of specified height and width.
     pub fn new_uninit(height: usize, width: usize) -> Matrix<MaybeUninit<T>> {
         let size = size_checked(height, width);
         let buf_vec = uninit_vec::<T>(size);
@@ -63,6 +88,15 @@ impl<T> Matrix<T> {
         Matrix::from_flat_vec(height, width, buf_vec)
     }
 
+    /// Creates a new matrix with a specified height and width, initialized with a provided closure.
+    ///
+    /// This function allocates an uninitialized matrix of with a given `height` and `width`. A
+    /// reference to this uninitialized array is passed to the user-provided closure, which is
+    /// responsible for initializing all elements.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe as `f` must initialization on every element of the matrix.
     pub unsafe fn new_with<F: FnOnce(&mut Matrix<MaybeUninit<T>>)>(
         height: usize,
         width: usize,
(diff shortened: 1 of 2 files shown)


util/src/matrix.rs
```diff
@@ -0,0 +1,541 @@
...
+        // memory rules. `MaybeUninit` doesn't need initialization, so the values are already valid.
+        debug_assert_eq!(res.capacity(), len);
+        res.set_len(len);
+    }
+    res
+}
+
+/// A non-resizeable matrix.
+pub struct Matrix<T> {
+    data: *mut T,
```
dlubarov, line 28: I think you mentioned wanting to avoid `Vec` since we don't need resizing. How about `Box<[T]>`? Seems like that could reduce unsafe code while avoiding the two fields of `Vec`.

nbgl:
A `Box` will also store the length of the buffer, which is redundant information since the matrix already stores a `height` and a `width` (although you're right that it's less redundant than a `Vec`).

This code does already use `Box` for allocation and deallocation, it just stores it in a decomposed format. Like, to allocate a `Matrix`, we get a `Box` (somehow) and then extract the pointer to the first element, which we then store. In `drop`, we construct a `Box` from the pointer and the buffer size (computed from `height` and `width`), which takes care of deallocation.

I'm happy to just store a `Box` if that's your strong preference, but it would mean redundant information, and all the possible inconsistencies/bugs that come with it.

util/src/matrix.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// Creates a new matrix with a specified height and width, initialized with a provided closure.
+    ///
+    /// This function allocates an uninitialized matrix of with a given `height` and `width`. A
+    /// reference to this uninitialized array is passed to the user-provided closure, which is
+    /// responsible for initializing all elements.
+    ///
+    /// # Safety
+    ///
+    /// This function is unsafe as `f` must initialization on every element of the matrix.
+    pub unsafe fn new_with<F: FnOnce(&mut Matrix<MaybeUninit<T>>)>(
```
dlubarov, line 77: nit: maybe make it private or crate-private? Initializing with `new_uninit` feels a bit more idiomatic to me (if the safe initializers can't be used). Not saying we should get rid of it but just hesitant for the public API.

util/src/matrix.rs
```diff
@@ -0,0 +1,527 @@
...
+    pub unsafe fn get_unchecked_mut(&mut self, index: usize) -> &mut [T] {
+        let width = self.width; // borrow checker again
+        self.as_flat_mut()
+            .get_unchecked_mut(index * width..(index + 1) * width)
+    }
+
+    /// Iterate over all rows of the matrix as references.
+    pub fn iter(
+        &self,
+    ) -> impl Iterator<Item = &[T]> + DoubleEndedIterator + ExactSizeIterator + FusedIterator {
```
dlubarov, line 152: Seems easiest to return `Chunks` (or `ChunksMut` below)

util/src/matrix.rs
```diff
@@ -0,0 +1,527 @@
...
+    /// Due to interactions with the allocator, this method has stricter requirements than the
+    /// `mem::transmute` function in the library.
+    ///
+    /// Same as `mem::transmute`:
+    /// - `T` and `U` must have the same size.
+    /// - All elements of the matrix must be safe to interpret bitwise as `U`.
+    ///
+    /// Furthermore:
+    /// - `T` and `U` must have the same alignment.
+    unsafe fn transmute<U>(self) -> Matrix<U> {
```
dlubarov, line 226: I'm probably missing something but can this use `mem::transmute` (to implement this)?

dlubarov (approved):
Looks good to me!

util/src/matrix.rs
```diff
@@ -0,0 +1,527 @@
...
+use std::ops::{Index, IndexMut};
+use std::ptr;
+
+/// Calculates the number of elements in a matrix with the given height and width, panicking on
+/// overflow.
+fn size_checked(height: usize, width: usize) -> usize {
+    height.checked_mul(width).expect("memory overflow")
+}
+
+/// Returns an uninitialized boxed slice of lenth `len`.
```
recmo, line 12: \* length

util/src/matrix.rs
recmo: For a public API I'd expect:

```
/// Creates a new matrix with a specified height and width, initialized with a provided closure.
///
/// The closure will be called in row-major order for each element.
pub fn new_with<F: FnMut(usize, usize) -> T>(height: usize, width: usize, mut f: F) -> Self {
    let mut res = Self::new_uninit(height, width);
    for i in 0..height {
        for j in 0..width {
            res[i][j].write(f(i, j));
        }
    }
    unsafe {
        // SAFETY: All elements are now initialized.
        res.assume_init()
    }
}
```

It's obviously not as flexible as the original, but it covers quite a few use-cases and is simple and safe:

```rust
#[test]
fn test_new_with() {
    let m = Matrix::new_with(HEIGHT, WIDTH, |i, j| (WIDTH * i + j) as u64);
    check_matrix(m);
}
```

util/src/matrix.rs
```diff
@@ -0,0 +1,527 @@
...
+        // place.
+        self.transmute::<T>()
+    }
+}
+
+impl<T: Clone> Clone for Matrix<T> {
+    fn clone(&self) -> Self {
+        Self::from_flat_vec(self.height, self.width, self.as_flat().to_vec())
+    }
+}
```
recmo, line 285: Why not `#[derive(Clone)]`?  `Box<T>` impls clone when T does, so afaik derive should do the right thing.

util/src/matrix.rs
recmo, line 77: For a public API I'd expect (in the principle-of-least-surprise sense):

```rust
/// Creates a new matrix with a specified height and width, initialized with a provided closure.
///
/// The closure will be called in row-major order for each element.
pub fn new_with<F: FnMut(usize, usize) -> T>(height: usize, width: usize, mut f: F) -> Self {
    let mut res = Self::new_uninit(height, width);
    for (i, row) in res.iter_mut().enumerate() {
        for (j, element) in row.iter_mut().enumerate() {
            element.write(f(i, j));
        }
    }
    unsafe {
        // SAFETY: All elements are now initialized.
        res.assume_init()
    }
}
```

It's obviously not as flexible as the original, but it covers quite a few use-cases and is simple and safe:

```rust
#[test]
fn test_new_with() {
    let m = Matrix::new_with(HEIGHT, WIDTH, |i, j| (WIDTH * i + j) as u64);
    check_matrix(m);
}
```

If you do need the old behaviour the existing public APIs already allow that cleanly

```rust
#[test]
fn test_new_with3() {
    let m = {
        let mut m = Matrix::new_uninit(HEIGHT, WIDTH);
        for i in 0..HEIGHT {
            for j in 0..WIDTH {
                m[i][j].write((WIDTH * i + j) as u64);
            }
        }
        unsafe { m.assume_init() }
    };
    check_matrix(m);
}
```

recmo:
This is really cool, I expect it to be a big improvement over `Vec<Vec<_>>`. Contiguous memory FTW.

util/src/matrix.rs
nbgl: We do return `Chunks` but that's not something I'd like the API to _promise_. I'd rather keep the flexibility to define our own iterator struct if we need extra functionality.

util/src/matrix.rs
nbgl: Not directly, because Rust doesn't actually guarantee how `Matrix` is laid out in memory. Although we could force a particular layout with `#[repr(C)]`.

util/src/matrix.rs
nbgl: Yeah, I kind of liked the functional style. But you're right: it's not very Rust-like. I'll remove it.

util/src/matrix.rs
nbgl: Thanks!

util/src/matrix.rs
nbgl: Oh true! That is nicer

util/src/matrix.rs
nbgl: You're right that it would be safer, but it also forces a particular evaluation order that's not always optimal (e.g. if you're doing a transpose). I don't think this constructor is actually needed for anything, so I'll just get rid of it and we can bring it back later.

pgebheim:
Hello!

dlubarov:
Let's close this since it's a central aspect of Plonky3, and doesn't seem worthwhile to rework a bunch of Plonky2 code also to work with matrices.

PR #450, opened 2022-01-27
Remove GMiMC and Rescue hash functions

### LINKED ISSUE #432: GMiMC, Rescue aren't compatible with `GoldilocksField`
Our GMiMC and Rescue code is using `x^3`, since when we wrote it we were using `p = 2^64 - 9 * 2^28 + 1`, but we're now using `p = 2^64 - 2^32 + 1`.

We could update it to `x^7` (at least Rescue; GMiMC specifies `x^3` though others might be okay). That said, maybe we should just delete them, since we have no plans to use GMiMC or Rescue.

Thanks to Vijay who noticed this incompatibility.

unzvfu:
Like it says in the title; closes #432.


plonky2/src/gates/gate_tree.rs
```diff
@@ -276,7 +276,7 @@ mod tests {
         assert!(
             gates_with_prefix
                 .iter()
-                .all(|(g, p)| g.0.degree() + g.0.num_constants() + p.len() <= 8),
+                .all(|(g, p)| g.0.degree() + g.0.num_constants() + p.len() <= 9),
```
unzvfu, line 279: Replacing GMiMC with Poseidon changes the prefix lengths obtained by the Tree::traversal() greedy search earlier in this test. Also (probably related) the Poseidon gate has degree 7 while the GMiMC gate has degree 3. These combine to cause the value referred to above to go from 8 to 9. I think this is expected, but would appreciate confirmation from @dlubarov or @wborgeaud.

wborgeaud: Yes no problem with this, the gate tree that is generated in the recursive verifier also has degree 9.

nbgl (approved):
Not sure I'm the best person to review this but it looks good to me!

unzvfu:
> Not sure I'm the best person to review this but it looks good to me!

No worries; it's an easy PR, just needed someone else to compile and run it to make sure I hadn't missed anything.

PR #451, opened 2022-01-27
Start of STARK prover

wborgeaud:
Start of STARK prover, including the computation of quotient polynomials and a toy STARK for testing.


starky/src/fibonacci_stark.rs
```diff
@@ -0,0 +1,98 @@
...
+    const PUBLIC_INPUTS: usize = 0;
+
+    fn eval_packed_generic<FE, P, const D2: usize>(
+        &self,
+        vars: StarkEvaluationVars<FE, P, { Self::COLUMNS }, { Self::PUBLIC_INPUTS }>,
+        yield_constr: &mut ConstraintConsumer<P>,
+    ) where
+        FE: FieldExtension<D2, BaseField = F>,
+        P: PackedField<Scalar = FE>,
+    {
```
dlubarov, line 56: Let's enforce the base case also -
```rust
        yield_constr.one_first_row(vars.local_values[0] - P::from(FE::from_basefield(self.x0)));
        yield_constr.one_first_row(vars.local_values[1] - P::from(FE::from_basefield(self.x1)));
```

starky/src/fibonacci_stark.rs
```diff
@@ -0,0 +1,98 @@
...
+/// Computes a Fibonacci sequence with inital values `x0, x1` using the transition
+/// `x0 <- x1, x1 <- x0 + x1`.
+pub struct FibonacciStark<F: RichField + Extendable<D>, const D: usize> {
+    x0: F,
+    x1: F,
+    _phantom: PhantomData<F>,
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> FibonacciStark<F, D> {
+    const NUM_COLUMNS: usize = 2;
```
dlubarov, line 22: nit: Could use `Self::COLUMNS` instead

system_zero/src/system_zero.rs
```diff
@@ -104,6 +104,6 @@ mod tests {
         let config = StarkConfig::standard_fast_config();
         let mut timing = TimingTree::new("prove", Level::Debug);
         let trace = system.generate_trace();
-        prove::<F, C, S, D>(system, config, trace, &mut timing);
+        prove::<F, C, S, D>(system, config, trace, &mut timing).unwrap();
```
dlubarov, line 107: nit: let's have the test return `Result` so errors will be displayed

starky/src/fibonacci_stark.rs
```diff
@@ -0,0 +1,98 @@
...
+use plonky2::plonk::circuit_builder::CircuitBuilder;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::stark::Stark;
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+/// Toy STARK system used for testing.
+/// Computes a Fibonacci sequence with inital values `x0, x1` using the transition
+/// `x0 <- x1, x1 <- x0 + x1`.
+pub struct FibonacciStark<F: RichField + Extendable<D>, const D: usize> {
```
dlubarov, line 15: nit: `pub` not needed

starky/src/fibonacci_stark.rs
wborgeaud: Oh I was thinking of using public inputs for that (once they're taken care of).

starky/src/fibonacci_stark.rs
```diff
@@ -0,0 +1,98 @@
...
+                acc[1] = tmp[0] + tmp[1];
+                Some(tmp)
+            })
+            .collect()
+    }
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for FibonacciStark<F, D> {
+    const COLUMNS: usize = Self::NUM_COLUMNS;
+    const PUBLIC_INPUTS: usize = 0;
```
dlubarov, line 47: It might be good to exercise public input logic too - we could set this to 1 and add

```rust
yield_constr.one_last_row(vars.local_values[1] - vars.public_inputs[0]);
```

starky/src/prover.rs
```diff
@@ -57,27 +64,149 @@ where
...
+
+    let z_h_on_coset = ZeroPolyOnCoset::new(degree_bits, rate_bits);
+
+    // Retrieve the LDE values at index `i`.
+    let get_at_index = |comm: &PolynomialBatch<F, C, D>, i: usize| -> [F; S::COLUMNS] {
+        comm.get_lde_values(i).try_into().unwrap()
+    };
+
+    alphas
+        .iter()
```
dlubarov, line 180: Sorry for the oversight but in hindsight I think I should have given `ConstraintConsumer` multiple alphas, like `alphas: Vec<P::Scalar>`. So we can call `eval_packed_base` only once per point, and have `ConstraintConsumer` apply similar logic as `reduce_with_powers_multi`.

starky/src/prover.rs
```diff
@@ -57,27 +64,149 @@ where
...
+                            lagrange_last.values[i],
+                        );
+                        let vars =
+                            StarkEvaluationVars::<F, F, { S::COLUMNS }, { S::PUBLIC_INPUTS }> {
+                                local_values: &get_at_index(trace_commitment, i),
+                                next_values: &get_at_index(
+                                    trace_commitment,
+                                    (i + 1) % (degree << rate_bits),
+                                ),
+                                public_inputs: &[F::ZERO; S::PUBLIC_INPUTS],
```
dlubarov, line 199: I guess this should be a param to `prove`, or maybe add a TODO for now

starky/src/prover.rs
dlubarov: (Or can add a TODO for later)

starky/src/fibonacci_stark.rs
dlubarov: Ok that works. I think we could go ahead and add constraints for public inputs, as per below comment, even though there's no verification yet

dlubarov (approved):
LGTM otherwise

starky/src/fibonacci_stark.rs
wborgeaud: I'll add 3 public inputs, 2 for the starting state, and one for the second element of the last state.
The prover will have to take a new parameter `pis: [F; S::PUBLIC_INPUTS]` though, right?

starky/src/prover.rs
wborgeaud: > Sorry for the oversight but in hindsight I think I should have given ConstraintConsumer multiple alphas, and an accumulator for each alpha (so alphas and constraint_acc would be Vecs)


Hmm would that be more efficient than using many `ConstraintConsumer`s? It seems like it would just move the loop on `alphas` elsewhere.

dlubarov: If I'm reading it right, this currently calls `eval_packed_base` once per `(alpha, x)`, but if we did it the other way it'd be once per `x`, right?

starky/src/fibonacci_stark.rs
dlubarov: Not that it really matters since it's a test, but on second thought since `x0`, `x1` are fields of the `Stark`, to me that makes it seem like they'd be baked into the constraint system.

If the `Stark` is more general and can take any `x0`, `x1` are public inputs, maybe it'd make sense to make them arguments to `generate_trace`, instead of fields?

starky/src/prover.rs
wborgeaud: Hmm yeah but then we'd have to change `ConstraintConsumer::one()` to something like
```rust
pub fn one(&mut self, constraint: P) {
    for (alpha, acc) in self.alphas.iter().zip(&mut self.constraint_accs) {
        *acc *=alpha;
        *acc += constraint;
    }
}
```
right? So we'd still have to loop on `alphas`, just somewhere else. Either way is fine for me, just want to make sure I understand correctly.

starky/src/fibonacci_stark.rs
wborgeaud: Good point! I'll implement that.

dlubarov (approved):

PR #454, opened 2022-02-02
Rename `PrimeField` -> `Field64`

dlubarov:
And add TODOs for moving around various methods which aren't well-defined in their current traits, or would be well-defined in a supertrait.

### DIFF
--- a/field/src/field_types.rs
+++ b/field/src/field_types.rs
@@ -264,9 +264,11 @@ pub trait Field:
         subgroup.into_iter().map(|x| x * shift).collect()
     }
 
-    // TODO: move these to a new `PrimeField` trait (for all prime fields, not just 64-bit ones)
+    // TODO: The current behavior for composite fields doesn't seem natural or useful.
+    // Rename to `from_noncanonical_biguint` and have it return `n % Self::characteristic()`.
     fn from_biguint(n: BigUint) -> Self;
 
+    // TODO: Move to a new `PrimeField` trait.
     fn to_biguint(&self) -> BigUint;
 
     fn from_canonical_u64(n: u64) -> Self;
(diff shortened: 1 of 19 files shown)


field/src/field_types.rs
```diff
@@ -264,9 +264,11 @@ pub trait Field:
...
 
-    // TODO: move these to a new `PrimeField` trait (for all prime fields, not just 64-bit ones)
+    // TODO: The current behavior for composite fields doesn't seem natural or useful.
+    // Rename to `from_noncanonical_biguint` and have it return `n % Self::characteristic()`.
     fn from_biguint(n: BigUint) -> Self;
 
+    // TODO: Move to a new `PrimeField` trait.
     fn to_biguint(&self) -> BigUint;
 
     fn from_canonical_u64(n: u64) -> Self;
```
unzvfu, line 274: This is a bit nit-picky and only tangentially related to this PR, but maybe the `from_canonical_*` functions should be marked `unsafe`, because we're trusting that the input is indeed reduced? This is the same rationale behind making `add_canonical_u96` unsafe, as it already is. In a sense, `from_canonical_T` and `from_noncanonical_T` both do the same thing, construct a field element from a T, but the former makes assumptions on the input that are only checked in debug mode. Certainly not a blocker for this PR.

unzvfu (commented):
Looks good to me.

unzvfu (approved):

field/src/field_types.rs
dlubarov: I think I agree. The number of calls is pretty large so I'm a bit concerned about clutter, but safety first I suppose 😄. I added TODOs.

PR #455, opened 2022-02-02
Start of the STARK verifier

wborgeaud:
Start of the STARK verifier. Also adds a `FriChallenges` struct used in `ProofChallenges` and `StarkProofChallenges`, and a distinction between wrapping and non-wrapping constraints.

### DIFF
--- a/field/src/polynomial/mod.rs
+++ b/field/src/polynomial/mod.rs
@@ -57,6 +57,11 @@ impl<F: Field> PolynomialValues<F> {
         fft_with_options(coeffs, Some(rate_bits), None)
     }
 
+    pub fn coset_lde(self, rate_bits: usize) -> Self {
+        let coeffs = ifft(self).lde(rate_bits);
+        coeffs.coset_fft_with_options(F::coset_shift(), Some(rate_bits), None)
+    }
+
     pub fn degree(&self) -> usize {
         self.degree_plus_one()
             .checked_sub(1)
--- a/field/src/polynomial/mod.rs
+++ b/field/src/polynomial/mod.rs
@@ -57,7 +57,8 @@ impl<F: Field> PolynomialValues<F> {
         fft_with_options(coeffs, Some(rate_bits), None)
     }
 
-    pub fn coset_lde(self, rate_bits: usize) -> Self {
+    /// Low-degree extend `Self` (seen as evaluations over the subgroup) onto a coset.
+    pub fn lde_onto_coset(self, rate_bits: usize) -> Self {
         let coeffs = ifft(self).lde(rate_bits);
         coeffs.coset_fft_with_options(F::coset_shift(), Some(rate_bits), None)
     }
--- a/starky/src/constraint_consumer.rs
+++ b/starky/src/constraint_consumer.rs
@@ -104,8 +104,8 @@ impl<F: RichField + Extendable<D>, const D: usize> RecursiveConstraintConsumer<F
         builder: &mut CircuitBuilder<F, D>,
         constraint: ExtensionTarget<D>,
     ) {
-        self.constraint_acc =
-            builder.scalar_mul_add_extension(self.alpha, self.constraint_acc, constraint);
+        let filtered_constraint = builder.mul_extension(constraint, self.z_last);
+        self.constraint(builder, filtered_constraint);
     }
 
     /// Add one constraint valid on all rows.
@@ -114,8 +114,8 @@ impl<F: RichField + Extendable<D>, const D: usize> RecursiveConstraintConsumer<F
         builder: &mut CircuitBuilder<F, D>,
         constraint: ExtensionTarget<D>,
     ) {
-        let filtered_constraint = builder.mul_extension(constraint, self.z_last);
-        self.constraint(builder, filtered_constraint);
+        self.constraint_acc =
+            builder.scalar_mul_add_extension(self.alpha, self.constraint_acc, constraint);
     }
 
     /// Add one constraint, but first multiply it by a filter such that it will only apply to the
--- a/starky/src/constraint_consumer.rs
+++ b/starky/src/constraint_consumer.rs
@@ -105,7 +105,7 @@ impl<F: RichField + Extendable<D>, const D: usize> RecursiveConstraintConsumer<F
         constraint: ExtensionTarget<D>,
     ) {
         let filtered_constraint = builder.mul_extension(constraint, self.z_last);
-        self.constraint(builder, filtered_constraint);
+        self.constraint_wrapping(builder, filtered_constraint);
     }
 
     /// Add one constraint valid on all rows.
--- a/starky/src/constraint_consumer.rs
+++ b/starky/src/constraint_consumer.rs
@@ -14,6 +14,9 @@ pub struct ConstraintConsumer<P: PackedField> {
     /// Running sums of constraints that have been emitted so far, scaled by powers of alpha.
     constraint_accs: Vec<P>,
 
+    /// The evaluation of `X - g^(n-1)`.
+    z_last: P,
+
     /// The evaluation of the Lagrange basis polynomial which is nonzero at the point associated
     /// with the first trace row, and zero at other points in the subgroup.
     lagrange_basis_first: P,
@@ -24,10 +27,16 @@ pub struct ConstraintConsumer<P: PackedField> {
 }
 
 impl<P: PackedField> ConstraintConsumer<P> {
-    pub fn new(alphas: Vec<P::Scalar>, lagrange_basis_first: P, lagrange_basis_last: P) -> Self {
+    pub fn new(
+        alphas: Vec<P::Scalar>,
+        z_last: P,
+        lagrange_basis_first: P,
+        lagrange_basis_last: P,
+    ) -> Self {
         Self {
             constraint_accs: vec![P::ZEROS; alphas.len()],
             alphas,
+            z_last,
             lagrange_basis_first,
             lagrange_basis_last,
         }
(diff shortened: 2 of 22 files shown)


starky/src/constraint_consumer.rs
```diff
@@ -88,42 +98,45 @@ pub struct RecursiveConstraintConsumer<F: RichField + Extendable<D>, const D: us
...
+    pub fn constraint_wrapping(
         &mut self,
         builder: &mut CircuitBuilder<F, D>,
-        constraints: impl IntoIterator<Item = ExtensionTarget<D>>,
+        constraint: ExtensionTarget<D>,
     ) {
-        constraints
-            .into_iter()
-            .for_each(|constraint| self.one(builder, constraint));
+        let filtered_constraint = builder.mul_extension(constraint, self.z_last);
```
dlubarov, line 117: I think the filter should be in `constraint` rather than `constraint_wrapping`

starky/src/proof.rs
```diff
@@ -28,6 +42,25 @@ pub struct CompressedStarkProof<
...
+    F: RichField + Extendable<D>,
+    C: GenericConfig<D, F = F>,
+    const D: usize,
+> {
+    pub proof: CompressedStarkProof<F, C, D>,
+    pub public_inputs: Vec<F>,
+}
+
+pub(crate) struct StarkProofChallenges<F: RichField + Extendable<D>, const D: usize> {
+    // Random values used to combine PLONK constraints.
```
dlubarov, line 55: - Comments need tweaking
- Let's use `///` so tools associate the comment with the field

starky/src/proof.rs
```diff
@@ -56,4 +89,34 @@ impl<F: RichField + Extendable<D>, const D: usize> StarkOpeningSet<F, D> {
             quotient_polys: eval_commitment(zeta, quotient_commitment),
         }
     }
+
+    // Note: Can't implement this directly on `Challenger` as it's in a different crate.
+    pub fn observe<H: Hasher<F>>(&self, challenger: &mut Challenger<F, H>) {
```
dlubarov, line 94: This is only somewhat related to your PR, so maybe it should be a separate change, but it seems like we could replace `Challenger`'s `observe_opening_set` method with one that takes `FriOpenings`. Then methods like this could just become

```rust
challenger.observe_fri_openings(self.to_fri_openings());
```

(It would change the order that openings are observed in; not sure if any other code would need updating to reflect that.)

starky/src/proof.rs
```diff
@@ -56,4 +89,34 @@ impl<F: RichField + Extendable<D>, const D: usize> StarkOpeningSet<F, D> {
...
+        let zeta_batch = FriOpeningBatch {
+            values: [
+                self.local_values.as_slice(),
+                self.quotient_polys.as_slice(),
+                self.permutation_zs.as_slice(),
+            ]
+            .concat(),
+        };
+        let zeta_right_batch = FriOpeningBatch {
+            values: self.next_values.to_vec(),
```
dlubarov, line 116: I think we'll also want to open `permutation_zs` at `g zeta` right?

starky/src/stark.rs
```diff
@@ -67,10 +67,11 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
         zeta: F::Extension,
         g: F::Extension,
         rate_bits: usize,
+        num_challenges: usize,
     ) -> FriInstanceInfo<F, D> {
```
dlubarov, line 71: Let's add a TODO to incorporate `permutation_zs` in this method, now that we're adding some `permutation_zs` code elsewhere

field/src/polynomial/mod.rs
```diff
@@ -57,6 +57,11 @@ impl<F: Field> PolynomialValues<F> {
         fft_with_options(coeffs, Some(rate_bits), None)
     }
 
+    pub fn coset_lde(self, rate_bits: usize) -> Self {
```
dlubarov, line 60: I think the name is a little unclear / possibly misleading; when I saw the calls I initially thought this would treat `self` as poly values over a coset.

Maybe call it something like `lde_onto_coset`? And/or just comment to clarify that the input is treated as points over a subgroup, while the output is treated as points over a coset.

starky/src/prover.rs
```diff
@@ -157,45 +163,57 @@ where
...
+                    trace_commitment,
+                    (i + (1 << rate_bits)) % (degree << rate_bits),
+                ),
                 public_inputs: &public_inputs,
             };
             stark.eval_packed_base(vars, &mut consumer);
-            // TODO: Fix this once we a genuine `PackedField`.
+            // TODO: Fix this once we use a genuine `PackedField`.
             let mut constraints_evals = consumer.accumulators();
+            // We divide the constraints evaluations by `Z_H(x) / x - last`, i.e., the vanishing
```
dlubarov, line 215: Hm is this comment from before we settled on adding the `x - last` filter in `ConstraintConsumer`?

starky/src/verifier.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    (z_x * invs[0], z_x * invs[1])
+}
+
+/// Recover the length of the trace from a STARK proof and a STARK config.
+fn recover_degree<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>(
+    proof: &StarkProof<F, C, D>,
+    config: &StarkConfig,
+) -> usize {
+    1 << (proof.opening_proof.query_round_proofs[0]
```
dlubarov, line 142: nit: Let's split it up for readability, maybe like

```rust
    let initial_merkle_proof = &proof.opening_proof.query_round_proofs[0]
        .initial_trees_proof
        .evals_proofs[0]
        .1;
    let lde_bits = config.fri_config.cap_height + initial_merkle_proof.siblings.len();
    1 << (lde_bits - config.fri_config.rate_bits)
```

starky/src/verifier.rs
```diff
@@ -0,0 +1,150 @@
...
+            .stark_alphas
+            .iter()
+            .map(|&alpha| F::Extension::from_basefield(alpha))
+            .collect::<Vec<_>>(),
+        z_last,
+        l_1,
+        l_last,
+    );
+    stark.eval_ext(vars, &mut consumer);
+    let acc = consumer.accumulators();
```
dlubarov, line 88: nit: not a very meaningful name outside of `ConstraintConsumer`, maybe call it `reduced_constraints` or similar

starky/src/verifier.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    // Check each polynomial identity, of the form `vanishing(x) = Z_H(x) quotient(x)`, at zeta.
+    let quotient_polys_zeta = &proof.openings.quotient_polys;
+    let zeta_pow_deg = challenges.stark_zeta.exp_power_of_2(degree_bits);
+    let z_h_zeta = zeta_pow_deg - F::Extension::ONE;
```
dlubarov, line 93: nit: maybe have `eval_l_1_and_l_last` also return this? Not really concerned about native performance, but it seems kinda natural, and might save a few gates if the recursive circuit follows the same structure

dlubarov: Oh yeah, I see we use `zeta_pow_deg` below... maybe ignore this then

starky/src/verifier.rs
```diff
@@ -0,0 +1,150 @@
...
+    // `quotient_polys_zeta` holds `num_challenges * quotient_degree_factor` evaluations.
+    // Each chunk of `quotient_degree_factor` holds the evaluations of `t_0(zeta),...,t_{quotient_degree_factor-1}(zeta)`
+    // where the "real" quotient polynomial is `t(X) = t_0(X) + t_1(X)*X^n + t_2(X)*X^{2n} + ...`.
+    // So to reconstruct `t(zeta)` we can compute `reduce_with_powers(chunk, zeta^n)` for each
+    // `quotient_degree_factor`-sized chunk of the original evaluations.
+    for (i, chunk) in quotient_polys_zeta
+        .chunks(1 << config.fri_config.rate_bits)
```
dlubarov, line 100: Do you think we should have a `quotient_degree_factor` param as in Plonky2? Looks like it's always `1 << rate_bits` currently. If we wanted to support other factors, we could have `Stark`s implement a `fn constraint_degree()` and subtract 1 from that.

I don't think it's very important to support other factors, just seems a little odd if we support it for circuits but not STARKs.

starky/src/verifier.rs
```diff
@@ -0,0 +1,150 @@
...
+        .enumerate()
+    {
+        ensure!(acc[i] == z_h_zeta * reduce_with_powers(chunk, zeta_pow_deg));
+    }
+
+    let merkle_caps = &[proof.trace_cap, proof.quotient_polys_cap];
```
dlubarov, line 106: nit: Maybe add a TODO for `permutation_zs`

dlubarov (approved):

starky/src/proof.rs
wborgeaud: Good point, I added a `permutation_zs_right` field to `StarkOpeningSet` and fixed that.

starky/src/stark.rs
wborgeaud: It's already on line 65 I think.

field/src/polynomial/mod.rs
wborgeaud: Good point, I changed the name and added a comment.

starky/src/prover.rs
wborgeaud: Oops yes forgot to delete this.

starky/src/verifier.rs
wborgeaud: Yeah I chose to set it to the rate by default for simplicity, but you're right that having more generality could be useful. I've added a TODO on `trait Stark`.

PR #457, opened 2022-02-03
Add `PrimeField`, `PrimeField64` traits

dlubarov:
And move some methods to them, as per the TODOs added in #454. No functional changes.

### DIFF
--- a/field/src/field_types.rs
+++ b/field/src/field_types.rs
@@ -268,9 +268,6 @@ pub trait Field:
     // Rename to `from_noncanonical_biguint` and have it return `n % Self::characteristic()`.
     fn from_biguint(n: BigUint) -> Self;
 
-    // TODO: Move to a new `PrimeField` trait.
-    fn to_biguint(&self) -> BigUint;
-
     /// Returns `n`. Assumes that `n` is already in canonical form, i.e. `n < Self::order()`.
     // TODO: Should probably be unsafe.
     fn from_canonical_u64(n: u64) -> Self;
@@ -407,15 +404,13 @@ pub trait Field:
     }
 }
 
+pub trait PrimeField: Field {
+    fn to_biguint(&self) -> BigUint;
+}
+
 /// A finite field of order less than 2^64.
 pub trait Field64: Field {
     const ORDER: u64;
-
-    // TODO: Only well-defined for prime 64-bit fields. Move to a new PrimeField64 trait?
-    fn to_canonical_u64(&self) -> u64;
-
-    // TODO: Only well-defined for prime 64-bit fields. Move to a new PrimeField64 trait?
-    fn to_noncanonical_u64(&self) -> u64;
 
     /// Returns `x % Self::CHARACTERISTIC`.
     // TODO: Move to `Field`.
@@ -454,6 +449,13 @@ pub trait Field64: Field {
         // Default implementation.
         *self - Self::from_canonical_u64(rhs)
     }
+}
+
+/// A finite field of prime order less than 2^64.
+pub trait PrimeField64: PrimeField + Field64 {
+    fn to_canonical_u64(&self) -> u64;
+
+    fn to_noncanonical_u64(&self) -> u64;
 }
 
 /// An iterator over the powers of a certain base element `b`: `b^0, b^1, b^2, ...`.
--- a/field/src/field_types.rs
+++ b/field/src/field_types.rs
@@ -405,7 +405,7 @@ pub trait Field:
 }
 
 pub trait PrimeField: Field {
-    fn to_biguint(&self) -> BigUint;
+    fn to_canonical_biguint(&self) -> BigUint;
 }
 
 /// A finite field of order less than 2^64.
--- a/field/src/extension_field/quadratic.rs
+++ b/field/src/extension_field/quadratic.rs
@@ -95,10 +95,6 @@ impl<F: Extendable<2>> Field for QuadraticExtension<F> {
         Self([F::from_biguint(low), F::from_biguint(high)])
     }
 
-    fn to_biguint(&self) -> BigUint {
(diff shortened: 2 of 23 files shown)


nbgl:
Have you considered making `PrimeField64` a struct instead of a trait?

Concretely, I'm thinking of something like
```rust
trait PrimeField64Attrs {
    const ORDER: usize;
    const POWER_OF_TWO_GENERATOR: usize;
    // ... anything else we need

    fn reduce_u128(x: u128) -> u64;
    fn add(x: u64, y: u64) -> u64 {
        // some default implementation that we can specialize
    }
}

struct PrimeField64<A: PrimeField64Attrs> {
    val: u64;
}

impl<A: PrimeField64Attrs> PrimeField for PrimeField64<A> {
    const ORDER: usize = A::ORDER;

    // blah blah blah
}
```

The benefit would be that the type system guarantees that `PrimeField64` is stored in a `u64` for all values of `PrimeField64Attrs`.

dlubarov:
Hm not sure I really see the advantage. With `from_noncanonical_u64` and `to_noncanonical_u64` we basically have a setter and getter for (what we would normally expect to be) a `u64`, so it seems similar as long those two are implemented properly right?

I don't see a major drawback either, but a couple small ones -
- types would be a little longer: `PrimeField64<F>` vs `F`
- the API might seem less consistent, since we wouldn't have such a struct for other types of fields

nbgl:
You're right that there's not much of a difference. I just feel that if we're making an assumption about a type, we should do as much to enforce it as the language will allow.

In particular, enforcing that `u64` is used for storage can make packed fields safer. Right now it is impossible for me to write a `PackedField` implementation that works for all `PrimeField64`s. The reasons are:
- A trait cannot enforce the underlying storage size, so I don't know how many values I can fit per vector. If I assume it's 8 bytes, then we will panic _on safe code_ if someone creates a `PrimeField64` that's smaller.
- A trait cannot enforce `#[repr(transparent)]`, which is technically needed for correctness.

I have come across this before, when Rust (correctly) wouldn't let me use `transmute` to convert between a `[PrimeField64; 4]` and a `__m256i`, complaining that they are not guaranteed to have the same size. I had to use a strictly less safe `transmute_copy`. Not a big hassle, but slight annoyance.

Another advantage of a struct is that it can take care of all the parts of `PrimeField64`s that are common between all primes. Functions like `from_canonical_u64`, `rand_from_rng`, `eq`, `sub_assign`, `mul_assign`, `product`, `div`, whose implementation doesn't change. With a trait, one has to write hundreds of line of boilerplate to implement a `PrimeField64`.

I agree that the type being longer is a little bit annoying. Aliases like `type F = PrimeField64<blah>` might help with that, although they would have to exist in a number of places.

The API inconsistency is, I would say, a feature, not a bug. True that the other field types don't come with a struct, but they also don't provide such strong guarantees. If you have a `PrimeField64`, you know it's going to take up exactly 8 bytes; with a generic `Field`, you're on your own. The difference helps reinforce that.

dlubarov:
How about both? We could have a `StandardFp64<A: StandardFp64Attrs>` which implements `PrimeField64`.

`StandardFp64` would make it easier to add other impls as you mentioned, but we could still keep our APIs generic. Ordinary users would interact with the slightly nicer generic API, while only field implementers would need to know about `StandardFp64Attrs`.

This would be reminiscent of "skeletal" implementations like `AbstractList` in the Java/Scala world. Most `List` impls should build off of it, though it's not strictly required. Ordinary users don't need to know about it; they can just read the `List` docs which don't talk about implementation mechanics.

(Granted, our field traits aren't pure interfaces like `List`, since they have some default methods. So our `Field` docs might need to mention some implementation mechanics that aren't relevant to ordinary users. I still think it's best to keep that to a minimum, though.)

nbgl:
Consider this concrete scenario: `transpose` is a function that accepts a matrix of `PrimeField64`. We can make it faster if we can guarantee that all elements are exactly eight bytes (and of the correct layout). Size 8 lets us write AVX2/NEON routines that will transpose a matrix quickly.

What happens if we don't have the guarantee that all elements have size 8?

Options:
1. Have slower code.
2. Assume that `PrimeField64` has the correct size and layout. Works, but makes `PrimeField64` an `unsafe` trait.
3. Have multiple versions of the same code, specializing for `StandardFp64`. Also works, but adds complexity and someone implementing `PrimeField64` may be surprised by the performance penalty.

None of these options are _terrible_ but they're not _great_ either. Making `PrimeField64` a struct would make it easier to write such optimizations. But a trait wouldn't be the end of the world 🤷‍♂️

dlubarov:
Discussed a bit offline, I think it makes sense to proceed with this for now, as a quick fix for the function scope issues. It might be best to go with a struct, but would prefer to do it as a separate change since the diff will be large.

field/src/field_types.rs
```diff
@@ -407,16 +404,14 @@ pub trait Field:
     }
 }
 
+pub trait PrimeField: Field {
+    fn to_biguint(&self) -> BigUint;
```
unzvfu, line 408: Not strictly related to this PR, but while I'm here: For `u64` we have two variants of this method, `to_canonical_u64` and `to_noncanonical_u64`. The implementations of `to_biguint` all provide "`to_canonical_biguint`" functionality, so I think either this method should be called `to_canonical_biguint` or there should be some documentation for the caller so they know the result is reduced. I would tend to favour renaming the method so that it's similar to the other "upcast" methods, but I'm happy either way.

unzvfu (approved):
Looks good to me. Only one fairly unimportant observation to consider.

field/src/field_types.rs
dlubarov: Yeah good point - `to_canonical_biguint` seems fitting.

PR #458, opened 2022-02-03
Separate methods for hashing with or without padding

dlubarov:
This should be a tad better for for performance, and lets us do padding in a generic way, rather than each hash reimplementing it.

This also disables padding for public inputs. It seems unnecessary since the number of public inputs is fixed for any given instance.

### DIFF
--- a/plonky2/src/hash/hashing.rs
+++ b/plonky2/src/hash/hashing.rs
@@ -18,7 +18,7 @@ pub fn hash_or_noop<F: RichField, P: PlonkyPermutation<F>>(inputs: Vec<F>) -> Ha
     if inputs.len() <= 4 {
         HashOut::from_partial(inputs)
     } else {
-        hash_n_to_hash::<F, P>(&inputs, false)
+        hash_n_to_hash_no_pad::<F, P>(&inputs)
     }
 }
 
@@ -28,34 +28,23 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         if inputs.len() <= 4 {
             HashOutTarget::from_partial(inputs, zero)
         } else {
-            self.hash_n_to_hash::<H>(inputs, false)
+            self.hash_n_to_hash_no_pad::<H>(inputs)
         }
     }
 
-    pub fn hash_n_to_hash<H: AlgebraicHasher<F>>(
+    pub fn hash_n_to_hash_no_pad<H: AlgebraicHasher<F>>(
         &mut self,
         inputs: Vec<Target>,
-        pad: bool,
     ) -> HashOutTarget {
-        HashOutTarget::from_vec(self.hash_n_to_m::<H>(inputs, 4, pad))
+        HashOutTarget::from_vec(self.hash_n_to_m_no_pad::<H>(inputs, 4))
     }
 
-    pub fn hash_n_to_m<H: AlgebraicHasher<F>>(
+    pub fn hash_n_to_m_no_pad<H: AlgebraicHasher<F>>(
         &mut self,
-        mut inputs: Vec<Target>,
+        inputs: Vec<Target>,
         num_outputs: usize,
-        pad: bool,
     ) -> Vec<Target> {
         let zero = self.zero();
-        let one = self.one();
-
-        if pad {
-            inputs.push(zero);
-            while (inputs.len() + 1) % SPONGE_WIDTH != 0 {
-                inputs.push(one);
-            }
-            inputs.push(zero);
-        }
 
         let mut state = [zero; SPONGE_WIDTH];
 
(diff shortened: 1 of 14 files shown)


plonky2/src/hash/keccak.rs
```diff
@@ -56,7 +56,7 @@ impl<F: RichField, const N: usize> Hasher<F> for KeccakHash<N> {
     type Hash = BytesHash<N>;
     type Permutation = KeccakPermutation;
 
-    fn hash(input: &[F], _pad: bool) -> Self::Hash {
```
dlubarov, line 59: This was ignoring the flag, which I guess was fine since the `keccak` call should pad. We might be able to skip padding here, but I don't think we should since we need consistency with the EVM's `KECCAK256` instruction, which I believe will always pad.

After this change, `hash_no_pad` will pad, and `hash_pad` will double-pad. I guess I should override `hash_pad` to call `hash_no_pad`, so that both will pad once, as before. Does that sound good?

nbgl:
I'm not sure that the current handling of padding has that much impact on the speed in the non-padded case. The branch should be very predictable, and if `hash_n_to_m` is inlined, it will go away altogether.

In the padded case, we allocate a vector for the padded input. @dlubarov, I think you mentioned that we don't really care about the speed in the padded case, so this is fine. But if we wanted to speed up padding, we could handle it by special-casing the last two (1) hashes after the for loop in `hash_n_to_m`. We cannot do this if the padding is delegated to a different function.

It is nice to see that `pad` argument gone from `hash_n_to_m`. It was always a bit confusing: you can't tell what it does just from the function name.

(1) two because the padding can go slightly over one chunk

dlubarov:
Public inputs was actually the last use of padding, so now that that's switched, we could consider just removing `hash_pad` altogether. The only hesitation I have is that some other projects might be interested in using it. WDYT?

If we do keep it, we could potentially avoid allocation by accepting iterators and chaining on the padding. I gave this a try in #460. It seems to be a bit slower though. I had to replace `copy_from_slice` with a simple for-loop copy, so I suspect the old version may have been compiled with some optimized memcpy implementation.

plonky2/src/hash/keccak.rs
wborgeaud: Sounds good to me.

plonky2/src/plonk/config.rs
```diff
@@ -31,7 +31,21 @@ pub trait Hasher<F: RichField>: Sized + Clone + Debug + Eq + PartialEq {
     /// Permutation used in the sponge construction.
     type Permutation: PlonkyPermutation<F>;
 
-    fn hash(input: &[F], pad: bool) -> Self::Hash;
+    /// Hash a message without any padding step. Note that this can enable length-extension attacks.
+    /// However, it is still collision-resistant in cases where the input has a fixed length.
+    fn hash_no_pad(input: &[F]) -> Self::Hash;
+
+    /// Pad the message using the `pad10*1` rule, then hash it.
```
wborgeaud, line 38: nit: I've always found this confusing, it seems like the padding is of the form `01*0`, not `10*1`.

plonky2/src/hash/hashing.rs
```diff
@@ -100,21 +89,10 @@ pub trait PlonkyPermutation<F: RichField> {
 /// If `pad` is enabled, the message is padded using the pad10*1 rule. In general this is required
 /// for the hash to be secure, but it can safely be disabled in certain cases, like if the input
 /// length is fixed.
```
wborgeaud, line 91: This comment should be updated/removed.

plonky2/src/hash/hashing.rs
```diff
@@ -136,9 +114,6 @@ pub fn hash_n_to_m<F: RichField, P: PlonkyPermutation<F>>(
     }
 }
```
wborgeaud, line 115: This function returns `[0,0,0,0]` when `inputs` is empty. In particular `permute` isn't called on empty inputs. I'm not sure if it's a big issue, but it's the reason some tests are failing. Since the public inputs are hashed without padding now, any circuit without public inputs doesn't use a `PoseidonGate` anymore which makes some buggy tests fail.

wborgeaud (approved):
Looks good to me!

> Public inputs were actually the last use of padding, so now that that's switched, we could consider just removing `hash_pad` altogether. The only hesitation I have is that some other projects might be interested in using it. WDYT?

Either way is fine for me, but I would probably just keep it since it's a natural function to have next to `hash_no_pad` and we (or other projects) might need it at some point.

plonky2/src/hash/keccak.rs
wborgeaud: Sounds good to me.

nbgl:
> Public inputs were actually the last use of padding, so now that that's switched, we could consider just removing `hash_pad` altogether. The only hesitation I have is that some other projects might be interested in using it. WDYT?

Tbh, the padding is simple enough to roll yourself. We've already done the hard bits!

> If we do keep it, we could potentially avoid allocation by accepting iterators and chaining on the padding. I gave this a try in #460. It seems to be a bit slower though. I had to replace `copy_from_slice` with a simple for-loop copy, so I suspect the old version may have been compiled with some optimized memcpy implementation.

Yeah, I guess the current solution can copy a whole vector at a time, whereas the iterator solution can only copy 64 bits at a time (because the iterator length may not be a multiple of vector length).

Another hypothesis is that currently we're not doing memory writes at all, storing the state array in registers. If Rust is inserting a bunch of if statements, to see which iterator in the chain we're up to, then LLVM might get confused and refuse to keep the array in registers, forcing writes and reads where none existed before. In this case loop unrolling might help.

plonky2/src/hash/hashing.rs
dlubarov: Done

plonky2/src/plonk/config.rs
dlubarov: Hm I think I just got it backwards (though either should be secure). Fixed.

plonky2/src/hash/hashing.rs
dlubarov: Oh thanks for investigating! I'll wait for #469 to land.

nbgl: #469 has shipped 😊

PR #459, opened 2022-02-03
Implement Poseidon in system_zero/permutation_unit

nbgl:


system_zero/src/column_layout.rs
```diff
@@ -24,35 +24,56 @@ pub(crate) const COL_STACK_PTR: usize = COL_FRAME_PTR + 1;
...
+}
+
+pub(crate) const fn col_permutation_full_second_after_mds(round: usize, i: usize) -> usize {
     debug_assert!(round <= poseidon::HALF_N_FULL_ROUNDS);
     debug_assert!(i < SPONGE_WIDTH);
-    START_PERMUTATION_FULL_SECOND + round * SPONGE_WIDTH + i
+    START_PERMUTATION_FULL_SECOND + (2 * round + 1) * SPONGE_WIDTH + i
 }
 
 pub(crate) const fn col_permutation_input(i: usize) -> usize {
```
dlubarov, line 69: Let's move this up near `START_PERMUTATION_UNIT` now that it's really part of the layout definition (before it was like an alias)

system_zero/src/permutation_unit.rs
```diff
@@ -2,36 +2,117 @@ use plonky2::field::extension_field::{Extendable, FieldExtension};
...
+fn constant_layer<F, FE, P, const D2: usize>(
+    mut state: [P; SPONGE_WIDTH],
+    round: usize,
+) -> [P; SPONGE_WIDTH]
+where
+    F: RichField,
+    FE: FieldExtension<D2, BaseField = F>,
+    P: PackedField<Scalar = FE>,
+{
+    // One day I might actually vectorize this, but today is not that day.
```
dlubarov, line 29: This should work right?

```rust
for i in 0..SPONGE_WIDTH {
    state[i] += P::from(FE::from_canonical_u64(ALL_ROUND_CONSTANTS[i + SPONGE_WIDTH * round]));
}
```

system_zero/src/permutation_unit.rs
```diff
@@ -2,36 +2,117 @@ use plonky2::field::extension_field::{Extendable, FieldExtension};
...
+    }
+    state
+}
+
 impl<F: RichField + Extendable<D>, const D: usize> SystemZero<F, D> {
-    pub(crate) fn generate_permutation_unit(&self, values: &mut [F; NUM_COLUMNS]) {
+    pub(crate) fn generate_permutation_unit(values: &mut [F; NUM_COLUMNS]) {
         // Load inputs.
-        let mut state = [F::ZERO; SPONGE_WIDTH];
+        let mut state = [F::ZEROS; SPONGE_WIDTH];
```
dlubarov, line 65: nit: feels a little odd to me to use the plural in scalar code (and I don't think we can vectorize this to compute multiple permutations, since we generally only know the preimage of the current cycle)

system_zero/src/permutation_unit.rs
```diff
@@ -2,36 +2,117 @@ use plonky2::field::extension_field::{Extendable, FieldExtension};
...
+        for j in 0..SPONGE_WIDTH {
+            state[j].as_slice_mut()[i] = unpacked_state[j];
+        }
+    }
+    state
+}
+
 impl<F: RichField + Extendable<D>, const D: usize> SystemZero<F, D> {
-    pub(crate) fn generate_permutation_unit(&self, values: &mut [F; NUM_COLUMNS]) {
+    pub(crate) fn generate_permutation_unit(values: &mut [F; NUM_COLUMNS]) {
```
dlubarov, line 63: Not that this should be a major cost anyway, but maybe it would be good to short-circuit if the input is zeros. I expect that will be common - maybe 50% or more of cycles won't use the permutation unit.

(I mean only for witness generation; the eval methods should generally receive "random" looking inputs since they come from evaluating witness polys on a different domain.)

system_zero/src/permutation_unit.rs
```diff
@@ -2,36 +2,117 @@ use plonky2::field::extension_field::{Extendable, FieldExtension};
...
-        // TODO: Partial rounds.
-        // TODO: Second full rounds.
+        for r in 0..HALF_N_FULL_ROUNDS {
+            F::constant_layer(&mut state, r);
 
-        // Write outputs.
-        for i in 0..SPONGE_WIDTH {
-            values[col_permutation_output(i)] = state[i];
+            for i in 0..SPONGE_WIDTH {
+                let state_cubed = state[i] * state[i].square();
```
dlubarov, line 74: nit: could use `cube()`

dlubarov:
Sorry for the conflicts, William and I changed `ConstraintConsumer`'s `one` to `constraint_wrapping`. (Actually having second thoughts about that naming, I might suggest something else later...)

system_zero/src/permutation_unit.rs
```diff
@@ -72,15 +195,152 @@ impl<F: RichField + Extendable<D>, const D: usize> SystemZero<F, D> {
...
+            assert_eq!(acc, GoldilocksField::ZERO);
         }
     }
+
+    #[test]
+    fn poseidon_result() {
+        const D: usize = 1;
+        type F = GoldilocksField;
+
+        let mut state = [F::default(); SPONGE_WIDTH];
```
dlubarov, line 322: Maybe use `F::rand_arr::<NUM_COLUMNS>()` to make sure it passes with other inputs?

system_zero/src/permutation_unit.rs
nbgl: Yeah, this is not intentional, just a copy/paste error. Nice catch!

dlubarov (approved):

system_zero/src/permutation_unit.rs
nbgl: I didn't do this to avoid copying the Poseidon logic into the permutation unit (as much as possible, anyway). I'd rather make the Poseidon function work with packed fields, which will be easy, but is a different PR.

system_zero/src/permutation_unit.rs
nbgl: I haven't added `cube` to packed fields, because I'm not convinced I can do it faster than a square and a mul. But that also makes it trivial to implement. Do you think it's worth adding?

system_zero/src/permutation_unit.rs
nbgl: Good idea!

system_zero/src/permutation_unit.rs
nbgl: I'd rather do this optimization if we find it actually speeds things up. I'd have to keep the hash of `[0; 12]` in a constant, which would then have to be kept up to date with Poseidon parameters.

dlubarov (approved):

PR #460, opened 2022-02-03
Have hash methods take iterators

dlubarov:


plonky2/src/hash/hashing.rs
```diff
@@ -90,14 +92,22 @@ pub trait PlonkyPermutation<F: RichField> {
...
+        if pos == SPONGE_RATE {
+            state = P::permute(state);
+            pos = 0;
+        }
+        state[pos] = *input.borrow();
+        pos += 1;
+    }
+    if pos != 0 {
         state = P::permute(state);
     }
```
nbgl, line 112:
```suggestion
    let inputs = inputs.into_inter().fused();
    while Some(state_0) = inputs.next() {
        state[0] = *state_0.borrow();
        for (state_i, i) in inputs.zip(1..SPONGE_RATE) {
            state[i] = *state_i.borrow();
        }
        state = P::permute(state);
    }
```

dlubarov: Good idea; changed. I'm getting similar performance, but it is nicer at least.

I just had to switch the `zip` order, to avoid zip consuming an item from `inputs` before realizing the other iterator was exhausted. Took me a while to notice that 😅

dlubarov: (Though take my perf claims with a grain of salt, haven't done proper side by side comparisons)

nbgl: > consuming an item from `inputs` before realizing the other iterator was exhausted

Oh true! oops

PR #464, opened 2022-02-03
Replace `observe_opening_set` by `observe_openings` taking a `FriOpenings`

wborgeaud:
Makes the challenger code more generic and simplifies the STARK code.


dlubarov (approved):
Nice, thanks for doing that

PR #466, opened 2022-02-03
Glv

npwardberkeley:


wborgeaud:
I think this should be ready to merge now. 
This implements gadgets for GLV, windowed scalar multiplication and fixed base multiplication. These methods are combined in the ECDSA verification circuit.

Issue #463, opened 2022-02-03
Issues with small circuits


wborgeaud:
- `MerkleTree::new` panics when `leaves.len() == 1 << cap_height`. The expected behavior is a Merkle tree with empty digest and with cap set to the hash of the leaves.
- `MerkleTree::new` rightfully panics when `leaves.len() < 1 << cap_height` with a subtraction overflow. We should probably make it panic with a clear error message like `panic!("Cap height should be at most log(leaves.len())")`.  For circuits, we should also probably catch this earlier, e.g., in `CircuitBuilder::build()`.
- In `compute_quotient_polys`, a hardcoded `BATCH_SIZE=32` is used. This leads to a panic if the size of the LDE is less than 32.

nbgl:
Not handling very small circuits in `compute_quotient_polys` was a deliberate choice for simplicity. I’m happy to add the special case if you think it’d be useful.

nbgl:
It’s more to do with how to the gates are vectorized. If the circuit size is not a multiple of the vector length, then you need extra code to deal with the leftovers.

I’ll make a PR for it today.

nbgl:
Fixed in #467 and #469. (I thought extra changes were needed to vectorized gate evaluation code to account for leftovers, but it turns out I've already implemented that.)

PR #467, opened 2022-02-03
Merkle tree bugfixes + tests

nbgl:
Addresses part of #463. Prints a better message when panicking because the cap height is too large. Fixes the panic that would occur when `cap_height == log2(leaves.len())`.

### DIFF
--- a/plonky2/src/hash/merkle_tree.rs
+++ b/plonky2/src/hash/merkle_tree.rs
@@ -93,13 +93,14 @@ fn fill_digests_buf<F: RichField, H: Hasher<F>>(
     // Special case of a tree that's all cap. The usual case will panic because we'll try to split
     // an empty slice into chunks of `0`. (We would not need this if there was a way to split into
     // `blah` chunks as opposed to chunks _of_ `blah`.)
-    if digests_buf.len() == 0 {
+    if digests_buf.is_empty() {
         debug_assert_eq!(cap_buf.len(), leaves.len());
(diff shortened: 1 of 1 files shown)


plonky2/src/hash/merkle_tree.rs
```diff
@@ -194,16 +214,45 @@ mod tests {
         const D: usize,
     >(
         leaves: Vec<Vec<F>>,
-        n: usize,
+        cap_height: usize,
```
nbgl, line 217: Changing the signature introduce `cap_height`. Got rid of `n` while I was at it because it was redundant.

plonky2/src/hash/merkle_tree.rs
```diff
@@ -90,6 +90,20 @@ fn fill_digests_buf<F: RichField, H: Hasher<F>>(
...
+    if digests_buf.is_empty() {
+        debug_assert_eq!(cap_buf.len(), leaves.len());
+        cap_buf
+            .par_iter_mut()
+            .zip(leaves)
+            .for_each(|(cap_buf, leaf)| {
+                cap_buf.write(H::hash(leaf, false));
+            });
+        return;
+    }
```
nbgl, line 105: A special case with an early return was the neatest way I found to do this. 

The closure in the normal case is correct when `subtree_digests.is_empty()` but Rust will not produce such a value. In other words, the _real_ bug here is Rust's well-meaning sanity checks…

wborgeaud (approved):
Looks good to me!

PR #468, opened 2022-02-04
Stark constraint degree

wborgeaud:
Adds a `fn degree(&self) -> usize` method to the `Stark` trait and use it to get the minimal number of quotient polynomial chunks.
Also adds a test `test_stark_low_degree` to check the degree of the STARK constraints, analogous to `test_low_degree` for gates.

### DIFF
--- a/starky/src/fibonacci_stark.rs
+++ b/starky/src/fibonacci_stark.rs
@@ -83,7 +83,7 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for FibonacciStar
     }
 
     fn degree(&self) -> usize {
-        3
+        2
     }
 }
 
--- a/starky/src/fibonacci_stark.rs
+++ b/starky/src/fibonacci_stark.rs
@@ -97,7 +97,7 @@ mod tests {
     use crate::config::StarkConfig;
     use crate::fibonacci_stark::FibonacciStark;
     use crate::prover::prove;
-    use crate::stark_testing::test_low_degree;
+    use crate::stark_testing::test_stark_low_degree;
     use crate::verifier::verify;
 
     fn fibonacci(n: usize, x0: usize, x1: usize) -> usize {
@@ -141,6 +141,6 @@ mod tests {
         let config = StarkConfig::standard_fast_config();
         let num_rows = 1 << 5;
         let stark = S::new(num_rows);
-        test_low_degree(stark)
+        test_stark_low_degree(stark)
     }
 }
--- a/starky/src/fibonacci_stark.rs
+++ b/starky/src/fibonacci_stark.rs
@@ -81,6 +81,10 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for FibonacciStar
     ) {
         todo!()
     }
+
+    fn degree(&self) -> usize {
+        2
+    }
 }
 
 #[cfg(test)]
@@ -93,6 +97,7 @@ mod tests {
     use crate::config::StarkConfig;
     use crate::fibonacci_stark::FibonacciStark;
     use crate::prover::prove;
+    use crate::stark_testing::test_low_degree;
     use crate::verifier::verify;
 
     fn fibonacci(n: usize, x0: usize, x1: usize) -> usize {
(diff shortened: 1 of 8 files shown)


starky/src/fibonacci_stark.rs
```diff
@@ -81,6 +81,10 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for FibonacciStar
     ) {
         todo!()
     }
+
+    fn constraint_degree(&self) -> usize {
+        2
```
dlubarov, line 86: Seems like it's degree 1, but I guess that would give a quotient degree of 0 which wouldn't work? How about if we did

```rust
fn quotient_degree_factor(&self) {
    1.max(self.constraint_degree() - 1)
}
```

dlubarov: (or maybe a global method so it's not public and can't be overridden)

starky/src/prover.rs
```diff
@@ -164,34 +162,44 @@ where
 {
     let degree = 1 << degree_bits;
 
+    let max_degree_bits = log2_ceil(stark.constraint_degree() - 1);
```
dlubarov, line 165: I know this matches the PLONK code but the name seems kinda unclear to me; maybe something like `evaluation_domain_bits` would be more fitting

starky/src/prover.rs
```diff
@@ -164,34 +162,44 @@ where
...
+    assert!(
+        max_degree_bits <= rate_bits,
+        "Having constraints of degree higher than the rate is not supported yet."
+    );
+    let step = 1 << (rate_bits - max_degree_bits);
+    // When opening the `Z`s polys at the "next" point, need to look at the point `next_step` steps away.
+    let next_step = 1 << max_degree_bits;
```
dlubarov, line 172: nit: maybe `step_size`? `next_step` kinda sounds like the absolute index vs the delta

starky/src/stark_testing.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn test_stark_low_degree<F: RichField + Extendable<D>, S: Stark<F, D>, const D: usize>(
+    stark: S,
+) -> Result<()>
+where
+    [(); S::COLUMNS]:,
+    [(); S::PUBLIC_INPUTS]:,
+{
+    let rate_bits = log2_ceil(stark.constraint_degree() + 1);
+
+    let wire_ldes = random_low_degree_matrix::<F>(S::COLUMNS, rate_bits);
```
dlubarov, line 27: Let's call it `trace_ldes` now that it's a STARK. I think it's also more common for STARK code to say "trace" instead of "witness", though either seems okay.

starky/src/stark_testing.rs
```diff
@@ -0,0 +1,98 @@
...
+        evals.values[0] = F::ONE;
+        evals.lde(rate_bits)
+    };
+    let lagrange_last = {
+        let mut evals = PolynomialValues::new(vec![F::ZERO; WITNESS_SIZE]);
+        evals.values[WITNESS_SIZE - 1] = F::ONE;
+        evals.lde(rate_bits)
+    };
+
+    let z_h_on_coset = ZeroPolyOnCoset::<F>::new(log2_strict(WITNESS_SIZE), rate_bits);
```
dlubarov, line 42: unused?

starky/src/fibonacci_stark.rs
wborgeaud: > Seems like it's degree 1, but I guess that would give a quotient degree of 0 which wouldn't work?

I think it's of degree 2. For example 
```rust
yield_constr.constraint_first_row(vars.local_values[0] - vars.public_inputs[Self::PI_INDEX_X0]);
```
has degree 2.

starky/src/prover.rs
wborgeaud: Hmm I agree `max_degree_bits` isn't a very good name, but I would interpret `evaluation_domain_bits` as the log of the LDE size. What do you think about `constraint_degree_bits` or `quotient_degree_bits`?

starky/src/stark_testing.rs
wborgeaud: Oops yeah. That's weird, it seems like this should get caught by Clippy...

dlubarov (approved):
LGTM otherwise

starky/src/prover.rs
dlubarov: `quotient_degree_bits` sounds good to me

starky/src/fibonacci_stark.rs
dlubarov: Oh yeah makes sense.

starky/src/stark_testing.rs
dlubarov: I did `#![allow(dead_code)]` for now, just until we fill in the `todo!()`s

wborgeaud: Ohh I forgot about that, makes sense 👍

PR #469, opened 2022-02-04
Permit small circuits in `compute_quotient_polys`

nbgl:
Note that this still does not permit batch sizes that are not a multiple of the packing length used for the field. This will require separate changes to `gates:packed_util::PackedEvaluableBase`.

### DIFF
--- a/plonky2/src/plonk/prover.rs
+++ b/plonky2/src/plonk/prover.rs
@@ -5,7 +5,7 @@ use anyhow::Result;
 use plonky2_field::extension_field::Extendable;
 use plonky2_field::polynomial::{PolynomialCoeffs, PolynomialValues};
 use plonky2_field::zero_poly_coset::ZeroPolyOnCoset;
-use plonky2_util::log2_ceil;
+use plonky2_util::{ceil_div_usize, log2_ceil};
 use rayon::prelude::*;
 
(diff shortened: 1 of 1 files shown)


nbgl:
Also, I'm not sure I'm testing this right. @wborgeaud would you mind writing a few tests to ensure that we handle small circuits correctly?

plonky2/src/plonk/prover.rs
```diff
@@ -357,12 +357,18 @@ fn compute_quotient_polys<
     let z_h_on_coset = ZeroPolyOnCoset::new(common_data.degree_bits, max_degree_bits);
 
     let points_batches = points.par_chunks(BATCH_SIZE);
+    let num_batches = (points.len() + (BATCH_SIZE - 1)) / BATCH_SIZE;
```
wborgeaud, line 360: nit: We have a `ceil_div_usize()` function for this.

wborgeaud (approved):
Looks good to me!

> Also, I'm not sure I'm testing this right. @wborgeaud would you mind writing a few tests to ensure that we handle small circuits correctly?

Hmm it's actually quite hard to get small circuits right now since every circuit hashes public inputs, but that will change with #458. Maybe we can wait until this PR lands to add small circuits tests.

PR #471, opened 2022-02-07
Move some FRI stuff into the FRI module

dlubarov:
I think it would be nice to eventually have separate crates for IOP, FRI, PLONK, etc. This is one step toward that. No functional changes.


wborgeaud (approved):
Looks good to me!

PR #472, opened 2022-02-07
`trim_to_len` helper function

dlubarov:
Seems a little nicer IMO to only remove a certain number of zeros, vs removing all trailing zeros then re-adding some.

### DIFF
--- a/field/src/polynomial/mod.rs
+++ b/field/src/polynomial/mod.rs
@@ -186,13 +186,13 @@ impl<F: Field> PolynomialCoeffs<F> {
         poly
     }
 
-    /// Removes any zero coefficients from the side associated with the highest exponents.
+    /// Removes any leading zero coefficients.
     pub fn trim(&mut self) {
         self.coeffs.truncate(self.degree_plus_one());
     }
 
-    /// Removes zero coefficients from the side associated with the highest exponents, such that a
-    /// desired length is reached. Fails if a nonzero coefficient is encountered before then.
+    /// Removes some leading zero coefficients, such that a desired length is reached. Fails if a
+    /// nonzero coefficient is encountered before then.
     pub fn trim_to_len(&mut self, len: usize) -> Result<()> {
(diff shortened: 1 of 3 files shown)


field/src/polynomial/mod.rs
```diff
@@ -186,11 +186,20 @@ impl<F: Field> PolynomialCoeffs<F> {
         poly
     }
 
-    /// Removes leading zero coefficients.
+    /// Removes any zero coefficients from the side associated with the highest exponents.
```
wborgeaud, line 189: nit: Hmm I find "coefficients from the side associated with the highest exponents" less clear than "leading coefficients". I think "leading coefficient" is well-established polynomial terminology, so I don't think there should be confusion about which side of the vector representation we're removing zeros from.

field/src/polynomial/mod.rs
```diff
@@ -186,11 +186,20 @@ impl<F: Field> PolynomialCoeffs<F> {
...
+    /// Removes zero coefficients from the side associated with the highest exponents, such that a
+    /// desired length is reached. Fails if a nonzero coefficient is encountered before then.
+    pub fn trim_to_len(&mut self, len: usize) -> Result<()> {
+        ensure!(self.len() >= len);
+        ensure!(self.coeffs[len..].iter().all(F::is_zero));
+        self.coeffs.truncate(len);
+        Ok(())
+    }
+
     /// Removes leading zero coefficients.
```
wborgeaud, line 203: nit: In case we keep the "coefficients from the side associated with the highest exponents" terminology, we should maybe update this comment too.

wborgeaud (approved):
Looks good to me!

field/src/polynomial/mod.rs
dlubarov: Sounds good; I agree it was clunky wording. I was just thinking some readers might interpret "leading" as the front of the `Vec`, but hopefully not.

PR #473, opened 2022-02-07
Standardize `set_*` method parameter order

wborgeaud:
Most functions used to fill the witness are of the form `set_X_target(xt: XTarget, x: X)`.
`set_proof_target`, `set_proof_with_pis_target`, and `set_fri_proof_target` currently use the opposite parameter order.
This PR fixes that.


dlubarov (approved):
👍

PR #474, opened 2022-02-07
Match native and recursive verifier structures

wborgeaud:
The native and recursive verifiers currently do not have the same structure. For example, the native verifier first computes all challenges, while the recursive verifier computes these challenges in the middle of other verification steps.
This PR unifies the code of the native and recursive verifier, making it easier to audit and providing a common structure also for the STARK (recursive) verifier.


plonky2/src/plonk/get_challenges.rs
```diff
@@ -219,3 +225,96 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
         FriInferredElements(fri_inferred_elements)
     }
 }
+
+impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
+    pub(crate) fn get_challenges<C: GenericConfig<D, F = F>>(
```
dlubarov, line 230: nit: seems like this can be private

dlubarov (approved):
LGTM

PR #478, opened 2022-02-10
`from_partial` now takes in a slice

BGluth:
This was causing a few calls in the client to allocate a `Vec` when we didn't need to.

### DIFF
--- a/plonky2/src/hash/hash_types.rs
+++ b/plonky2/src/hash/hash_types.rs
@@ -31,14 +31,12 @@ impl<F: Field> HashOut<F> {
         }
     }
 
-    pub fn from_partial(mut elements: Vec<F>) -> Self {
-        debug_assert!(elements.len() <= 4);
-        while elements.len() < 4 {
-            elements.push(F::ZERO);
-        }
-        Self {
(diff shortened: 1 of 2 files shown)


plonky2/src/hash/hash_types.rs
```diff
@@ -31,14 +31,12 @@ impl<F: Field> HashOut<F> {
...
-    pub fn from_partial(mut elements: Vec<F>) -> Self {
-        debug_assert!(elements.len() <= 4);
-        while elements.len() < 4 {
-            elements.push(F::ZERO);
-        }
-        Self {
-            elements: [elements[0], elements[1], elements[2], elements[3]],
-        }
+    pub fn from_partial(elements_in: &[F]) -> Self {
+        debug_assert!(elements_in.len() <= 4);
```
nbgl, line 35: I don't think this `debug_assert` is necessary anymore: the bounds checking on `elements` should handle it now.

BGluth: Oh nice good to know!

dlubarov (approved):

PR #483, opened 2022-02-12
Logic unit stubs

dlubarov:


nbgl:
What do these stubs cover. Bitwise operations is a given. Anything else? E.g., bitshifts, sign extension, arithmetic?

What determines what operation is performed?

dlubarov:
At the time I was thinking anything we compute in a "bit oriented" way would go in the logic unit. Might need to rethink that after our chat though, if we decide to use lookup tables.

For arithmetic we have the separate unit, but I didn't really have a plan yet for how to handle or where to put bitshifts or sign extension. Let's talk about it in the sync.

I guess there isn't a clear line between arithmetic and logic computations. Do you think one larger "ALU" would make more sense? I think it shouldn't affect how we implement things, just code organization.

> What determines what operation is performed?

Nothing implemented for this yet, but it seems like for most ALU operations, we'll just want to enable a certain circuit (e.g. division) based on whether the current opcode (after reading an instruction from memory and decoding it) has a certain value. I wonder if there will be exceptions though.

nbgl (approved):

dlubarov:
We decided to put logic operations in an ALU instead: #496

PR #484, opened 2022-02-12
Move some methods outside `impl System`

dlubarov:
I didn't really have a good reason for putting there; seems more idiomatic to make them global since they don't need `self`/`Self`.

Also tweak some generic bounds that seemed more restrictive than necessary.

No functional changes.


nbgl (approved):
👍

PR #485, opened 2022-02-13
Prover code for permutation argument

dlubarov:
It's slightly complex because we batch `constraint_degree - 1` permutation arguments into a single `Z` polynomial. This is a slight generalization of the [technique](https://zcash.github.io/halo2/design/proving-system/lookup.html) described in the Halo2 book.

Without this batching, we would simply have `num_challenges` random challenges (betas and gammas). With this batching, however, we need to use different randomness for each permutation argument within the same batch. Hence we end up generating `batch_size * num_challenges` challenges for all permutation arguments.

### DIFF
--- a/starky/src/prover.rs
+++ b/starky/src/prover.rs
@@ -1,5 +1,8 @@
+use std::iter::once;
+
 use anyhow::{ensure, Result};
 use itertools::Itertools;
+use plonky2::field::batch_util::batch_multiply_inplace;
 use plonky2::field::extension_field::Extendable;
 use plonky2::field::field_types::Field;
 use plonky2::field::polynomial::{PolynomialCoeffs, PolynomialValues};
@@ -16,8 +19,9 @@ use rayon::prelude::*;
 
 use crate::config::StarkConfig;
 use crate::constraint_consumer::ConstraintConsumer;
+use crate::get_challenges::get_n_permutation_challenge_sets;
 use crate::proof::{StarkOpeningSet, StarkProof, StarkProofWithPublicInputs};
-use crate::stark::Stark;
+use crate::stark::{PermutationChallenge, PermutationInstance, PermutationPair, Stark};
 use crate::vars::StarkEvaluationVars;
 
 pub fn prove<F, C, S, const D: usize>(
@@ -38,7 +42,7 @@ where
     let degree = trace.len();
     let degree_bits = log2_strict(degree);
 
-    let trace_vecs = trace.into_iter().map(|row| row.to_vec()).collect_vec();
+    let trace_vecs = trace.iter().map(|row| row.to_vec()).collect_vec();
     let trace_col_major: Vec<Vec<F>> = transpose(&trace_vecs);
 
     let trace_poly_values: Vec<PolynomialValues<F>> = timed!(
@@ -56,7 +60,9 @@ where
         timing,
         "compute trace commitment",
         PolynomialBatch::<F, C, D>::from_values(
-            trace_poly_values,
+            // TODO: Cloning this isn't great; consider having `from_values` accept a reference,
+            // or having `compute_z_poly` read trace values from the `PolynomialBatch`.
+            trace_poly_values.clone(),
             rate_bits,
             false,
             cap_height,
@@ -68,6 +74,32 @@ where
     let trace_cap = trace_commitment.merkle_tree.cap.clone();
     let mut challenger = Challenger::new();
     challenger.observe_cap(&trace_cap);
+
+    // Permutation arguments.
+    let z_commitment = if stark.uses_permutation_args() {
+        let z_polys =
+            compute_z_polys::<F, C, S, D>(&stark, config, &mut challenger, &trace_poly_values);
+        timed!(
+            timing,
+            "compute permutation Z commitments",
+            Some(PolynomialBatch::from_values(
+                z_polys,
+                rate_bits,
+                false,
+                config.fri_config.cap_height,
+                timing,
+                None,
+            ))
+        )
+    } else {
+        None
+    };
+    let permutation_zs_cap = z_commitment
+        .as_ref()
+        .map(|commit| commit.merkle_tree.cap.clone());
+    for cap in &permutation_zs_cap {
+        challenger.observe_cap(cap);
+    }
 
     let alphas = challenger.get_n_challenges(config.num_challenges);
     let quotient_polys = compute_quotient_polys::<F, C, S, D>(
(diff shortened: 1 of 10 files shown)


starky/src/stark.rs
```diff
@@ -96,4 +97,36 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+    /// Pairs of lists of columns that should be permutations of one another. A permutation argument
+    /// will be used for each such pair. Empty by default.
+    fn permutation_pairs(&self) -> Vec<PermutationPair> {
+        vec![]
+    }
+}
+
+/// A pair of lists of columns, `lhs` and `rhs`, that should be permutations of one another.
+/// In particular, there should exist some permutation `pi` such that for any `i`,
+/// `lhs[i] = rhs[pi(i)]`.
```
wborgeaud, line 110: Hmm I'm not sure I understand this. From this comment, it seems like `lhs` is a set of column indices, and `rhs` is a permutation of `lhs`. What information does that give us about permutations of columns?

dlubarov: Oh yeah I wrote this wrong. Does it make sense now?

wborgeaud: Yeah makes sense! So the permutation `pi` would be the same for all pairs in `PermutationPair`? Where would we use a `PermutationPair` with lists of length `>1`?

field/src/field_types.rs
```diff
@@ -91,6 +92,13 @@ pub trait Field:
         *self * (Self::ONE + Self::TWO)
     }
 
+    fn elementwise_product(lhs: &[Self], rhs: &[Self]) -> Vec<Self> {
```
nbgl, line 95: We already have `field::batch_util::batch_multiply_inplace`, which internally uses vectorization where possible.

starky/src/prover.rs
```diff
@@ -68,6 +71,10 @@ where
     let mut challenger = Challenger::new();
     challenger.observe_cap(&trace_cap);
 
+    // Permutation arguments.
+    let z_polys =
+        compute_z_polys::<F, C, S, D>(&stark, config, &mut challenger, &trace_poly_values);
```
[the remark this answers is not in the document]
dlubarov: Done. It expanded this PR a fair amount, since I needed to add the cap it the proof in order for the verifier to include it in their opening arg.

field/src/field_types.rs
dlubarov: Oh good point, done

starky/src/stark.rs
dlubarov: Yeah exactly. I think we'll need that property for the memory argument, where we'll have timestamp, address, and value columns, plus a sorted version of the same columns. I.e. we'd want the same permutation to map each unsorted col to each sorted col.

wborgeaud: Oh yeah that sounds good!

starky/src/prover.rs
```diff
@@ -142,6 +177,81 @@ where
...
+        .cartesian_product(0..config.num_challenges)
+        .chunks(stark.permutation_batch_size())
+        .into_iter()
+        .flat_map(|batch| {
+            batch.enumerate().map(|(i, (pair, chal))| {
+                let challenge = permutation_challenge_sets[i].challenges[chal];
+                PermutationInstance { pair, challenge }
+            })
+        })
+        .collect_vec();
```
wborgeaud, line 210: nit: It's not immediately clear what this does. Maybe add a comment explaining how `permutation_instances` is constructed.

starky/src/stark.rs
```diff
@@ -76,24 +79,99 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+}
+
+/// A pair of lists of columns, `lhs` and `rhs`, that should be permutations of one another.
+/// In particular, there should exist some permutation `pi` such that for any `i`,
+/// `trace[lhs[i]] = pi(trace[rhs[i]])`. Here `trace` denotes the trace in column-major form, so
+/// `trace[col]` is a column vector.
+pub struct PermutationPair {
+    pub lhs_columns: Vec<usize>,
+    pub rhs_columns: Vec<usize>,
+}
```
wborgeaud, line 157: nit: Maybe this could be of the form
```rust
pub struct PermutationPair {
    pub pairs: Vec<(usize, usize)>,
}
```
since `lhs_columns` and `rhs_columns` are only used in a `lhs_columns.zip_eq(rhs_columns)` anyways.

starky/src/prover.rs
```diff
@@ -142,6 +177,81 @@ where
...
+
+    // Compute Z, which contains partial products of the quotients.
+    let mut partial_products = Vec::with_capacity(degree);
+    let mut acc = F::ONE;
+    for q in quotients {
+        partial_products.push(acc);
+        acc *= q;
+    }
+    PolynomialValues::new(partial_products)
+}
```
wborgeaud, line 253: Maybe I'm missing something, but it seems like this could be done over batches of `quotient_degree_factor` instances.

If I understand correctly, this does (in the case of two lhs columns `l_0, l_1` and two rhs columns `r_0, r_1`):
```
Z(gx) = Z(x) * (gamma + l_0 + beta * l_1) / (gamma + r_0 + beta * r_1)
=>
Z(gx) * (gamma + r_0 + beta * r_1) - Z(x) * (gamma + l_0 + beta * l_1) = 0.
```
This last constraint is of degree 2, whereas the vanishing polynomial can be of degree up to 3 (with `rate_bits=1`). 

So it seems like having `Z` take batches of `quotient_degree_factor` instances would yield a constraint of degree `constraint_degree`, and would result in fewer `Z` polynomials.

wborgeaud (approved):
Looks good to me otherwise!

starky/src/stark.rs
```diff
@@ -76,24 +79,99 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+    }
+
+    fn uses_permutation_args(&self) -> bool {
+        !self.permutation_pairs().is_empty()
+    }
+
+    fn permutation_batch_size(&self) -> usize {
+        // TODO explain
+        self.constraint_degree() - 1
+    }
```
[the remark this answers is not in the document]
dlubarov: Ah yeah I forgot. I'll change it to call `quotient_degree_factor()` then, I just didn't want to use it without explanation as I feel like it's sort of coincidental that `permutation_batch_size() == constraint_degree()`. I'll add a note.

starky/src/prover.rs
dlubarov: Done

starky/src/stark.rs
dlubarov: Good idea

starky/src/prover.rs
dlubarov: Ah yes, that was my intention but I forgot to actually implement the batching logic here. I'll just merge the non-batch version for now, as I'm not sure when I'll have time to finish this and want to minimize conflicts, but I added a TODO and I'll come back to this when I have a chance.

wborgeaud: Sounds good 👍

PR #486, opened 2022-02-14
Recursive STARK verifier

wborgeaud:
Add gadgets to recursively verify a STARK proof using Plonky2.


starky/src/recursive_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+use crate::config::StarkConfig;
+use crate::constraint_consumer::RecursiveConstraintConsumer;
+use crate::proof::{
+    StarkOpeningSetTarget, StarkProof, StarkProofChallengesTarget, StarkProofTarget,
+    StarkProofWithPublicInputs, StarkProofWithPublicInputsTarget,
+};
+use crate::stark::Stark;
+use crate::vars::StarkEvaluationTargets;
+
+pub fn recursively_verify_stark_proof<
```
wborgeaud, line 21: Random thought: Maybe we should have a naming convention for gadgets not in an `impl CircuitBuilder` block. 
Here I've used `recursively_*` to distinguish from the native stark verifier. Other options would be `circuit_*` or `gadget_*`. Can also work as suffixes.

plonky2/src/fri/mod.rs
```diff
@@ -30,6 +30,20 @@ impl FriConfig {
     pub fn rate(&self) -> f64 {
         1.0 / ((1 << self.rate_bits) as f64)
     }
+
+    pub fn fri_params(&self, degree_bits: usize, hiding: bool) -> FriParams {
```
dlubarov, line 34: There's also a `StarkConfig::fri_params` that could call this

plonky2/src/iop/witness.rs
```diff
@@ -201,64 +202,28 @@ pub trait Witness<F: Field> {
         );
         self.set_cap_target(&proof_target.quotient_polys_cap, &proof.quotient_polys_cap);
 
-        for (&t, &x) in proof_target
```
dlubarov, line 204: Nice improvement 😎

starky/src/fibonacci_stark.rs
```diff
@@ -79,7 +79,25 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for FibonacciStar
...
+        let pis_constraints = [
+            builder.sub_extension(vars.local_values[0], vars.public_inputs[Self::PI_INDEX_X0]),
+            builder.sub_extension(vars.local_values[1], vars.public_inputs[Self::PI_INDEX_X1]),
+            builder.sub_extension(vars.local_values[1], vars.public_inputs[Self::PI_INDEX_RES]),
+        ];
+        yield_constr.constraint_first_row(builder, pis_constraints[0]);
+        yield_constr.constraint_first_row(builder, pis_constraints[1]);
+        yield_constr.constraint_last_row(builder, pis_constraints[2]);
+
+        // x0 <- x1
```
dlubarov, line 92: nit: maybe use `'` to distinguish local/next? `x0' <- x1`

starky/src/recursive_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+        local_values: builder.add_virtual_extension_targets(S::COLUMNS),
+        next_values: builder.add_virtual_extension_targets(S::COLUMNS),
+        permutation_zs: vec![/*TODO*/],
+        permutation_zs_right: vec![/*TODO*/],
+        quotient_polys: builder
+            .add_virtual_extension_targets(stark.quotient_degree_factor() * num_challenges),
+    }
+}
+
+pub fn set_startk_proof_with_pis_target<F, C: GenericConfig<D, F = F>, W, const D: usize>(
```
dlubarov, line 217: typo? `startk`

starky/src/get_challenges.rs
```diff
@@ -94,6 +101,89 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
dlubarov, line 104: nit: could move to lib.rs as we did for other crates

starky/src/proof.rs
```diff
@@ -1,12 +1,22 @@
...
+use plonky2::hash::hash_types::{MerkleCapTarget, RichField};
 use plonky2::hash::merkle_tree::MerkleCap;
+use plonky2::iop::ext_target::ExtensionTarget;
+use plonky2::iop::target::Target;
 use plonky2::plonk::config::GenericConfig;
 use rayon::prelude::*;
 
+use crate::config::StarkConfig;
+
+// TODO: Permutation polynomials.
```
dlubarov, line 18: nit: maybe not needed since it'll be added shortly in #485

starky/src/recursive_verifier.rs
dlubarov: Yeah agreed. I think suffixes can be nice when we have native and recursive variants of something, to emphasize their similarity, e.g. they'd be listed side by side on docs.rs.

Looks like we have some existing methods with `_recursive` and some with `_recursively`, so we should probably change one to match the other, or `_circuit` seems reasonable too if you prefer it.

Separately, maybe rename the native verification function to `verify_stark_proof` to match this one?

starky/src/recursive_verifier.rs
```diff
@@ -0,0 +1,261 @@
...
+    [(); S::PUBLIC_INPUTS]:,
+{
+    let one = builder.one_extension();
+
+    let StarkProofWithPublicInputsTarget {
+        proof,
+        public_inputs,
+    } = proof_with_pis;
+    let local_values = &proof.openings.local_values;
+    let next_values = &proof.openings.local_values;
```
dlubarov, line 75: `.local_values` -> `.next_values`

dlubarov: Oh seems these are overwritten by the line below?

starky/src/recursive_verifier.rs
wborgeaud: > Yeah agreed. I think suffixes can be nice when we have native and recursive variants of something, to emphasize their similarity, e.g. they'd be listed side by side on docs.rs.

Good idea!

> Looks like we have some existing methods with `_recursive` and some with `_recursively`, so we should probably change one to match the other, or `_circuit` seems fine as well if you prefer it.

I have a slight preference for `_circuit`, since some gadgets might not be directly used in recursion. But I'd be fine with the other suffixes too. 

> Separately, maybe rename the native verification function to `verify_stark_proof` to match this one?

Good idea, done.

I'll leave it like this for now and make another PR where I'll standardize the gadgets suffix.

plonky2/src/fri/mod.rs
wborgeaud: Oh I must have missed this one. I think it's kind of natural to have a way to generate the `FriParams` directly from the `FriConfig` though. It also makes the `CircuitBuilder::fri_params` function more succinct.
 I've similarly changed `StarkConfig::fri_params` to call `FriConfig::fri_params`.

starky/src/recursive_verifier.rs
wborgeaud: Oops yeah don't know what these 2 lines were for. They were also in the native verifier. I've removed them.

starky/src/recursive_verifier.rs
dlubarov: Yeah true, `_circuit` sounds good then.

dlubarov (approved):

PR #487, opened 2022-02-14
Derived more traits for ecdsa types

BGluth:
The client needed a few more derives on the ecdsa types. `Serialize` & `Deserialze` were the critical ones, but I also thought it was a good idea to also derive `PartialEq`, `Eq`, and `Hash` where I could.

Few notes:

- I didn't add any of the derives on the circuit versions since it was going to require adding a lot more derives on types that I wasn't sure should be implementing the traits.
- Not sure if it makes sense to derive `PartialEq` & `Eq` on `Secp256K1`, a unit struct. It seems wrong to declare that you can compare two unit structs against each other, but this seems to be required if we want to construct `ECDSASecretKey<Secp256K1>`.
- Right now since we're deriving these traits directly on the ecdsa types, we're enforcing that any curves used also implement these traits. If we don't want to require this, we can replace the derive impls with more verbose trait impls that only impl if the underlying trait impls the trait as well:
```rust
impl<C: Curve + PartialEq> PartialEq for ECDSASecretKey<C> {
    fn eq(&self, other: &Self) -> bool {
        self.0 == other.0
    }
}
```

### DIFF
--- a/plonky2/src/curve/curve_types.rs
+++ b/plonky2/src/curve/curve_types.rs
@@ -3,6 +3,7 @@ use std::ops::Neg;
 
 use plonky2_field::field_types::{Field, PrimeField};
 use plonky2_field::ops::Square;
+use serde::{Deserialize, Serialize};
 
 // To avoid implementation conflicts from associated types,
 // see https://github.com/rust-lang/rust/issues/20400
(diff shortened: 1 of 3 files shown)


npwardberkeley (approved):
Looks good to me! I don't see any reason we wouldn't want these

PR #489, opened 2022-02-15
Quintic extension fields

unzvfu:
A straight-forward adaptation of the existing extension field code to handle quintic extensions.


field/src/goldilocks_field.rs
```diff
@@ -317,6 +318,31 @@ impl Extendable<4> for GoldilocksField {
...
+    // DTH_ROOT = W^((ORDER - 1)/5)
+    const DTH_ROOT: Self = Self(1041288259238279555);
+
+    const EXT_MULTIPLICATIVE_GROUP_GENERATOR: [Self; 5] = [
+        Self(1931274660132142120),
+        Self(1092386509894096633),
+        Self(1605533804202809407),
+        Self(14704369562396645516),
+        Self(1750907431983753016),
+    ];
```
wborgeaud, line 335: nit: Was this generator chosen for performance reasons? In other extension fields the generator `g` satisfies
```
g^((p^D - 1) >> F::Extension::TWO_ADICITY) = F::EXT_POWER_OF_TWO_GENERATOR
```
I can't remember now if we use this property somewhere in the code, but maybe it would be good to have it for the quintic extension too. 
Here's an example of generator that satisfies the relation:
```rust
[2899034827742553394, 13012057356839176729, 14593811582388663055, 7722900811313895436, 4557222484695340057]
```

wborgeaud (approved):
Looks good to me!

field/src/extension_field/quintic.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        // Writing 'a' for self:
+        let d = self.frobenius(); // d = a^p
+        let e = d * d.frobenius(); // e = a^(p + p^2)
+        let f = e * e.repeated_frobenius(2); // f = a^(p + p^2 + p^3 + p^4)
+
+        // f contains a^(r-1) and g = a^r is in the base field.
+        let g = *self * f;
+        debug_assert!(FieldExtension::<5>::is_in_basefield(&g));
+        let g = g.0[0];
```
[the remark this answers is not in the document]
unzvfu: Oh, good catch! I doubt the compiler would be that clever, so I'll make the change. This observation is also true for the quadratic and quartic implementations of `try_inverse()`. Fixing them will fit neatly in my next PR, which includes the multiplication speedups mentioned over chat.

field/src/goldilocks_field.rs
unzvfu: No, that generator was selected at random, I'm happy to use the one you suggest.

I can't find that requirement documented anywhere, so I'll write up the requirement in the `Extendable` trait.

PR #490, opened 2022-02-15
Simplify batched gate code

wborgeaud:
Use generators to count the number of operations performed in a gate and add state to the circuit builder to keep track of the available slots for each gate. As a result, all the `BatchedGate` and gate-specific `find_*_gate` and `fill_*_gate` functions can be removed.


npwardberkeley (approved):
Looks great apart from that one minor comment! What an improvement :D

plonky2/src/gates/gmimc.rs
```diff
@@ -0,0 +1 @@
+
```
dlubarov, line 1: outdated?

plonky2/src/hash/gmimc.rs
```diff
@@ -0,0 +1 @@
+
```
dlubarov, line 1: outdated?

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -211,7 +206,12 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
     }
 
     /// Adds a gate to the circuit, and returns its index.
-    pub fn add_gate<G: Gate<F, D>>(&mut self, gate_type: G, constants: Vec<F>) -> usize {
+    pub fn add_gate<G: Gate<F, D>>(
+        &mut self,
+        gate_type: G,
+        constants: Vec<F>,
+        params: Vec<F>,
```
dlubarov, line 213: Could you explain the use case for this?

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -392,6 +397,48 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         })
     }
 
+    /// Find an available slot, of the form `(gate_index, op)` for gate `G`.
+    pub fn find_slot<G: Gate<F, D> + Clone>(
+        &mut self,
+        gate: G,
+        params: &[F],
```
npwardberkeley, line 404: Oh, a comment to explain what `params` is would be helpful here too. Also, very much a nit, but maybe keep `params` and `constants` in the same order here as elsewhere?

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -392,6 +397,48 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+        let slot = gate_slot.current_slot.get(params);
+        let res = if let Some(&s) = slot {
+            s
+        } else {
+            self.add_gate(gate, constants.to_vec(), params.to_vec());
+            (num_gates, 0)
+        };
+        if res.1 == num_ops - 1 {
+            // We've filled up the slots at this index.
+            self.current_slots
```
dlubarov, line 425: nit: to minimize redundancy with the else case we could do

```rust
let current_slot = &mut self.current_slots.get_mut(&gate_ref).unwrap().current_slot;
```

plonky2/src/gates/gate.rs
```diff
@@ -171,10 +178,18 @@ impl<F: RichField + Extendable<D>, const D: usize> Debug for GateRef<F, D> {
...
+pub struct CurrentSlot<F: RichField + Extendable<D>, const D: usize> {
+    pub current_slot: HashMap<Vec<F>, (usize, usize)>,
+}
+
 /// A gate along with any constants used to configure it.
+#[derive(Clone)]
 pub struct GateInstance<F: RichField + Extendable<D>, const D: usize> {
     pub gate_ref: GateRef<F, D>,
     pub constants: Vec<F>,
+    pub params: Vec<F>,
```
dlubarov, line 192: I know params are used in `find_slot`, but it seems like this field and the argument to `add_gate` aren't used, unless I missed something.

plonky2/src/plonk/circuit_builder.rs
dlubarov: I think I get it now after reading more of the code. Params are like the key used to decide whether two operations can share a gate, correct?

As Nick said elsewhere, I think it could use a comment explaining that, though not sure if it should be here. (As per other comment, maybe we can delete params here and only take it in `find_slot`?)

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -695,12 +740,26 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+
         // Add gate generators.
         self.add_generators(
             self.gate_instances
                 .iter()
                 .enumerate()
-                .flat_map(|(index, gate)| gate.gate_ref.0.generators(index, &gate.constants))
+                .flat_map(|(index, gate)| {
+                    let mut gens = gate.gate_ref.0.generators(index, &gate.constants);
+                    // Remove unused generators, if any.
```
dlubarov, line 757: This is a nice solution, my only concern is it seems like "batch" gates now must have one generator per slot. I think it's okay, but to avoid any surprises it might be good to document the requirement (in `generators()`?), and/or maybe assert `gens.len() == num_ops`.

Or to make it even safer, maybe we could change `generators()` to return a single generator for a particular slot index. That would be a major change though so probably not now.

plonky2/src/gates/gate.rs
```diff
@@ -171,10 +178,18 @@ impl<F: RichField + Extendable<D>, const D: usize> Debug for GateRef<F, D> {
     }
 }
 
+/// Map between gate parameters and available slots.
+#[derive(Clone, Debug)]
+pub struct CurrentSlot<F: RichField + Extendable<D>, const D: usize> {
+    pub current_slot: HashMap<Vec<F>, (usize, usize)>,
+}
+
 /// A gate along with any constants used to configure it.
```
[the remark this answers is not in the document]
wborgeaud: Oh yeah good idea, I'll add comments.

plonky2/src/plonk/circuit_builder.rs
wborgeaud: > I think I get it now after reading more of the code. Params are like the key used to decide whether two operations can share a gate, correct?

Yeah exactly. At first I just wanted to use constants to differentiate gates. That works for gates like `ArithmeticGate`, but not for gates like `SwitchGate` which have differentiating parameters that are not in constants. So I added this `params` field which can sometimes be redundant with constants.

> As Nick said elsewhere, I think it could use a comment explaining that, though not sure if it should be here. (As per other comment, maybe we can delete params here and only take it in `find_slot`?)

Sounds good, I'll add comments.

plonky2/src/plonk/circuit_builder.rs
wborgeaud: So the default `num_ops()` just returns the number of generators:
```rust
fn num_ops(&self) -> usize {
    self.generators(0, &vec![F::ZERO; self.num_constants()])
        .len()
}
```
I agree it would be nice to disallow overwriting this function. Maybe just adding a comment: "Don't overwrite!" would be ok?

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -392,6 +397,48 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+        params: &[F],
+        constants: &[F],
+    ) -> (usize, usize) {
+        let num_gates = self.num_gates();
+        let num_ops = gate.num_ops();
+        let gate_ref = GateRef::new(gate.clone());
+        let gate_slot = self
+            .current_slots
+            .entry(gate_ref.clone())
+            .or_insert(CurrentSlot {
```
dlubarov, line 413: nit: I think it would be natural to have `CurrentSlot` `#[derive(Default)]`, and then we could replace `or_insert(...)` with `or_default()`

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -392,6 +397,48 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+                current_slot: HashMap::new(),
+            });
+        let slot = gate_slot.current_slot.get(params);
+        let res = if let Some(&s) = slot {
```
dlubarov, line 417: nit: it's a bit more verbose but I think doing `let (gate_idx, slot_idx)` would make it easier to read

plonky2/src/plonk/circuit_builder.rs
wborgeaud: > Or to make it even safer, maybe we could change `generators()` to return a single generator for a particular slot index. That would be a major change though so probably not now.

Yeah I've thought about doing this, I think that would also work but I kind of like the current solution where `generators()` is the only source of truth and dictates the number of operations. I think that makes it really easy to write non-batchable gate without having to think about any batch-related logic.

plonky2/src/gates/gate.rs
wborgeaud: Oh good catch! I used it in a previous version, but you're right that it's unnecessary now. I fixed it 👍

plonky2/src/plonk/circuit_builder.rs
dlubarov: Oh I see. I wouldn't really worry about `num_ops` being overridden, but I think the other risk is still there, e.g. someone could write a gate that supported `n` operations but used `2n` generators.

I feel like the API change would be safest, though a comment that gates should return one generator per slot that they actually support could also work.

plonky2/src/gates/gate.rs
```diff
@@ -171,7 +178,14 @@ impl<F: RichField + Extendable<D>, const D: usize> Debug for GateRef<F, D> {
     }
 }
 
+/// Map between gate parameters and available slots.
+#[derive(Clone, Debug)]
+pub struct CurrentSlot<F: RichField + Extendable<D>, const D: usize> {
+    pub current_slot: HashMap<Vec<F>, (usize, usize)>,
```
dlubarov, line 184: nit: please comment what the tuple is

plonky2/src/plonk/circuit_builder.rs
wborgeaud: Hmm is there a use case for a gate with more than one generator per slot?

dlubarov (approved):
LGTM otherwise

plonky2/src/plonk/circuit_builder.rs
dlubarov: I don't think it's something important we need to support, but I think there's a decent chance someone could attempt to write it that way, e.g. if they were trying to reuse some existing generators and compose them together.

PR #491, opened 2022-02-16
Fix reduction strategy

wborgeaud:
Fix bug where the reduction strategy could make the final polynomial degree be less than the cap height, which results in undefined behavior when constructing the FRI Merkle trees.

### DIFF
--- a/plonky2/src/fri/reduction_strategies.rs
+++ b/plonky2/src/fri/reduction_strategies.rs
@@ -8,11 +8,12 @@ pub enum FriReductionStrategy {
     /// Specifies the exact sequence of arities (expressed in bits) to use.
     Fixed(Vec<usize>),
 
-    /// `ConstantArityBits(arity_bits, final_poly_bits)` applies reductions of arity `2^arity_bits`
-    /// until the polynomial degree is `2^final_poly_bits` or less. This tends to work well in the
-    /// recursive setting, as it avoids needing multiple configurations of gates used in FRI
-    /// verification, such as `InterpolationGate`.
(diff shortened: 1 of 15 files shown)


plonky2/src/fri/reduction_strategies.rs
```diff
@@ -8,11 +8,12 @@ pub enum FriReductionStrategy {
...
-    /// until the polynomial degree is `2^final_poly_bits` or less. This tends to work well in the
-    /// recursive setting, as it avoids needing multiple configurations of gates used in FRI
-    /// verification, such as `InterpolationGate`.
-    ConstantArityBits(usize, usize),
+    /// `ConstantArityBits(arity_bits, final_poly_bits, cap_height)` applies reductions of arity `2^arity_bits`
+    /// until the polynomial degree is less than or equal to `2^final_poly_bits` or until any further
+    /// `arity_bits`-reduction makes the last FRI tree have height less than `cap_height`.
+    /// This tends to work well in the recursive setting, as it avoids needing multiple configurations
+    /// of gates used in FRI verification, such as `InterpolationGate`.
+    ConstantArityBits(usize, usize, usize),
```
dlubarov, line 16: It seems like there would be a bit of redundancy in the config if we add `cap_height` here; could we make it a param to `reduction_arity_bits` that comes from the existing config value?

wborgeaud: Oh yeah good idea, done.

dlubarov (approved):
👍

PR #493, opened 2022-02-17
Update MDS matrix and round consts in Poseidon; disable vectorization

nbgl:
Changes the MDS matrix used by Poseidon to one that has a nice form under complex FFT.

Also changes round constants. This has been on the backlog for a while. The round constant generation code looked a bit sus because it was actually generating constants in the Crandall field for historical reasons. We now sample directly from Goldilocks so our method shouldn't raise any eyebrows.

Lastly, vectorized versions of Poseidon are temporarily disabled. They need updating with the new MDS matrix and round constants, which is better in separate PRs. This will temporarily make proofs much slower, but once the vector changes are merged, they'll be even faster than before.

### DIFF
--- a/plonky2/src/bin/generate_constants.rs
+++ b/plonky2/src/bin/generate_constants.rs
@@ -7,22 +7,15 @@ use plonky2_field::goldilocks_field::GoldilocksField;
 use rand::{Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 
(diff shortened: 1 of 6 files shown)


nbgl:
I had to use @unzvfu's script to generate new fast partial round constants. Kudos to him for making it easy to run and simple to adapt to a new matrix!

PR #497, opened 2022-02-20
Rename constraint methods

dlubarov:
Most of our constraints apply to all rows, and it seems safest to make that the "default".

cc @nbgl fyi


wborgeaud (approved):
👍

PR #499, opened 2022-02-22
Stark verifier permutation checks

wborgeaud:
Add the verifier checks for the permutation argument.


starky/src/verifier.rs
```diff
@@ -88,8 +90,18 @@ where
         l_1,
         l_last,
     );
-    stark.eval_ext(vars, &mut consumer);
-    // TODO: Add in constraints for permutation arguments.
+    let permutation_data = stark.uses_permutation_args().then(|| PermutationCheckData {
+        local_zs: permutation_zs.as_ref().unwrap().clone(),
```
dlubarov, line 94: I think we should return an error if the STARK uses permutation args but those openings are missing (or the challenges are missing because the cap was missing), vs panicking.

starky/src/vanishing_poly.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::config::StarkConfig;
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::permutation::{
+    eval_permutation_checks, eval_permutation_checks_recursively, PermutationCheckData,
+    PermutationCheckDataTarget,
+};
+use crate::stark::Stark;
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+pub(crate) fn eval_vanishing_poly<F, FE, C, S, const D: usize, const D2: usize>(
```
dlubarov, line 15: Let's have this supported packed fields - it's a small change, cb749ed7554e9f1cd249b1f8ce207e4e49a2709c

starky/src/permutation.rs
```diff
@@ -178,3 +166,222 @@ pub(crate) fn get_n_permutation_challenge_sets<F: RichField, H: Hasher<F>>(
...
+                    let challenge = permutation_challenge_sets[i].challenges[chal];
+                    PermutationInstance { pair, challenge }
+                })
+                .collect_vec()
+        })
+        .collect()
+}
+
+// TODO: Use slices.
+pub struct PermutationCheckData<F: Field, FE: FieldExtension<D2, BaseField = F>, const D2: usize> {
```
dlubarov, line 242: nit: maybe "Vars" would be fitting to emphasize the similarity with `StarkEvaluationVars`

starky/src/permutation.rs
```diff
@@ -178,3 +166,222 @@ pub(crate) fn get_n_permutation_challenge_sets<F: RichField, H: Hasher<F>>(
...
+    pub(crate) local_zs: Vec<FE>,
+    pub(crate) next_zs: Vec<FE>,
+    pub(crate) permutation_challenge_sets: Vec<PermutationChallengeSet<F>>,
+}
+
+pub(crate) fn eval_permutation_checks<F, FE, C, S, const D: usize, const D2: usize>(
```
dlubarov, line 248: Likewise here, I think we should support packed fields (though if you run into any complications we can add it later)

starky/src/permutation.rs
```diff
@@ -178,3 +166,222 @@ pub(crate) fn get_n_permutation_challenge_sets<F: RichField, H: Hasher<F>>(
...
+        &permutation_pairs,
+        &permutation_challenge_sets,
+        config.num_challenges,
+        stark.permutation_batch_size(),
+    );
+
+    // Each zs value corresponds to a permutation batch.
+    for (i, instances) in permutation_batches.iter().enumerate() {
+        // Z(gx) * down = Z x  * up
+        let (reduced_lhs, reduced_rhs): (Vec<FE>, Vec<FE>) = instances
```
dlubarov, line 285: It looks like there are quite a few allocations here due to unzips. I think it's fine for now, but just something to keep an eye on as we gather performance data.

starky/src/permutation.rs
```diff
@@ -178,3 +166,222 @@ pub(crate) fn get_n_permutation_challenge_sets<F: RichField, H: Hasher<F>>(
...
+        // Z(gx) * down = Z x  * up
+        let (reduced_lhs, reduced_rhs): (Vec<ExtensionTarget<D>>, Vec<ExtensionTarget<D>>) =
+            instances
+                .iter()
+                .map(|instance| {
+                    let PermutationInstance {
+                        pair: PermutationPair { column_pairs },
+                        challenge: PermutationChallenge { beta, gamma },
+                    } = instance;
+                    let zero = builder.zero_extension();
```
dlubarov, line 362: seems unused

starky/src/permutation.rs
```diff
@@ -178,3 +166,222 @@ pub(crate) fn get_n_permutation_challenge_sets<F: RichField, H: Hasher<F>>(
...
+    let permutation_batches = get_permutation_batches(
+        &permutation_pairs,
+        &permutation_challenge_sets,
+        config.num_challenges,
+        stark.permutation_batch_size(),
+    );
+
+    // Each zs value corresponds to a permutation batch.
+    for (i, instances) in permutation_batches.iter().enumerate() {
+        // Z(gx) * down = Z x  * up
```
dlubarov, line 284: nit: I don't quite get what down/up signify and notation seems unusual

starky/src/permutation.rs
```diff
@@ -49,40 +56,21 @@ pub(crate) fn compute_permutation_z_polys<F, C, S, const D: usize>(
     config: &StarkConfig,
     challenger: &mut Challenger<F, C::Hasher>,
```
dlubarov, line 57: It seems `challenger` is no longer needed since `get_n_permutation_challenge_sets` was moved outside of this method

starky/src/verifier.rs
wborgeaud: What do you think about adding a function
```rust
fn check_permutation_options<
    F: RichField + Extendable<D>,
    C: GenericConfig<D, F = F>,
    S: Stark<F, D>,
    const D: usize,
>(
    stark: &S,
    proof_with_pis: &StarkProofWithPublicInputs<F, C, D>,
    challenges: &StarkProofChallenges<F, D>,
) -> Result<()> {
    let options = [
        proof_with_pis.proof.permutation_zs_cap.is_some(),
        proof_with_pis.proof.openings.permutation_zs.is_some(),
        proof_with_pis.proof.openings.permutation_zs_right.is_some(),
        challenges.permutation_challenge_sets.is_some(),
    ];
    ensure!(
        options
            .into_iter()
            .all(|b| b == stark.uses_permutation_args()),
        "Permutation data doesn't match with Stark configuration."
    );
    Ok(())
}
```
at the top of the verifier?

starky/src/prover.rs
```diff
@@ -80,28 +84,39 @@ where
...
+            (
+                PolynomialBatch::from_values(
+                    permutation_z_polys,
+                    rate_bits,
+                    false,
+                    config.fri_config.cap_height,
+                    timing,
+                    None,
+                ),
+                permutation_challenge_sets
```
dlubarov, line 113: nit: feels a little unusual to have this inside `timed!`, I would suggest rearranging to

```rust
        let permutation_commit = timed!(
            timing,
            "compute permutation Z commitments",
            PolynomialBatch::from_values(
                permutation_z_polys,
                rate_bits,
                false,
                config.fri_config.cap_height,
                timing,
                None,
            )
        );
        (permutation_commit, permutation_challenge_sets)
```

starky/src/prover.rs
```diff
@@ -228,9 +249,8 @@ where
     let z_h_on_coset = ZeroPolyOnCoset::<F>::new(degree_bits, quotient_degree_bits);
 
     // Retrieve the LDE values at index `i`.
-    let get_at_index = |comm: &PolynomialBatch<F, C, D>, i: usize| -> [F; S::COLUMNS] {
-        comm.get_lde_values(i * step).try_into().unwrap()
-    };
+    let get_at_index =
+        |comm: &'a PolynomialBatch<F, C, D>, i: usize| -> &'a [F] { comm.get_lde_values(i * step) };
```
dlubarov, line 253: nit: might be nice to also add

```rust
let get_trace_at_index = |i| get_at_index(trace_commitment, i).try_into().unwrap();
```

to tidy up the calls below a little

starky/src/recursive_verifier.rs
```diff
@@ -187,24 +202,29 @@ pub fn add_virtual_stark_proof<F: RichField + Extendable<D>, S: Stark<F, D>, con
...
 
-    let num_leaves_per_oracle = &[
-        S::COLUMNS,
-        // TODO: permutation polys
-        stark.quotient_degree_factor() * config.num_challenges,
-    ];
-
-    let permutation_zs_cap = if stark.uses_permutation_args() {
-        Some(builder.add_virtual_cap(cap_height))
+    let num_leaves_per_oracle = if stark.uses_permutation_args() {
```
dlubarov, line 205: What do you think of

```rust
    let num_leaves_per_oracle = once(S::COLUMNS)
        .chain(
            stark
                .uses_permutation_args()
                .then(|| stark.num_permutation_batches(config)),
        )
        .chain(once(stark.quotient_degree_factor() * config.num_challenges))
        .collect_vec();
```

A little shorter and it would match the `merkle_caps` code, though it might be a bit weird to read, so either way.

starky/src/verifier.rs
dlubarov: Sounds good, good idea 👍

dlubarov (approved):
LGTM otherwise

starky/src/recursive_verifier.rs
wborgeaud: Looks good 👍

PR #505, opened 2022-02-28
Recover public key

wborgeaud:
Add function `recover_public_key` to recover an ECDSA public key from a message, signature and recovery id.
I've made the recovery id a separate type to keep `ECDSASignature` similar to `ECDSASignatureTarget`, but it could also be a field of the signature if it's more convenient.

cc @dlubarov @npwardberkeley

### DIFF
--- a/field/src/secp256k1_base.rs
+++ b/field/src/secp256k1_base.rs
@@ -244,6 +244,12 @@ impl DivAssign for Secp256K1Base {
     }
 }
 
+impl Secp256K1Base {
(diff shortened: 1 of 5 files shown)


wborgeaud:
We're moving to an external ECDSA crate, so this won't be used.

BGluth:
Also decided to keep the branch around and later I'll create some benches to compare it against the major [ECDSA](https://docs.rs/ecdsa/latest/ecdsa/) crate.

PR #508, opened 2022-03-02
System Zero subtraction operation

unzvfu:
Implementation of subtraction for the System Zero VM.

### DIFF
--- a/system_zero/src/alu/subtraction.rs
+++ b/system_zero/src/alu/subtraction.rs
@@ -38,13 +38,10 @@ pub(crate) fn eval_subtraction<F: Field, P: PackedField<Scalar = F>>(
     let base = F::from_canonical_u64(1 << 16);
     let base_sqr = F::from_canonical_u64(1 << 32);
 
-    // Note that this can't overflow. Since each output limb has been
-    // range checked as 16-bits
-    let out_diff = out_1 + out_2 * base;
     let out_br = out_br * base_sqr;
-
     let lhs = (out_br + in_1) - in_2;
-    let rhs = out_diff;
+    let rhs = out_1 + out_2 * base;
+
     yield_constr.constraint(is_sub * (lhs - rhs));
 
     // We don't need to check that out_br is in {0, 1} because it's
@@ -66,12 +63,12 @@ pub(crate) fn eval_subtraction_recursively<F: RichField + Extendable<D>, const D
     let base = builder.constant_extension(F::Extension::from_canonical_u64(1 << 16));
     let base_sqr = builder.constant_extension(F::Extension::from_canonical_u64(1 << 32));
 
+    // lhs = (out_br + in_1) - in_2
+    let lhs = builder.add_extension(out_br, in_1);
+    let lhs = builder.sub_extension(lhs, in_2);
+
     // rhs = out_1 + base * out_2
     let rhs = builder.mul_add_extension(out_2, base, out_1);
-
-    // lhs = (out_br + in_2) - in_1
-    let lhs = builder.add_extension(out_br, in_2);
-    let lhs = builder.sub_extension(lhs, in_1);
 
     // filtered_diff = is_sub * (lhs - rhs)
     let diff = builder.sub_extension(lhs, rhs);
--- a/system_zero/src/alu/subtraction.rs
+++ b/system_zero/src/alu/subtraction.rs
@@ -21,7 +21,7 @@ pub(crate) fn generate_subtraction<F: PrimeField64>(values: &mut [F; NUM_COLUMNS
 
     values[COL_SUB_OUTPUT_0] = F::from_canonical_u16(diff as u16);
     values[COL_SUB_OUTPUT_1] = F::from_canonical_u16((diff >> 16) as u16);
-    values[COL_SUB_OUTPUT_2] = F::from_canonical_u16(br as u16);
+    values[COL_SUB_OUTPUT_BORROW] = F::from_canonical_u16(br as u16);
 }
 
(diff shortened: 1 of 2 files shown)


system_zero/src/registers/alu.rs
```diff
@@ -34,6 +34,19 @@ pub(crate) const COL_ADD_OUTPUT_1: usize = super::range_check_16::col_rc_16_inpu
...
+pub(crate) const COL_SUB_INPUT_0: usize = shared_col(0);
+pub(crate) const COL_SUB_INPUT_1: usize = shared_col(1);
+
+// FIXME: Check whether comment above for addition outputs applies here too.
+/// The first 16-bit chunk of the output, based on little-endian ordering.
+pub(crate) const COL_SUB_OUTPUT_0: usize = super::range_check_16::col_rc_16_input(0);
+/// The second 16-bit chunk of the output, based on little-endian ordering.
+pub(crate) const COL_SUB_OUTPUT_1: usize = super::range_check_16::col_rc_16_input(1);
+/// The borrow output
+pub(crate) const COL_SUB_OUTPUT_2: usize = super::boolean::col_bit(2);
```
dlubarov, line 48: nit: would prefer if we call it `COL_SUB_OUTPUT_BORROW` (or `_BR` etc)

system_zero/src/alu/subtraction.rs
```diff
@@ -10,15 +10,45 @@ use crate::registers::alu::*;
...
+    let out = (out_br * base_sqr + out_1) - out_2 * base;
+
+    // NB: Not clear how to compute in_1 - in_2 in one expression for
+    // PackedFields: If in_1 < in_2 then the sign extension will
+    // happen inside the big field, which is probably wrong. Instead,
+    // we *first* subtract in_2 from the expected output, and then
+    // subtract in_1; the result then should be zero.
+    yield_constr.constraint(is_sub * ((out + in_2) - in_1));
+
+    yield_constr.constraint(is_sub * out_br * (P::ONES - out_br));
```
dlubarov, line 51: I think we can omit this, since the borrow output register is an alias for a `col_bit` register, and those will be constrained to be 0 or 1 by the "boolean unit".

(We haven't actually added those constraints yet, but that's the plan at least.)

dlubarov: Yeah I think so 👍

system_zero/src/registers/alu.rs
```diff
@@ -34,6 +34,19 @@ pub(crate) const COL_ADD_OUTPUT_1: usize = super::range_check_16::col_rc_16_inpu
...
+pub(crate) const COL_SUB_INPUT_0: usize = shared_col(0);
+pub(crate) const COL_SUB_INPUT_1: usize = shared_col(1);
+
+// FIXME: Check whether comment above for addition outputs applies here too.
+/// The first 16-bit chunk of the output, based on little-endian ordering.
+pub(crate) const COL_SUB_OUTPUT_0: usize = super::range_check_16::col_rc_16_input(0);
+/// The second 16-bit chunk of the output, based on little-endian ordering.
+pub(crate) const COL_SUB_OUTPUT_1: usize = super::range_check_16::col_rc_16_input(1);
+/// The borrow output
+pub(crate) const COL_SUB_OUTPUT_BORROW: usize = super::boolean::col_bit(2);
```
dlubarov, line 48: I think we can start at `col_bit(0)`, since these bool registers are separate from the ALU shared registers and the range check registers. (Though `col_bit(2)` would also work - it won't be used for anything else during subtraction - just feels less natural to me.)

system_zero/src/alu/subtraction.rs
```diff
@@ -27,5 +57,25 @@ pub(crate) fn eval_subtraction_recursively<F: RichField + Extendable<D>, const D
...
+
+    let base = builder.constant_extension(F::Extension::from_canonical_u64(1 << 16));
+    let base_sqr = builder.constant_extension(F::Extension::from_canonical_u64(1 << 32));
+
+    // rhs = out_1 + base * out_2
+    let rhs = builder.mul_add_extension(out_2, base, out_1);
+
+    // lhs = (out_br + in_2) - in_1
+    let lhs = builder.add_extension(out_br, in_2);
+    let lhs = builder.sub_extension(lhs, in_1);
```
dlubarov, line 74: I think `in_1` and `in_2` may have gotten swapped?

system_zero/src/alu/subtraction.rs
```diff
@@ -27,5 +57,25 @@ pub(crate) fn eval_subtraction_recursively<F: RichField + Extendable<D>, const D
...
+    let in_2 = local_values[COL_SUB_INPUT_1];
+    let out_1 = local_values[COL_SUB_OUTPUT_0];
+    let out_2 = local_values[COL_SUB_OUTPUT_1];
+    let out_br = local_values[COL_SUB_OUTPUT_BORROW];
+
+    let base = builder.constant_extension(F::Extension::from_canonical_u64(1 << 16));
+    let base_sqr = builder.constant_extension(F::Extension::from_canonical_u64(1 << 32));
+
+    // rhs = out_1 + base * out_2
+    let rhs = builder.mul_add_extension(out_2, base, out_1);
```
dlubarov, line 70: We generally try to make the two eval functions (native & recursive) very similar, so it's easy to audit one and then check that the other matches. A couple nits to increase similarity:
- Let's move `rhs` below `lhs`
- Maybe inline `out_diff` in the native version to match this version

dlubarov (approved):
LGTM otherwise!

PR #511, opened 2022-03-03
Gadget for curve MSM, fixed-base scalar multiplication and ECDSA verification in under `2^17` or `2^16` gates

wborgeaud:
Add a gadget for curve MSM using a windowed MSM technique and a gadget for fixed-base scalar multiplication with a public base.

Combined, the two gadgets can be used to bring the cost of ECDSA verification down to `101202 < 2^17` gates. The prover running time is around 15s.
By simply increasing the number of wires to 234 (instead of 136), the number of gates goes to `65531 < 2^16`. With this, the prover running time is around 9s.
The recursion cost of increasing the number of wires is quite small, with 5023 vs 5359 gates for a recursive proof of an ECDSA verification, using 136 vs 234 wires respectively.


plonky2/src/gadgets/curve_msm.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
+    /// Computes `n*p + m*q` using windowed MSM, with a 2-bit window.
+    /// See Algorithm 9.23 in Handbook of Elliptic and Hyperelliptic Curve Cryptography for a
+    /// description.
+    /// Note: Doesn't work if `p == q`.
+    pub fn curve_msm<C: Curve>(
+        &mut self,
+        p: &AffinePointTarget<C>,
+        q: &AffinePointTarget<C>,
+        n: &BigUintTarget,
```
npwardberkeley, line 22: Should we make it consistent with `curve_scalar_mul`? I feel like both should take a biguint target, or both should take a scalar field nonnative target?

plonky2/src/gadgets/ecdsa.rs
```diff
@@ -72,7 +74,15 @@ mod tests {
 
         type Curve = Secp256K1;
 
-        let config = CircuitConfig::standard_ecc_config();
+        const WIDE: bool = true;
```
npwardberkeley, line 77: Should we have two different tests maybe?

npwardberkeley (approved):
Looks great!!

plonky2/src/gadgets/curve_msm.rs
wborgeaud: Good idea, done.

plonky2/src/gadgets/ecdsa.rs
wborgeaud: Good idea, done

PR #513, opened 2022-03-06
Halo2 style lookup arguments in System Zero

dlubarov:
It's a really nice and simple protocol, particularly for the verifier since the constraints are trivial (aside from the underlying batched permutation checks, which we already support). See the [Halo2 book](https://zcash.github.io/halo2/design/proving-system/lookup.html) and this [talk](https://www.youtube.com/watch?v=YlTt12s7vGE&t=5237s) by @daira.

Previously we generated the whole trace in row-wise form, but it's much more efficient to generate these "permuted" columns column-wise. So I changed our STARK framework to accept the trace in column-wise form. STARK impls now have the flexibility to do some generation row-wise and some column-wise (without extra costs; there's a single transpose as before).


system_zero/src/lookup.rs
```diff
@@ -0,0 +1,114 @@
...
+    for i in 1..n {
+        let is_repeat = permuted_inputs[i] == permuted_inputs[i - 1];
+        permuted_table.push(if is_repeat {
+            unused_table_vals
+                .next()
+                .expect("No more unused table values; this should never happen")
+        } else {
+            permuted_inputs[i]
+        });
+    }
```
wborgeaud, line 57: So I was wondering if there's a way to make this more efficient using sorting instead of hash maps, and I came up with this (works on `usize` for now, and assumes `table` is sorted):
```rust
let mut permuted_inputs = inputs.clone();
permuted_inputs.sort_unstable();
let mut unused_table_inds = Vec::with_capacity(n);
let mut unused_table_vals = Vec::with_capacity(n);
let mut permuted_table = vec![0; n];
let mut i = 0;
let mut j = 0;
while (j < n) && (i < n) {
    match permuted_inputs[i].cmp(&table[j]) {
        Ordering::Greater => {
            unused_table_vals.push(table[j]);
            j += 1;
        }
        Ordering::Less => {
            if let Some(x) = unused_table_vals.pop() {
                permuted_table[i] = x;
            } else {
                unused_table_inds.push(i);
            }
            i += 1;
        }
        Ordering::Equal => {
            permuted_table[i] = table[j];
            i += 1;
            j += 1;
        }
    }
}
for jj in j..n {
    unused_table_vals.push(table[jj]);
}
for ii in i..n {
    unused_table_inds.push(ii);
}
for (ind, val) in unused_table_inds.into_iter().zip(unused_table_vals) {
    permuted_table[ind] = val;
}
```
In my benchmarks, it's 3x faster than using hash maps with `n = 2^20` (30ms vs 100ms). I don't know if this will be a bottleneck, so not sure if it's worth the added complexity.

starky/src/util.rs
```diff
@@ -0,0 +1,16 @@
...
+/// A helper function to transpose a row-wise trace and put it in the format that `prove` expects.
+pub fn trace_rows_to_poly_values<F: Field, const COLUMNS: usize>(
+    trace_rows: Vec<[F; COLUMNS]>,
+) -> Vec<PolynomialValues<F>> {
+    let trace_row_vecs = trace_rows.iter().map(|row| row.to_vec()).collect_vec();
+    let trace_col_vecs: Vec<Vec<F>> = transpose(&trace_row_vecs);
+    trace_col_vecs
+        .into_iter()
+        .map(|column| PolynomialValues::new(column))
+        .collect()
```
wborgeaud, line 15: nit: I think the corresponding method in the Plonky2 prover uses `par_iter`. It probably doesn't change much in terms of performance but maybe it's best if they match.

system_zero/src/lookup.rs
```diff
@@ -0,0 +1,114 @@
...
+pub(crate) fn eval_lookups<F: Field, P: PackedField<Scalar = F>>(
+    vars: StarkEvaluationVars<F, P, NUM_COLUMNS, NUM_PUBLIC_INPUTS>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    for i in 0..NUM_LOOKUPS {
+        let local_perm_input = vars.local_values[col_permuted_input(i)];
+        let next_perm_table = vars.next_values[col_permuted_table(i)];
+        let next_perm_input = vars.next_values[col_permuted_input(i)];
+
+        // A "vertical" diff between this permuted input and the one in the previous row.
```
wborgeaud, line 77: nit: previous -> next ?

system_zero/src/lookup.rs
```diff
@@ -0,0 +1,114 @@
...
+        let diff_input_prev = next_perm_input - local_perm_input;
+        // A "horizontal" diff between this permuted input and the associated permuted table value.
```
wborgeaud, line 79: nit: this permuted input -> the next permuted input ?

wborgeaud (approved):
Looks good to me!

system_zero/src/lookup.rs
dlubarov: Interesting! I thought sorting might be a little faster but didn't expect the gap to be that significant. Did your `Ord` impl use `to_canonical`, similar to the `Hash` impl?

wborgeaud: So I didn't benchmark with field elements, just with `usize`s for now. But yes, to implement `Ord` on field elements I would probably compare the `to_canonical` calls. I wouldn't expect this to change the benchmark results, since the `Hash` impl also calls `to_canonical`.

dlubarov: Oh got it. It's an interesting problem and seems like the costs are nontrivial, so I'll try adding a benchmark with field elements.

I think the sort method may call cmp more than we currently call hash, so potentially more `to_canonical` calls, but we could pre-canonicalize to avoid that.

I think it would be nice to support tables that aren't initially ordered, though I guess we could always pass a flag to indicate if they're already ordered.

wborgeaud: > I think the sort method may call cmp more than we currently call hash, so potentially more `to_canonical` calls, but we could pre-canonicalize to avoid that.

Yeah we could canonicalize the inputs and table (maybe with a new `fn canonicalize(&mut self)` function on `Field`) and then do comparison with `to_noncanonical` calls.

> I think it would be nice to support tables that aren't initially ordered, though I guess we could always pass a flag to indicate if they're already ordered.

Good point, I think in most cases we can make sure the table is sorted but I agree it makes sense to also support unsorted tables. 
Btw in my benchmark, sorting the inputs takes 75% of the time for the whole snippet (somewhat expected since the other operations are linear), so if we have to sort the table too, using sorting becomes only 2x faster than hash maps.

wborgeaud: I've pushed my benchmarks on the `lookups_hashmap_vs_sorting` branch if you want to have a look. A few things to note:

- I've changed the benchmarks to use field elements instead of `usize`s. The result is that using sorting now takes 55ms vs 100ms with hash maps. This is without pre-canonicalization though.
- I've benchmarked what happens when the table is the one for 16-bit range checks, i.e. `t = list(range(2**16)) + [2**16-1 for _ in range(2**16, n)]`, and interestingly sorting is way faster, taking only 10ms. I suspect it's because Rust uses some kind of linear counting sort when all values are small. 
- Again using the 16-bit range check table, the hash map based method fails. I think in general the method is buggy when the table has repeated elements. I'll add a PR comment on the relevant lines.

system_zero/src/lookup.rs
```diff
@@ -0,0 +1,114 @@
...
+pub(crate) fn permuted_cols<F: PrimeField64>(inputs: &[F], table: &[F]) -> (Vec<F>, Vec<F>) {
+    let n = inputs.len();
+
+    // In the permuted inputs, copies of the same value must be grouped together. We accomplish this
+    // by building a hash-based bag (aka multiset) from the inputs, then flattening it.
+    let input_bag = create_hash_bag(inputs);
+    let permuted_inputs = flatten_hash_bag(&input_bag);
+
+    // Enumerate tables values that do not appear in the input list.
+    let mut unused_table_vals = table.iter().filter(|v| !input_bag.contains_key(v)).copied();
```
wborgeaud, line 42: I think this is wrong when `table` has duplicate elements. For example, if `table = [0; n]`, `unused_table_vals` will be empty and we'll get a panic in the `next().expect()` below.

A working version would be:
```rust
let mut unused_table_vals = create_hash_bag(&table);
for &x in input_bag.keys() {
    unused_table_vals.entry(x).and_modify(|v| *v -= 1);
}
```

system_zero/src/lookup.rs
dlubarov: Good point about duplicate table elements, I forgot we needed to support that.

I pushed my benchmark also; I read through yours and I think our methods were similar, except that I was pre-canonicalizing, and including the time to sort the table (though that's debatable since it'd be possible to skip it in our use case).

It seems sorting is slightly faster even when sorting a table of random elements, so it seems like the better option.

wborgeaud: Sounds good 👍 
I've tried your benchmark (modified to avoid the duplicate elements bug and assuming the table is sorted) and I get 7.7ms for sorting vs 26ms for hash maps with size 2^18. With the 16-bit range check table, I get 2.3ms vs 17ms.

system_zero/src/lookup.rs
dlubarov: I was thinking of "next" as the frame of reference (since the horizontal check involves both "next" values), but I see how it's confusing, will reword it.

system_zero/src/lookup.rs
dlubarov: Reworded, hopefully it's somewhat more clear now?

It might be a little inherently weird since we don't usually have constraints involving all "next" values, but here I think that's necessary, as the logic is "either this input is a duplicate of the one **above** it, or it's next to a matching table value". The Halo2 doc references the "above" value like that, with g^-1 in the constraint.

If we reframe to use "local" and "next" instead of "previous" and "local", the logic becomes "either the next input is a duplicate of the local one, or the next input is next to a matching (next) table value". I reworded the comments to frame it like that.

system_zero/src/lookup.rs
dlubarov: Fixed by deleting that method :)

starky/src/util.rs
dlubarov: Hm it seems the Plonky2 code has to clone columns because it uses the witness grid for something else later, and I think copying memory can be faster with multiple threads.

In this case I think the `Vec` should be taking ownership of the array rather than copying, making this almost free, but I'm not certain. I'll add a TODO to check.

dlubarov: I did a bit of timing, it seems to take around 0.04s either way (serial or parallel iterator) in our `SystemZero` test, though it was very noisy so hard to compare.

dlubarov: Slight correction, it's converting arrays to Vecs that's taking 0.04s - I guess it does need to copy because of the way arrays are stored as contiguous values (might be avoidable if we box them?).

Converting to `PolynomialValues` seems to be free - makes sense since it's a trivial wrapper around the `Vec`, so shouldn't change the way data is stored.

wborgeaud: Oh ok makes sense. In that case I agree it's fine not to use rayon.

PR #515, opened 2022-03-15
Fix salt issues

### LINKED ISSUE #514: Proof serialization/deserialization bug
When running:
```
let proof_bytes = proof.to_bytes()?;
let proof_from_bytes = ProofWithPublicInputs::from_bytes(proof_bytes, cd)?;
```
With (from what I can tell) any proof with the `CircuitConfig::standard_recursion_zk_config()`.
It fails with some "itertools: .zip_eq() reached end of one iterator before the other" and "failed to fill whole buffer" on a different proof.

To recreate, modify [here](https://github.com/mir-protocol/plonky2/blob/310493c293436e73096a75d0e5755816fbfe3d58/plonky2/src/plonk/recursive_verifier.rs#L203) to use the `CircuitConfig::standard_recursion_zk_config()` and run `cargo test --package plonky2 --lib --release -- plonk::recursive_verifier::tests::test_recursive_verifier --exact`

Without modification the test passes.

This also occurred with a custom circuit I was building, and seems to occur with the other test circuits in `recursive_verifier.rs`.

Am trying to work this out myself but would love some help from the dev team.

wborgeaud:
The `add_virtual_proof` and `write_fri_initial_proof` functions do not take salting into account, which leads to bugs when using ZK.
Thanks to @Shalevos for catching this! Fixes #514.

### DIFF
--- a/plonky2/src/plonk/recursive_verifier.rs
+++ b/plonky2/src/plonk/recursive_verifier.rs
@@ -4,6 +4,7 @@ use crate::hash::hash_types::{HashOutTarget, RichField};
 use crate::plonk::circuit_builder::CircuitBuilder;
 use crate::plonk::circuit_data::{CommonCircuitData, VerifierCircuitTarget};
 use crate::plonk::config::{AlgebraicHasher, GenericConfig};
+use crate::plonk::plonk_common::salt_size;
 use crate::plonk::proof::{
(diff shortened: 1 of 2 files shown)


dlubarov (approved):
Hm it would be nice if this could use the existing FRI config (`FriOracleInfo::blinding`), but that seems non-trivial. I think we'll need to split off `FriProof` serialization later and make it more generic (so Starky can use it), so maybe we can do it then.

PR #516, opened 2022-03-16
Move nonnative/curve/ecdsa code to `ecdsa` module

wborgeaud:
No functional changes. 
Uses the same trick as in the `insertion` module to keep the gadgets as `CircuitBuilder` methods, by adding `CircuitBuilderCurve, CircuitBuilderGlv,...` traits.


ecdsa/src/gadgets/curve.rs
```diff
@@ -21,19 +21,68 @@ impl<C: Curve> AffinePointTarget<C> {
     }
 }
 
-impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
-    pub fn constant_affine_point<C: Curve>(
+pub trait CircuitBuilderCurve<F: RichField + Extendable<D>, const D: usize> {
```
npwardberkeley, line 24: Hmm this feels a little hacky 🤔 But I feel like it's probably the best solution!

ecdsa/src/gadgets/ecdsa.rs
```diff
@@ -22,48 +24,48 @@ pub struct ECDSASignatureTarget<C: Curve> {
...
-        let point2 = self.glv_mul(&pk.0, &u2);
-        let point = self.curve_add(&point1, &point2);
-
-        let x = NonNativeTarget::<Secp256K1Scalar> {
-            value: point.x.value,
-            _phantom: PhantomData,
-        };
-        self.connect_nonnative(&r, &x);
-    }
+pub fn verify_message_circuit<F: RichField + Extendable<D>, const D: usize>(
```
npwardberkeley, line 27: Just curious, is this because we're expecting ECDSA verification to be done by a dedicated circuit?

npwardberkeley (approved):

ecdsa/src/gadgets/curve.rs
wborgeaud: Yeah definitely hacky haha. I think it's either that or changing all gadgets to take a `builder: &mut CircuitBuilder` which would be quite annoying.

dlubarov:
Thanks for doing that! I think it may be nice to move the U32 gates/gadgets as well, maybe into a plonky2_u32 crate, so that the "core" plonky2 crate only has the essentials needed for recursion.

PR #517, opened 2022-03-17
u32 division

nbgl:


system_zero/src/registers/alu.rs
```diff
@@ -66,4 +66,26 @@ pub(crate) const COL_MUL_ADD_OUTPUT_2: usize = super::range_check_16::col_rc_16_
...
+pub(crate) const COL_DIV_OUTPUT_REM_0: usize = super::range_check_16::col_rc_16_input(2);
+/// The second 16-bit chunk of the remainder, based on little-endian ordering.
+pub(crate) const COL_DIV_OUTPUT_REM_1: usize = super::range_check_16::col_rc_16_input(3);
+
+/// The first 16-bit chunk of a temporary value (divisor - remainder - 1).
+pub(crate) const COL_DIV_RANGE_CHECKED_TMP_0: usize = super::range_check_16::col_rc_16_input(4);
+/// The second 16-bit chunk of a temporary value (divisor - remainder - 1).
+pub(crate) const COL_DIV_RANGE_CHECKED_TMP_1: usize = super::range_check_16::col_rc_16_input(5);
+
+pub(super) const END: usize = START_SHARED_COLS + NUM_SHARED_COLS;
```
nbgl, line 91: I think this was a bug. It seems like `END` should be `START_SHARED_COLS + NUM_SHARED_COLS`, not `super::START_ALU + NUM_SHARED_COLS`. @dlubarov can you confirm.

system_zero/src/registers/alu.rs
```diff
@@ -66,4 +66,26 @@ pub(crate) const COL_MUL_ADD_OUTPUT_2: usize = super::range_check_16::col_rc_16_
...
+pub(crate) const COL_DIV_OUTPUT_QUOT_1: usize = super::range_check_16::col_rc_16_input(1);
+/// The first 16-bit chunk of the remainder, based on little-endian ordering.
+pub(crate) const COL_DIV_OUTPUT_REM_0: usize = super::range_check_16::col_rc_16_input(2);
+/// The second 16-bit chunk of the remainder, based on little-endian ordering.
+pub(crate) const COL_DIV_OUTPUT_REM_1: usize = super::range_check_16::col_rc_16_input(3);
+
+/// The first 16-bit chunk of a temporary value (divisor - remainder - 1).
+pub(crate) const COL_DIV_RANGE_CHECKED_TMP_0: usize = super::range_check_16::col_rc_16_input(4);
+/// The second 16-bit chunk of a temporary value (divisor - remainder - 1).
+pub(crate) const COL_DIV_RANGE_CHECKED_TMP_1: usize = super::range_check_16::col_rc_16_input(5);
```
nbgl, line 89: This column name is not at all descriptive, but any description would be long and not really within the purview of `alu.rs`.

system_zero/src/registers/alu.rs
dlubarov: Good catch 👍

system_zero/src/registers/alu.rs
dlubarov: Yeah I think "tmp" is reasonable for hard-to-name things like this. I didn't really understand the `divisor_rem_diff_m1` name until seeing that it was set to `divisor - remainder - 1`.

system_zero/src/alu/div-constraints-proof.txt
```diff
@@ -0,0 +1,94 @@
+Constraints A:
+A1. 0 <= dividend <= u32::MAX
```
dlubarov, line 2: Just curious why the `0 <=`s? I think it's implied that these variables are in the field `Z/pZ`, in which case negatives just don't exist right? Though doesn't hurt if you prefer it.

dlubarov: Oh I see, steps like `then divisor - remainder - 1 < 0` (in B7) are written in terms of integer arithmetic, not field arithmetic. I think to be fully rigorous we would want the latter, which would slightly complicate steps like that (since it holds for sufficiently large fields). Maybe it's alright to skip those details though.

system_zero/src/registers/alu.rs
```diff
@@ -66,4 +66,26 @@ pub(crate) const COL_MUL_ADD_OUTPUT_2: usize = super::range_check_16::col_rc_16_
...
 pub(crate) const COL_MUL_ADD_OUTPUT_3: usize = super::range_check_16::col_rc_16_input(3);
 
-pub(super) const END: usize = super::START_ALU + NUM_SHARED_COLS;
+/// Dividend for division, as an unsigned u32
+pub(crate) const COL_DIV_INPUT_DIVIDEND: usize = shared_col(0);
+/// Divisor for division, as an unsigned u32
+pub(crate) const COL_DIV_INPUT_DIVISOR: usize = shared_col(1);
+/// Inverse of the divisor in the prime field (or an arbitrary value if one does not exist).
+pub(crate) const COL_DIV_INVDIVISOR: usize = shared_col(2);
+pub(crate) const COL_DIV_NONZERO_DIVISOR: usize = shared_col(3);
```
dlubarov, line 75: nit: comment please (there's one in `generate_division`, but I think it's most useful here, or can keep both)

dlubarov (approved):
LGTM!

system_zero/src/alu/div-constraints-proof.txt
unzvfu: Hmm, maybe this works as a fully rigorous proof if we take our field representatives to be {-(p-1)/2, ..., -1, 0, 1, ..., (p-1)/2} rather than the usual {0, 1, ..., p-1}? Then lifting field elements to integers and comparing them makes sense with negatives.

unzvfu (changes requested):
Looks good. Would just like clarification on the A9 constraint as noted.

system_zero/src/alu/div-constraints-proof.txt
```diff
@@ -0,0 +1,94 @@
...
+A2. 0 <= divisor <= u32::MAX
+A3. 0 <= quotient <= u32::MAX
+A4. 0 <= remainder <= u32::MAX
+A5. 0 <= divisor_rem_diff_m1 <= u32::MAX
+A6. divisor * div_inverse = div_div_inverse
+A7. (div_div_inverse - 1) * (remainder - quotient - u32::MAX) = 0
+A8. divisor * (div_div_inverse - 1) = 0
+A9. div_inverse * (div_div_inverse - 1) = 0
```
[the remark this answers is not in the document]
nbgl: Huh, good catch! The logic there was to force `div_inverse = 0` if `divisor = 0`, but you're totally right: that's not actually necessary 😊

nbgl:
@unzvfu would you mind having another look? I've removed the redundant constraint and made the proof more formal by using proper set notation.

PR #520, opened 2022-03-20
Vectorize constraint evaluation in Starky

dlubarov:
cc @nbgl

### DIFF
--- a/starky/src/permutation.rs
+++ b/starky/src/permutation.rs
@@ -246,19 +246,23 @@ pub(crate) fn get_permutation_batches<'a, T: Copy>(
         .collect()
 }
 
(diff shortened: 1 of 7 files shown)


starky/src/permutation.rs
```diff
@@ -246,19 +246,23 @@ pub(crate) fn get_permutation_batches<'a, T: Copy>(
         .collect()
 }
 
-// TODO: Use slices.
```
dlubarov, line 249: (These need to be copied now since `get_lde_values_packed` does a transpose, so I don't think slices would improve performance.)

wborgeaud (approved):
Looks good to me!

PR #521, opened 2022-03-20
Fix inv mod xn

### LINKED ISSUE #519: `test_inv_mod_xn` is flaky
I've seen it fail randomly at least once. I assume the PRNG is to blame.

wborgeaud:
Fix `inv_mod_xn` so that it works on constant polynomials.
Fixes #519.

### DIFF
--- a/field/src/polynomial/mod.rs
+++ b/field/src/polynomial/mod.rs
@@ -526,9 +526,9 @@ mod tests {
     fn test_inv_mod_xn() {
         type F = GoldilocksField;
         let mut rng = thread_rng();
-        let a_deg = rng.gen_range(1..1_000);
+        let a_deg = rng.gen_range(0..1_000);
         let n = rng.gen_range(1..1_000);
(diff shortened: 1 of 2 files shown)


field/src/polynomial/mod.rs
```diff
@@ -526,12 +526,17 @@ mod tests {
...
         let n = rng.gen_range(1..1_000);
-        let a = PolynomialCoeffs::new(F::rand_vec(a_deg));
+        let mut a = PolynomialCoeffs::new(F::rand_vec(a_deg + 1));
+        if a.coeffs[0].is_zero() {
+            a.coeffs[0] = F::ONE; // First coefficient needs to be nonzero.
+        }
         let b = a.inv_mod_xn(n);
         let mut m = &a * &b;
-        m.coeffs.drain(n..);
+        if m.coeffs.len() > n {
```
dlubarov, line 537: I think `truncate(n)` would have the same effect

dlubarov (approved):

PR #524, opened 2022-03-25
Change selector scheme

wborgeaud:
Change the selector scheme from tree-based selectors to chain-based selectors. 
For the recursive verifier circuit, this scheme uses 3 selector polynomials instead of 7 currently. Constants need to be added in separate polynomials though, so to keep the current 5 constants per gate, we need to add another 5 polynomials.
This can be solved by changing how we handle constants in the circuit.


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -669,27 +664,16 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
-        // We find the value that minimizes `num_partial_product + quotient_degree_factor`.
-        let min_quotient_degree_factor = (max_filtered_constraint_degree - 1).max(2);
-        let max_quotient_degree_factor = self.config.max_quotient_degree_factor.min(1 << rate_bits);
-        let quotient_degree_factor = (min_quotient_degree_factor..=max_quotient_degree_factor)
-            .min_by_key(|&q| num_partial_products(self.config.num_routed_wires, q) + q)
-            .unwrap();
-        debug!("Quotient degree factor set to: {}.", quotient_degree_factor);
+        let quotient_degree_factor = self.config.max_quotient_degree_factor;
+        let mut gates = self.gates.iter().cloned().collect::<Vec<_>>();
+        gates.sort_unstable_by_key(|g| (g.0.degree(), g.0.id()));
```
dlubarov, line 669: It's a little hard to follow since the reason for sorting by degree has to do with the implementation of `selector_polynomials`. What do you think of just sorting by ID here, and then stable-sorting by degree in `selector_polynomials`? It'd be slightly wasteful but the list should be tiny, and I think it'd be easy to follow the reasoning.

In any case I think we should keep some comment; what we had before was

```rust
        // The HashSet of gates will have a non-deterministic order. When converting to a Vec, we
        // sort by ID to make the ordering deterministic.
```

dlubarov: Oh I see, `SelectorsInfo` is storing gate indices so I guess `gates` here needs the same ordering involving degree. Let's just explain why it's needed then.

plonky2/src/gates/selectors.rs
```diff
@@ -0,0 +1,82 @@
...
+    for (j, g) in instances.iter().enumerate() {
+        let GateInstance { gate_ref, .. } = g;
+        let i = index(gate_ref.0.id());
+        let gr = group(i);
+        for g in 0..groups.len() {
+            polynomials[g].values[j] = if g == gr {
+                F::from_canonical_usize(i)
+            } else {
+                unused
+            };
```
wborgeaud, line 70: Just realized that this isn't necessary if there's only one selector polynomial. I'll make the change.

plonky2/src/gates/selectors.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::gates::gate::{GateInstance, GateRef};
+use crate::hash::hash_types::RichField;
+
+/// Placeholder value to indicate that a gate doesn't use a selector polynomial.
+pub(crate) const UNUSED_SELECTOR: usize = u32::MAX as usize;
+
+#[derive(Debug, Clone)]
+pub(crate) struct SelectorsInfo {
+    pub(crate) selector_indices: Vec<usize>,
+    pub(crate) groups: Vec<(usize, usize)>,
```
dlubarov, line 13: nit: I think `Vec<Range<usize>>` may be nice - it would be a bit more self-commenting, and would slightly simplify `group` (using `contains`) and the `compute_filter`s (no literal range)

plonky2/src/gates/selectors.rs
```diff
@@ -0,0 +1,82 @@
...
+pub(crate) fn selector_polynomials<F: RichField + Extendable<D>, const D: usize>(
+    gates: &[GateRef<F, D>],
+    instances: &[GateInstance<F, D>],
+    max_degree: usize,
+) -> (Vec<PolynomialValues<F>>, SelectorsInfo) {
+    let n = instances.len();
+
+    // Greedily construct the groups.
+    let mut groups = Vec::new();
+    let mut pos = 0;
```
dlubarov, line 38: Couple nits,
- I think the loop structure may be slightly more clear if we change `pos`, `i` to `start`, `size` or similar
- Should we panic if `i` never increments because some gate is too high-degree? It seems like the current code might infinite loop in that case, unless we check for it elsewhere (I forget if we do)

plonky2/src/gates/selectors.rs
```diff
@@ -0,0 +1,82 @@
...
+        }
+        groups.push((pos, pos + i));
+        pos += i;
+    }
+
+    let index = |id| gates.iter().position(|g| g.0.id() == id).unwrap();
+    let group = |i| groups.iter().position(|&(a, b)| a <= i && i < b).unwrap();
+
+    // `selector_indices[i] = j` iff the `i`-th gate uses the `j`-th selector polynomial.
+    let selector_indices = gates
```
dlubarov, line 52: nit: can be

```rust
let selector_indices = (0..gates.len()).map(group).collect_vec();
```

plonky2/src/gates/selectors.rs
wborgeaud: Good point, I added a check.

plonky2/src/gates/selectors.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::hash::hash_types::RichField;
+
+/// Placeholder value to indicate that a gate doesn't use a selector polynomial.
+pub(crate) const UNUSED_SELECTOR: usize = u32::MAX as usize;
+
+#[derive(Debug, Clone)]
+pub(crate) struct SelectorsInfo {
+    pub(crate) selector_indices: Vec<usize>,
+    pub(crate) groups: Vec<Range<usize>>,
+    pub(crate) num_selectors: usize,
```
dlubarov, line 16: nit: since this equals `groups.len()` maybe make it a method that returns that?

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -669,27 +664,17 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
-        // We find the value that minimizes `num_partial_product + quotient_degree_factor`.
-        let min_quotient_degree_factor = (max_filtered_constraint_degree - 1).max(2);
-        let max_quotient_degree_factor = self.config.max_quotient_degree_factor.min(1 << rate_bits);
-        let quotient_degree_factor = (min_quotient_degree_factor..=max_quotient_degree_factor)
-            .min_by_key(|&q| num_partial_products(self.config.num_routed_wires, q) + q)
-            .unwrap();
-        debug!("Quotient degree factor set to: {}.", quotient_degree_factor);
+        let quotient_degree_factor = self.config.max_quotient_degree_factor;
+        let mut gates = self.gates.iter().cloned().collect::<Vec<_>>();
+        // Gates need to be sorted by their degrees to compute the selector polynomials.
```
dlubarov, line 669: I think we should also mention that `g.0.id()` is in the ordering because it's needed to make the ordering deterministic.

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -669,27 +664,17 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+        gates.sort_unstable_by_key(|g| (g.0.degree(), g.0.id()));
+        let (mut constant_vecs, selectors_info) =
+            selector_polynomials(&gates, &self.gate_instances, quotient_degree_factor + 1);
+        constant_vecs.extend(self.constant_polys());
```
dlubarov, line 673: Hm the naming is a little confusing, not your fault though and no need to change it here, but I think we should consider a big rename at some point. "Preprocessed polys" seems fitting for the whole set of polys, and then we could have "gate constants", "selectors" and "sigmas", or something like that.

dlubarov (approved):
Sorry for the slow review, it LGTM!

plonky2/src/plonk/circuit_builder.rs
wborgeaud: Yeah good point, I like the idea of renaming to "preprocessed polys" 👍

PR #525, opened 2022-03-31
Generate constants in `RandomAccessGate`

wborgeaud:
This is a PoC for having constants (as in targets returned by `CircuitBuilder::constant()`) be generated in gates other than `ConstantGate`. This is only done in `RandomAccessGate` in this PR, but this can be done generally for every gate with unused routed wires and unused constants.

Pros:
- Can use only 2 preprocessed polynomials to hold constants (we use 5 right now (kind of)). That's kind of the lower bound since gates like `ArithmeticGate` use 2 constants.
- Removes a bunch of `ConstantGate`s which makes circuits smaller.

For example, with this the recursive verifier circuit uses 3996 gates and the proof is 127184 bytes, as opposed to 4083 gates and 127904 bytes now.

Cons:
- Very ugly.
- Doesn't improve anything in the case of circuits not using gates with free routed space for these extra constants.


dlubarov:
This is not as ugly as I expected! That's nice that we only need `RandomAccessGate`, and that can do away with `ConstantGate` entirely in the recursive circuit, saving quite a few rows and also reducing the overhead of constraint evaluation. It makes me think, maybe we can also do away with NoopGate when ZK is off, and use some other gate (like ArithmeticGate with virtual targets, later connected to zero) to do no-ops.

Since `extra_constants` is in the public API I just want to be sure it's as simple as it can be. I made a small PR with a variant to consider: #528.

I was hoping the constant constraints could go in some "core" code like `eval_filtered` rather than gate-specific code, but currently the gate-specific code is allocating a `Vec` of the proper length, so unless we change that or do an unnecessary resize, it seems the gate-specific needs to account for those constraints.

dlubarov (approved):

wborgeaud:
> It makes me think, maybe we can also do away with NoopGate when ZK is off, and use some other gate (like ArithmeticGate with virtual targets, later connected to zero) to do no-ops.

Good idea!

> I was hoping the constant constraints could go in some "core" code like `eval_filtered` rather than gate-specific code, but currently the gate-specific code is allocating a `Vec` of the proper length, so unless we change that or do an unnecessary resize, it seems the gate-specific needs to account for those constraints.

Yeah ideally this would all be done in the `build()` function. We could look at all the gates and see if there's space for extra-constants, and if so somehow wrap the gate in a `GateWithExtraConstants<G>` that automatically adds the constraints and route the constants. But I think this would require quite a large refactor and be overly complex since in practice we probably won't need that many gates with extra constants.

PR #528, opened 2022-04-04
Change gate method for extra constants

dlubarov:
Potential variant of #525.


plonky2/src/iop/generator.rs
```diff
@@ -328,35 +328,27 @@ impl<F: Field> SimpleGenerator<F> for NonzeroTestGenerator {
 
 /// Generator used to fill an extra constant.
 #[derive(Debug, Clone)]
-pub struct ConstantGenerator<F: Field> {
+pub(crate) struct ConstantGenerator<F: Field> {
     pub gate_index: usize,
     pub constant_index: usize,
-    pub target_index: usize,
+    pub input_index: usize,
```
wborgeaud, line 334: nit: Hm why the "input" name here? I would actually see this index as more of an output index, which would be consistent with the naming `wire_output()` used in `ConstantGate`.

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -219,10 +219,17 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         );
         constants.resize(gate_type.num_constants(), F::ZERO);
 
-        let index = self.gate_instances.len();
+        let gate_index = self.gate_instances.len();
 
         self.constant_generators
-            .extend(gate_type.extra_constants(index));
+            .extend(gate_type.extra_constant_wires().enumerate().map(
+                |(constant_index, input_index)| ConstantGenerator {
```
wborgeaud, line 226: I think this wouldn't work for a gate that uses (non-extra) constants. For example, if we wanted to add extra-constants to `ArithmeticGate`, the first two constants would be the `c0,c1` used in the arithmetic operation. So the `enumerate` here would yield incorrect indices.

I'm not sure we'll ever need extra-constants on gates already using constants, and this could be fixed by using the convention that extra-constants should be placed before other constants. But I think having `extra_constant_wires()` return a `Vec<(usize, usize)>` would be less confusing.

wborgeaud (approved):
Looks good to me otherwise!

plonky2/src/iop/generator.rs
dlubarov: Hm I see what you mean, I just thought I should match the `Wire::input` naming - in that sense each wire is an input to some gate. But I see how it can be misleading.

Do you think "wire" would be more clear? Or maybe "column" - it's not often used with circuits but at least it couldn't be confused with anything else.

wborgeaud: Ah I see, that makes sense. Yeah I think "wire" might be clearer.

plonky2/src/plonk/circuit_builder.rs
dlubarov: Good point! Changed to `Vec<(usize, usize)>`.

PR #529, opened 2022-04-06
System Zero binary bitwise operations

unzvfu:
This PR adds support for bitwise AND, IOR, XOR, and ANDNOT of u64 operands to the System Zero VM.


system_zero/src/alu/bitops.rs
```diff
@@ -0,0 +1,473 @@
...
+
+use crate::registers::alu::*;
+use crate::registers::NUM_COLUMNS;
+
+/// Interpret the first 32 elements of `bits` as bits from low to high
+/// of a u32 and return \sum_i bits[i] 2^i as an element of P.
+///
+/// TODO: This should probably raise an error if there are more than
+/// 32 elements in `bits`.
+fn binary_to_u32<F, P, I>(bits: I) -> P
```
nbgl, line 17: I would suggest defining `I` as `&[P; 32]`, so the type system enforces the length.

system_zero/src/registers/alu.rs
```diff
@@ -89,4 +93,150 @@ pub(crate) const COL_DIV_RANGE_CHECKED_TMP_0: usize = super::range_check_16::col
 /// The second 16-bit chunk of a temporary value (divisor - remainder - 1).
 pub(crate) const COL_DIV_RANGE_CHECKED_TMP_1: usize = super::range_check_16::col_rc_16_input(5);
 
+///
+/// Bitwise logic operations
+///
+
+/// Bit decomposition of 64-bit values
+pub(crate) const COL_BIT_DECOMP_INPUT_A_LO_00: usize = shared_col(0);
```
nbgl, line 101: Maybe let’s make these an array 🙂

system_zero/src/registers/alu.rs
```diff
@@ -89,4 +93,34 @@ pub(crate) const COL_DIV_RANGE_CHECKED_TMP_0: usize = super::range_check_16::col
...
+    regs
+}
+
+pub(crate) const COL_BIT_DECOMP_INPUT_A_LO_BIN_REGS: [usize; 32] = gen_bitop_32bit_input_regs(0);
+pub(crate) const COL_BIT_DECOMP_INPUT_A_HI_BIN_REGS: [usize; 32] = gen_bitop_32bit_input_regs(32);
+pub(crate) const COL_BIT_DECOMP_INPUT_B_LO_BIN_REGS: [usize; 32] = gen_bitop_32bit_input_regs(64);
+pub(crate) const COL_BIT_DECOMP_INPUT_B_HI_BIN_REGS: [usize; 32] = gen_bitop_32bit_input_regs(96);
+
+/// The first 16-bit chunk of the output, based on little-endian ordering.
+pub(crate) const COL_BITOP_OUTPUT_0: usize = super::range_check_16::col_rc_16_input(0);
```
dlubarov, line 118: Hm I would have expected outputs to be in 32-bit chunks. For mul etc. we used 16-bit output chunks since they needed to be range-checked, but here the outputs' range should be enforced by the constraints, I think.

I suppose there's no harm in this though, since it's using registers that would otherwise be unused for bit ops. It's nice not to need extra registers.

In the future, maybe we can write outputs like this directly to the memory bus (though it's not implemented yet).

system_zero/src/alu/bitops.rs
```diff
@@ -0,0 +1,319 @@
...
+
+/// As for `binary_to_u32` but uses `builder`.
+fn binary_to_u32_recursively<F, const D: usize>(
+    builder: &mut CircuitBuilder<F, D>,
+    bits: [ExtensionTarget<D>; 32],
+) -> ExtensionTarget<D>
+where
+    F: RichField + Extendable<D>,
+{
+    let terms = bits
```
dlubarov, line 33: We also have a method `reduce_with_powers_ext_recursive` which could be used. It might use fewer gates since it uses `ReducingExtensionGate`, vs the more general-purpose `ArithmeticExtensionGate`.

system_zero/src/registers/alu.rs
unzvfu: I've updated this to return two 32-bit chunks, taking two extra registers (total 130 for bit ops). We can always change back if we want to save those two extra registers.

system_zero/src/alu/bitops.rs
```diff
@@ -134,28 +118,40 @@ pub(crate) fn eval_bitop<F: Field, P: PackedField<Scalar = F>>(
...
 
+fn constrain_all_to_bits<F: RichField + Extendable<D>, const D: usize>(
+    builder: &mut CircuitBuilder<F, D>,
+    yield_constr: &mut RecursiveConstraintConsumer<F, D>,
+    filter: ExtensionTarget<D>,
+    vals: &[ExtensionTarget<D>]
+) {
+    for v in vals.into_iter() {
+        let one = builder.one_extension();
+        let t0 = builder.sub_extension(one, *v);
```
dlubarov, line 141: Normally this would be too minor to bother, but since we'll have 128 iterations here, I think it'd be good to change `v * (1 - v)` to `v * v - v`, so we can use `mul_sub(v, v, v)` to save one arithmetic op.

(Since that switches `1 - v` to `v - 1`, I think the same change would need to be made in `eval_bitop_32`.)

dlubarov (approved):
LGTM!

system_zero/src/alu/bitops.rs
unzvfu: Good catch!

PR #530, opened 2022-04-06
Move u32 gates and gadgets to `plonky2_u32` crate

wborgeaud:
Create a `plonky2_u32` crate and move all u32 related stuff there. 
Also move the `SwitchGate` to `waksman` crate.


plonky2_u32/src/gadgets/multiple_comparison.rs
```diff
@@ -0,0 +1,149 @@
...
+use plonky2::plonk::circuit_builder::CircuitBuilder;
+use plonky2_field::extension_field::Extendable;
+use plonky2_util::ceil_div_usize;
+
+use super::arithmetic_u32::U32Target;
+use crate::gates::comparison::ComparisonGate;
+
+/// Returns true if a is less than or equal to b, considered as base-`2^num_bits` limbs of a large value.
+/// This range-checks its inputs.
+pub fn list_le_circuit<F: RichField + Extendable<D>, const D: usize>(
```
wborgeaud, line 12: I wasn't too sure about where to put the comparison gadget and gate. It is currently only used in the biguint code, so it could also make sense to move it there.

Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [workspace]
-members = ["field", "insertion", "plonky2", "starky", "system_zero", "util", "waksman", "ecdsa"]
+members = ["field", "insertion", "plonky2", "starky", "system_zero", "util", "waksman", "ecdsa", "plonky2_u32"]
```
dlubarov, line 2: nit: existing convention is to include `plonky2_` in the crate name but not the directory

wborgeaud: For some reason I felt that having a directory called `u32` would be a bit weird, but I agree we should stick to the convention 👍

dlubarov (approved):
LGTM, thanks for doing that 👍

PR #533, opened 2022-04-07
Do not export global allocator

### LINKED ISSUE #532: Do not set a global allocator
I agree that jemalloc is great and can outperform the default allocator (especially on musl builds). In fact I use mimalloc for these reasons, which is smaller and faster according to the authors.

But it is not appropriate for a library to set the global allocator. The global allocator can only be set once, or else the build will fail. If plonky2 appears as a dependency anywhere in a project that uses a custom allocator, the build will fail. If another library where to do the same, those two could never co-exist, etc.

Downstreams projects may want to use mimalloc instead or may want to wrap it with custom logic to add metrics and logging to allocations.

recmo:
Fixes #532.

Jemalloc is still used for tests and benchmarks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,11 @@ in the Plonky2 directory.
 
 ## Running
 
-To see recursion performance, one can run this test, which generates a chain of three recursion proofs:
+To see recursion performance, one can run this bench, which generates a chain of three recursion proofs:
 
 ```sh
-RUST_LOG=debug RUSTFLAGS=-Ctarget-cpu=native cargo test --release test_recursive_recursive_verifier
+RUSTFLAGS=-Ctarget-cpu=native cargo run --release --example bench_recursion -- -vv
 ```
-
 
 ## Jemalloc
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,15 @@ RUST_LOG=debug RUSTFLAGS=-Ctarget-cpu=native cargo test --release test_recursive
 
 ## Jemalloc
 
-By default, Plonky2 uses the [Jemalloc](http://jemalloc.net) memory allocator due to its superior performance. Currently, it changes the default allocator of any binary to which it is linked. You can disable this behavior by removing the corresponding lines in [`plonky2/src/lib.rs`](https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/lib.rs).
+Plonky2 prefers the [Jemalloc](http://jemalloc.net) memory allocator due to its superior performance. To use it, include `jemallocator = "0.3.2"` in`Cargo.toml`and add the following lines
+to your `main.rs`:
+
+```rust
+use jemallocator::Jemalloc;
+
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
+```
 
 Jemalloc is known to cause crashes when a binary compiled for x86 is run on an Apple silicon-based Mac under [Rosetta 2](https://support.apple.com/en-us/HT211861). If you are experiencing crashes on your Apple silicon Mac, run `rustc --print target-libdir`. The output should contain `aarch64-apple-darwin`. If the output contains `x86_64-apple-darwin`, then you are running the Rust toolchain for x86; we recommend switching to the native ARM version.
 
--- a/field/src/field_testing.rs
+++ b/field/src/field_testing.rs
@@ -9,8 +9,7 @@ macro_rules! test_field_arithmetic {
         mod field_arithmetic {
             use num::bigint::BigUint;
             use rand::Rng;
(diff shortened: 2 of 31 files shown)


nbgl:
Wow, that was fast! Thanks for this PR 🙂 Looks good to me.

(I’m sorry, we haven’t got a policy on how to accept pull requests yet. They’re most certainly welcome and I do want to merge this! I just need to check if there’s a CLA involved. Hopefully this will be quick.)

recmo:
Do you think it makes sense to have jemalloc active in the tests? It adds boilerplate to all the `lib.rs` and `Cargo.toml` files. We could use the default allocator, I don't think the test time will noticably change. We would still use jemalloc for benchmarks, which gives a 1.5% boost on my machine for the ffts. WDYT?

nbgl:
Unfortunately, [our main benchmark](https://github.com/mir-protocol/plonky2/blob/1558b89f1a9d64a9e20766e3f78c929dc1e38750/plonky2/src/plonk/recursive_verifier.rs#L216) is actually a test that logs the timings (@dlubarov might know why that decision was made)

recmo:
I just did a quick bench comparing default, `jemalloc` and `mimalloc`:

```
# jemalloc
[DEBUG plonky2::util::timing] 0.2571s to prove
[DEBUG plonky2::util::timing] 0.2600s to prove
[DEBUG plonky2::util::timing] 0.2522s to prove
[DEBUG plonky2::util::timing] 0.2555s to prove
[DEBUG plonky2::util::timing] 0.2772s to prove

# mimalloc
[DEBUG plonky2::util::timing] 0.2636s to prove
[DEBUG plonky2::util::timing] 0.2774s to prove
[DEBUG plonky2::util::timing] 0.2832s to prove
[DEBUG plonky2::util::timing] 0.2802s to prove
[DEBUG plonky2::util::timing] 0.2696s to prove

# std
[DEBUG plonky2::util::timing] 0.3220s to prove
[DEBUG plonky2::util::timing] 0.2981s to prove
[DEBUG plonky2::util::timing] 0.2913s to prove
[DEBUG plonky2::util::timing] 0.2963s to prove
[DEBUG plonky2::util::timing] 0.3136s to prove
```

Looks like jemalloc is the fastest.

dlubarov:
Thanks for the PR! 

> Unfortunately, [our main benchmark](https://github.com/mir-protocol/plonky2/blob/1558b89f1a9d64a9e20766e3f78c929dc1e38750/plonky2/src/plonk/recursive_verifier.rs#L216) is actually a test that logs the timings (@dlubarov might know why that decision was made)

We didn't want to use a Criterion benchmark since it's a pretty large operation, so we wanted to run it just once, and I think Criterion has a minimum iteration count.

We can convert some of our recursion tests to bins though, if that would that avoid the boilerplate? We were thinking about doing that anyway. I'm happy to make the change later.

(We might keep some as tests, e.g. `test_recursive_verifier_multi_hash`, since we want that to make sure it passes CI and aren't too interested in its performance. So we'd be splitting up perf tests into perf bins and regular tests.)

recmo:
I tried to get Criterion to run a one-shot benchmark in some hacky ways. I got it to work, but the plot generation has some errors. Criterion really doesn't want to do one-shot. I also tried the `iai` crate from the same author, which does do one-shot. But this requires `valgrind` to be installed which is not even available on MacOS.

So the solution seems to be a custom benchmarking bein. Easy enough. I was planning to write such a tool anyway to automatically run a long sequence of proofs of different sizes and different number of cores. This is very useful for investigating performance issues with very large proofs.

I tried copying over the test, but I'm hitting issues with certain functions and types not being public. (I've hit this issue before, for example it doesn't seem to be possible to produce a `CompressedProof` out side of the crate because some data was private).

At this point I feel that it's getting out of scope for this PR, so I propose splitting the work up in

1. Merge this PR in it's current state.
2. Merge PR: Implement a benchmarking tool fixing any visibility issues.
3. Merge PR: Chore to remove the now redudant boilerplate.

nbgl:
Hi @recmo, I’m sorry it took me so long to get back to you. We still haven’t quite settled on a license for Plonky2, but we do intend to release it under an open-source license. We would love to merge your contributions if you are happy to license them under MIT, BSD, or some other permissive (and GPL-compatible) license; this lets us retain flexibility while we evaluate different options.

recmo:
Of course! Any permissive license works for me.

To state it explicitly: my PRs and contributions to plonky2 are available under an MIT license.

Let me know if this statement is enough or if further action is required on my end.

dlubarov (approved):
Sorry for the delay, this seems like a good approach to me!

PR #534, opened 2022-04-08
Move benches to bins

recmo:
🪑➡️🗑️


recmo:
This is a straightforward copy from the tests, but requires most of the `circuit_data` structs to be made public. I'm not sure if this is desirable.

recmo:
I've added some CLI arguments to do sequences of bench runs:

```

```

recmo:
Added some CLI arguments to facilitate running benchmarks over ranges of sizes/threads:

```
USAGE:
    bench_recursion [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information
    -v, --verbose    Verbose mode (-v, -vv, -vvv, etc.)

OPTIONS:
        --log-filter <log-filter>    Apply an env_filter compatible log filter [env: LOG_FILTER=]  [default: ]
        --seed <seed>                Random seed for deterministic runs. If not specified a new seed is generated from
                                     OS entropy
        --size <size>                Log2 gate count of the inner proof. Can be a single value or a rust style range
                                     [default: 14]
        --threads <threads>          Number of compute threads to use. Defaults to number of cores. Can be a single
                                     value or a rust style range
```

Next step: add CSV output support for further perf analysis.

dlubarov:
Sorry for the delay! I've been busy moving, but taking a look now.

plonky2/src/plonk/circuit_data.rs
```diff
@@ -247,40 +247,40 @@ pub struct CommonCircuitData<
     C: GenericConfig<D, F = F>,
     const D: usize,
 > {
-    pub(crate) config: CircuitConfig,
+    pub config: CircuitConfig,
 
-    pub(crate) fri_params: FriParams,
+    pub fri_params: FriParams,
```
dlubarov, line 252: For this struct, I see why `config` and `degree_bits` need to be public now but it seems like the other fields can stay private, no?

dlubarov (commented):
This is very nifty! Thanks for adding it.

The code looks good to me apart from a few minor comments and the warnings. The only other thing is the license question (see #533).

dlubarov:
Look great. CI is just red due to formatting but I'll fmt after merging. I sent you an invite to the repo so CI can auto-run in any future PRs.

PR #535, opened 2022-04-13
System Zero bit rotate and shift operations

unzvfu:
This PR adds support for the ROTATE_LEFT, ROTATE_RIGHT, SHIFT_LEFT and SHIFT_RIGHT instructions to the System Zero VM.

### DIFF
--- a/system_zero/src/alu/bitops.rs
+++ b/system_zero/src/alu/bitops.rs
@@ -10,9 +10,9 @@ use starky::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsume
 use crate::registers::alu::*;
 use crate::registers::NUM_COLUMNS;
 
-/// Interpret the 32 elements of `bits` as bits from low to high of a
+/// Interpret the N <= 32 elements of `bits` as bits from low to high of a
 /// u32 and return \sum_i bits[i] 2^i as an element of P.
-fn binary_to_u32<F, P>(bits: [P; 32]) -> P
+pub(crate) fn binary_to_u32<const N: usize, F, P>(bits: [P; N]) -> P
 where
     F: Field,
     P: PackedField<Scalar = F>,
@@ -130,7 +130,7 @@ pub(crate) fn eval_bitop<F: Field, P: PackedField<Scalar = F>>(
     );
 }
 
-fn constrain_all_to_bits<F: RichField + Extendable<D>, const D: usize>(
+pub(crate) fn constrain_all_to_bits<F: RichField + Extendable<D>, const D: usize>(
     builder: &mut CircuitBuilder<F, D>,
     yield_constr: &mut RecursiveConstraintConsumer<F, D>,
     filter: ExtensionTarget<D>,
--- a/system_zero/src/alu/bitops.rs
+++ b/system_zero/src/alu/bitops.rs
@@ -12,7 +12,7 @@ use crate::registers::NUM_COLUMNS;
 
 /// Interpret the N <= 32 elements of `bits` as bits from low to high of a
 /// u32 and return \sum_i bits[i] 2^i as an element of P.
-pub(crate) fn binary_to_u32<const N: usize, F, P>(bits: [P; N]) -> P
+pub(crate) fn binary_to_u32<F, P>(bits: [P; 32]) -> P
 where
     F: Field,
     P: PackedField<Scalar = F>,
--- a/system_zero/src/alu/mod.rs
+++ b/system_zero/src/alu/mod.rs
@@ -11,6 +11,11 @@ use crate::alu::addition::{eval_addition, eval_addition_recursively, generate_ad
 use crate::alu::bitops::{eval_bitop, eval_bitop_recursively, generate_bitop};
 use crate::alu::division::{eval_division, eval_division_recursively, generate_division};
 use crate::alu::mul_add::{eval_mul_add, eval_mul_add_recursively, generate_mul_add};
+use crate::alu::rotate_shift::{
+    eval_rotate_left, eval_rotate_left_recursively,
+    eval_shift_left, eval_shift_left_recursively,
+    generate_rotate_shift,
+};
 use crate::alu::subtraction::{
     eval_subtraction, eval_subtraction_recursively, generate_subtraction,
 };
@@ -23,11 +28,14 @@ mod bitops;
 mod canonical;
 mod division;
 mod mul_add;
+mod rotate_shift;
 mod subtraction;
 
 // TODO: This probably belongs in a more easily accessible location.
-const ALL_OPERATIONS: [usize; 8] = [
+const ALL_OPERATIONS: [usize; 13] = [
     IS_ADD, IS_SUB, IS_MUL_ADD, IS_DIV, IS_AND, IS_IOR, IS_XOR, IS_ANDNOT,
+    IS_ROTATE_LEFT, IS_SHIFT_LEFT, IS_ROTATE_RIGHT, IS_SHIFT_RIGHT,
+    IS_ARITH_SHIFT_RIGHT,
 ];
 
 pub(crate) fn generate_alu<F: PrimeField64>(values: &mut [F; NUM_COLUMNS]) {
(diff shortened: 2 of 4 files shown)


system_zero/src/alu/rotate_shift.rs
```diff
@@ -0,0 +1,596 @@
...
+    //       = pow_exp
+    //
+    // on the conditions that:
+    //
+    // pow_exp_aux_0 = \prod_i=0^1 ((2^i - 1) * exp_bits[i] + 1)
+    // pow_exp_aux_1 = \prod_i=2^3 ((2^i - 1) * exp_bits[i] + 1)
+    // pow_exp_aux_2 = pow_exp_aux_0 * pow_exp_aux_1
+    // pow_exp_mod32 = pow_exp_aux_2 * ((2^(2^4) - 1) * exp_bits[4] + 1)
+
+    let one = P::ONES;
```
nbgl, line 163: Redundant variable? Can just use `P::ONES`, like two lines below this one.

system_zero/src/alu/rotate_shift.rs
```diff
@@ -0,0 +1,596 @@
...
+    let c = [1, 2, 3, 4].map(|i| P::from(F::from_canonical_u64(1u64 << (1u32 << i))) - P::ONES);
+
+    let constr1 = (exp_bits[0] + one) * (c[0] * exp_bits[1] + one);
+    yield_constr.constraint(filter * (constr1 - pow_exp_aux_0));
+    let constr2 = (c[1] * exp_bits[2] + one) * (c[2] * exp_bits[3] + one);
+    yield_constr.constraint(filter * (constr2 - pow_exp_aux_1));
+    let constr3 = pow_exp_aux_0 * pow_exp_aux_1;
+    yield_constr.constraint(filter * (constr3 - pow_exp_aux_2));
+    let constr4 = pow_exp_aux_2 * (c[3] * exp_bits[4] + one);
+    yield_constr.constraint(filter * (constr4 - pow_exp));
```
nbgl, line 174: I really suggest making this its own function. It’s kinda difficult to parse, and has a very different purpose from the rest of the function.

system_zero/src/alu/rotate_shift.rs
```diff
@@ -0,0 +1,596 @@
...
+) -> (P, P, P, P, P, P, P) {
+    // Input
+    let input_lo = lv[COL_ROTATE_SHIFT_INPUT_LO];
+    let input_hi = lv[COL_ROTATE_SHIFT_INPUT_HI];
+
+    // Delta is the shift/rotate displacement; exp is delta mod 32 or
+    // (32 - delta mod 32) mod 32, depending on whether the operation
+    // direction is left or right.
+    let exp_bits = COL_ROTATE_SHIFT_EXP_BITS.map(|r| lv[r]);
+    let top_bit = lv[COL_ROTATE_SHIFT_DELTA_DIV32];
```
nbgl, line 125: The name `top_bit` isn’t very descriptive. `top_exp_bit`? But then `exp_bits` would have to become something like `other_exp_bits`… 🤔

system_zero/src/alu/rotate_shift.rs
```diff
@@ -0,0 +1,596 @@
...
+
+    let pow_exp_aux_0 = lv[COL_ROTATE_SHIFT_POW_EXP_AUX_0];
+    let pow_exp_aux_1 = lv[COL_ROTATE_SHIFT_POW_EXP_AUX_1];
+    let pow_exp_aux_2 = lv[COL_ROTATE_SHIFT_POW_EXP_AUX_2];
+    let pow_exp = lv[COL_ROTATE_SHIFT_POW_EXP];
+
+    let shifted_lo_0 = lv[COL_ROTATE_SHIFT_DISPLACED_INPUT_LO_0];
+    let shifted_lo_1 = lv[COL_ROTATE_SHIFT_DISPLACED_INPUT_LO_1];
+    let shifted_hi_0 = lv[COL_ROTATE_SHIFT_DISPLACED_INPUT_HI_0];
+    let shifted_hi_1 = lv[COL_ROTATE_SHIFT_DISPLACED_INPUT_HI_1];
```
nbgl, line 135: When I read `shifted_lo`, I expect the `lo` part of the `shifted` value. But this is the `lo` part that has been `shifted`. I think `lo_shifted` would be less confusing, at least for me.

system_zero/src/alu/rotate_shift.rs
```diff
@@ -0,0 +1,596 @@
...
+    let constr1 = shifted_lo_aux_0 * (u32_max - shifted_lo_1);
+    yield_constr.constraint(filter * (constr1 - shifted_lo_aux_1));
+    let constr2 = shifted_hi_aux_0 * (u32_max - shifted_hi_1);
+    yield_constr.constraint(filter * (constr2 - shifted_hi_aux_1));
+
+    let shifted_lo_is_valid = shifted_lo_0 * (one - shifted_lo_aux_1);
+    let shifted_hi_is_valid = shifted_hi_0 * (one - shifted_hi_aux_1);
+
+    yield_constr.constraint(filter * shifted_lo_is_valid);
+    yield_constr.constraint(filter * shifted_hi_is_valid);
```
nbgl, line 201: Again, this checking validation of `shifted_lo_0` and `shifted_lo_1` (and `_hi`) could be done in a separate function just for clarity

nbgl (approved):
Looks good! A few nits.

I like how this doesn’t require any range checks, but I wonder if they could simplify some of the logic. For example, if we split the input into 16-bit chunks, then when it is shifted it is guaranteed to be a valid Goldilocks element. This would eliminate the need for the check-if-valid-element trick. In any case, this can be done later.

system_zero/src/alu/bitops.rs
```diff
@@ -10,9 +10,9 @@ use starky::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsume
 use crate::registers::alu::*;
 use crate::registers::NUM_COLUMNS;
 
-/// Interpret the 32 elements of `bits` as bits from low to high of a
+/// Interpret the N <= 32 elements of `bits` as bits from low to high of a
 /// u32 and return \sum_i bits[i] 2^i as an element of P.
-fn binary_to_u32<F, P>(bits: [P; 32]) -> P
+pub(crate) fn binary_to_u32<const N: usize, F, P>(bits: [P; N]) -> P
```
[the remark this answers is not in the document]
unzvfu: Yeah, it's a bit ridiculous that `static_assertions::const_assert!(N <= 32)` doesn't work. Anyway, turns out I don't need the generic version anyway (I did need it during one iteration of the code), so I'm just going to explicitly specify the array length as 32.

system_zero/src/alu/rotate_shift.rs
unzvfu: Good call; found a significant bug in the recursive version when I refactored!

system_zero/src/alu/rotate_shift.rs
unzvfu: Renamed to `exp_ge32_bit`.

system_zero/src/alu/mod.rs
```diff
@@ -47,6 +65,14 @@ pub(crate) fn generate_alu<F: PrimeField64>(values: &mut [F; NUM_COLUMNS]) {
...
         generate_bitop(values, IS_ANDNOT);
+    } else if values[IS_ROTATE_LEFT].is_one() {
+        generate_rotate_shift(values, IS_ROTATE_LEFT);
+    } else if values[IS_ROTATE_RIGHT].is_one() {
+        generate_rotate_shift(values, IS_ROTATE_RIGHT);
+    } else if values[IS_SHIFT_LEFT].is_one() {
+        generate_rotate_shift(values, IS_SHIFT_LEFT);
+    } else if values[IS_SHIFT_RIGHT].is_one() {
+        generate_rotate_shift(values, IS_SHIFT_RIGHT);
     }
```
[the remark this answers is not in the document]
unzvfu: I agree, but the `system_zero::tests::run` test relies on the fall-through case doing nothing. I've put in a commented `todo!(...)` as a compromise.

Issue #538, opened 2022-04-28
Order matters when evaluating constraints to `RecursiveConstraintConsumer`


Sladuca:
When implementing `eval_ext_recursively`, it seems that the order in which the constraints added to the `RecursiveConstraintConsumer` matters, specifically that the order matches that in `eval_packed_generic`. One order works just fine, while for the other order it fails saying something like `Partition containing ... was set twice with different values`, even with the same exact constraints.

Is this intended behavior?

wborgeaud:
Yes this is intended. 
The actual constraint that is used to compute the STARK is a linear combination of the constraints given in the `ConstraintConsumer` with powers of some random challenge `alpha`. So constraints `f_0, f_1, ..., f_n` will be combined into `f_0 + alpha * f_1 + ... + alpha^n * f_n`.
So a different constraint order will result in a different linear combination.

Sladuca:
Oh I see! Thank you.

PR #539, opened 2022-04-29
Make `recover_degree_bits` public

Sladuca:
`degree_bits` is needed by `add_virtual_stark_proof_with_pis` when building a plonky2 SNARK that verifies a starky STARK proof. Therefore the user needs `recover_degree_bits` to be public.

I also added a doc comment to `eval_ext_recursively` to clear up #538.

### DIFF
--- a/starky/src/proof.rs
+++ b/starky/src/proof.rs
@@ -53,7 +53,7 @@ pub struct StarkProofTarget<const D: usize> {
 
 impl<const D: usize> StarkProofTarget<D> {
     /// Recover the length of the trace from a STARK proof and a STARK config.
-    pub(crate) fn recover_degree_bits(&self, config: &StarkConfig) -> usize {
+    pub fn recover_degree_bits(&self, config: &StarkConfig) -> usize {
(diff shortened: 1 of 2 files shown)


starky/src/proof.rs
```diff
@@ -33,7 +33,7 @@ pub struct StarkProof<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>,
 
 impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize> StarkProof<F, C, D> {
     /// Recover the length of the trace from a STARK proof and a STARK config.
-    pub(crate) fn recover_degree_bits(&self, config: &StarkConfig) -> usize {
+    pub fn recover_degree_bits(&self, config: &StarkConfig) -> usize {
```
wborgeaud, line 36: Maybe we can also make the `recover_degree_bits` function on `StarkProofTarget` public as well.

wborgeaud (approved):
Looks good to me, thanks for the PR 👍

starky/src/proof.rs
Sladuca: good idea!

PR #541, opened 2022-05-06
Start of Starky multitable implementation

wborgeaud:
Start of (the prover side of the) Starky multitable implementation.
Since this has a lot of breaking changes with `starky`, I've put everything in a new `starky2` crate. The files in `starky2` that differ with the corresponding `starky` files are: `prover.rs`, `all_stark.rs`, `cross_table_lookup.rs`, `vanishing_polynomial.rs` just to add the cross-table check eval, and `mock_stark.rs` in which I plan to put some tests. 

TLDR of changes:

- Add a struct to hold cross-table lookup info:
```rust
pub struct CrossTableLookup {
    pub looking_table: Table,
    pub looking_columns: Vec<usize>,
    pub looked_table: usize,
    pub looked_columns: Vec<usize>,
}
```

- Add a struct to hold every STARK used in a tx proof:
```rust
pub struct AllStark<F: RichField + Extendable<D>, const D: usize> {
    pub cpu: CpuStark<F, D>,
    pub keccak: KeccakStark<F, D>,
    // ...
}
```

- Modify the prover to do the following
  - Takes `AllStark`, the corresponding tables, and the cross-table lookup data as input.
  - Commit to all the traces.
  - Compute the lookup `Z` polynomials.
  - Compute the proof of every STARK as in `starky`.

Still TODO:
- Implement openings of lookup `Z` polynomials. Q: Is it more efficient to open them like the other polynomials using the FRI LDT, or commit to `Z(x) - Z(w^n-1) * L_{w^n-1}(x)` instead and let the verifier fill the gap.
- Implement verifier and recursive verifier
- Fix challenges
- Tests
- Lots of cleaning
- Maybe: Different config for every STARK in the proof


starky2/src/cross_table_lookups.rs
```diff
@@ -0,0 +1,176 @@
...
+use crate::config::StarkConfig;
+use crate::constraint_consumer::ConstraintConsumer;
+use crate::permutation::PermutationChallenge;
+use crate::stark::Stark;
+use crate::vars::StarkEvaluationVars;
+
+pub struct CrossTableLookup {
+    pub looking_table: Table,
+    pub looking_columns: Vec<usize>,
+    pub looked_table: usize,
```
dlubarov, line 21: Was there a reason to use Table for looking but usize for looked?

wborgeaud: Oops, not at all just a typo. Fixed.

starky2/src/stark.rs
```diff
@@ -0,0 +1,229 @@
...
+        config: &StarkConfig,
+    ) -> FriInstanceInfo<F, D> {
+        let no_blinding_oracle = FriOracleInfo { blinding: false };
+        let mut oracle_indices = 0..;
+
+        let trace_info =
+            FriPolynomialInfo::from_range(oracle_indices.next().unwrap(), 0..Self::COLUMNS);
+
+        let num_permutation_batches = self.num_permutation_batches(config);
+        let permutation_ctl_zs_info = (num_permutation_batches + num_ctl_zs > 0).then(|| {
```
dlubarov, line 100: I think `num_ctl_zs > 0` should always hold? (Except maybe if this was used for single-table STARKs, but I think we can ignore other use cases since this is sort of EVM specific.)

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,294 @@
+use anyhow::{ensure, Result};
```
dlubarov, line 1: In terms of naming do you think "cross-table permutation" would be okay? Lookup might be misinterpreted as a subset type lookup

starky2/src/proof.rs
```diff
@@ -0,0 +1,259 @@
...
+}
+
+/// Purported values of each polynomial at the challenge point.
+#[derive(Debug, Clone)]
+pub struct StarkOpeningSet<F: RichField + Extendable<D>, const D: usize> {
+    pub local_values: Vec<F::Extension>,
+    pub next_values: Vec<F::Extension>,
+    pub permutation_ctl_zs: Option<Vec<F::Extension>>,
+    pub permutation_ctl_zs_right: Option<Vec<F::Extension>>,
+    pub ctl_zs_last: Vec<F>,
```
dlubarov, line 149: This is the same poly as the two fields above (at a different point) right? If so I think the names should be similar.

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+            looking_columns,
+            looked_table,
+            looked_columns,
+        }
+    }
+}
+
+/// Lookup data for one table.
+#[derive(Clone)]
+pub struct CtlData<F: Field> {
```
dlubarov, line 49: I didn't really get the meaning of "control" in this context, do you think PermutationData or LookupData can work, or since both are not 100% accurate (per the other comment), maybe GrandProductData?

starky2/src/stark.rs
wborgeaud: I guess in some designs we would have tables that aren't connected (via a lookup) to any other table. For example, an ECDSA table would be disconnected from the EVM tables (assuming the ECDSA table only verify the tx signature and that it doesn't use a lookup to e.g. a bignum table). Though in this case we could also just compute the proof as a regular `starky` proof.
What do you think? Do you think we should assume that every STARK in `AllStark` is part of a cross-table lookup?

starky2/src/cross_table_lookup.rs
wborgeaud: Hmm I think permutation might be confusing since the tables don't have the same dimensions.
Technically, it's a multiset subset check `looked ⊂ looking` + a check ` looking \ looked = default^(n-m)`. So maybe "cross-table multiset check` ? 
Also, right now only one-to-one lookups are supported, but in the future it might be useful to support many-to-one lookups. For example, having many tables use a single bignum table for their bignum arithmetic. In that context, I think "lookup" makes more sense.

starky2/src/proof.rs
wborgeaud: So this is the openings of the cross-table lookups `Z` polynomials at the last point of the trace, whereas the openings above are for the permutation + cross-table lookups `Z` polynomials.
That's why I only kept the `ctl_zs` part of the name `permutation_ctl_zs`.

starky2/src/cross_table_lookup.rs
wborgeaud: Oh sorry I should have made this explicit somewhere, CTL is the acronym of Cross-Table Lookup. I used this to make names less verbose, like `permutation_cross_table_lookup_zs` -> `permutation_ctl_zs`.

starky2/src/stark.rs
dlubarov: Hm yeah I guess we could use starky if we did have an "isolated" table. Also I think we'll probably want to connect ECDSA to the CPU table, to support ecrecover and maybe a variable number of txns in one proof.

starky2/src/cross_table_lookup.rs
dlubarov: Hm ok seems reasonable. I guess lookup doesn't have to mean that the table contains no duplicates. Let's keep lookup/CTL then.

starky2/src/proof.rs
dlubarov: Oh makes sense now, I just didn't know to read it as permutation *and* ctl. The names seem fine then, I would just comment to make the "and" and the meaning of ctl clear.

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+    for i in 0..degree {
+        partial_prod *=
+            gamma + reduce_with_powers(columns.iter().map(|&j| &trace[j].values[i]), beta);
+        res.push(partial_prod);
+    }
+    res.into()
+}
+
+#[derive(Clone)]
+pub struct CTLCheckVars<'a, F, FE, P, const D2: usize>
```
dlubarov, line 138: nit: I think acronyms usually aren't capitalized in Rust, and `CtlCheckVars` would match `CtlData`

starky2/src/stark.rs
wborgeaud: Makes sense, so do you think we should assume `num_ctl_zs > 0` or keep the current level of generality?

starky2/src/proof.rs
wborgeaud: Sounds good, I added some comments 👍

starky2/src/stark.rs
dlubarov: Hm it seems like assuming it could lead to several small simplifications (like removing a few conditions and unwraps in prover), so I think it's probably worth doing unless/until a use case comes up where we wanted an isolated table (I can't think of one right now).

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+            looked_table,
+            looked_columns,
+        }
+    }
+}
+
+/// Cross-table lookup data for one table.
+#[derive(Clone)]
+pub struct CtlData<F: Field> {
+    // Challenges used in the argument.
```
dlubarov, line 50: nit: `//` -> `///`

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,294 @@
...
+    {
+        let looking_degree = 1 << degrees_bits[looking_table as usize];
+        let looked_degree = 1 << degrees_bits[looked_table as usize];
+        let looking_z = *ctl_zs_openings[looking_table as usize].next().unwrap();
+        let looked_z = *ctl_zs_openings[looked_table as usize].next().unwrap();
+        ensure!(
+            looking_z
+                == looked_z
+                    * challenges.challenges[i % config.num_challenges]
+                        .gamma
```
dlubarov, line 287: It looks like the default tuple is the all-zero tuple here? I think this can work, but will lead to a couple complications for tables like Keccak, where `0 -> 0` isn't a valid input-output pair.

- We won't be able to pad the Keccak table to a power of two with `default`, since it won't satisfy constraints (unless we suppress them)
- On CPU rows where Keccak is not performed, we would need constraints to enforce that `default` data is present, to ensure that rows where Keccak is performed can only contain valid (non-`default`) data

So I think it's probably easiest to define a default tuple for each CTL.

starky2/src/stark.rs
wborgeaud: Sounds good, I'll make the change.

starky2/src/cross_table_lookup.rs
wborgeaud: Oh yeah I initially put zeros with the plan to change it afterwards, but forgot to make the change. I agree it's definitely better to have a default value 👍

starky2/src/all_stark.rs
```diff
@@ -0,0 +1,192 @@
...
+
+#[derive(Copy, Clone)]
+pub struct CpuStark<F, const D: usize> {
+    #[allow(dead_code)]
+    num_rows: usize,
+    f: PhantomData<F>,
+}
+
+#[derive(Copy, Clone)]
+pub struct KeccakStark<F, const D: usize> {
```
dlubarov, line 38: Would you mind moving these to `cpu` and `keccak` modules/directories, as per our sync? Or we can do it as a followup

starky2/src/all_stark.rs
```diff
@@ -0,0 +1,192 @@
...
+            self.cpu_stark.num_permutation_batches(config),
+            self.keccak_stark.num_permutation_batches(config),
+        ]
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct CpuStark<F, const D: usize> {
+    #[allow(dead_code)]
+    num_rows: usize,
```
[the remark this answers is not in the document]
wborgeaud: Good point, I'll remove these fields 👍

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,299 @@
...
+                    .iter()
+                    .skip(num_permutation)
+                    .zip(
+                        p.proof
+                            .openings
+                            .permutation_ctl_zs_right
+                            .iter()
+                            .skip(num_permutation),
+                    )
+            })
```
dlubarov, line 189: nit: I think it'd be easier to read with a few intermediate vars -

```suggestion
            .map(|(p, &num_perms)| {
                let openings = &p.proof.openings;
                let ctl_zs = openings.permutation_ctl_zs.iter().skip(num_perms);
                let ctl_zs_right = openings.permutation_ctl_zs_right.iter().skip(num_perms);
                ctl_zs.zip(ctl_zs_right)
            })
```

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,299 @@
...
+            .collect::<Vec<_>>();
+
+        cross_table_lookups
+            .iter()
+            .fold(vec![vec![]; proofs.len()], |mut acc, ctl| {
```
dlubarov, line 194: nit: I feel like it's a little unusual to use fold when the accumulator is being mutated. A for-loop is a tad less code and I like that there's indentation for the large body. Maybe it's debatable though.

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+pub fn cross_table_lookup_data<F: RichField, C: GenericConfig<D, F = F>, const D: usize>(
+    config: &StarkConfig,
+    trace_poly_values: &[Vec<PolynomialValues<F>>],
+    cross_table_lookups: &[CrossTableLookup<F>],
+    challenger: &mut Challenger<F, C::Hasher>,
+) -> Vec<CtlData<F>> {
+    let challenges = get_grand_product_challenge_set(challenger, config.num_challenges);
+    cross_table_lookups.iter().fold(
```
dlubarov, line 87: nit: likewise here, to me it seems a little harder to read a fold (vs a for loop) with mutations and a large body

starky2/src/cross_table_lookup.rs
```diff
@@ -0,0 +1,299 @@
...
+    columns: &[usize],
+    beta: F,
+    gamma: F,
+) -> PolynomialValues<F> {
+    let mut partial_prod = F::ONE;
+    let degree = trace[0].len();
+    let mut res = Vec::with_capacity(degree);
+    for i in 0..degree {
+        partial_prod *=
+            gamma + reduce_with_powers(columns.iter().map(|&j| &trace[j].values[i]), beta);
```
dlubarov, line 144: I think it may be nice to have a combine method on `GrandProductChallenge`, since we do this in a few places, and it would save us from unpacking `GrandProductChallenge` into its parts

starky2/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+#![allow(incomplete_features)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::type_complexity)]
+#![feature(generic_const_exprs)]
+
+pub mod config;
+pub mod constraint_consumer;
+pub mod cross_table_lookup;
+mod get_challenges;
+// pub mod mock_stark;
```
dlubarov, line 10: Is this a work in progress?

starky2/src/mock_stark.rs
```diff
@@ -0,0 +1,362 @@
...
+    ) {
+        todo!()
+    }
+
+    fn constraint_degree(&self) -> usize {
+        2
+    }
+}
+
+#[cfg(test)]
```
dlubarov, line 231: Seems like the whole `mock_stark` module should probably be `#[cfg(test)]`, or in examples or what not

dlubarov:
General nit: `_recursively` -> `_circuit`

starky2/src/get_challenges.rs
```diff
@@ -0,0 +1,367 @@
...
+            stark,
+            trace_cap,
+            permutation_zs_cap.as_ref(),
+            quotient_polys_cap,
+            openings,
+            commit_phase_merkle_caps,
+            final_poly,
+            *pow_witness,
+            config,
+        )
```
dlubarov, line 257: I think it may be a bit nicer to inline `get_challenges_target`, to avoid the extra function with the long param list. If we skip the destructuring (can keep it if you prefer though), the combined function is only slightly longer than `get_challenges_target`, overall it saves 48 lines -

```suggestion
        let proof = &self.proof;
        let opening_proof = &proof.opening_proof;
        let num_challenges = config.num_challenges;

        let mut challenger = RecursiveChallenger::<F, C::Hasher, D>::new(builder);

        challenger.observe_cap(&proof.trace_cap);

        let permutation_challenge_sets =
            proof.permutation_zs_cap.as_ref().map(|permutation_zs_cap| {
                let tmp = get_n_permutation_challenge_sets_target(
                    builder,
                    &mut challenger,
                    num_challenges,
                    stark.permutation_batch_size(),
                );
                challenger.observe_cap(permutation_zs_cap);
                tmp
            });

        let stark_alphas = challenger.get_n_challenges(builder, num_challenges);

        challenger.observe_cap(&proof.quotient_polys_cap);
        let stark_zeta = challenger.get_extension_challenge(builder);

        challenger.observe_openings(&proof.openings.to_fri_openings());

        StarkProofChallengesTarget {
            permutation_challenge_sets,
            stark_alphas,
            stark_zeta,
            fri_challenges: challenger.fri_challenges::<C>(
                builder,
                &opening_proof.commit_phase_merkle_caps,
                &opening_proof.final_poly,
                opening_proof.pow_witness,
                &config.fri_config,
            ),
        }
```

starky2/src/get_challenges.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+}
+
+// TODO: Deal with the compressed stuff.
```
dlubarov, line 261: I think we can omit any compression code since we'll wrap these proofs with recursive proofs

dlubarov (approved):
LGTM otherwise! It looks like this is mostly finished, but just the multi-table recursive verifier and example isn't finished yet? I think we can merge though and add those things later.

starky2/src/cross_table_lookup.rs
wborgeaud: I wanted the functional style points :P But I agree the for-loop is better, I've made change 👍

starky2/src/cross_table_lookup.rs
wborgeaud: Good idea! I've added it

starky2/src/lib.rs
wborgeaud: It's a relic from the beginning of the branch, when I wanted to do things generically. I've removed it.
It's going to be a bit tricky to do testing with the non-generic approach. Like right now I've just added dummy columns to the CPU and Keccak STARKs, but once we add real columns and constraints, the test will fail if just one of the table still has TODOs.

starky2/src/mock_stark.rs
wborgeaud: I removed `mock_stark`.

starky2/src/get_challenges.rs
wborgeaud: Yeah I agree. I think it might be nice to add it at some point to `starky`, just for completeness if someone else wants to use the library.

starky2/src/get_challenges.rs
wborgeaud: Good idea! I've done the same for the native version. Having the separate `get_challenges` function is useful for compression, but since we won't need it I agree inlining is nice.

wborgeaud (commented):
> LGTM otherwise! It looks like this is mostly finished, but just the multi-table recursive verifier and example isn't finished yet? I think we can merge though and add those things later.

Yeah I'll add the recursive stuff in the next PR. For the example, I've removed `mock_stark` and think some more on how to best do testing.

starky2/src/lib.rs
dlubarov: Makes sense, we can just test with the real thing. I think we can use TODO comments (with no code) rather than `todo!()` etc. so that tests will generally be green, unless we add constraints before witness generation or what not.

PR #546, opened 2022-05-17
Document gate layouts

wborgeaud:
Document wire, constant and constraint layouts for every gate.

### DIFF
--- a/plonky2/src/gates/arithmetic_base.rs
+++ b/plonky2/src/gates/arithmetic_base.rs
@@ -16,8 +16,27 @@ use crate::plonk::vars::{
     EvaluationVarsBasePacked,
 };
 
-/// A gate which can perform a weighted multiply-add, i.e. `result = c0 x y + c1 z`. If the config
+/// A gate which can perform a weighted multiply-add, i.e. `output = c0 x y + c1 z`. If the config
 /// supports enough routed wires, it can support several such operations in one gate.
+//
+// Layout:
+//
+// +-----------+
+// | Constants |
+// +-----+-----+
+// |  c0 | c1  |
+// +-----+-----+
+//
+// +-------+---------------------------+
+// |       |        Routed wires       |
+// +-------+----+------+------+--------+
+// | Wires |  x |   y  |   z  | output |
+// +-------+----+------+------+--------+
+// | Index | 4i | 4i+1 | 4i+2 |  4i+3  |
+// +-------+----+------+------+--------+
+//
+// Constraints:
+// - `output - (c0 * x * y + c1 * z)`
 #[derive(Debug, Clone)]
 pub struct ArithmeticGate {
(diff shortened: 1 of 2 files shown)


plonky2/src/gates/arithmetic_base.rs
```diff
@@ -16,8 +16,27 @@ use crate::plonk::vars::{
...
+// +-------+---------------------------+
+// |       |        Routed wires       |
+// +-------+----+------+------+--------+
+// | Wires |  x |   y  |   z  | output |
+// +-------+----+------+------+--------+
+// | Index | 4i | 4i+1 | 4i+2 |  4i+3  |
+// +-------+----+------+------+--------+
+//
+// Constraints:
+// - `output - (c0 * x * y + c1 * z)`
```
dlubarov, line 39: nit: I'd add `= 0` to be extra clear

plonky2/src/gates/arithmetic_base.rs
```diff
@@ -16,8 +16,27 @@ use crate::plonk::vars::{
...
+// +-----+-----+
+// |  c0 | c1  |
+// +-----+-----+
+//
+// +-------+---------------------------+
+// |       |        Routed wires       |
+// +-------+----+------+------+--------+
+// | Wires |  x |   y  |   z  | output |
+// +-------+----+------+------+--------+
+// | Index | 4i | 4i+1 | 4i+2 |  4i+3  |
```
dlubarov, line 35: It's nice and simple here, but the indices might be a bit complicated to write out for things like `InterpolationGate`, and I'm not sure if it's that useful? (Maybe somewhat useful for auditors, but not sure since they'd still need to check that the code matches.)

What do you think of doing something similar to this old [post](https://mirprotocol.org/blog/Fast-recursive-arguments-based-on-Plonk-and-Halo) (under `Explicit formulae`)? I.e.

> Constants:
> - `c0`
> - `c1`
>
> Routed wires:
> - `x_i`
> - `y_i`
> - `z_i`
> - `output_i`
>
> Constraints:
> - `output_i - (c0 * x_i * y_i + c1 * z_i) = 0`

wborgeaud: Hmm I'm not sure I see the difference between bullet points vs table form. I agree that the indices might be a bit complex for some gates, but hopefully once they're written down they won't change. 
I think it could definitely be useful for auditors, but also for external people wanting to get a quick overview of our gates' layout and constraints, without having to parse any code. It can also be useful for us when debugging.

dlubarov: Sorry for the delay. Ok we can include that info if you think it's useful 👍

pgebheim:
Hello!

wborgeaud:
Done in #692.

PR #549, opened 2022-05-19
Make some `evm` structs more generic

wborgeaud:
Make some `evm` structs more generic to reduce friction when adding new tables.

### DIFF
--- a/evm/src/get_challenges.rs
+++ b/evm/src/get_challenges.rs
@@ -1,3 +1,4 @@
+use itertools::izip;
 use plonky2::field::extension_field::Extendable;
 use plonky2::fri::proof::FriProof;
 use plonky2::hash::hash_types::RichField;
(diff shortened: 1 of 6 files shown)


evm/src/get_challenges.rs
```diff
@@ -52,10 +52,11 @@ where
     C: GenericConfig<D, F = F>,
 {
     /// Computes all Fiat-Shamir challenges used in the STARK proof.
-    pub(crate) fn get_challenges<S: Stark<F, D>>(
+    pub(crate) fn get_challenges(
         &self,
         challenger: &mut Challenger<F, C::Hasher>,
-        stark: &S,
+        // Should correspond to `(stark.use_permutation_args(), stark.permutation_batch_size())`.
+        stark_permutation_info: (bool, usize),
```
dlubarov, line 59: nit: I think it'd be slightly easier to read with separate `uses_permutation_args: bool, permutation_batch_size: usize`, then it'd be self-commenting

dlubarov (approved):
LGTM

evm/src/get_challenges.rs
wborgeaud: Good point, done

PR #550, opened 2022-05-19
Add test to check that native and circuit STARK constraints are consistent

wborgeaud:


evm/src/cpu/cpu_stark.rs
```diff
@@ -44,3 +44,39 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for CpuStark<F, D
...
+#[cfg(test)]
+mod tests {
+    use anyhow::Result;
+    use plonky2::plonk::config::{GenericConfig, PoseidonGoldilocksConfig};
+
+    use crate::cpu::cpu_stark::CpuStark;
+    use crate::stark_testing::{test_stark_circuit_constraints, test_stark_low_degree};
+
+    #[test]
+    #[should_panic] // TODO: fix this when constraints are no longer all 0.
```
dlubarov, line 57: nit: maybe best to `#[ignore]` so it shows as ignored rather than passing

wborgeaud: Good point, done

dlubarov (approved):
Sorry for the delay, it LGTM

PR #551, opened 2022-05-27
`evm` recursive verifier

wborgeaud:
Implement circuit for the `evm` verifier.

### DIFF
--- a/evm/src/permutation.rs
+++ b/evm/src/permutation.rs
@@ -190,7 +190,7 @@ pub(crate) fn get_n_grand_product_challenge_sets<F: RichField, H: Hasher<F>>(
         .collect()
 }
 
-fn get_permutation_challenge_target<
+fn get_grand_product_challenge_target<
     F: RichField + Extendable<D>,
     H: AlgebraicHasher<F>,
     const D: usize,
@@ -203,7 +203,7 @@ fn get_permutation_challenge_target<
     GrandProductChallenge { beta, gamma }
 }
 
-fn get_permutation_challenge_set_target<
+pub(crate) fn get_grand_product_challenge_set_target<
     F: RichField + Extendable<D>,
     H: AlgebraicHasher<F>,
     const D: usize,
@@ -213,12 +213,12 @@ fn get_permutation_challenge_set_target<
     num_challenges: usize,
 ) -> GrandProductChallengeSet<Target> {
     let challenges = (0..num_challenges)
-        .map(|_| get_permutation_challenge_target(builder, challenger))
+        .map(|_| get_grand_product_challenge_target(builder, challenger))
         .collect();
     GrandProductChallengeSet { challenges }
 }
 
-pub(crate) fn get_n_permutation_challenge_sets_target<
+pub(crate) fn get_n_grand_product_challenge_sets_target<
     F: RichField + Extendable<D>,
     H: AlgebraicHasher<F>,
     const D: usize,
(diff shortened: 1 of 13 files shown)


evm/src/permutation.rs
```diff
@@ -68,6 +68,32 @@ impl<F: Field> GrandProductChallenge<F> {
...
+impl GrandProductChallenge<Target> {
+    pub(crate) fn combine_circuit<F: RichField + Extendable<D>, const D: usize>(
+        &self,
+        builder: &mut CircuitBuilder<F, D>,
+        terms: &[ExtensionTarget<D>],
+    ) -> ExtensionTarget<D> {
+        let zero = builder.zero();
+        let mut factor = ReducingFactorTarget::new(self.beta.to_ext_target(zero));
+        let reduced = factor.reduce(terms, builder);
+        builder.add_extension(reduced, self.gamma.to_ext_target(zero))
```
dlubarov, line 80: nit: a tad simpler

```suggestion
        let reduced = reduce_with_powers_ext_circuit(terms, self.beta);
        let gamma = builder.convert_to_ext(self.gamma);
        builder.add_extension(reduced, gamma)
```

evm/src/permutation.rs
```diff
@@ -68,6 +68,32 @@ impl<F: Field> GrandProductChallenge<F> {
...
+    }
+
+    pub(crate) fn combine_base_circuit<F: RichField + Extendable<D>, const D: usize>(
+        &self,
+        builder: &mut CircuitBuilder<F, D>,
+        terms: &[Target],
+    ) -> Target {
+        let zero = builder.zero();
+        let combination = terms
```
dlubarov, line 89: Could use `reduce_base`, to save a little code and potentially use `ReducingGate` in some cases

evm/src/permutation.rs
dlubarov: Or maybe better to stick with `ReducingFactorTarget` if you take the suggestion below, so that the two functions can be similar.

evm/src/permutation.rs
wborgeaud: Hm so `reduce_base` takes a reducing factor that is an `ExtensionTarget`, whereas here the factor `self.beta` is just a `Target`. `reduce_base` will either use `ArithmeticExtensionGate` operations or add a new `ReducingGate`.
I think if the number of terms is low (like `< num_routed_wires/4`), using `ArithmeticGate` operation is cheaper. I think I'll add a util function for this kind of reduction.

plonky2/src/plonk/plonk_common.rs
```diff
@@ -138,6 +139,25 @@ where
...
+    terms: &[Target],
+    alpha: Target,
+) -> Target {
+    if terms.len() <= ArithmeticGate::new_from_config(&builder.config).num_ops + 1 {
+        terms
+            .iter()
+            .rev()
+            .fold(builder.zero(), |acc, &t| builder.mul_add(alpha, acc, t))
+    } else {
+        let terms_ext = terms
```
dlubarov, line 153: nit: `reduce_with_powers_ext_circuit` will lead to a `ReducingFactorTarget::reduce` call, but seems like it could be preferable to call `ReducingFactorTarget::reduce_base` so that it can potentially use `ReducingGate` rather than `ReducingExtensionGate`

dlubarov (approved):
LGTM!

PR #552, opened 2022-06-01
Multi-table CTL

wborgeaud:
Implement multi-table CTL allowing a table to be looked up from several other tables.

### DIFF
--- a/plonky2/src/gadgets/arithmetic.rs
+++ b/plonky2/src/gadgets/arithmetic.rs
@@ -25,7 +25,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
 
     /// Computes `x^3`.
     pub fn cube(&mut self, x: Target) -> Target {
-        self.mul_many(&[x, x, x])
+        self.mul_many([x, x, x])
     }
 
     /// Computes `const_0 * multiplicand_0 * multiplicand_1 + const_1 * addend`.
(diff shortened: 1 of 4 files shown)


plonky2/src/gadgets/arithmetic.rs
```diff
@@ -206,12 +206,16 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
-            .iter()
-            .copied()
-            .reduce(|acc, t| self.mul(acc, t))
-            .unwrap_or_else(|| self.one())
+    pub fn mul_many<T>(&mut self, terms: impl IntoIterator<Item = T>) -> Target
+    where
+        T: Borrow<Target>,
+    {
+        let mut iter = terms.into_iter();
+        if let Some(first) = iter.next() {
```
dlubarov, line 214: I think it's also fine to start with an acc of 1 since `1 * first` should be optimized away, though we can keep this style if you prefer it

dlubarov (approved):
LGTM

plonky2/src/gadgets/arithmetic.rs
wborgeaud: Good point, done.

Btw, what would you think of changing all the other `mul_many_extension, add_many,...` functions to also take a generic iterator instead of a slice? I think it would be more versatile and possibly save a few `collect()`s.

wborgeaud:
> Not really about this PR, but I notice we're using `_next` and `_right`, with the same meaning it seems. We should probably standardize it. I think `_next` seems more fitting since it's a "vertical" offset.

Yeah I agree `next` makes more sense. I can make a PR to standardize this 👍

plonky2/src/gadgets/arithmetic.rs
dlubarov: Sounds good to me! I think the only downside is that the signatures in our docs become a bit harder to read, but probably worth it for ease of calling.

PR #556, opened 2022-06-03
Change some functions to take iterators instead of slices

wborgeaud:


plonky2/src/gadgets/arithmetic_extension.rs
```diff
@@ -301,12 +307,16 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
     }
 
     /// Multiply `n` `ExtensionTarget`s.
-    pub fn mul_many_extension(&mut self, terms: &[ExtensionTarget<D>]) -> ExtensionTarget<D> {
-        terms
-            .iter()
-            .copied()
-            .reduce(|acc, t| self.mul_extension(acc, t))
```
dlubarov, line 308: I think `fold` was previously changed to `reduce`, as an optimization to avoid the `1 * t` step. Probably not very important but might be best to keep `reduce` if it can work with the updated API?

dlubarov (approved):
LGTM

plonky2/src/gadgets/arithmetic_extension.rs
dlubarov: My mistake, I was thinking this was the native function. I agree this is fine (and slightly nicer) for circuit arithmetic.

PR #557, opened 2022-06-06
Keccak Stark

npwardberkeley:


evm/src/keccak/registers.rs
```diff
@@ -0,0 +1,252 @@
...
+    0x8000000000000080,
+    0x000000000000800A,
+    0x800000008000000A,
+    0x8000000080008081,
+    0x8000000000008080,
+    0x0000000080000001,
+    0x8000000080008008,
+];
+
+const RC_BITS: [[u8; 64]; 24] = [
```
dlubarov, line 44: Would you mind moving these into a `constants` module? Would prefer if we keep `registers` just for register indices

evm/src/all_stark.rs
```diff
@@ -85,31 +87,40 @@ mod tests {
...
-
-        let default = vec![F::ZERO; 2];
-        let cross_table_lookups = vec![CrossTableLookup {
-            looking_tables: vec![Table::Cpu],
-            looking_columns: vec![vec![cpu::columns::OPCODE]],
-            looked_table: Table::Keccak,
-            looked_columns: vec![keccak_looked_col],
+        let cross_table_lookups = vec![CrossTableLookup::new(
+            vec![Table::Cpu],
+            vec![vec![cpu::columns::OPCODE]],
```
dlubarov, line 119: Let's just add a TODO that this is a temporary thing until CTL filters are working

evm/src/keccak/registers.rs
npwardberkeley: Ah yes, good idea!

evm/src/all_stark.rs
npwardberkeley: Ah what exactly is temporary? The use of `default`?

dlubarov: I mean that the `OPCODE` register won't be part of the actual CTL right?

evm/src/keccak/keccak_stark.rs
```diff
@@ -1,60 +1,522 @@
...
+    /// Generate the rows of the trace. Note that this does not generate the permuted columns used
+    /// in our lookup arguments, as those are computed after transposing to column-wise form.
+    pub(crate) fn generate_trace_rows(
+        &self,
+        inputs: Vec<[u64; INPUT_LIMBS]>,
+    ) -> Vec<[F; NUM_REGISTERS]> {
+        let num_rows = (inputs.len() * NUM_ROUNDS).next_power_of_two();
+        info!("{} rows", num_rows);
+        let mut rows = Vec::with_capacity(num_rows);
+        for input in inputs.iter().take(1) {
```
dlubarov, line 50: Why `take(1)`, was that for testing?

evm/src/all_stark.rs
npwardberkeley: Ah yes

evm/src/keccak/keccak_stark.rs
npwardberkeley: Oh right — thanks, good catch

evm/src/keccak/keccak_stark.rs
```diff
@@ -1,60 +1,522 @@
...
+                    .collect();
+
+                for z in 0..64 {
+                    next_row[reg_a(x, y, z)] = F::from_canonical_u64(bit_values[z]);
+                }
+            }
+        }
+    }
+
+    fn generate_trace_rows_for_round(&self, row: &mut [F; NUM_REGISTERS], round: usize) {
```
dlubarov, line 111: nit: seems like it should be singular "row" (sorry if it's my typo)

npwardberkeley: Ah yes

evm/src/keccak/keccak_stark.rs
```diff
@@ -1,60 +1,522 @@
...
+    fn copy_output_to_input(
+        &self,
+        prev_row: [F; NUM_REGISTERS],
+        next_row: &mut [F; NUM_REGISTERS],
+    ) {
+        for x in 0..5 {
+            for y in 0..5 {
+                let cur_lo = prev_row[reg_a_prime_prime_prime(x, y)];
+                let cur_hi = prev_row[reg_a_prime_prime_prime(x, y) + 1];
+                let cur_u64 = cur_lo.to_canonical_u64() + (1 << 32) * cur_hi.to_canonical_u64();
```
dlubarov, line 95: nit: I think it's slightly more idiomatic to combine them with logic operators like `cur_lo.to_canonical_u64() | (cur_hi.to_canonical_u64() << 32)`

dlubarov: And optionally `*acc % 2` -> `*acc & 1`, `*acc /= 2` -> `*acc >>= 1`, but I guess it's just a matter of preference, fine to keep as is if you prefer

npwardberkeley: Ah yes that looks better!

npwardberkeley: Hm in that case should I also replace `val % (1 << 32)` with `val & (1 << 32 - 1)`?

dlubarov: Hm up to you. Another option is `val as u32 as u64`

npwardberkeley: Oh interesting. I think I'll just go with the bit-logic one

PR #558, opened 2022-06-07
Filtered CTL

wborgeaud:
Add option to add filters in the CTL argument. Filters have the form `filter_columns: Vec<usize>` which is a vector of columns `[c_1,...,c_k]` used as a filter using the sum `c_1 + ... + c_k`. An empty vector corresponds to no filter.


evm/src/all_stark.rs
```diff
@@ -87,40 +86,74 @@ mod tests {
         let keccak_stark = KeccakStark::<F, D> {
             f: Default::default(),
         };
-        let keccak_rows = (2 * NUM_ROUNDS + 1).next_power_of_two();
-        let keccak_looked_col = 3;
 
-        let mut rng = ChaCha8Rng::seed_from_u64(0x6feb51b7ec230f25);
+        let mut rng = thread_rng();
         let num_inputs = 2;
```
dlubarov, line 91: Can we call this something like `num_keccak_perms`? A bit confusing with this and `NUM_INPUTS`

evm/src/cpu/columns.rs
```diff
@@ -132,4 +132,116 @@ pub const OPCODE_BITS: [usize; 8] = [
     END_INSTRUCTION_FLAGS + 7,
 ];
 
-pub const NUM_CPU_COLUMNS: usize = OPCODE_BITS[OPCODE_BITS.len() - 1] + 1;
+/// Filter. 1 iff a Keccak permutation is computed on this row.
+pub const IS_KECCAK: usize = OPCODE_BITS[OPCODE_BITS.len() - 1] + 1;
+
+pub const START_KECCAK_INPUT: usize = IS_KECCAK + 1;
+#[allow(dead_code)] // TODO: Remove when used
+pub const KECCAK_INPUT_LIMBS: [usize; 50] = [
```
dlubarov, line 140: Could these be ranges?

evm/src/cross_table_lookup.rs
```diff
@@ -21,33 +22,61 @@ use crate::proof::{StarkProofWithPublicInputs, StarkProofWithPublicInputsTarget}
...
+    columns: Vec<usize>,
+    /// Vector of columns `[c_1,...,c_k]` used as a filter using the sum `c_1 + ... + c_k`.
+    /// An empty vector corresponds to no filter.
+    filter_columns: Vec<usize>,
+}
+
+impl TableWithColumns {
+    pub fn new(table: Table, columns: Vec<usize>, filter_columns: Vec<usize>) -> Self {
+        debug_assert_eq!(
+            filter_columns.iter().collect::<HashSet<_>>().len(),
```
dlubarov, line 37: nit: could use itertools, `filter_columns.iter().unique().count()`

evm/src/cross_table_lookup.rs
```diff
@@ -21,33 +22,61 @@ use crate::proof::{StarkProofWithPublicInputs, StarkProofWithPublicInputsTarget}
...
-        assert!(looking_columns
+        assert!(looking_tables
             .iter()
-            .all(|cols| cols.len() == looked_columns.len()));
-        assert!(default.len() == looked_columns.len());
+            .all(|twc| twc.columns.len() == looked_table.columns.len()));
+        assert!(
+            looking_tables
+                .iter()
+                .all(|twc| twc.filter_columns.is_empty())
```
dlubarov, line 69: nit: might be best to move the `== default.is_some()` inside `all`, so that we do the check for each looking table.

evm/src/cross_table_lookup.rs
```diff
@@ -143,13 +188,25 @@ pub fn cross_table_lookup_data<F: RichField, C: GenericConfig<D, F = F>, const D
...
+        } else {
+            filter_columns.iter().map(|&j| trace[j].values[i]).sum()
+        };
+        partial_prod *= if filter.is_zero() {
+            F::ONE
+        } else if filter.is_one() {
+            challenge.combine(columns.iter().map(|&j| &trace[j].values[i]))
+        } else {
+            panic!("Non-binary filter?")
+        };
```
dlubarov, line 209: Since this might be slightly performance sensitive, might be best to avoid multiplication by 1 like
```suggestion
        if filter.is_one() {
            partial_prod *= challenge.combine(columns.iter().map(|&j| &trace[j].values[i]));
        } else {
            assert_eq!(filter, F::ZERO, "Non-binary filter?");
        }
```

dlubarov:
What do you think of removing default later? I'm not sure if we'll still have a use for it; at least in the CPU table we'll have flags for each operation so it seems easiest to use filters. We could wait and see though.

dlubarov (approved):
LGTM!

wborgeaud:
> What do you think of removing default later? I'm not sure if we'll still have a use for it; at least in the CPU table we'll have flags for each operation so it seems easiest to use filters. We could wait and see though.

Yeah my idea was to keep `default` for now until we've finished designing all the tables and CTLs, and if then all of them use filters we can remove `default`.

PR #561, opened 2022-06-09
Keccak input registers

wborgeaud:
Add input registers to Keccak for wiring with Cpu table.

### DIFF
--- a/evm/src/keccak/keccak_stark.rs
+++ b/evm/src/keccak/keccak_stark.rs
@@ -17,7 +17,7 @@ use crate::keccak::logic::{
 };
 use crate::keccak::registers::{
     reg_a, reg_a_prime, reg_a_prime_prime, reg_a_prime_prime_0_0_bit, reg_a_prime_prime_prime,
-    reg_b, reg_c, reg_c_partial, reg_step, NUM_REGISTERS,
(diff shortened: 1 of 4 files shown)


evm/src/keccak/keccak_stark.rs
```diff
@@ -565,16 +608,10 @@ mod tests {
...
         let base = F::from_canonical_u64(1 << 32);
-        for x in 0..5 {
-            for y in 0..5 {
-                output.push(
-                    last_row[reg_a_prime_prime_prime(x, y)]
-                        + base * last_row[reg_a_prime_prime_prime(x, y) + 1],
-                );
-            }
-        }
+        let output = (0..INPUT_LIMBS)
```
npwardberkeley, line 612: Should we change `INPUT_LIMBS` --> `NUM_LIMBS`? Since it's equally the number of outputs

npwardberkeley (approved):
Looks good1

evm/src/keccak/keccak_stark.rs
wborgeaud: What do you think about `NUM_INPUTS`, to avoid confusion since there are `2*INPUT_LIMBS` 32-bit limbs?

npwardberkeley:
Ah yes, that makes sense!

PR #562, opened 2022-06-10
Logic stark

nbgl:
I’ll connect this to the CPU table in a separate PR to avoid merge conflicts.

### DIFF
--- a/evm/src/logic.rs
+++ b/evm/src/logic.rs
@@ -1,0 +1,304 @@
+use std::marker::PhantomData;
+
+use plonky2::field::extension_field::{Extendable, FieldExtension};
+use plonky2::field::packed_field::PackedField;
+use plonky2::hash::hash_types::RichField;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::stark::Stark;
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+// Total number of bits per input/output.
+const VAL_BITS: usize = 256;
+// Number of bits stored per field element. Ensure that this fits; it is not checked.
+const PACKED_LIMB_BITS: usize = 16;
+// Number of field elements needed to store each input/output at the specified packing.
+const PACKED_LEN: usize = (VAL_BITS + PACKED_LIMB_BITS - 1) / PACKED_LIMB_BITS;
+// Number of bits in the last limb, which may be smaller than `PACKED_LIMB_BITS`.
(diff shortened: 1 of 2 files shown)


evm/src/logic.rs
```diff
@@ -0,0 +1,304 @@
...
+                    .take(PACKED_LIMB_BITS);
+                for (j, col) in limb_bits_cols.enumerate() {
+                    let bit = (limb >> j) & 1;
+                    lv[col] = F::from_canonical_u64(bit);
+                }
+            }
+        }
+
+        for (res_col, (limb_in0_col, limb_in1_col)) in
+            columns::RESULT.zip(columns::INPUT0_PACKED.zip(columns::INPUT1_PACKED))
```
dlubarov, line 83: nit: consider itertools' `izip!`

evm/src/logic.rs
```diff
@@ -0,0 +1,304 @@
...
+        for (packed_input_cols, bit_cols) in [
+            (columns::INPUT0_PACKED, columns::INPUT0_BITS),
+            (columns::INPUT1_PACKED, columns::INPUT1_BITS),
+        ] {
+            for (i, limb_col) in packed_input_cols.enumerate() {
+                let limb = lv[limb_col].to_canonical_u64();
+                let limb_bits_cols = bit_cols
+                    .clone()
+                    .skip(i * PACKED_LIMB_BITS)
+                    .take(PACKED_LIMB_BITS);
```
dlubarov, line 74: nit: might be nice to have a couple `input{0,1}_limb_bits(limb: usize)` methods, since this is a little verbose and we do it a few times

evm/src/logic.rs
```diff
@@ -0,0 +1,304 @@
...
+
+pub(crate) mod columns {
+    use std::ops::Range;
+
+    use super::{PACKED_LEN, VAL_BITS};
+
+    pub const IS_AND: usize = 0;
+    pub const IS_OR: usize = IS_AND + 1;
+    pub const IS_XOR: usize = IS_OR + 1;
+    pub const AND_COEFF: usize = IS_XOR + 1;
```
dlubarov, line 28: It seems like this isn't needed (since it's degree 1) and you didn't end up using it?

dlubarov (approved):
LGTM

evm/src/logic.rs
nbgl: Good catch! That’s a remnant of an old approach that I later dumped :)

evm/src/logic.rs
nbgl: Good idea!

PR #564, opened 2022-06-14
Linear combinations as columns in CTL 

wborgeaud:
h/t @nbgl for the idea.

Modify the CTL argument to allow lookups of linear combinations of columns.

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -46,7 +46,7 @@ impl<F: Field> Column<F> {
         debug_assert_eq!(
             v.iter().map(|(c, _)| c).unique().count(),
             v.len(),
-            "Duplicate filter columns."
+            "Duplicate columns."
         );
         Self::LinearCombination(v)
     }
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -1,3 +1,5 @@
+use std::iter::repeat;
+
 use anyhow::{ensure, Result};
 use itertools::Itertools;
 use plonky2::field::extension_field::{Extendable, FieldExtension};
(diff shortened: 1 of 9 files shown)


evm/src/cross_table_lookup.rs
```diff
@@ -21,42 +23,127 @@ use crate::proof::{StarkProofWithPublicInputs, StarkProofWithPublicInputsTarget}
...
+
+    pub fn linear_combination<I: IntoIterator<Item = (usize, F)>>(iter: I) -> Self {
+        Self::linear_combination_with_constant(iter, F::ZERO)
+    }
+
+    pub fn le_bits<I: IntoIterator<Item = usize>>(cs: I) -> Self {
+        Self::linear_combination(cs.into_iter().zip(F::TWO.powers()))
+    }
+
+    pub fn sum(cs: &[usize]) -> Self {
```
nbgl, line 70: Nit: How about `pub fn sum<I: IntoIterator<Item = usize>>(cs: I) -> Self {`? Would be slightly more flexible.

evm/src/cross_table_lookup.rs
```diff
@@ -21,42 +23,127 @@ use crate::proof::{StarkProofWithPublicInputs, StarkProofWithPublicInputsTarget}
...
+    pub fn single(c: usize) -> Self {
+        Self {
+            linear_combination: vec![(c, F::ONE)],
+            constant: F::ZERO,
+        }
+    }
+
+    pub fn singles(cs: Vec<usize>) -> Vec<Self> {
+        cs.into_iter().map(Self::single).collect()
+    }
```
nbgl, line 43: I suggest making changing this signature to `singles<I: IntoIterator<Item = usize>>(cs: I) -> impl Iterator<Item = Self>`. It would be slightly more flexible (and saves you a `.collect()`).

nbgl (approved):
Looks good! Just some nits.

Although I should note that I’m not very familiar with this code, so this shouldn’t be the only review.

evm/src/cross_table_lookup.rs
```diff
@@ -293,14 +384,17 @@ pub(crate) fn eval_cross_table_lookup_checks<F, FE, P, C, S, const D: usize, con
             next_z,
             challenges,
             columns,
-            filter_columns,
+            filter_column,
         } = lookup_vars;
-        let combine = |v: &[P]| -> P { challenges.combine(columns.iter().map(|&i| &v[i])) };
+        let combine = |v: &[P]| -> P {
+            let evals = columns.iter().map(|c| c.eval(v)).collect::<Vec<_>>();
```
dlubarov, line 390: Can we skip collecting here?

dlubarov (approved):
LGTM also

evm/src/cross_table_lookup.rs
wborgeaud: I'm not sure. `challenge.combine` takes a `IntoIterator<Item = &'a P>` (same as `reduce_with_powers`). So if I try changing this to 
```rust
let evals = columns.iter().map(|c| &c.eval(v));
challenges.combine(evals)
```
I get a "cannot return reference to temporary value" error. One solution would be to change `reduce_with_powers` to take a `IntoIterator<Item = impl Borrow<P>>`. I've tried this and the downside is that we then have to add explicit generics to every call of `reduce_with_powers`. If you think it's worth it in terms of performance, that could be a solution.

dlubarov: Oh I see. I thought `Borrow` params could typically be inferred? But in any case feel free to merge as is, I can play with it later.

PR #565, opened 2022-06-14
NOT stark

nbgl:
The NOT instruction is super simple, so we can just do it in the CPU table.

It does not require any additional columns. The input/output columns have to be there anyway for all the other operations, so they are not actually additional.

I made `simple_logic` its own directory, because I also plan to put EQ and ISZERO there.

### DIFF
--- a/evm/src/cpu/simple_logic/not.rs
+++ b/evm/src/cpu/simple_logic/not.rs
@@ -1,0 +1,71 @@
+use plonky2::field::extension_field::Extendable;
+use plonky2::field::field_types::Field;
+use plonky2::field::packed_field::PackedField;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns;
+
+const LIMB_SIZE: usize = 16;
+const ALL_1_LIMB: u64 = (1 << LIMB_SIZE) - 1;
+
+pub fn generate<F: RichField>(lv: &mut [F; columns::NUM_CPU_COLUMNS]) {
+    let cycle_filter = lv[columns::IS_CPU_CYCLE].to_canonical_u64();
+    if cycle_filter == 0 {
+        return;
+    }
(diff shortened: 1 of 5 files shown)


evm/src/cpu/simple_logic/not.rs
```diff
@@ -0,0 +1,71 @@
...
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns;
+
+const LIMB_SIZE: usize = 16;
+const ALL_1_LIMB: u64 = (1 << LIMB_SIZE) - 1;
+
+pub fn generate<F: RichField>(lv: &mut [F; columns::NUM_CPU_COLUMNS]) {
+    let cycle_filter = lv[columns::IS_CPU_CYCLE].to_canonical_u64();
+    if cycle_filter == 0 {
```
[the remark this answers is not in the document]
nbgl: Good idea!

evm/src/cpu/simple_logic/not.rs
```diff
@@ -0,0 +1,65 @@
...
+    let is_not_filter = lv[columns::IS_NOT].to_canonical_u64();
+    if is_not_filter == 0 {
+        return;
+    }
+    assert_eq!(is_not_filter, 1);
+
+    for (input_col, output_col) in columns::SIMPLE_LOGIC_INPUT0.zip(columns::SIMPLE_LOGIC_OUTPUT) {
+        let input = lv[input_col].to_canonical_u64();
+        assert_eq!(input >> LIMB_SIZE, 0);
+        let output = input ^ ALL_1_LIMB;
```
dlubarov, line 23: nit: `!input` seems a bit more direct

dlubarov (approved):
LGTM

evm/src/cpu/simple_logic/not.rs
nbgl: That inverts all the bits though, whereas I want to invert the bottom `LIMB_SIZE` bits.

dlubarov: Oh right!

PR #566, opened 2022-06-14
EQ and ISZERO

nbgl:


evm/src/cpu/decode.rs
```diff
@@ -160,6 +163,10 @@ pub fn eval_packed_generic<P: PackedField>(
 ) {
     let cycle_filter = lv[columns::IS_CPU_CYCLE];
 
+    // If cycle_filter = 0, then we require is_eq = 0 and is_iszero = 0.
```
dlubarov, line 166: I'm not sure this is needed? If `cycle_filter = 0`, it seems like we could leave them unconstrained since they'd be unused

evm/src/cpu/simple_logic/eq_iszero.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    // Check SIMPLE_LOGIC_DIFF
+    let diffs = lv[columns::SIMPLE_LOGIC_DIFF];
+    let diffs_inv = lv[columns::SIMPLE_LOGIC_DIFF_INV];
+    {
+        let input0_sum: P = columns::SIMPLE_LOGIC_INPUT0.map(|i| lv[i]).sum();
+        yield_constr.constraint(iszero_filter * (diffs - input0_sum));
+
+        let sum_squared_diffs: P = columns::SIMPLE_LOGIC_INPUT0
```
dlubarov, line 75: Clever way to do this!

dlubarov (approved):
LGTM

evm/src/cpu/decode.rs
nbgl: They are used though: I have some degree-2 constraints, so the filter must be degree-1. I can’t filter by, e.g., `cycle_filter * is_eq` and must filter by just `is_eq` instead. That means that `is_eq` must always be `0` if we’re not executing a cycle. (`is_iszero` is analogous.)

PR #569, opened 2022-06-15
Connect logic stark to CPU

nbgl:


evm/src/all_stark.rs
```diff
@@ -146,31 +236,79 @@ mod tests {
...
+            let mut res = vec![
+                logic::columns::IS_AND,
+                logic::columns::IS_OR,
+                logic::columns::IS_XOR,
+            ];
+            res.extend(logic::columns::INPUT0_PACKED);
+            res.extend(logic::columns::INPUT1_PACKED);
+            res.extend(logic::columns::RESULT);
+            res
+        };
```
wborgeaud, line 261: Maybe we should have these be constructed in the relevant `columns` module, like `cpu::columns::logic_input_output()`. Also goes for the Keccak stuff, so if you want I can also do it in a later PR.

wborgeaud (approved):
Looks good to me!

evm/src/all_stark.rs
nbgl: Good idea! I’m happy to do it for both Keccak and logic—should be quick :)

PR #571, opened 2022-06-16
EVM memory stark

npwardberkeley:


evm/src/cpu/columns.rs
```diff
@@ -7,10 +7,12 @@ pub const IS_CPU_CYCLE: usize = 0;
 // If CPU cycle: The opcode being decoded, in {0, ..., 255}.
 pub const OPCODE: usize = IS_CPU_CYCLE + 1;
 
+pub const MEMORY_DUMMY: usize = OPCODE + 1;
```
dlubarov, line 10: Seems like you're using the real columns in the CTL now, so is this no longer needed?

evm/src/cpu/columns.rs
```diff
@@ -138,7 +140,7 @@ pub const OPCODE_BITS: [usize; 8] = [
 pub const IS_KECCAK: usize = OPCODE_BITS[OPCODE_BITS.len() - 1] + 1;
 
 pub const START_KECCAK_INPUT: usize = IS_KECCAK + 1;
-#[allow(dead_code)] // TODO: Remove when used
+// TODO: Remove when used
```
dlubarov, line 143: Can this be removed now?

evm/src/cpu/columns.rs
```diff
@@ -153,4 +155,40 @@ pub const SIMPLE_LOGIC_OUTPUT: Range<usize> = SIMPLE_LOGIC_INPUT1.end..SIMPLE_LO
...
 pub const SIMPLE_LOGIC_DIFF_INV: usize = SIMPLE_LOGIC_DIFF + 1;
 
-pub const NUM_CPU_COLUMNS: usize = SIMPLE_LOGIC_DIFF_INV + 1;
+const NUM_MEMORY_OPS: usize = 4;
+const NUM_MEMORY_VALUE_LIMBS: usize = 8;
+
+pub(crate) const MEMORY_TIMESTAMP: usize = SIMPLE_LOGIC_DIFF_INV + 1;
+
+const USES_MEMOP_START: usize = MEMORY_TIMESTAMP + 1;
+pub const fn uses_memop(op: usize) -> usize {
```
dlubarov, line 164: Please comment to explain briefly

evm/src/cpu/columns.rs
```diff
@@ -153,4 +155,40 @@ pub const SIMPLE_LOGIC_OUTPUT: Range<usize> = SIMPLE_LOGIC_INPUT1.end..SIMPLE_LO
...
+    debug_assert!(op < NUM_MEMORY_OPS);
```
dlubarov, line 165: nit: let's do this in the below methods also

starky/src/factorial_stark.rs
```diff
@@ -0,0 +1,244 @@
+use std::marker::PhantomData;
```
dlubarov, line 1: Hm do you know why this shows up in this PR?

plonky2/src/gadgets/curve_msm.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::gadgets::curve::AffinePointTarget;
+use crate::gadgets::nonnative::NonNativeTarget;
+use crate::hash::hash_types::RichField;
+use crate::plonk::circuit_builder::CircuitBuilder;
+
+const DIGITS_PER_CHUNK: usize = 80;
+
+const WINDOW_SIZE: usize = 4;
+
+pub struct MsmPrecomputationTarget<C: Curve> {
```
dlubarov, line 14: Was this accidentally added to this branch?

starky/src/factorial_stark.rs
dlubarov, line 1: Did you mean to include these? Maybe it's good to have, but just checking

evm/src/all_stark.rs
```diff
@@ -146,7 +180,30 @@ mod tests {
         keccak_keccak_input_output.extend(Column::singles(
             (0..2 * NUM_INPUTS).map(keccak::registers::reg_output_limb),
         ));
-        let cross_table_lookups = vec![CrossTableLookup::new(
+
+        let cpu_memory_values: Vec<Vec<_>> = (0..4)
```
dlubarov, line 184: nit: I'd say `cols` or `regs` instead of `values` (likewise for `memory_memory_values`)

evm/src/cpu/columns.rs
```diff
@@ -153,4 +155,40 @@ pub const SIMPLE_LOGIC_OUTPUT: Range<usize> = SIMPLE_LOGIC_INPUT1.end..SIMPLE_LO
 pub const SIMPLE_LOGIC_DIFF: usize = SIMPLE_LOGIC_OUTPUT.end;
 pub const SIMPLE_LOGIC_DIFF_INV: usize = SIMPLE_LOGIC_DIFF + 1;
 
-pub const NUM_CPU_COLUMNS: usize = SIMPLE_LOGIC_DIFF_INV + 1;
+const NUM_MEMORY_OPS: usize = 4;
+const NUM_MEMORY_VALUE_LIMBS: usize = 8;
+
+pub(crate) const MEMORY_TIMESTAMP: usize = SIMPLE_LOGIC_DIFF_INV + 1;
```
dlubarov, line 161: nit: I'd prefer to call it `CLOCK` to make it clear it's measured in cycles, and it might have other uses besides memory

evm/src/cpu/columns.rs
npwardberkeley: Oh yes!

evm/src/all_stark.rs
```diff
@@ -146,7 +180,30 @@ mod tests {
...
+            .map(|op| {
+                let mut vals = vec![
+                    cpu::columns::MEMORY_TIMESTAMP,
```
dlubarov, line 187: I think our plan was to pass something like `NUM_MEMORY_OPS * clock + op` here, so the `NUM_MEMORY_OPS` ops occur in a specific sequence. Maybe add a TODO to do that later?

The CTL code doesn't support linear combinations like that currently, but maybe it wouldn't be too hard to add since they're supported for filters. cc @wborgeaud

starky/src/factorial_stark.rs
npwardberkeley: Oooops, no I didn't

evm/src/all_stark.rs
dlubarov: Let's use `NUM_MEMORY_OPS` vs 4, since we might change it a few times

evm/src/all_stark.rs
npwardberkeley: Ohh I could try to work on that! I'd like to get my hands dirty with CTL stuff a little

evm/src/memory/registers.rs
```diff
@@ -0,0 +1,45 @@
+//! Memory unit.
```
dlubarov, line 1: Everything looks good here but I think some comments here would be useful, since readers might click through to the register definitions if they don't know what they represent. E.g. noting that the `_FIRST_CHANGE` ones are 1 iff that was the most significant address component that changed, otherwise 0. And that COUNTER just starts at 0 and increments by 1.

evm/src/all_stark.rs
wborgeaud: Hm normally the linear combinations supported are the same for filters and columns now. Here `NUM_MEMORY_OPS` and `op` are constants and `clock` is a column, right? If so you can use `Column::linear_combination_with_constant([(clock, NUM_MEMORY_OPS)], ops)`.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,577 @@
...
+        timestamp.iter().cloned(),
+        is_read.iter().cloned(),
+        context.iter().cloned(),
+        segment.iter().cloned(),
+        virtuals.iter().cloned(),
+        values.iter().cloned(),
+    )
+    .collect();
+
+    ops.sort_by_key(|&(t, _, c, s, v, _)| {
```
wborgeaud, line 105: nit: I think we don't need the sorting to be stable here, so we can use `sort_unstable_by_key` which should be faster.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,577 @@
...
+
+pub fn generate_first_change_flags<F: RichField>(
+    context: &[F],
+    segment: &[F],
+    virtuals: &[F],
+) -> (Vec<F>, Vec<F>, Vec<F>) {
+    let num_ops = context.len();
+    let mut context_first_change = Vec::new();
+    let mut segment_first_change = Vec::new();
+    let mut virtual_first_change = Vec::new();
```
wborgeaud, line 125: nit: Can be `Vec::with_capacity(num_ops)`.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,577 @@
...
+    for idx in 0..num_ops - 1 {
+        let this_context_first_change = if context[idx] != context[idx + 1] {
+            F::ONE
+        } else {
+            F::ZERO
+        };
```
wborgeaud, line 131: nit: Can also be `F::from_bool(context[idx] != context[idx + 1])` if you prefer.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,577 @@
...
+        let this_segment_first_change = if segment[idx] != segment[idx + 1] {
+            F::ONE * (F::ONE - this_context_first_change)
+        } else {
+            F::ZERO
+        };
```
wborgeaud, line 136: nit: Totally subjective, but I would find
```rust
let this_segment_first_change = if segment[idx] != segment[idx + 1] && this_context_first_change.is_zero() {
    F::ONE
} else {
    F::ZERO
};
```
a bit clearer. Or you could just keep these variables as bools (and convert in the `push` below) and just have 
```rust
let this_segment_first_change = segment[idx] != segment[idx + 1] && !this_context_first_change;
```

npwardberkeley: Ohh yes I like that better, thanks!

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,577 @@
...
+    COUNTER, COUNTER_PERMUTED, IS_READ, NUM_REGISTERS, RANGE_CHECK, RANGE_CHECK_PERMUTED,
+    SEGMENT_FIRST_CHANGE, SORTED_ADDR_CONTEXT, SORTED_ADDR_SEGMENT, SORTED_ADDR_VIRTUAL,
+    SORTED_IS_READ, SORTED_TIMESTAMP, TIMESTAMP, VIRTUAL_FIRST_CHANGE,
+};
+use crate::stark::Stark;
+use crate::util::{permuted_cols, trace_rows_to_poly_values};
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+#[derive(Default)]
+pub struct TransactionMemory {
```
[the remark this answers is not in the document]
npwardberkeley: Ah okay! So I should remove them from this PR right?

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+        let timestamp = &trace_cols[TIMESTAMP];
+        let is_read = &trace_cols[IS_READ];
+        let context = &trace_cols[ADDR_CONTEXT];
+        let segment = &trace_cols[ADDR_SEGMENT];
+        let virtuals = &trace_cols[ADDR_VIRTUAL];
+        let values: Vec<[F; 8]> = (0..num_trace_rows)
+            .map(|i| {
+                let arr: [F; 8] = (0..8)
+                    .map(|j| &trace_cols[value_limb(j)][i])
+                    .cloned()
```
wborgeaud, line 209: Can be replaced by
```rust
.map(|j| trace_cols[value_limb(j)][i])
```

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+                    .collect_vec()
+                    .try_into()
+                    .unwrap();
+                arr
+            })
+            .collect();
```
wborgeaud, line 215: Transpose doesn't seem necessary since the transpose is taken again below. So we could just have `values: [Vec<F>, 8]`.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+                trace_cols[sorted_value_limb(j)][i] = sorted_values[i][j];
+            }
+        }
+
+        trace_cols[CONTEXT_FIRST_CHANGE] = context_first_change;
+        trace_cols[SEGMENT_FIRST_CHANGE] = segment_first_change;
+        trace_cols[VIRTUAL_FIRST_CHANGE] = virtual_first_change;
+
+        trace_cols[RANGE_CHECK] = range_check_value;
+        trace_cols[COUNTER] = (0..trace_cols[0].len())
```
wborgeaud, line 255: nit: `trace_cols[0].len()` -> `num_trace_rows`.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+        trace_cols[COUNTER_PERMUTED] = permuted_table;
+    }
+
+    pub fn generate_trace(
+        &self,
+        memory_ops: Vec<(F, F, F, F, F, [F; 8])>,
+    ) -> Vec<PolynomialValues<F>> {
+        let mut timing = TimingTree::new("generate trace", log::Level::Debug);
+
+        // Generate the witness, except for permuted columns in the lookup argument.
```
wborgeaud, line 271: "except for permuted columns in the lookup argument" Is that up to date? Seems like these columns are generated.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+            "convert to PolynomialValues",
+            trace_rows_to_poly_values(trace_rows)
+        );
+
+        timing.print();
+        trace_polys
+    }
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F, D> {
```
wborgeaud, line 289: Isn't it missing the `permutation_pairs()` for the range-checks?

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+        // Enumerate purportedly-ordered log.
+        for i in 0..8 {
+            yield_constr
+                .constraint(next_is_read * address_unchanged * (next_values[i] - values[i]));
+        }
+
+        // Lookup argument for the range check.
+        let local_perm_input = vars.local_values[RANGE_CHECK_PERMUTED];
+        let next_perm_table = vars.next_values[COUNTER_PERMUTED];
+        let next_perm_input = vars.next_values[COUNTER_PERMUTED];
```
wborgeaud, line 366: Is it normal that these two values are the same?

evm/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #![allow(clippy::too_many_arguments)]
 #![allow(clippy::type_complexity)]
 #![feature(generic_const_exprs)]
+#![allow(dead_code)]
```
wborgeaud, line 6: I don't think we should had that lint globally. Maybe add it locally to unused functions along with a todo to remove it once used.

wborgeaud (approved):
Looks good to me!

evm/src/all_stark.rs
npwardberkeley: Nice, thanks!!

evm/src/memory/registers.rs
npwardberkeley: Good idea!

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,551 @@
...
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+pub(crate) const NUM_PUBLIC_INPUTS: usize = 0;
+
+#[derive(Copy, Clone)]
+pub struct MemoryStark<F, const D: usize> {
+    pub(crate) f: PhantomData<F>,
+}
+
+pub fn generate_random_memory_ops<F: RichField>(num_ops: usize) -> Vec<(F, F, F, F, F, [F; 8])> {
```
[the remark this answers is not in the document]
npwardberkeley: Ohh yes I like that!

evm/src/memory/memory_stark.rs
npwardberkeley: Hm I think it's necessary for it to be by row, so we can sort it in `sort_memory_ops`?

evm/src/memory/memory_stark.rs
npwardberkeley: Oh right, the second is supposed to be `RANGE_CHECK_PERMUTED`. Good catch

evm/src/lib.rs
npwardberkeley: Oh right — I think that was for debugging and I forgot to remove it

evm/src/cpu/columns.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
dlubarov, line 1: nit: let's add a TODO (empty is fine) to make sure we don't forget it, and maybe scope it more narrowly if it's not too many items?

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+    timestamp: &[F],
+    context_first_change: &[F],
+    segment_first_change: &[F],
+    virtual_first_change: &[F],
+) -> Vec<F> {
+    let num_ops = context.len();
+    let mut range_check = Vec::new();
+
+    for idx in 0..num_ops - 1 {
+        let this_address_unchanged = F::ONE
```
dlubarov, line 166: Maybe call this `address_first_change`? In my mind it has the same semantics as the other flags, just this one is derived instead of having its own column

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
+
+pub(crate) const NUM_PUBLIC_INPUTS: usize = 0;
+
+#[derive(Copy, Clone)]
+pub struct MemoryStark<F, const D: usize> {
+    pub(crate) f: PhantomData<F>,
+}
+
+pub struct MemoryOp<F> {
```
dlubarov, line 32: I think it would be useful to also include the channel index in here, so that the generation code here can handle populating the `is_memop` flags.

Or if it's hard to change with the way the test is written, we can add a TODO for later. I think `generate_random_memory_ops` etc. can go away later when we have more "real" tests.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+        let next_addr_segment = vars.next_values[SORTED_ADDR_SEGMENT];
+        let next_addr_virtual = vars.next_values[SORTED_ADDR_VIRTUAL];
+        let next_values: Vec<_> = (0..8)
+            .map(|i| vars.next_values[sorted_value_limb(i)])
+            .collect();
+
+        let context_first_change = vars.local_values[CONTEXT_FIRST_CHANGE];
+        let segment_first_change = vars.local_values[SEGMENT_FIRST_CHANGE];
+        let virtual_first_change = vars.local_values[VIRTUAL_FIRST_CHANGE];
+        let address_unchanged =
```
dlubarov, line 345: Likewise here, maybe call it `address_first_change`?

evm/src/memory/memory_stark.rs
dlubarov: Er I meant `timestamp_first_change`. But on second thought your name makes sense also, fine to keep it if you like.

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+        yield_constr
+            .constraint_transition(virtual_first_change * (next_addr_context - addr_context));
+        yield_constr
+            .constraint_transition(virtual_first_change * (next_addr_segment - addr_segment));
+        yield_constr.constraint_transition(address_unchanged * (next_addr_context - addr_context));
+        yield_constr.constraint_transition(address_unchanged * (next_addr_segment - addr_segment));
+        yield_constr.constraint_transition(address_unchanged * (next_addr_virtual - addr_virtual));
+
+        // Third set of ordering constraints: range-check difference in the column that should be increasing.
+        let range_check_value = context_first_change * (next_addr_context - addr_context - one)
```
dlubarov, line 373: nit: we usually name it like `computed_range_check`

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+            yield_constr
+                .constraint(next_is_read * address_unchanged * (next_values[i] - values[i]));
+        }
+
+        // Lookup argument for the range check.
+        let local_perm_input = vars.local_values[RANGE_CHECK_PERMUTED];
+        let next_perm_table = vars.next_values[COUNTER_PERMUTED];
+        let next_perm_input = vars.next_values[RANGE_CHECK_PERMUTED];
+
+        // A "vertical" diff between the local and next permuted inputs.
```
dlubarov, line 390: I think it would be nice to have this logic in a lookup module, similar to what we had in system_zero, so we can have a module doc with links to the lookup spec etc.

We could keep `eval_lookups` and `eval_lookups_circuit` as before, except `col_permuted_input` and `col_permuted_table` would become parameters that the caller passes in.

If so we could move `permuted_cols` there also; it also seems hard to understand without documentation (which we could have at the module level so we don't need to repeat it).

evm/src/memory/memory_stark.rs
```diff
@@ -0,0 +1,580 @@
...
+            trace_cols[ADDR_VIRTUAL][i] = virt;
+            for j in 0..8 {
+                trace_cols[value_limb(j)][i] = value[j];
+            }
+        }
+
+        self.generate_memory(&mut trace_cols);
+
+        let mut trace_rows = vec![[F::ZERO; NUM_REGISTERS]; num_ops];
+        for (i, col) in trace_cols.iter().enumerate() {
```
dlubarov, line 214: Looks like this is transposing? Seems unnecessary since the next step would be to transpose back to col-major form.

It's not a big issue though, and I wanted to experiment with some other changes to witness generation, namely doing more of it in row-major form to avoid sparse use of memory (for cache efficiency). So I think we can leave it for now and I can make a PR with various generation tweaks later.

dlubarov (approved):
LGTM otherwise

evm/src/memory/memory_stark.rs
npwardberkeley: Hm yeah it used to be `timestamp_first_change`, I forget exactly why I decided to change it. It might have been because it's also used in constraints to check read/write consistency, so it's more clear that it corresponds to having the same address

evm/src/cpu/columns.rs
npwardberkeley: Hm it's all six of the functions, so I thought it was better to just do it throughout the file

PR #572, opened 2022-06-17
Fix `check_constraints` to only look at subgroup points

dlubarov:
`get_lde_values_packed` was loading points from the Merkle tree, which contains evaluations on a coset, rather than the subgroup.


npwardberkeley (approved):
Looks good, thanks!

PR #573, opened 2022-06-18
Define columns for CTL closer to the constraints

nbgl:
Just cleaning up the code structure, as @wborgeaud suggested in a recent PR comment


wborgeaud (approved):
Looks good to me!

PR #574, opened 2022-06-22
Exponentiation kernel function

nbgl:
This doesn’t have a unit test, but I have tested it in an EVM simulator and it works

### DIFF
--- a/evm/src/cpu/kernel/asm/exp.asm
+++ b/evm/src/cpu/kernel/asm/exp.asm
@@ -1,0 +1,70 @@
+global exp:
+// we don't seem to handle global labels yet, so this function has a local label too for now:
+exp:
+    // stack: x, e, retdest
+    dup2
+    // stack: e, x, e, retdest
+    push step_case
+    // stack: step_case, e, x, e, retdest
+    jumpi
+    // stack: x, e, retdest
+    pop
+    // stack: e, retdest
+    pop
+    // stack: retdest
+    push 1
+    // stack: 1, retdest
+    swap1
+    // stack: retdest, 1
+    jump
+
+step_case:
+    // stack: x, e, retdest
+    push recursion_return
+    // stack: recursion_return, x, e, retdest
+    push 2
+    // stack: 2, recursion_return, x, e, retdest
+    dup4
+    // stack: e, 2, recursion_return, x, e, retdest
+    div
+    // stack: e / 2, recursion_return, x, e, retdest
+    dup3
+    // stack: x, e / 2, recursion_return, x, e, retdest
+    dup1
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/asm/exp.asm
```diff
@@ -0,0 +1,70 @@
...
+    mul
+    // stack: (x - 1) * (e % 2), exp(x * x, e / 2), x, e, retdest
+    push 1
+    // stack: 1, (x - 1) * (e % 2), exp(x * x, e / 2), x, e, retdest
+    add
+    // stack: 1 + (x - 1) * (e % 2), exp(x * x, e / 2), x, e, retdest
+    mul
+    // stack: (1 + (x - 1) * (e % 2)) * exp(x * x, e / 2), x, e, retdest
+    swap3
+    // stack: retdest, x, e, (x / (e % 2)) * exp(x * x, e / 2)
```
unzvfu, line 63: Should be `exp(x, e % 2)`, not `x / (e % 2)`, right? (And similarly for the next three comments.)

evm/src/cpu/kernel/asm/exp.asm
```diff
@@ -0,0 +1,70 @@
+global exp:
+// we don't seem to handle global labels yet, so this function has a local label too for now:
+exp:
```
unzvfu, line 3: Could probably do with a simple comment along the lines of "Implementation using recursive expression `exp(x, e) = exp(x, e % 2) * exp(x * x, e / 2)`."

unzvfu (approved):
Looks good! Nice and straight forward. There are alternative methods of course that would probably take less stack space or possibly fewer/cheaper instructions (not certain, never implemented them on a stack machine), but they would be more complicated and not necessarily a net win in our context; this approach is already very short and elegant. :+1:

evm/src/cpu/kernel/asm/exp.asm
nbgl: I wrote out the pseudocode for the function, which should hopefully help :)

nbgl:
> There are alternative methods of course that would probably take less stack space or possibly fewer/cheaper instructions (not certain, never implemented them on a stack machine)

My first attempt was indeed iterative like that! @dlubarov later played with it too and managed to shave off a few instructions.

You do save a _lot_ of stack space that way, but (pls correct me if I’m wrong) I don’t think the stack space is actually expensive in the proof settings (remember that the number of reads/writes remains ~the same, even if the max space used increases).

The iterative approach makes the stack manipulation (imo) a lot harder to follow. I like the recursive version purely for readability.

PR #576, opened 2022-06-24
Fix EVM dependency list

nbgl:
Upgrades the `itertools` dependency to version `0.10.3`. That’s the first release that includes `multiunzip`. The build fails with older versions.


npwardberkeley:
Nice, thanks!

npwardberkeley (approved):

nbgl:
No worries :) It’s a really bad design decision for Cargo to ignore an _explicitly pinned_ version number…

PR #577, opened 2022-06-24
CTL: limbs (CPU) <-> bits (logic)

nbgl:
Uses the linear combination `Column` API to do CTL directly between limbs in the CPU table and bits in the logic table. Removes 32 columns from the logic table.

### DIFF
--- a/evm/src/all_stark.rs
+++ b/evm/src/all_stark.rs
@@ -116,11 +116,11 @@ mod tests {
             let mut row = [F::ZERO; logic::columns::NUM_COLUMNS];
 
             assert_eq!(logic::PACKED_LIMB_BITS, 16);
-            for col in logic::columns::INPUT0_PACKED {
-                row[col] = F::from_canonical_u16(rng.gen());
+            for col in logic::columns::INPUT0 {
+                row[col] = F::from_bool(rng.gen());
             }
-            for col in logic::columns::INPUT1_PACKED {
-                row[col] = F::from_canonical_u16(rng.gen());
+            for col in logic::columns::INPUT1 {
+                row[col] = F::from_bool(rng.gen());
             }
             let op: usize = rng.gen_range(0..3);
             let op_col = [
@@ -148,6 +148,7 @@ mod tests {
         memory_stark.generate_trace(memory_ops)
     }
 
+    // JNTODO
     fn make_cpu_trace(
         num_keccak_perms: usize,
         num_logic_rows: usize,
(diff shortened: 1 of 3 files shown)


evm/src/all_stark.rs
```diff
@@ -116,11 +116,11 @@ mod tests {
...
             assert_eq!(logic::PACKED_LIMB_BITS, 16);
-            for col in logic::columns::INPUT0_PACKED {
-                row[col] = F::from_canonical_u16(rng.gen());
+            for col in logic::columns::INPUT0 {
+                row[col] = F::from_bool(rng.gen());
             }
-            for col in logic::columns::INPUT1_PACKED {
-                row[col] = F::from_canonical_u16(rng.gen());
+            for col in logic::columns::INPUT1 {
+                row[col] = F::from_bool(rng.gen());
```
nbgl, line 123: Previously we sampled from 16-bit limbs (which were later decomposed to bits). Now sample the bits directly.

evm/src/logic.rs
```diff
@@ -145,8 +127,7 @@ fn make_result<F: RichField>(lv: &mut [F; columns::NUM_COLUMNS], op: Op) {
 impl<F: RichField, const D: usize> LogicStark<F, D> {
     pub fn generate(&self, lv: &mut [F; columns::NUM_COLUMNS]) {
         let op = check_op_flags(lv);
-        check_limb_length(lv);
-        make_bit_decomposition(lv);
+        check_bits(lv);
         make_result(lv, op);
```
nbgl, line 131: We no longer decompose the limbs into bits. Instead, we read the bits directly.

evm/src/logic.rs
```diff
@@ -66,6 +66,7 @@ pub struct LogicStark<F, const D: usize> {
 }
 
 enum Op {
+    Zero,
```
wborgeaud, line 69: nit: Maybe add a comment saying what this op does and why we need it.

evm/src/logic.rs
```diff
@@ -78,62 +79,43 @@ fn check_op_flags<F: RichField>(lv: &[F; columns::NUM_COLUMNS]) -> Op {
...
-        columns::INPUT1_PACKED
+        columns::limb_bit_cols_for_input(columns::INPUT0),
+        columns::limb_bit_cols_for_input(columns::INPUT1),
     ) {
-        let limb_in0 = lv[limb_in0_col].to_canonical_u64();
-        let limb_in1 = lv[limb_in1_col].to_canonical_u64();
+        let limb_in0: u64 = limb_in0_cols
+            .enumerate()
+            .map(|(i, col)| lv[col].to_canonical_u64() << i)
+            .sum();
```
wborgeaud, line 112: nit: It seems like this conversion `bits -> u64/usize` is done in a few places. Maybe we could add a function for it.

wborgeaud (approved):
Looks good to me 👍

nbgl:
Those were both good points, @wborgeaud! I’ve made those changes. Would you mind having a look at them? They’re just big enough that I wouldn’t be comfortable merging without another review.

wborgeaud (approved):
Thanks, looks good 👍

PR #578, opened 2022-06-25
Move CTL config out of test

dlubarov:
Since it's not specific to the test, but will be used in any EVM proof.


evm/src/all_stark.rs
```diff
@@ -56,6 +62,50 @@ impl Table {
     }
 }
 
+#[allow(unused)] // TODO: Should be used soon.
+pub(crate) fn all_cross_table_lookups<F: Field>() -> Vec<CrossTableLookup<F>> {
+    let mut cross_table_lookups = vec![
+        CrossTableLookup::new(
```
wborgeaud, line 68: I think this function will get quite large once all the tables are ready. Maybe we could add comments to specify what each `CrossTableLookup` is doing. 
Or maybe we could split it up in smaller functions `X_ctl() ->  Vec<CrossTableLookup<F>>` and concatenate them in this function.
Doesn't have to be in this PR though since the size is still manageable.

wborgeaud (approved):
Looks good to me!

PR #580, opened 2022-06-26
ASM macro support

dlubarov:
Also recognize global labels as a `PUSH` target; previously it only considered local labels.


nbgl (approved):
Looks good!

The way I’m reading this code, macros can’t invoke other macros? This is not a problem, but sounds like a nice feature to add at some point.

dlubarov:
It should be supported :) I just added a test.

PR #581, opened 2022-06-26
`extension_field` -> `extension`

dlubarov:
It seems redundant in most contexts, e.g. `use plonky2::field::extension_field::Extendable;`. One could import `extension_field`, but it's not that common in Rust, and `field::extension` is now roughly as short.


wborgeaud (approved):
👍

PR #583, opened 2022-06-27
`field_types` -> `types`

dlubarov:
Here too, I think "field" is usually clear from context, e.g. in `use plonky2::field::types::Field;`.


wborgeaud (approved):
👍 (once CI is green)

PR #587, opened 2022-06-28
`registers` -> `columns`

wborgeaud:
~~Rename some `columns` modules to `registers` for consistency.~~
Edit: Rename some `registers` modules to `columns` for consistency.


dlubarov (approved):
LGTM

nbgl:
I am 100% for consistency but I do think “columns” is actually more descriptive. When you say “register”, I think of EVM registers (i.e., the stack and the program counter).

wborgeaud:
> I am 100% for consistency but I do think “columns” is actually more descriptive. When you say “register”, I think of EVM registers (i.e., the stack and the program counter).

Yeah makes sense. I don't have a strong opinion on this so either works for me. I just think we need to be consistent across the crate.
Does anybody else have a preference on this?

bpfarmer:
I actually like columns for what it's worth.

unzvfu:
> > I am 100% for consistency but I do think “columns” is actually more descriptive. When you say “register”, I think of EVM registers (i.e., the stack and the program counter).
> 
> Yeah makes sense. I don't have a strong opinion on this so either works for me. I just think we need to be consistent across the crate. Does anybody else have a preference on this?

I don't have strong feelings either way, but I agree with @nbgl's point that we should try to clearly distinguish the concepts of EVM register and our columns/registers. Using "column" for our side seems like a good option.

wborgeaud:
By popular demand I changed this to use "columns" instead of "registers".

PR #589, opened 2022-06-28
Structured wrapper over CPU table row

nbgl:


evm/src/cpu/columns.rs
```diff
@@ -1,207 +1,204 @@
...
-/// Filter. 1 if the row corresponds to a cycle of execution and 0 otherwise.
-/// Lets us re-use decode columns in non-cycle rows.
-pub const IS_CPU_CYCLE: usize = IS_BOOTSTRAP_CONTRACT + 1;
-
-/// If CPU cycle: The opcode being decoded, in {0, ..., 255}.
-pub const OPCODE: usize = IS_CPU_CYCLE + 1;
-
-// If CPU cycle: flags for EVM instructions. PUSHn, DUPn, and SWAPn only get one flag each. Invalid
-// opcodes are split between a number of flags for practical reasons. Exactly one of these flags
-// must be 1.
```
nbgl, line 23: NB: I meant to bring back these comments before making a PR. I’ll copy them over

evm/src/cpu/columns.rs
```diff
@@ -1,207 +1,204 @@
...
 }
 
-const MEM_ADDR_VALUE_START: usize = MEM_ADDR_VIRTUAL_START + memory::NUM_CHANNELS;
-pub const fn mem_value(channel: usize, limb: usize) -> usize {
-    debug_assert!(channel < memory::NUM_CHANNELS);
-    debug_assert!(limb < memory::VALUE_LIMBS);
-    MEM_ADDR_VALUE_START + channel * memory::VALUE_LIMBS + limb
+const fn make_col_map() -> CpuColumnsView<usize> {
+    let mut indices_arr = [0; NUM_CPU_COLUMNS];
+    let mut i = 0;
```
dlubarov, line 196: nit: `for i in 0..NUM_CPU_COLUMNS` seems more succinct, or could do `(0..NUM_CPU_COLUMNS).try_into().unwrap()`

dlubarov (approved):
LGTM, pretty neat!

cc @wborgeaud @npwardberkeley @unzvfu FYI. We could use this pattern in other tables later.

evm/src/cpu/columns.rs
unzvfu: I don't think `for` loops are permitted in a `const fn` are they?

nbgl: Yeah, neither of those approaches is permitted in a `const` context

dlubarov: Oh ok then 👍

npwardberkeley:
Wow nice! Should we just go ahead and change the other tables too, for consistency?

nbgl:
> Should we just go ahead and change the other tables too, for consistency?

Yeah, I think that would be good! This PR only changes the CPU table (for the sake of incrementalizm) but the same thing will work for other tables too

PR #590, opened 2022-06-28
`CommonCircuitData` serialization

Sladuca:
Add support for serializing / deserializing `CommonCircuitData` to the serialization [`Buffer`](https://github.com/mir-protocol/plonky2/blob/dec0765fb554f76e80b8e314b5991318be20609b/plonky2/src/util/serialization.rs#L26).

### DIFF
--- a/plonky2/src/gates/gate.rs
+++ b/plonky2/src/gates/gate.rs
@@ -39,7 +39,6 @@ pub enum GateKind {
     Reducing,
 }
 
-
 /// A custom gate.
 pub trait Gate<F: RichField + Extendable<D>, const D: usize>: 'static + Send + Sync {
     fn id(&self) -> String;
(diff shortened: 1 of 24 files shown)


plonky2/src/gates/gate.rs
```diff
@@ -17,11 +18,38 @@ use crate::plonk::circuit_builder::CircuitBuilder;
...
+    RandomAccess,
+    ReducingExt,
+    Reducing,
+}
 
 /// A custom gate.
 pub trait Gate<F: RichField + Extendable<D>, const D: usize>: 'static + Send + Sync {
     fn id(&self) -> String;
 
+    fn kind(&self) -> GateKind;
```
wborgeaud, line 46: One drawback of this approach is that new gates cannot be defined in external crates (as can be seen from the failing test). 
Do you really need this `GateKind` enum in your applications, or would using the `id()` string as a proxy also work?

wborgeaud: Maybe we could just have the serialization function take a vector of supported gates `gates: Vec<Box<dyn Gate<F, D>>>` and have the tag of a gate be its index in the vector. That way we don't need the `GateKind` enum nor the `kind()` function. 
Serialization would be dependent on an agreed-upon set of gates for the application, but I think that's a feature as different applications might require different sets of gates.

Sladuca: Good idea

wborgeaud: Hm good point. It's a bit hacky but I guess we could have `deserialize` be non-static.

Sladuca:
Gonna circle back and do this a different way.

PR #591, opened 2022-06-29
Begin work on witness generation and kernel bootstrapping

dlubarov:
Witness generation is divided into two phases:
- First, the CPU table is generated row-by-row as we simulate the EVM. While doing so we record some info for the next phase, like a list of Keccak inputs.
- Second, we generating all the other tables. This can be done in parallel, though I left it as a future optimization.


evm/src/all_stark.rs
```diff
@@ -163,32 +164,20 @@ mod tests {
...
-
-        for _ in num_rows..num_rows.next_power_of_two() {
-            trace_rows.push([F::ZERO; logic::columns::NUM_COLUMNS])
-        }
-        trace_rows_to_poly_values(trace_rows)
+        let ops = (0..num_rows)
+            .map(|_| {
+                let input0 = U256(rng.gen());
+                let input1 = U256(rng.gen());
+                let result = input0 ^ input1;
```
dlubarov, line 171: I can add back AND/OR if you think it's important, but this "random" test should be going away soonish, replaced by real transaction tests.

evm/tests/transfer_to_new_addr.rs
```diff
@@ -0,0 +1,46 @@
...
+use plonky2_evm::verifier::verify_proof;
+
+type F = GoldilocksField;
+const D: usize = 2;
+type C = PoseidonGoldilocksConfig;
+
+/// Test a simple token transfer to a new address.
+#[test]
+#[ignore] // TODO: Won't work until txn parsing, storage, etc. are implemented.
+fn test_simple_transfer() -> anyhow::Result<()> {
```
dlubarov, line 18: Added this test to illustrate what we're working toward. Once we get storage etc working, we can add more realistic tests like this, and eventually delete the "fake" test stuff like `make_cpu_trace`, `generate_random_memory_ops`, etc.

evm/src/generation/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::all_stark::AllStark;
+use crate::cpu::bootstrap_kernel::generate_bootstrap_kernel;
+use crate::cpu::columns::NUM_CPU_COLUMNS;
+use crate::generation::state::GenerationState;
+use crate::util::trace_rows_to_poly_values;
+
+mod memory;
+pub(crate) mod state;
+
+pub type RlpMerkleProof = Vec<Vec<u8>>;
```
wborgeaud, line 15: nit: Maybe add a comment explaining how this is structured. Intuitively I would have thought the whole Merkle proof would be RLP encoded.

wborgeaud (approved):
Looks good to me!

evm/src/generation/mod.rs
dlubarov: Good point, added a comment. Packing them in a single RLP blob could work also, but one RLP blob per node seems to be conventional. I think the convention comes from EIP-1186 (it wasn't formally accepted, but some clients seem to support `eth_getProof` based on it anyway).

evm/src/generation/state.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Compute logical AND, and record the operation to be added in the logic table later.
+    #[allow(unused)] // TODO: Should be used soon.
+    pub(crate) fn and(&mut self, input0: U256, input1: U256) -> U256 {
+        let result = input0 & input1;
+        self.logic_ops.push(Operation {
+            operator: logic::Op::And,
+            input0,
+            input1,
+            result,
+        });
```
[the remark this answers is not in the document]
dlubarov: Good idea, done.

evm/src/generation/state.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub(crate) fn xor(&mut self, input0: U256, input1: U256) -> U256 {
+        self.logic_op(Op::Xor, input0, input1)
+    }
+
+    /// Compute logical AND, and record the operation to be added in the logic table later.
+    pub(crate) fn logic_op(&mut self, op: Op, input0: U256, input1: U256) -> U256 {
+        let operation = Operation::new(op, input0, input1);
+        let result = operation.result;
+        self.logic_ops.push(operation);
+        result
```
nbgl, line 50: Logic ops (also arithmetic) rows could potentially be reused, if an operation is executed multiple times with identical inputs. Don’t worry about this now, but potentially a thing to do in a future change.

nbgl (commented):
Sorry this review took so long. Here are my comments for what it’s worth

evm/src/generation/memory.rs
```diff
@@ -0,0 +1,52 @@
...
+            .copied()
+            .unwrap_or([F::ZERO; VALUE_LIMBS])
+    }
+
+    pub(super) fn set(&mut self, virtual_addr: usize, value: [F; VALUE_LIMBS]) {
+        if virtual_addr + 1 > self.content.len() {
+            self.content
+                .resize(virtual_addr + 1, [F::ZERO; VALUE_LIMBS]);
+        }
+        self.content[virtual_addr] = value;
```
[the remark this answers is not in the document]
dlubarov: We don't have data on this, but Ethereum's gas model should discourage sparse memory use, as there's a gas fee based on the largest address that (at some point) had a nonzero value.

evm/src/generation/memory.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[derive(Default, Debug)]
+pub(crate) struct MemoryContextState<F: Field> {
+    /// The content of each memory segment.
+    pub segments: [MemorySegmentState<F>; NUM_SEGMENTS],
+}
+
+#[derive(Default, Debug)]
+pub(crate) struct MemorySegmentState<F: Field> {
+    pub content: Vec<[F; VALUE_LIMBS]>,
```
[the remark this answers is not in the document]
dlubarov: Yeah it's a bit much 😅 but I think the access pattern should be reasonably cache-friendly. There should only be a few `MemorySegmentState`s that we're likely to interact with around the same time, like the current context's stack and memory (both should be a few kilobytes generally) and any frequently-used bits of kernel code.

I think we might end up with the property that all memory values fit in 256 bits; if so we could ~halve memory use by storing `U256`s instead of `[F; 8]`. I'm not certain though, so let's revisit a bit later.

evm/src/memory/segments.rs
```diff
@@ -0,0 +1,22 @@
...
+pub const CALLDATA: usize = 4;
+
+/// Data returned to the current context by its latest callee.
+pub const RETURNDATA: usize = 5;
+
+/// A segment which contains a few fixed-size metadata fields, such as the caller's context, or the
+/// size of `CALLDATA` and `RETURNDATA`.
+pub const METADATA: usize = 6;
+
+pub const NUM_SEGMENTS: usize = 7;
```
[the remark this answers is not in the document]
dlubarov: Hm would we list them out separately in the assert? What do you think of using an enum instead, like

```
pub enum Segment {
    Code = 0,
    Stack = 1,
    ...
}
```

Seems a bit safer since it wouldn't rely on a separate list. It wouldn't check `NUM_SEGMENTS` though; if we wanted that we could use one of those crates with a procedural macro for number of variants.

PR #593, opened 2022-07-05
Find global labels before assembly

wborgeaud:
Current implementation builds the global labels map incrementally, making it impossible to call a global label defined in a subsequent file.
Fixes this by building the labels map before calling the `assemble_file` function.


dlubarov (approved):
LGTM

PR #594, opened 2022-07-05
Elliptic curve assembly

wborgeaud:
Kernel functions for `ecAdd` and `ecMul` precompiles, performing addition and scalar multiplication on BN254.


evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,417 @@
...
+    // stack: x1, y1, x0, y0, x1, y1, retdest
+    DUP4
+    // stack: y0, x1, y1, x0, y0, x1, y1, retdest
+    DUP4
+    // stack: x0, y0, x1, y1, x0, y0, x1, y1, retdest
+    %ec_check
+    // stack: isValid(x0, y0), x1, y1, x0, y0, x1, y1, retdest
+    PUSH ec_add_valid_first_point
+    // stack: ec_add_valid_first_point, isValid(x0, y0), x1, y1, x0, y0, x1, y1, retdest
+    JUMPI
```
dlubarov, line 56: I think it would be slightly easier to read if we computed `isValid(x0, y0) && isValid(x1, y1)` in a branch-free way, so that we can combine two branches into one, and combine the two "return invalid" cases.

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,417 @@
...
+    // PUSH 0xe7313274bb29566ff0c8220eb9841de1d96c2923c6a4028f7dd3c6a14cee770
+    JUMPDEST
+    // stack: x0, y0, x1, y1, retdest
+    DUP2
+    // stack: y0, x0, y0, x1, y1, retdest
+    DUP2
+    // stack: x0, y0, x0, y0, x1, y1, retdest
+    ISZERO
+    // stack: x0==0, y0, x0, y0, x1, y1, retdest
+    SWAP1
```
dlubarov, line 18: nit: could avoid swapping if we checked `y0 == 0` before dup'ing `x0`

dlubarov: Although on the other hand, it might be nice to add a little macro to check if the top point on the stack is the identity point, since we seem to do that 3x. If so disregard the above.

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,417 @@
...
+    POP
+    // stack: x0, y0, x1, y1, retdest
+    POP
+    // stack: y0, x1, y1, retdest
+    POP
+    // stack: x1, y1, retdest
+    POP
+    // stack: y1, retdest
+    POP
+    // stack: retdest
```
dlubarov, line 69: It looks like this returns nothing if the input is invalid? I think it would be hard for the caller (i.e. the wrapper we'll add which reads inputs from memory etc) to recognize error cases. We could return some special value such as `(u256::MAX, u256::MAX)`

evm/src/cpu/kernel/asm/curve_add.asm
dlubarov: Also I think it would be easiest if we checked input validity first, before checking for any exceptional cases, so we don't need to check validity in `ec_add_first_zero` etc.

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,417 @@
...
+    POP
+    // stack: y, retdest
+    POP
+    // stack: retdest
+    PUSH 0
+    // stack: 0, retdest
+    PUSH 0
+    // stack: 0, 0, retdest
+    SWAP2
+    // stack:  0, retdest, 0
```
dlubarov, line 161: Shouldn't it be `retdest, 0, 0` here?

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,417 @@
...
+    ADDMOD
+    // stack: y1 + y0, x0, y0, x1, y1, retdest
+    PUSH ec_add_equal_points
+    // stack: ec_add_equal_points, y1 + y0, x0, y0, x1, y1, retdest
+    JUMPI
+    // stack: x0, y0, x1, y1, retdest
+    POP
+    // stack: y0, x1, y1, retdest
+    POP
+    // stack: x1, y1, retdest
```
dlubarov, line 293: Can we use `ret_zero` for the rest?

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,383 @@
...
+    JUMPDEST
+    // stack: x0, y0, x1, y1, retdest
+    DUP2
+    // stack: y0, x0, y0, x1, y1, retdest
+    DUP2
+    // stack: x0, y0, x0, y0, x1, y1, retdest
+    %ec_isidentity
+    // stack: (x0,y0)==(0,0), x0, y0, x1, y1, retdest
+    PUSH ec_add_first_zero
+    // stack: ec_add_first_zero, y0==0 & x0==0, x0, y0, x1, y1, retdest
```
dlubarov, line 51: nit: let's change `y0==0 & x0==0` to `(x0,y0)==(0,0)` to match above (and on line 61)

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,383 @@
...
+    // stack: x1, y1, retdest
+    DUP2
+    // stack: y1, x1, y1, retdest
+    DUP2
+    // stack: x1, y1, x1, y1, retdest
+    %ec_isidentity
+    // stack: (x1,y1)==(0,0), x1, y1, retdest
+    PUSH ret_zero
+    // stack: ret_zero, (x1,y1)==(0,0), x1, y1, retdest
+    JUMPI
```
dlubarov, line 108: Do we need to handle this separately? Seems like the general case of returning (x1, y1), below, would still work if it's the identity?

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,383 @@
...
+ec_add_equal_first_coord:
+    JUMPDEST
+    // stack: x0, y0, x1, y1, retdest with x0 == x1
+    %bn_base
+    // stack: N, x0, y0, x1, y1, retdest
+    DUP3
+    // stack: y0, N, x0, y0, x1, y1, retdest
+    DUP6
+    // stack: y1, y0, N, x0, y0, x1, y1, retdest
+    ADDMOD
```
dlubarov, line 214: nit: `EQ` seems a bit more intuitive, and should be a tad cheaper since it's done directly on the CPU

evm/src/cpu/kernel/asm/curve_add.asm
```diff
@@ -0,0 +1,383 @@
...
+    // stack: ec_add_equal_points, x0, y0, x0, y0, retdest
+    JUMP
+
+// Push the order of the BN254 base field.
+%macro bn_base
+    PUSH 0x30644e72e131a029b85045b68181585d97816a916871ca8d3c208c16d87cfd47
+%endmacro
+
+// Assumption: x, y < N and 2N < 2^256.
+// Note: Doesn't hold for Secp256k1 base field.
```
dlubarov, line 273: Just thinking about secp for later, we could detect underflow as `x < x - y` (an idea Jakub used a lot in SIMD code), though the code gets a bit longer

```
%macro submod
    // stack: x, y
    SWAP1
    // stack: y, x
    DUP2
    // stack: x, y, x
    SUB
    // stack: x - y, x
    DUP1
    // stack: x - y, x - y, x
    SWAP2
    // stack: x, x - y, x - y
    LT
    // stack: x < x - y, x - y
    %secp_base
    // stack: N, x < x - y, x - y
    MUL
    // stack: N * (x < x - y), x - y
    ADD
    // (x-y) % N
%endmacro
```

If performance is an issue we can discuss adding a `SUBMOD` instruction, or some other feature like carry flags.

dlubarov (approved):
Looks good otherwise!

PR #595, opened 2022-07-07
Macros with arguments

dlubarov:
See `basic_macros.rs` for an example. Will also let us write macros like `%mul_by(2)` which I think will be handy.

cc @nbgl

### DIFF
--- a/evm/src/cpu/kernel/assembler.rs
+++ b/evm/src/cpu/kernel/assembler.rs
@@ -19,6 +19,20 @@ pub struct Kernel {
     global_labels: HashMap<String, usize>,
 }
 
+struct Macro {
+    params: Vec<String>,
+    items: Vec<Item>,
+}
+
+impl Macro {
+    fn get_param_index(&self, param: &str) -> usize {
+        self.params
+            .iter()
+            .position(|p| p == param)
+            .unwrap_or_else(|| panic!("No such param: {} {:?}", param, &self.params))
+    }
+}
+
 pub(crate) fn assemble(files: Vec<File>) -> Kernel {
     let macros = find_macros(&files);
     let mut global_labels = HashMap::new();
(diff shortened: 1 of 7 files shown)


evm/src/cpu/kernel/assembler.rs
```diff
@@ -41,33 +55,31 @@ pub(crate) fn assemble(files: Vec<File>) -> Kernel {
...
 
-fn find_macros(files: &[File]) -> HashMap<String, Vec<Item>> {
+fn find_macros(files: &[File]) -> HashMap<String, Macro> {
     let mut macros = HashMap::new();
     for file in files {
         for item in &file.body {
-            if let Item::MacroDef(name, items) = item {
-                macros.insert(name.clone(), items.clone());
+            if let Item::MacroDef(name, params, items) = item {
+                let _macro = Macro {
```
wborgeaud, line 63: nit: Why the underscore?

evm/src/cpu/kernel/ast.rs
```diff
@@ -29,6 +29,7 @@ pub(crate) enum Item {
 pub(crate) enum PushTarget {
     Literal(Literal),
     Label(String),
+    Var(String),
```
wborgeaud, line 32: nit: Maybe call it `MacroVar` to make it clear that it should only be used in macros, or add a comment.

wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/assembler.rs
dlubarov: `macro` seems to be reserved in Rust (i.e. not a keyword yet, but reserved for future use)

wborgeaud: Ah I see, I've usually seen people put the underscore after the name in these cases. You can also escape keywords using `r#macro`, although it's pretty ugly. Anyways, definitely not important.

PR #596, opened 2022-07-07
EVM interpreter

wborgeaud:
Small EVM interpreter to test kernel functions.


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -0,0 +1,301 @@
...
+    stack: Vec<U256>,
+}
+
+pub fn run(code: &[u8], initial_offset: usize, initial_stack: Vec<U256>) -> Vec<U256> {
+    let mut interpreter = Interpreter {
+        code,
+        offset: initial_offset,
+        stack: initial_stack,
+    };
+    // Halt the execution if a jump to 0xdeadbeef was done.
```
dlubarov, line 15: I think it would be nice if we handle `STOP` (`0x00`) instead, and then interpret out-of-range offsets as zero, to match the EVM behavior. The tests could still terminate by jumping to `0xdeadbeef`, since it should always be out of range (unless our kernel grows to ~4mb), resulting in `STOP`.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -0,0 +1,301 @@
...
+        let x = self.pop();
+        let y = self.pop();
+        self.push(x.overflowing_pow(y).0);
+        self.incr(1);
+    }
+
+    fn run_lt(&mut self) {
+        let x = self.pop();
+        let y = self.pop();
+        self.push(if x < y { U256::one() } else { U256::zero() });
```
dlubarov, line 190: nit: I think a `self.push_bool` may be nice since we'll have several

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -0,0 +1,301 @@
...
+    fn run_pop(&mut self) {
+        self.pop();
+        self.incr(1);
+    }
+
+    fn run_jump(&mut self) {
+        let x = self.pop().as_usize();
+        self.offset = x;
+        if let Some(&landing_opcode) = self.code.get(self.offset) {
+            assert_eq!(landing_opcode, 0x5b, "Destination is not a JUMPDEST.");
```
dlubarov, line 254: Hm technically we're supposed to also check that the 0x5b isn't inside some PUSH data, but it's nontrivial and I guess we don't need to strictly conform here

evm/src/cpu/kernel/interpreter.rs
wborgeaud: Hm can you jump to an out-of-range location in the EVM? It would seem that this would panic since an out-of-range location can't be a JUMPDEST. 
But yes it's a good idea to support STOP. I thought about adding a global function that we can call in tests to stop execution and examine the stack
```
global stop:
  JUMPDEST
  STOP
```
but one issue is that we probably don't want this function in the production code (?) so we'd have to add it only for tests.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -0,0 +1,301 @@
...
+        let x = self.pop();
+        let y = self.pop();
+        self.push(x.overflowing_sub(y).0);
+        self.incr(1);
+    }
+
+    fn run_div(&mut self) {
+        let x = self.pop();
+        let y = self.pop();
+        self.push(x / y);
```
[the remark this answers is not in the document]
wborgeaud: Oh no it doesn't, good catch!

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -0,0 +1,301 @@
...
+        self.push(U256([0, 1, 2, 3].map(|i| res.0[i])));
+        self.incr(1);
+    }
+
+    fn run_mulmod(&mut self) {
+        let x = self.pop();
+        let y = self.pop();
+        let z = U512::from(self.pop());
+        let res = x.full_mul(y) % z;
+        self.push(U256([0, 1, 2, 3].map(|i| res.0[i])));
```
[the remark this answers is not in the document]
wborgeaud: I don't see a `from` but there's a `try_from`. I changed to use it.

evm/src/cpu/kernel/interpreter.rs
wborgeaud: Interesting, I didn't know this was also checked. Shouldn't be too hard to add with a preprocessing step. I'll do it in a later PR.

PR #598, opened 2022-07-07
Allow constants to be passed from Rust into our assembly

dlubarov:
Roughly like environment variables. So we don't have to declare things like segment IDs twice.


wborgeaud (approved):
Looks good to me!

PR #599, opened 2022-07-07
Separate timestamps per memory operation

npwardberkeley:


evm/src/memory/columns.rs
```diff
@@ -36,20 +38,23 @@ pub(crate) const CONTEXT_FIRST_CHANGE: usize = SORTED_VALUE_START + VALUE_LIMBS;
 pub(crate) const SEGMENT_FIRST_CHANGE: usize = CONTEXT_FIRST_CHANGE + 1;
 pub(crate) const VIRTUAL_FIRST_CHANGE: usize = SEGMENT_FIRST_CHANGE + 1;
 
+// Flags to indicate if this operation came from the `i`th channel of the memory bus.
+const IS_CHANNEL_START: usize = VIRTUAL_FIRST_CHANGE + 1;
+#[allow(dead_code)]
```
dlubarov, line 43: Doesn't seem like there's dead code?

evm/src/memory/memory_stark.rs
```diff
@@ -200,27 +201,34 @@ pub fn generate_range_check_value<F: RichField>(
...
     let num_ops = context.len();
     let mut range_check = Vec::new();
 
+    let mut max_timestamp_diff = 0;
     for idx in 0..num_ops - 1 {
         let this_address_unchanged = F::ONE
             - context_first_change[idx]
             - segment_first_change[idx]
             - virtual_first_change[idx];
+        let timestamp_diff = timestamp[idx + 1] - timestamp[idx] - F::ONE;
```
dlubarov, line 214: I think we could also have cases where the largest range check is for a virtual address delta, rather than a timestamp delta. Contexts and segments shouldn't have very large deltas, but might be best to consider those deltas also just in case.

I think the easiest way to find the largest delta (for any field) would be to do `let max_diff = range_check.iter().max();` at the end.

evm/src/memory/memory_stark.rs
```diff
@@ -304,14 +315,16 @@ impl<F: RichField + Extendable<D>, const D: usize> MemoryStark<F, D> {
             &segment_first_change,
             &virtual_first_change,
         );
+        let to_pad_to = max_timestamp_diff.next_power_of_two().max(num_trace_rows);
```
dlubarov, line 318: I think we may need to use `max_timestamp_diff + 1`, to account for zero. Like if the max diff is `d`, we want to support `0..=d` as range check inputs, which is `d + 1` values.

evm/src/memory/columns.rs
npwardberkeley: Ah yes!

evm/src/memory/memory_stark.rs
```diff
@@ -321,17 +334,23 @@ impl<F: RichField + Extendable<D>, const D: usize> MemoryStark<F, D> {
...
             permuted_cols(&trace_cols[RANGE_CHECK], &trace_cols[COUNTER]);
         trace_cols[RANGE_CHECK_PERMUTED] = permuted_inputs;
         trace_cols[COUNTER_PERMUTED] = permuted_table;
     }
 
-    pub fn generate_trace(&self, memory_ops: Vec<MemoryOp<F>>) -> Vec<PolynomialValues<F>> {
+    pub fn generate_trace(
+        &self,
+        memory_ops: Vec<MemoryOp<F>>,
+    ) -> (Vec<PolynomialValues<F>>, usize) {
```
dlubarov, line 353: Could the caller get the number of ops based on the length of the first polynomial instead? Would prefer to avoid changing this method signature if we can, since it matches similar methods for the other tables.

evm/src/memory/memory_stark.rs
npwardberkeley: Oh wow yes, that's much better!

evm/src/memory/memory_stark.rs
```diff
@@ -409,15 +434,21 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F
             .constraint_transition(virtual_first_change * (next_addr_context - addr_context));
         yield_constr
             .constraint_transition(virtual_first_change * (next_addr_segment - addr_segment));
-        yield_constr.constraint_transition(address_unchanged * (next_addr_context - addr_context));
-        yield_constr.constraint_transition(address_unchanged * (next_addr_segment - addr_segment));
-        yield_constr.constraint_transition(address_unchanged * (next_addr_virtual - addr_virtual));
+        yield_constr.constraint_transition(
```
dlubarov, line 437: Hmm I'm not sure if this is safe, it seems like our ordering could be "broken" by some invalid rows in between valid rows. E.g.

- valid row, time=6
- invalid row, time=8
- invalid row, time=2
- valid row, time=4

Maybe there's a way to make padding rows satisfy the regular constraints, e.g. we could record the last "real" memory timestamp, and add padding `MemoryOp`s at the end which start off at that timestamp. The padding `MemoryOp`s could perform some read that we know will succeed, e.g. repeating the last real read, or reading just beyond the last used address, which must be zero by default.

Or if not, we could add constraints to force padding rows to be together in a contiguous block, either at the beginning or the end.

In any case this seems nontrivial, so we could add a TODO for later.

evm/src/memory/memory_stark.rs
```diff
@@ -556,7 +600,24 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F
...
+            SORTED_TIMESTAMP,
+            SORTED_IS_READ,
+            SORTED_ADDR_CONTEXT,
+            SORTED_ADDR_SEGMENT,
+            SORTED_ADDR_VIRTUAL,
+        ];
+        sorted_cols.extend((0..VALUE_LIMBS).map(sorted_value_limb));
+        let column_pairs: Vec<_> = unsorted_cols
+            .iter()
+            .cloned()
```
dlubarov, line 608: nit: I think we can avoid cloning if we consume the `Vec`s with `into_iter()`

evm/src/memory/memory_stark.rs
dlubarov: Also is `num_trace_rows` guaranteed to be a power of two here? If not we could move `next_power_of_two()` after the `max`

evm/src/memory/memory_stark.rs
npwardberkeley: Oh yes I was wondering about that, good point

evm/src/memory/memory_stark.rs
npwardberkeley: Hm well for now I think the only padding rows are entirely 0 values, except for the COUNTER column (and the COUNTER_PERMUTED and RANGE_CHECK_PERMUTED), and they're all at the beginning!

dlubarov: Yeah our prover code puts them at the beginning, but seems like it'd be possible for a malicious prover to put them in the middle? If we wanted to enforce it we could constraint `is_valid` to have deltas in `{0, 1}`

evm/src/memory/memory_stark.rs
npwardberkeley: Oh right!

dlubarov (approved):
LGTM, besides the `next_power_of_two` question and the TODO about `is_valid`

evm/src/memory/memory_stark.rs
npwardberkeley: Ohh right

npwardberkeley: Yeah that sounds great, thanks!

PR #600, opened 2022-07-09
Transaction (RLP) parsing

dlubarov:


dlubarov:
I changed approaches so this is obsolete; will make another for the new RLP branch later.

PR #601, opened 2022-07-09
`U32ArithmeticGate`: constrain canonicity of output.

### LINKED ISSUE #526: `U32ArithmeticGate` permits non-canonical outputs
For example, when computing `0 * 0 + 0`, it seems outputs of either `0` or `p` (our native field order) would pass the constraints.

The largest permissible element is `p - 1 = 0xFFFFFFFF00000000`, or in `u32` limbs, it's `(u32::MAX, 0)`. So we can check `x < p` as `(x_lo == 0) OR (x_hi != u32::MAX)`; see this [code](https://github.com/mir-protocol/plonky2/blob/main/system_zero/src/alu/canonical.rs) in Starky.

therealyingtong:
Closes #526. This PR introduces a canonicity check for the `U32ArithmeticGate` output, along with a test using a non-canonical output. (Best reviewed commit-by-commit.)

### DIFF
--- a/u32/src/gates/arithmetic_u32.rs
+++ b/u32/src/gates/arithmetic_u32.rs
@@ -36,31 +36,30 @@ impl<F: RichField + Extendable<D>, const D: usize> U32ArithmeticGate<F, D> {
     }
 
     pub(crate) fn num_ops(config: &CircuitConfig) -> usize {
-        let wires_per_op = 5 + Self::num_limbs();
-        let routed_wires_per_op = 5;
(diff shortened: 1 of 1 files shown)


dlubarov:
Thank you for the fix! The code looks great 👍 

Typically we would store inverses like this in advice wires, though I think it's fine in this case since most configurations will have more routed wires than this gate uses, and the way you have it seems slightly nicer.

Since we plan to add an open-source license in the future, is it okay to treat your contributions as MIT-licensed? If so I think it's good to merge.

therealyingtong:
Thank you! Yes, MIT-licensed is good.

PR #602, opened 2022-07-11
Better CTL errors

wborgeaud:
Provide better errors when the CTL check fails by manually performing the multiset check.

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -682,6 +682,9 @@ pub(crate) mod testutils {
             looked_table,
             default,
         } = ctl;
+
+        // Maps `m` with `(table, i) in m[row]` iff the `i`-th row of `table` is equal to `row` and
+        // the filter is 1. Without default values, the CTL check holds iff `looking_multiset == looked_multiset`.
         let mut looking_multiset = MultiSet::<F>::new();
         let mut looked_multiset = MultiSet::<F>::new();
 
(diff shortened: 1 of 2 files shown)


evm/src/cross_table_lookup.rs
```diff
@@ -649,3 +649,123 @@ pub(crate) fn verify_cross_table_lookups_circuit<
...
+            if let Some(default) = default {
+                if row == default {
+                    *extra_default_count.as_mut().unwrap() +=
+                        looking_locations.len() - looked_locations.len();
+                    continue;
+                }
+            }
+            check_locations(looking_locations, looked_locations, ctl_index, row);
+        }
+        // Check that the number of extra default rows is correct.
```
dlubarov, line 711: It seems like this could fail if the default row occurs naturally (with a matching row in the looked table). I would suggest adding the table size delta to `looked_multiset[default]` and then proceeding with the normal multiset comparison.

dlubarov (approved):
LGTM otherwise

evm/src/cross_table_lookup.rs
dlubarov: Oh I see, I misread the `extra_default_count` calculation.

Seems right then, though I think it'd be slightly easier to read if we calculate it upfront instead of in the loop, like
```rust
        let extra_default_count = default.as_ref().map(|d| {
            let looking_defaults = looking_multiset.get(d).unwrap_or(empty).len();
            let looked_defaults = looked_multiset.get(d).unwrap_or(empty).len();
            looking_defaults - looked_defaults
        });
```

wborgeaud: Good idea, done!

PR #603, opened 2022-07-11
Simplify memory table

dlubarov:
By no longer storing unsorted operations; they are effectively stored in the CPU table already.

I ran into some issues with sorting, since the existing sort method didn't include `is_channel` columns. Rather than update the existing method, I removed it and added a sort on the `MemoryOp`s, which I think seems cleaner.

### DIFF
--- a/evm/src/memory/columns.rs
+++ b/evm/src/memory/columns.rs
@@ -4,7 +4,7 @@ use std::ops::Range;
 
 use crate::memory::{NUM_CHANNELS, VALUE_LIMBS};
 
-// Columns for the same memory operations, ordered by (addr, timestamp).
+// Columns for memory operations, ordered by (addr, timestamp).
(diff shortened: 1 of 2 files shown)


evm/src/memory/columns.rs
```diff
@@ -4,6 +4,7 @@ use std::ops::Range;
 
 use crate::memory::{NUM_CHANNELS, VALUE_LIMBS};
 
+// Columns for the same memory operations, ordered by (addr, timestamp).
```
npwardberkeley, line 7: Nit: "same memory operations" is out of date

npwardberkeley (approved):
Looks great, thanks!! Much cleaner

evm/src/memory/columns.rs
dlubarov: Good catch

PR #604, opened 2022-07-12
Use FFTs to get subgroup evaluations in `check_constraints`

dlubarov:
Instead of quadratic evaluation. Should speed up `test_all_stark`.


wborgeaud (approved):
👍

PR #605, opened 2022-07-13
More realistic padding rows in memory table

dlubarov:
This adds padding rows which satisfy the ordering checks. To ensure that they also satisfy the value consistency checks, I just copied the address and value from the last operation.

I think this method of padding feels more natural, though it is a big more code since we need to calculate the max range check in a different way. But on the plus side, the constraints are a bit smaller and simpler.

Also added a few constraints that I think we need for soundness:
- Each `is_channel` flag is bool.
- Sum of `is_channel` flags is bool.
- Dummy operations must be reads (otherwise the prover could put writes in the memory table which aren't in the CPU table).


npwardberkeley (approved):
Looks great!

PR #606, opened 2022-07-13
Fix interpreter JUMPDEST check + change stopping behavior

wborgeaud:
Check that JUMPDEST is not in push data and implement STOP opcode.


dlubarov (approved):
LGTM

PR #608, opened 2022-07-13
Have `make_kernel` log the size of each (assembled) file

dlubarov:
For now it doesn't log filenames, but we can compare against the list of filenames in `combined_kernel`.

Current output:
```
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 0 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 49 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 387 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 27365 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 0 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 11 bytes
[DEBUG plonky2_evm::cpu::kernel::assembler] Assembled file size: 7 bytes
[DEBUG plonky2_evm::cpu::kernel::aggregator::tests] Total kernel size: 27819 bytes
```

This shows that most of our kernel code is from `curve_add.asm`, which makes sense since it invovles a couple uses of the large `inverse` macro.  Thankfully that will be replaced at some point.

### DIFF
--- a/evm/src/cpu/kernel/aggregator.rs
+++ b/evm/src/cpu/kernel/aggregator.rs
@@ -36,6 +36,7 @@ mod tests {
 
     use anyhow::Result;
     use ethereum_types::U256;
+    use log::debug;
     use rand::{thread_rng, Rng};
 
     use crate::cpu::kernel::aggregator::combined_kernel;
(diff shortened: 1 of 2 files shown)


wborgeaud (approved):
👍

PR #609, opened 2022-07-13
Generate most of the memory table while it's in row-wise form

dlubarov:
This should improve cache locality - since we generally access several values at a time in a given row, we want them to be close together in memory. Also avoids the extra transposes we were doing previously.

There are a few steps that make more sense column-wise, though, such as generating the `COUNTER` column. I put those after the transpose.


npwardberkeley (approved):
Wow, much cleaner! Thanks

PR #611, opened 2022-07-14
`ecrecover` kernel function

wborgeaud:
Implement the `ecrecover` kernel function minus ε.
ε = Final Keccak hashing of the public key to get the address. Will do that in a later PR.


evm/src/cpu/kernel/asm/ecrecover.asm
```diff
@@ -0,0 +1,193 @@
...
+    MOD
+    // stack: hash%p, r^(-1), Y, X, retdest
+    %secp_scalar
+    // stack: p, hash%p, r^(-1), Y, X, retdest
+    SUB
+    // stack: -hash, r^(-1), Y, X, retdest
+    %mulmodn_secp_scalar
+    // stack: u2, Y, X, retdest
+
+    // Compute u2 * GENERATOR and chain the call to `ec_mul` with a call to `ec_add` to compute PUBKEY = (X,Y) + u2 * GENERATOR,
```
wborgeaud, line 97: So here `PUBKEY` is computed like 
```
(X,Y) = u1 * P
(X', Y') = u2 * GENERATOR
PUBKEY = (X,Y) + (X',Y')
```
It would probably be more efficient to compute this using a MSM function. If we end up merging the `ec_mul` functions for Secp256k1 and BN254 then the MSM function would add a bunch of kernel code, so the tradeoffs are a bit unclear for now.

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -0,0 +1,825 @@
...
+    ADDMOD
+%endmacro
+
+// Returns sqrt(x). Computed as x^(q+1)/4, with q the Secp base field order.
+/// To replace with more efficient method using non-determinism later.
+%macro sqrt_secp_base
+    // stack: x
+    DUP1
+    %squaremodn_secp_base
+    DUP2
```
dlubarov, line 77: I just made #613 to add a `%rep` syntax, if it looks good to you maybe you could hit merge and use it here

evm/src/cpu/kernel/asm/secp256k1/curve_add.asm
```diff
@@ -0,0 +1,338 @@
+// #define N 0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141 // Secp256k1 scalar field order
+
+// Secp256k1 elliptic curve addition.
+// Uses the standard affine addition formula.
+global ec_add_secp:
```
dlubarov, line 5: Hm not sure if we'll use this, since there isn't a precompile for secp add? But we can keep it for now if you like

evm/src/cpu/kernel/asm/secp256k1/curve_add.asm
```diff
@@ -0,0 +1,338 @@
...
+    // stack: 3/2, x0^2, N, x0, y0, x1, y1, retdest
+    MULMOD
+    // stack: 3/2 * x0^2, x0, y0, x1, y1, retdest
+    DUP3
+    // stack: y0, 3/2 * x0^2, x0, y0, x1, y1, retdest
+    %moddiv_secp_base
+    // stack: lambda, x0, y0, x1, y1, retdest
+    %jump(ec_add_valid_points_with_lambda)
+
+// BN254 elliptic curve doubling.
```
dlubarov, line 235: Comment out of date

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
wborgeaud: So I generate these exponentiation functions with a small Python script:
```python
for b in reversed(exponent.bits()[:-1]):
    print("%squaremodn")
    if b:
        print("DUP2\n%mulmodn")
```
Using `%rep` would make this script slightly more complicated and shouldn't change much since these functions should go away once we have non-determinism. Is there another reason to use `%rep` that I'm missing?

evm/src/cpu/kernel/asm/secp256k1/curve_add.asm
wborgeaud: Do you mean we delete this one and keep only `ec_add_valid_points_secp`?

dlubarov: Yeah that's what I mean

evm/src/cpu/kernel/asm/secp256k1/curve_add.asm
```diff
@@ -0,0 +1,338 @@
...
+    // stack: y, x
+    DUP2
+    // stack: x, y, x
+    SUB
+    // stack: x - y, x
+    DUP1
+    // stack: x - y, x - y, x
+    SWAP2
+    // stack: x, x - y, x - y
+    LT
```
dlubarov, line 264: This makes sense (after seeing the method in @nbgl's SIMD code) but please add a brief note, since it's not obvious that this is an underflow check

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
dlubarov: Ok - we can keep it as is for now

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -0,0 +1,825 @@
+// Returns y such that (x,y) is on Secp256k1 and y&1 = v - 27,
```
dlubarov, line 1: Hm I think we'll also want to use this when verifying newer transactions which don't use the `v = 27 + y_parity` format. Might be best to take `y_parity` as a bool here, and move `v - 27` to some code that's specifically for the ECREC precompile? Or we can tweak it later when we add transaction verification.

wborgeaud: Good point, done.

evm/src/cpu/kernel/asm/secp256k1/curve_add.asm
wborgeaud: Ok I deleted it and also `ec_mul_secp` since it's also unused.

wborgeaud: Ok I deleted it and also `ec_mul_secp` since it's also unused.

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -0,0 +1,817 @@
...
+%macro addmodn_secp_base
+    // stack: x, y
+    %secp_base
+    // stack: N, x, y
+    SWAP2
+    // stack: y, x, N
+    ADDMOD
+%endmacro
+
+// Returns sqrt(x). Computed as x^(q+1)/4, with q the Secp base field order.
```
dlubarov, line 63: To clarify a bit more, can we say it returns one of the square roots if it exists, otherwise an undefined value?

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -0,0 +1,817 @@
...
+    // stack: 1, y, parity, y, sqrtOk
+    AND
+    // stack: 1 & y, parity, y, sqrtOk
+    EQ
+    // stack: correctParity, y, sqrtOk
+    DUP2
+    // stack: y, correctParity, y, sqrtOk
+    %secp_base
+    // stack: N, y, correctParity, y, sqrtOk
+    SUB
```
dlubarov, line 37: Seems like this could give a non-canonical `-y` if `y=0`. But I guess it would be ignored since `sqrtOk` will be false in that case, correct?

dlubarov: Oh I see, you also check `r != 0` in `ecrecover_input_check`

evm/src/cpu/kernel/asm/ecrecover.asm
dlubarov: Yeah seems fine for now 👍

dlubarov (approved):
LGTM

PR #612, opened 2022-07-14
Continue work on bootstrapping

dlubarov:
The kernel is hashed using a Keccak based sponge for now. We could switch to Poseidon later if our kernel grows too large.

Note that we use simple zero-padding (pad0*) instead of the standard pad10* rule. It's simpler, and we don't care that the prover can add extra 0s at the end of the code. The program counter can never reach those bytes, and even if it could, they'd be 0 anyway given the EVM's zero-initialization rule.

In one CPU row, we can do a whole Keccak hash (via the CTL), absorbing 136 bytes. But we can't actually bootstrap that many bytes of kernel code in one row, because we're also limited by memory bandwidth. Currently we can write 4 bytes of the kernel to memory in one row.

So we treat the `keccak_input_limbs` columns as a buffer. We gradually fill up this buffer, 4 bytes (one `u32` word) at a time. Every `136 / 4 = 34` rows, the buffer will be full, so at that point we activate the Keccak CTL to absorb the buffer.

### DIFF
--- a/evm/src/cpu/bootstrap_kernel.rs
+++ b/evm/src/cpu/bootstrap_kernel.rs
@@ -92,6 +92,7 @@ pub(crate) fn eval_bootstrap_kernel<F: Field, P: PackedField<Scalar = F>>(
     // - If CLOCK is a multiple of KECCAK_RATE_LIMBS, activate the Keccak CTL, and ensure the output
     //   is copied to the next row (besides the first limb which will immediately be overwritten).
     // - Otherwise, ensure that the Keccak input is copied to the next row (besides the next limb).
+    // - The next limb we add to the buffer is also written to memory.
 
     // If IS_BOOTSTRAP_KERNEL changed (from 1 to 0), check that
     // - the clock is a multiple of KECCAK_RATE_LIMBS (TODO)
@@ -131,6 +132,7 @@ pub(crate) fn eval_bootstrap_kernel_circuit<F: RichField + Extendable<D>, const 
     // - If CLOCK is a multiple of KECCAK_RATE_LIMBS, activate the Keccak CTL, and ensure the output
     //   is copied to the next row (besides the first limb which will immediately be overwritten).
     // - Otherwise, ensure that the Keccak input is copied to the next row (besides the next limb).
+    // - The next limb we add to the buffer is also written to memory.
 
     // If IS_BOOTSTRAP_KERNEL changed (from 1 to 0), check that
     // - the clock is a multiple of KECCAK_RATE_LIMBS (TODO)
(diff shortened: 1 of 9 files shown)


evm/src/cpu/bootstrap_kernel.rs
```diff
@@ -1,33 +1,74 @@
...
+
+        // Write this chunk to memory, while simultaneously packing its bytes into a u32 word.
+        let mut packed_bytes: u32 = 0;
         for (addr, byte) in chunk {
             let mut value = [F::ZERO; memory::VALUE_LIMBS];
             value[0] = F::from_canonical_u8(byte);
 
-            let channel = addr % memory::NUM_CHANNELS;
+            let channel = addr % NUM_CHANNELS;
             state.set_mem_current(channel, segments::CODE, addr, value);
```
wborgeaud, line 56: Probably a stupid idea, but couldn't we pack opcodes (in the kernel and elsewhere) by 4, so that the memory taken by the code is divided by 4. Though we'd need to do some annoying unpacking when executing the opcodes, but could be useful if the memory table becomes a bottleneck.

evm/src/cpu/bootstrap_kernel.rs
```diff
@@ -36,30 +77,48 @@ pub(crate) fn eval_bootstrap_kernel<F: Field, P: PackedField<Scalar = F>>(
...
     yield_constr.constraint_transition(delta_is_bootstrap * (delta_is_bootstrap + P::ONES));
 
-    // If IS_BOOTSTRAP_KERNEL changed (from 1 to 0), check that the current kernel hash matches a
-    // precomputed one.
-    let hash_diff = F::ZERO; // TODO
-    yield_constr.constraint_transition(delta_is_bootstrap * hash_diff)
+    // TODO: Constraints to enforce that, if IS_BOOTSTRAP_KERNEL,
+    // - If CLOCK is a multiple of KECCAK_RATE_LIMBS, activate the Keccak CTL, and ensure the output
+    //   is copied to the next row (besides the first limb which will immediately be overwritten).
+    // - Otherwise, ensure that the Keccak input is copied to the next row (besides the next limb).
```
wborgeaud, line 94: Do we also need to check that the new element in `keccak_input_limbs` matches the packed opcodes added to memory?

wborgeaud (approved):
Looks good to me!

evm/src/cpu/bootstrap_kernel.rs
dlubarov: I think it's a very reasonable idea :) @nbgl and I discussed it a while back, and wanted to hold off for now just to keep it simple, but we could do it if bootstrapping ends up being a bottleneck. I think there's a tradeoff since unpacking would add some cost (unless we pipeline it so unpacking happens over 4 rows), but definitely worth considering.

evm/src/cpu/bootstrap_kernel.rs
dlubarov: Good point, added a bullet for it.

PR #614, opened 2022-07-15
Implement memory for the EVM interpreter

wborgeaud:


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -306,13 +365,40 @@ fn find_jumpdests(code: &[u8]) -> Vec<usize> {
...
+        //         PUSH1 0x27
+        //         MSTORE8
+        let code = vec![
+            0x60, 0xff, 0x60, 0x0, 0x52, 0x60, 0, 0x51, 0x60, 0x1, 0x51, 0x60, 0x42, 0x60, 0x27,
+            0x53,
+        ];
+        let run = run(&code, 0, vec![]);
+        let Interpreter { stack, memory, .. } = run;
+        assert_eq!(stack, vec![0xff.into(), 0xff00.into()]);
+        assert_eq!(&memory.memory, &hex!("00000000000000000000000000000000000000000000000000000000000000ff0000000000000042000000000000000000000000000000000000000000000000"));
```
dlubarov, line 402: It seems like the 42 is in `memory[39]` in the hex, but shouldn't it be at `memory[27]`? I'm probably missing something since the code looks right

dlubarov (approved):
LGTM. I think later we'll want to support other (kernel-owned) segments, and it's probably best to just reuse `MemoryState` then. (It would include the stack; we could add a convenience method for tests to get the last context's stack.) But we can generalize it later.

evm/src/cpu/kernel/interpreter.rs
wborgeaud: It should be at location 0x27 = 39.

dlubarov: Ah of course :)

PR #616, opened 2022-07-16
Store memory values as `U256`s

dlubarov:
Ultimately they're encoded as `[F; 8]`s in the table, but I don't anticipate that we'll have any use cases where we want to store more than 256 bits. Might as well store `U256` until we actually build the table since they're more compact.


npwardberkeley (approved):
Looks good! I agree it seems slightly more meaningful this way

PR #617, opened 2022-07-16
Organize segments in an enum

dlubarov:
It's a bit more type-safe (can't mix up segment with context or virtual addr), and this way uniqueness of ordinals is enforced, partially addressing a concern raised in #591.

To avoid making `Segment` public (which I don't think would be appropriate), I had to make some other visibility changes, and had to move `generate_random_memory_ops` into the test module.

### DIFF
--- a/evm/src/memory/segments.rs
+++ b/evm/src/memory/segments.rs
@@ -1,22 +1,61 @@
-/// Contains EVM bytecode.
-pub const CODE: usize = 0;
+#[allow(dead_code)] // TODO: Not all segments are used yet.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
+pub(crate) enum Segment {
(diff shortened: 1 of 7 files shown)


npwardberkeley (approved):
I like it!

evm/src/memory/segments.rs
```diff
@@ -1,22 +1,61 @@
...
+    /// Contains transaction data (after it's parsed and converted to a standard format).
+    TxnData = 7,
+    /// Raw RLP data.
+    RlpRaw = 8,
+    /// RLP data that has been parsed and converted to a more "friendly" format.
+    RlpParsed = 9,
+}
 
-pub const STACK: usize = 1;
+impl Segment {
```
nbgl, line 28: Oh! I didn’t know you could have an `impl` for an `enum`! :O

nbgl (approved):
This is much neater!

PR #618, opened 2022-07-17
More basic ASM macros

dlubarov:
To be used in upcoming RLP code.


wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/asm/assertions.asm
```diff
@@ -0,0 +1,82 @@
+// It is convenient to have a single panic routine, which we can jump to from
+// anywhere.
+global panic:
+    JUMPDEST
+    PANIC
```
wborgeaud, line 5: Could also be
```
global panic:
  STOP
```
which would also panic (because destination isn't a JUMPDEST) without having to add a new opcode. Debug info would be worse though.

dlubarov: Discussed in #619; I'll update it later if we decide to give invalid jumps the same behavior

PR #619, opened 2022-07-17
Add custom opcodes

dlubarov:
- `GET_STATE_ROOT` and `SET_STATE_ROOT` deal with the root of the state trie, and will be called from storage routines. Similarly `GET_RECEIPT_ROOT` and `SET_RECEIPT_ROOT` deal with the root of the receipt trie.
- `PANIC` enables an unsatisfiable constraint, so no proof can be generated.
- `GET_CONTEXT` and `SET_CONTEXT`, used when calling and returning
- `CONSUME_GAS` charges the sender gas; useful for cases where gas calculations are nontrivial and best implemented in assembly.
- `EXIT_KERNEL` simply clears the CPU flag indicating that we're in kernel mode; it would be used just before a jump to return to the (userspace) caller.
- `MLOAD_GENERAL` and `MSTORE_GENERAL` are for reading and writing memory, but they're not limited to the main memory segment of the current context; they can access any context and any segment. I added a couple macros to show how the they would typically be used.

There may be more later, but these are the ones I think we need for now.  I tried to fill in smaller invalid sections of the decoder's tree, as Jacqui suggested, while keeping related opcodes together. We can fine tune it when the opcode list is more stable.

These are all intended to be priviledged, i.e. they will be treated as invalid if used from userspace, for compatibility as well as (in some cases) security reasons.


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -123,13 +127,20 @@ impl<'a> Interpreter<'a> {
             0xa2 => todo!(),                                           // "LOG2",
             0xa3 => todo!(),                                           // "LOG3",
             0xa4 => todo!(),                                           // "LOG4",
+            0xa5 => panic!("Executed PANIC"),                          // "PANIC",
```
nbgl, line 130: I don’t think a program panicking should cause the _interpreter_ to panic. A panicking program is a perfectly valid input. Maybe consider returning a `Result`?

nbgl (approved):

evm/src/cpu/kernel/interpreter.rs
dlubarov: Hm fair point. Changed it to return a `Result`.

evm/src/cpu/kernel/asm/memory.asm
```diff
@@ -0,0 +1,27 @@
+// Load a byte from the given segment of the current context's memory space.
+// Note that main memory values are one byte each, but in general memory values
+// can be 256 bits. This macro deals with a single address (unlike MSTORE), so
+// if it is used with main memory, it will load a single byte.
+%macro mload_current(segment)
```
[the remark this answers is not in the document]
dlubarov: I see how this could be confusing, since it sounds like `MLOAD` but doesn't have the same behavior of combining 32 values. But I think `MLOAD8` could also be confusing, since certain segments (like the stack) contain values larger than 8 bits, unlike main memory. E.g. `%mload_current(@SEGMENT_STACK)` could give 256 bits. (Not sure if we'd ever have a reason to use this macro with the stack, but we could also use it with general Kernel memory which could contain large values.)

Maybe we could call it `%mload_single_current` to emphasize that it deals with a single address, but it might be nice to keep it short since it may be used a lot.

dlubarov:
> When would we need to set the state/receipt roots? I would have thought the new roots would be given as public inputs. Or are we updating the roots after each storage/logging operation?

I was thinking we'd have public inputs for the initial and final roots, but also have registers for roots of the trie's intermediate states. We'd update those registers after observing a valid Merkle proof.

> Also, what about the transaction trie?

I think it may be okay to omit it (i.e. only prove the state root and receipt root), at least for now, since it can't be used to steal funds from the bridge. But let me know if you think it's important.

If we do end up needing to compute transaction hashes, it will just make transaction RLP logic slightly more complex, since we'll need the RLP of both the full transaction (including signature fields) and the signed data.

> As mentioned in #618, couldn't we just jump to a non JUMPDEST location instead of adding this opcode?

Hm, my goal with `PANIC` was to trigger a "non-provable" error, in contrast to reverts where there's still a well-defined post-state that can be proven. We could handle invalid jumps differently in user mode and kernel mode, but I think a dedicated opcode for non-provable errors should make the semantics a bit more clear.

The other thing is, since kernel code is bootloaded by "hardware" logic, I was hoping to skip the the `JUMPDEST` analysis to keep the constraints simple. We definitely need `JUMPDEST` analysis for user code, but I was thinking user code could be bootloaded by some kernel assembly.

> Is this CPU flag already implemented?

Not yet, I can add it soon.

PR #620, opened 2022-07-18
Implement SHA3 in interpreter and use it in ecrecover

wborgeaud:


evm/src/cpu/kernel/opcodes.rs
```diff
@@ -35,6 +35,7 @@ pub(crate) fn get_opcode(mnemonic: &str) -> u8 {
         "SHR" => 0x1c,
         "SAR" => 0x1d,
         "KECCAK256" => 0x20,
+        "SHA3" => 0x20,
```
wborgeaud, line 38: I haven't been able to find out what's the consensus for the 0x20 mnemonic.

Sources using `SHA3`:
- Official Ethereum website https://ethereum.org/en/developers/docs/evm/opcodes/
- OpenEthereum
- evm.codes
- ethervm.io

Sources using `KECCAK256`:
- Yellow paper
- Geth

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -119,7 +125,7 @@ impl<'a> Interpreter<'a> {
             0x1b => todo!(),                                           // "SHL",
             0x1c => todo!(),                                           // "SHR",
             0x1d => todo!(),                                           // "SAR",
-            0x20 => todo!(),                                           // "KECCAK256",
+            0x20 => self.run_sha3(),                                   // "SHA3",
```
dlubarov, line 128: nit: I think `KECCAK256` is the preferred name, the yellowpaper changed the mnemonic from `SHA3` at some point.

evm/src/cpu/kernel/opcodes.rs
dlubarov: I guess there was never a formal decision about it (EIP-1803 proposed the rename but it says stagnant), but it seems like `KECCAK256` is generally preferred, and it's just taking a while for all the `SHA3` uses to go away. The yellowpaper recently switched to `KECCAK256`, and that seems most authoritative IMO.

evm/src/cpu/kernel/asm/ecrecover.asm
```diff
@@ -107,33 +107,53 @@ ecrecover_with_first_point:
...
-// TODO
-final_hashing:
+// Take a public key (PKx, PKy) and return the associated address SHA3(PKx || PKy)[-20:].
+pubkey_to_addr:
     JUMPDEST
-    PUSH 0xdeadbeef
+    // stack: PKx, PKy, retdest
+    PUSH 0
+    // stack: 0, PKx, PKy, retdest
+    MSTORE
```
dlubarov, line 139: Let's add a TODO to switch to kernel memory (like `%mstore_current(@SEGMENT_KERNEL_GENERAL)`) once it's supported.

dlubarov (approved):
LGTM

PR #621, opened 2022-07-18
Shared CPU columns

dlubarov:
I was thinking we could have two sets of shared columns:
- First, a set of "core" columns which would contain instruction decoding registers during an execution cycle, or some counter data during a kernel bootloading cycle.
- Second, a set of "general" columns which would be more general-purpose. For now it could contain "looking" columns for most CTLs (Keccak, arithmetic and logic; NOT memory since memory can be used simultaneously with the others). It could potentially be reused for other things too, such as the registers used for `EQ` and `IS_ZERO` (but I know it's nontrivial to share those since we would need to use lower-degree constraints, so I wouldn't bother for now).

This PR implements just the latter. If it looks good I'll proceed with the former afterward.


evm/src/cpu/columns/mod.rs
```diff
@@ -2,14 +2,18 @@
...
 
+use crate::cpu::columns::shared::CpuSharedColumnsView;
 use crate::memory;
 
+mod shared;
+
 #[repr(C)]
 #[derive(Eq, PartialEq, Debug)]
-pub struct CpuColumnsView<T> {
+pub struct CpuColumnsView<T: Copy> {
```
dlubarov, line 16: I think we need this to deal with dropping `CpuGeneralColumnsView`, since it's union, unless we wrap stuff in `ManuallyDrop`.

evm/src/cpu/columns/general.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::fmt::{Debug, Formatter};
+use std::mem::{size_of, transmute};
+
+/// General purpose columns, which can have different meanings depending on what CTL or other
+/// operation is occurring at this row.
+pub(crate) union CpuGeneralColumnsView<T: Copy> {
+    keccak: CpuKeccakView<T>,
+    arithmetic: CpuArithmeticView<T>,
+    logic: CpuLogicView<T>,
+}
```
wborgeaud, line 11: That's really neat 🤯

wborgeaud (approved):
Looks good to me!

nbgl:
I was thinking of doing it in a way that’s quite different. It’s neater in _some_ ways (but also involves some ugly metaprogramming). My flight to the US is soon; I’ll make a PoC on the plane and we can decide which way we prefer :)

PR #622, opened 2022-07-18
Implement memcpy

dlubarov:
This can be used, for example, to copy `CALL` data (which is a slice of the caller's main memory) to the callee's `CALLDATA` segment.

I'll add a test once we update the interpreter to handle other memory segments.

### DIFF
--- a/evm/src/cpu/kernel/asm/memory.asm
+++ b/evm/src/cpu/kernel/asm/memory.asm
@@ -1,4 +1,4 @@
-// Load a byte from the given segment of the current context's memory space.
+// Load a value from the given segment of the current context's memory space.
 // Note that main memory values are one byte each, but in general memory values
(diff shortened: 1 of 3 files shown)


evm/src/cpu/kernel/asm/memory.asm
```diff
@@ -25,3 +25,55 @@
...
+    // Increment src_addr.
+    SWAP5
+    %add_const(1)
+    SWAP5
+    // Decrement count.
+    SWAP6
+    %sub_const(1)
+    SWAP6
+
+    // Recurse!
```
nbgl, line 71: This is kind of a technicality, but this is _tail_-recursion. In assembly, it looks the same as a loop, so I’d just say loop :)

nbgl (approved):

wborgeaud (approved):
👍

PR #624, opened 2022-07-19
Add a `mload_kernel_code_u32` macro

dlubarov:
Intended for loading constants in SHA2, and maybe RIPEMD (cc @typ3c4t).

Sample usage
```
// Loads the i'th K256 constant.
%macro k256
  // stack: i
  %mul_const(4)
  // stack: 4*i
  PUSH k256_data
  // stack: k256_data, 4*i
  ADD
  // stack: k256_data + 4*i
  %mload_kernel_code_u32
  // stack: K256[4*i]
%endmacro

k256_data:
    BYTES 0x42, 0x8a, 0x2f, 0x98
    BYTES 0x71, 0x37, 0x44, 0x91
    ...
```

Untested for now since our interpreter doesn't have the needed memory support quite yet.

### DIFF
--- a/evm/src/cpu/kernel/asm/memory.asm
+++ b/evm/src/cpu/kernel/asm/memory.asm
@@ -26,6 +26,47 @@
     // stack: (empty)
 %endmacro
 
+// Load a single byte from kernel code.
+%macro mload_kernel_code
+    // stack: offset
+    PUSH @SEGMENT_CODE
+    // stack: segment, offset
+    PUSH 0 // kernel has context 0
+    // stack: context, segment, offset
+    MLOAD_GENERAL
+    // stack: value
+%endmacro
+
+// Load a big-endian u32, consisting of 4 bytes (c_3, c_2, c_1, c_0),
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/asm/memory.asm
```diff
@@ -26,6 +26,47 @@
     // stack: (empty)
 %endmacro
 
+// Load a single byte from kernel code.
+%macro mload_kernel_code
```
npwardberkeley, line 30: Just curious, would it be more "idiomatic" to make this take `offset` as an argument? Or maybe that's an undefined notion haha

npwardberkeley (approved):
Looks good!! Thanks a lot

evm/src/cpu/kernel/asm/memory.asm
dlubarov: Hm I don't think we can with the current asm syntax; for now only int literals are allowed as macro args.

PR #625, opened 2022-07-19
Stack manipulation macro

dlubarov:
Uses a variant of Dijkstra's to find a path of instructions between the two stack states. We don't explicitly store the graph though.

The Dijkstra implementation is somewhat inspired by the `pathfinding` crate. That crate doesn't quite fit our needs though.

If we need to make it faster later, there are a lot of allocations and clones that we could probably eliminate.


evm/src/cpu/kernel/stack_manipulation.rs
```diff
@@ -0,0 +1,259 @@
...
+    fn cost(&self) -> u32 {
+        let (cpu_rows, memory_rows) = match self {
+            StackOp::Push(n) => {
+                let bytes = n.to_trimmed_be_bytes().len() as u32;
+                // This is just a rough estimate; we can update it after implementing PUSH.
+                (bytes, bytes)
+            }
+            Pop => (1, 1),
+            StackOp::Dup(_) => (1, 2),
+            StackOp::Swap(_) => (1, 4),
```
wborgeaud, line 219: nit: Maybe add a small explanation for these numbers.

evm/src/cpu/kernel/stack_manipulation.rs
```diff
@@ -0,0 +1,259 @@
...
+    Literal(Literal),
+}
+
+#[derive(Clone, Debug)]
+enum StackOp {
+    Push(Literal),
+    Pop,
+    Dup(u8),
+    Swap(u8),
+}
```
wborgeaud, line 163: Not sure if we'll need it, but could be also cool in the future to support MLOAD either with a constant offset or one present on the stack.

wborgeaud (approved):
Very nice!

evm/src/cpu/kernel/stack_manipulation.rs
dlubarov: Done 👍

evm/src/cpu/kernel/stack_manipulation.rs
dlubarov: Ah like using memory to help rearrange the stack? Could be interesting though using `MLOAD_GENERAL` may be a bit expensive since we need to push the context/segment

PR #626, opened 2022-07-20
A few ASM fixes

dlubarov:


evm/src/cpu/kernel/assembler.rs
```diff
@@ -132,13 +132,29 @@ fn expand_macro_call(
...
+
     let expanded_item = _macro
         .items
         .iter()
         .map(|item| {
             if let Item::Push(PushTarget::MacroVar(var)) = item {
-                let param_index = _macro.get_param_index(var);
-                Item::Push(args[param_index].clone())
+                Item::Push(get_arg(var))
+            } else if let Item::MacroCall(name, args) = item {
```
dlubarov, line 146: When inlining a macro call, we were only looking for vars (to replace with our args) in PUSHs, not in nested macro calls.

evm/src/memory/segments.rs
```diff
@@ -21,12 +21,10 @@ pub(crate) enum Segment {
     TxnData = 7,
     /// Raw RLP data.
     RlpRaw = 8,
-    /// RLP data that has been parsed and converted to a more "friendly" format.
-    RlpParsed = 9,
```
dlubarov, line 25: Ended up not needing this; will make sense with the upcoming RLP branch.

PR #628, opened 2022-07-21
Keccak sponge sha256 config

Sladuca:
Add a config that uses the keccak sponge directly for the challenger and sha256 as the hasher along with a recursion benchmark using it.

### DIFF
--- a/plonky2/src/hash/keccak_sponge_sha256.rs
+++ b/plonky2/src/hash/keccak_sponge_sha256.rs
@@ -27,9 +27,6 @@ impl<F: RichField> PlonkyPermutation<F> for KeccakSpongePermutation {
         // keep squeezingu until we have SPONGE_WIDTH words that fit in the field
         let mut elems = 0;
         while elems < SPONGE_WIDTH {
-            #[cfg(target_os = "solana")]
-            solana_program::keccak_permutation::keccak_permutation(&mut state);
-            #[cfg(not(target_os = "solana"))]
             keccak(&mut state);
 
             for i in 0..SPONGE_WIDTH {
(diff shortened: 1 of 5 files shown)


plonky2/src/hash/keccak_sponge_sha256.rs
```diff
@@ -0,0 +1,72 @@
+use keccak::f1600 as keccak;
+use sha2::{Digest, Sha256};
+
+use crate::hash::hash_types::{BytesHash, RichField};
+use crate::hash::hashing::{PlonkyPermutation, SPONGE_WIDTH};
+use crate::plonk::config::Hasher;
+
+/// Keccak sponge pseudo-permutation (not necessarily one-to-one) used in the challenger.
```
dlubarov, line 8: Hm I'm a little confused by this. Can you not use the Keccak permutation directly? We used a pseudo-permutation in `KeccakPermuation` for compatibility with EVM's `Keccak256`; do you have some constraint like that?

It doesn't seem like this is technically a sponge, since it absorbs all `SPONGE_WIDTH` inputs in the first round, rather than absorbing `SPONGE_RATE` inputs at a time.

dlubarov:
I don't fully understand the context, but my impression is that this seems fairly specific to your use case, so I'm not sure if it'd be useful to have it in the main library.

Come to think of it `KeccakGoldilocksConfig` is also a bit specific to our setting, so perhaps it could be moved somewhere else.

Sladuca:
Sure, it can totally live somewhere else.

plonky2/src/hash/keccak_sponge_sha256.rs
Sladuca: > Hm I'm a little confused by this. Can you not use the Keccak permutation directly? We used a pseudo-permutation in `KeccakPermuation` for compatibility with EVM's `Keccak256`; do you have some constraint like that?

How do we guarantee the 64-bit state elements fit in the field? Is it OK to just mod by `F::ORDER`? I was under the impression that would introduce some bias into the distribution, though I'm not sure by how much.

> It doesn't seem like this is technically a sponge, since it absorbs all `SPONGE_WIDTH` inputs in the first round, rather than absorbing `SPONGE_RATE` inputs at a time.

Thanks, that should be `SPONGE_RATE` not `SPONGE_WDITH` there.

Issue #615, opened 2022-07-15
improve portability


Sladuca:
There are a lot of dependencies that are either 1) only used in tests/examples (e.g. `rand`), or 2) do not compile in rust-based smart contract runtimes (e.g. `env_logger`, `log`, `std::time`, `rayon`). If we intend this library to be used outside a standard environment, we need to be able to build the library without these dependencies.

To address this, I suggest the following:
1) feature-gate `log` / `TimingTree` behind a feature that's enabled by default, and use conditional compilation to remove `timed!`, `with_context!`, `TimingTree`, and `ContextTree` if disabled
2) Wrap `rayon`'s `par_iter` with a macro `cfg_iter` that conditionally switches between `iter()` and `par_iter` like arkworks does [here](https://github.com/arkworks-rs/std/blob/05b7a19a99551bdd737eb2ffb1c9eac7be0dd11f/src/lib.rs#L84) 
3) add a `test_utils` feature to deal with stuff like the [`gate_testing.rs`](https://github.com/mir-protocol/plonky2/blob/2e3ad0142e81a58418f8436c6ba2652f10c56612/plonky2/src/gates/mod.rs#L27) in the interrim until cargo supports this functionality

Sladuca:
Happy to open a PR for this (I'm already working on this because I need it)

Sladuca:
I've got a branch I'm using for this at https://github.com/proxima-one/plonky2/tree/feature-gate-stuff

dlubarov:
For logging, shouldn't `log` itself be portable since it no-ops by default? I see why `env_logger` could be an issue though; would moving it to `dev-dependencies` address the concern?

For `rand`, I think a feature gate sounds good, like `num` has.

For Rayon, I think the feature would ideally be in Rayon itself, see rayon-rs/rayon#861. If we don't want to wait for that we can add a shim temporarily, but I don't really like the arkworks approach since it changes all the call sites to use macros. I'd prefer either
- conditionally importing a `rayon` module with shims for `ParallelIterator` etc, or
- having a `maybe_rayon` module, containing a `MaybeParallelIterator` with a `maybe_par_iter()`, etc

A feature gate for `timed!` and `with_context!` sounds fine, as long as it doesn't change the API; seems like the changes can be internal to those macros.

Sladuca:
I personally need a shim for rayon now, but I can keep that part in our branch for now if others don't need it too urgently. 

[...]

PR #629, opened 2022-07-21
add rayon shim

Sladuca:
Part of https://github.com/mir-protocol/plonky2/issues/615

### DIFF
--- a/maybe_rayon/Cargo.toml
+++ b/maybe_rayon/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "maybe_rayon"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
+parallel = ["rayon"]
+
+[dependencies]
+rayon = { version = "1.5.3", optional = true }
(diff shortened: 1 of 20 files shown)


maybe_rayon/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "maybe_rayon"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[features]
```
dlubarov, line 7: I'd say let's add `default = ["parallel"]` to make it opt-out.

dlubarov: Oh I see it's enabled by default in the plonky2 crate. Makes sense.

Could you do the same for `starky` and `evm` though? Otherwise I guess by default they'd get a mix of parallel code (in starky/evm) and serial code (in plonky2 methods they depend on)?

dlubarov:
The shim itself looks good to me!

maybe_rayon/Cargo.toml
Sladuca: Yes I can do that.

dlubarov (approved):
Thanks, it LGTM now 👍 

Just to confirm, is it okay to treat your contributions as MIT licensed?

Sladuca:
Yes! MIT is good.

PR #630, opened 2022-07-21
Logging feature gates

Sladuca:
Part of https://github.com/mir-protocol/plonky2/issues/615, also feature-gated `ContextTree`

### DIFF
--- a/evm/Cargo.toml
+++ b/evm/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-plonky2 = { path = "../plonky2", features = ["timing"] }
+plonky2 = { path = "../plonky2" }
 plonky2_util = { path = "../util" }
 anyhow = "1.0.40"
 env_logger = "0.9.0"
--- a/evm/Cargo.toml
+++ b/evm/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-plonky2 = { path = "../plonky2" }
+plonky2 = { path = "../plonky2", features = ["timing"] }
 plonky2_util = { path = "../util" }
 anyhow = "1.0.40"
 env_logger = "0.9.0"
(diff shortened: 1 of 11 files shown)


dlubarov:
Could you explain the motivation for these? Is the concern that `timed!` might have a performance impact, or is it to minimize binary size?

I don't like the duplication this brings; if we do add the features I think we should only have guards within the `timed!`/`with_context!` macros.

Sladuca:
`TimingTree` doesn't compile without `std::time`, so the motivation was to remove the type entirely (including the argument being passed around)

I can make `TimingTree` and `ContextTree` empty stubs instead without the featurs enabled.

dlubarov:
Hm I see. I think empty stubs sound good then.

dlubarov:
How about `with_context!` though? I wouldn't expect it to cause any portability issues, and performance-wise it only affects preprocessing.

Sladuca:
I can leave `with_context` in.

evm/Cargo.toml
```diff
@@ -5,7 +5,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-plonky2 = { path = "../plonky2" }
+plonky2 = { path = "../plonky2", features = ["timing"] }
```
dlubarov, line 8: Are these needed? I thought the feature was enabled by default.

plonky2/src/util/timing.rs
```diff
@@ -135,16 +171,26 @@ impl TimingTree {
 #[macro_export]
 macro_rules! timed {
     ($timing_tree:expr, $level:expr, $ctx:expr, $exp:expr) => {{
+        #[cfg(feature = "timing")]
```
dlubarov, line 174: Could we keep these calls and stub them out, so that we don't need to suppress unused warnings (except in the stub methods) or not-mutated warnings?

evm/Cargo.toml
Sladuca: It is enabled by default, so no they're not needed.

plonky2/src/util/timing.rs
Sladuca: sure

Sladuca: the warnings still occur if I stub out the methods. I actually have to mutate in the methods for the warning to go away.

PR #631, opened 2022-07-21
Feature-gate rand

Sladuca:
Part of https://github.com/mir-protocol/plonky2/issues/615.


field/src/extension/quadratic.rs
```diff
@@ -4,6 +4,7 @@ use std::ops::{Add, AddAssign, Div, DivAssign, Mul, MulAssign, Neg, Sub, SubAssi
 
 use num::bigint::BigUint;
 use num::Integer;
+#[cfg(feature = "rand")]
 use rand::Rng;
```
dlubarov, line 8: Just a nit, but what do you think of having this import in the method(s) that use it? Either generally or just in files like this with a single rand method. It's a bit unusual, but might be nice to reduce the cfg guards.

plonky2/Cargo.toml
```diff
@@ -10,15 +10,21 @@ categories = ["cryptography"]
 edition = "2021"
 default-run = "generate_constants"
 
+[features]
+default = ["rand", "gate_testing"]
+gate_testing = []
+rand = ["dep:rand", "plonky2_field/rand"]
+rand_chaha = []
```
dlubarov, line 17: Typo? `chaha` -> `chacha`

dlubarov: And should it refer to the dep? Like

   rand_chacha = ["dep:rand_chacha"]

plonky2/src/gates/mod.rs
```diff
@@ -24,4 +24,5 @@ pub mod util;
 
 // Can't use #[cfg(test)] here because it needs to be visible to other crates.
 // See https://github.com/rust-lang/cargo/issues/8379
+#[cfg(any(feature = "rand", feature = "gate_testing", test))]
```
dlubarov, line 27: Hm this doesn't seem quite right to me, like if rand is true but the others are false, then we don't need this right?

I think we can do `any(feature = "gate_testing", test)`, and then have `gate_testing` require `rand` in Cargo.toml.

plonky2/src/plonk/circuit_data.rs
```diff
@@ -94,6 +94,10 @@ impl CircuitConfig {
     }
 
     pub fn standard_recursion_zk_config() -> Self {
+        #[cfg(not(feature = "rand"))]
```
dlubarov, line 97: I don't think it's useful to have another assert here, the one you added in `check_config` seems like the best place IMO (it'll give early detection and it's in a code path that's always hit).

plonky2/src/plonk/prover.rs
```diff
@@ -36,6 +36,13 @@ where
     [(); C::Hasher::HASH_SIZE]:,
 {
     let config = &common_data.config;
+
+    #[cfg(not(feature = "rand"))]
```
dlubarov, line 40: Likewise here; seems like this shouldn't be reachable since you added the assert in `check_config`.

field/src/extension/quadratic.rs
Sladuca: The method signature (takes generic type parameter `R: Rng`) depends on the trait being in-scope. I can just move this import to the bottom of the imports so it's separated and looks cleaner if that would be preferrable.

dlubarov (approved):
LGTM 👍

PR #633, opened 2022-07-22
set cap for challenger's input buffer size

Sladuca:
Challenger's input buffer can get pretty large and overflow small heaps. I've set it so that it automatically absorbs all of the buffered elements when it reaches a length of `MAX_INPUT_BUFFER`, which is currently set to `128`.

I've also removed some unnecessary `Vec` allocations when setting the output buffer.

### DIFF
--- a/plonky2/src/iop/challenger.rs
+++ b/plonky2/src/iop/challenger.rs
@@ -91,8 +91,7 @@ impl<F: RichField, H: Hasher<F>> Challenger<F, H> {
             // Evaluate the permutation to produce `r` new outputs.
             self.sponge_state = H::Permutation::permute(self.sponge_state);
             self.output_buffer.clear();
-            self.output_buffer
-                .extend_from_slice(&self.sponge_state[0..SPONGE_RATE]);
+            self.output_buffer.extend_from_slice(&self.sponge_state[0..SPONGE_RATE]);
         }
 
         self.output_buffer
@@ -152,8 +151,7 @@ impl<F: RichField, H: Hasher<F>> Challenger<F, H> {
         }
 
         self.output_buffer.clear();
-        self.output_buffer
-            .extend_from_slice(&self.sponge_state[0..SPONGE_RATE]);
+        self.output_buffer.extend_from_slice(&self.sponge_state[0..SPONGE_RATE]);
 
         self.input_buffer.clear();
     }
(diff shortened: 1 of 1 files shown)


plonky2/src/iop/challenger.rs
```diff
@@ -143,7 +151,9 @@ impl<F: RichField, H: Hasher<F>> Challenger<F, H> {
             self.sponge_state = H::Permutation::permute(self.sponge_state);
         }
 
-        self.output_buffer = self.sponge_state[0..SPONGE_RATE].to_vec();
+        self.output_buffer.clear();
```
dlubarov, line 154: Is this to avoid an allocation? Maybe we can do just in the (more performance sensitive) native verifier since it's sligthly more verbose.

dlubarov:
What do you think of doing it like this? #636

I figure we might as well use `SPONGE_RATE` as the buffer size. We could even use arrays for the input and output `Vec`s, though it would be a bit awkward. (`Cursor` fits the use case, at least for the input buffer, but it's focused on `u8`s and lacks more general features.)

It's simpler in a way since we no longer need the loop in `absorb_buffered_inputs`, though it's annoying that we can't make the same change in `RecursiveChallenger` (or `observe_element` would need the builder).

plonky2/src/iop/challenger.rs
Sladuca: Yes, it's to avoid the allocation.

Sladuca:
> What do you think of doing it like this? #636
Looks totally good to me.
> I figure we might as well use `SPONGE_RATE` as the buffer size. It's simpler in a way since we no longer need the loop in `absorb_buffered_inputs`, though it's annoying that we can't make the same change in `RecursiveChallenger` (or `observe_element` would need the builder).
> 
This is also good.

> We could even use arrays for the input and output buffers, though Rust makes it a bit cumbersome. (`Cursor` fits the use case, at least for the input buffer, but it's focused on `u8`s and lacks more general features.)
Works for me as long as the buffer size isn't too large. Solana's stack is only 4K.

plonky2/src/iop/challenger.rs
Sladuca: we can also make a helper function if we want to improve readability.

dlubarov:
Ok thanks, I merged #636.

PR #634, opened 2022-07-23
Fix bug where `constant_affine_point` is called on zero

wborgeaud:
Could also be fixed by adding a `constant_affine_point_unchecked` function, but I think it's good practice to check for zero input.


dlubarov (approved):
LGTM 👍

PR #635, opened 2022-07-23
Use non-determinism in EC ops

wborgeaud:


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -326,11 +332,13 @@ impl<'a> Interpreter<'a> {
...
-        let input = self
-            .prover_inputs
-            .pop()
-            .ok_or_else(|| anyhow!("Out of prover inputs"))?;
-        self.stack.push(input);
+        let prover_input_fn = self
+            .prover_inputs_map
+            .get(&(self.offset - 1))
+            .ok_or_else(|| anyhow!("Offset not in prover inputs."))?;
+        let output = prover_input_fn.run(self.stack.clone());
```
dlubarov, line 339: I know using `Vec::pop` could be more convenient, but I think this should probably pass a slice to avoid cloning the stack. Would also help emphasize that it's read-only; at first glance I thought the `pop()`s were actually mutating the stack.

evm/src/cpu/kernel/prover_input.rs
```diff
@@ -0,0 +1,139 @@
...
+        modexp(x, q, n)
+    }
+}
+
+fn modexp(x: U256, e: U256, n: U256) -> U256 {
+    let mut current = x;
+    let mut product = U256::one();
+
+    for j in 0..256 {
+        if !(e >> j & U256::one()).is_zero() {
```
dlubarov, line 133:
```suggestion
        if e.bit(j) {
```

evm/src/cpu/kernel/asm/moddiv.asm
```diff
@@ -25,480 +25,19 @@
...
+    // stack: x
+    PROVER_INPUT(ff::bn254_base::inverse)
+    // stack: x^-1, x
+    %stack (inv, x) -> (inv, x, @BN_BASE, inv, x)
+    // stack: x^-1, x, N, x^-1, x
+    MULMOD
+    // stack: x^-1 * x, x^-1, x
+    PUSH 1
+    // stack: 1, x^-1 * x, x^-1, x
+    %assert_eq
```
dlubarov, line 39:
```suggestion
    %assert_eq_const(1)
```

evm/src/cpu/kernel/asm/moddiv.asm
```diff
@@ -25,480 +25,19 @@
...
-    DUP2
-    %mulmodn
-    %squaremodn
-    %squaremodn
-    DUP2
-    %mulmodn
+    // stack: x
+    PROVER_INPUT(ff::bn254_base::inverse)
+    // stack: x^-1, x
+    %stack (inv, x) -> (inv, x, @BN_BASE, inv, x)
```
dlubarov, line 33: Seems the last `x` gets popped later, so we can omit it?

evm/src/cpu/kernel/asm/secp256k1/inverse_scalar.asm
```diff
@@ -17,654 +17,19 @@
...
-    %squaremodn_secp_scalar
-    DUP2
-    %mulmodn_secp_scalar
-    %squaremodn_secp_scalar
-    DUP2
-    %mulmodn_secp_scalar
+    // stack: x
+    PROVER_INPUT(ff::secp256k1_scalar::inverse)
+    // stack: x^-1, x
+    %stack (inv, x) -> (inv, x, @SECP_SCALAR, inv, x)
```
dlubarov, line 25: Same couple comments here - seems we can omit the `x` at the end of the stack, and use `%assert_eq_const(1)` below

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -60,759 +60,13 @@
     ADDMOD
 %endmacro
 
-// Returns a square root of x if one exists, otherwise an undefined value.
-// Computed as x^(q+1)/4, with q the Secp base field order.
-// To replace with more efficient method using non-determinism later.
+// Non-deterministically provide the square root modulo N.
```
dlubarov, line 63: It might sound like this witnesses a square root and verifies it, but it's the caller's responsibility to verify it right? (Like `sqrt_secp_base` could return a bad result even if `x` is square?) Let's emphasize it in the comment, or maybe add "unsafe" in the name.

evm/src/cpu/kernel/asm/secp256k1/moddiv.asm
```diff
@@ -25,760 +25,19 @@
...
-    DUP2
-    %mulmodn_secp_base
-    %squaremodn_secp_base
-    %squaremodn_secp_base
-    DUP2
-    %mulmodn_secp_base
+    // stack: x
+    PROVER_INPUT(ff::secp256k1_base::inverse)
+    // stack: x^-1, x
+    %stack (inv, x) -> (inv, x, @SECP_BASE, inv, x)
```
dlubarov, line 33: Same couple comments here about pruning the final `x` and `%assert_eq_const(1)`

evm/src/cpu/kernel/mod.rs
```diff
@@ -1,14 +1,14 @@
 pub mod aggregator;
 pub mod assembler;
 mod ast;
+pub mod interpreter;
```
dlubarov, line 4: Hm is the idea to use `Interpreter` to help with witness generation? (If so I'm not sure it's useful there, but we can discuss later.)

evm/src/cpu/kernel/prover_input.rs
wborgeaud: Much nicer indeed :)

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
wborgeaud: Yeah exactly, the square root is checked in `secp_lift_x` which is the only function using `sqrt_secp_base`. I added a comment and renamed it `sqrt_secp_base_unsafe` 👍

evm/src/cpu/kernel/mod.rs
wborgeaud: Oh I think this was just to silence a bunch of dead code warnings. I changed it back to a `cfg(test)` module.

dlubarov (approved):

evm/src/cpu/kernel/asm/secp256k1/lift_x.asm
```diff
@@ -61,7 +61,8 @@
 %endmacro
 
 // Non-deterministically provide the square root modulo N.
-%macro sqrt_secp_base
+// Note: The square root is not checked and the macro doesn't not panic if `x` is not a square.
```
dlubarov, line 64: Little typo here - `doesn't not`

PR #636, opened 2022-07-24
Use a fixed input buffer size in `Challenger`.

dlubarov:
Alternate implementation of #633, using `SPONGE_RATE` as the buffer size.


Sladuca:
Thanks!

PR #638, opened 2022-07-25
Contexts and segments in interpreter memory

wborgeaud:


evm/src/memory/segments.rs
```diff
@@ -54,4 +54,19 @@ impl Segment {
...
+            Segment::Stack => 256,
+            Segment::MainMemory => 8,
+            Segment::Calldata => 8,
+            Segment::Returndata => 8,
+            Segment::Metadata => 8,
+            Segment::KernelGeneral => 8,
+            Segment::TxnData => 8,
+            Segment::RlpRaw => 8,
+        }
+    }
```
wborgeaud, line 71: Not entirely sure about these values, needs double-checking.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -2,50 +2,32 @@ use anyhow::{anyhow, bail};
...
+use crate::memory::segments::Segment;
+
 /// Halt interpreter execution whenever a jump to this offset is done.
 const HALT_OFFSET: usize = 0xdeadbeef;
 
-#[derive(Debug, Default)]
-pub(crate) struct EvmMemory {
-    memory: Vec<u8>,
+#[derive(Debug)]
+pub(crate) struct ContextMemory {
```
dlubarov, line 12: nit: I feel like "context" would be better in the field name than the struct name, since the fact that the outer Vec is for the context dimension seems like an internal thing; in the API context is just one of three.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -54,7 +36,8 @@ pub(crate) struct Interpreter<'a> {
     jumpdests: Vec<usize>,
     offset: usize,
     pub(crate) stack: Vec<U256>,
```
dlubarov, line 38: Let's add a TODO to remove stack + code? Since they're also segments within `memory`. Using the segments in `memory` will mean we can use `mload_general` to load constants from code, or for stack-based arrays etc.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -404,6 +410,33 @@ impl<'a> Interpreter<'a> {
...
+        let x = self.pop();
+        self.context = x.as_usize();
+    }
+
+    fn run_mload_general(&mut self) {
+        let context = self.pop().as_usize();
+        let segment = Segment::all()[self.pop().as_usize()];
+        let offset = self.pop().as_usize();
+        let value = self.memory.mload_general(context, segment, offset);
+        assert!(value < U256::one() << segment.bit_range());
```
dlubarov, line 428:
```suggestion
        assert!(value.bits() <= segment.bit_range());
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -404,6 +410,33 @@ impl<'a> Interpreter<'a> {
...
+        self.push(value);
+    }
+
+    fn run_mstore_general(&mut self) {
+        let context = self.pop().as_usize();
+        let segment = Segment::all()[self.pop().as_usize()];
+        let offset = self.pop().as_usize();
+        let value = self.pop();
+        assert!(value < U256::one() << segment.bit_range());
```
dlubarov, line 437:
```suggestion
        assert!(value.bits() < segment.bit_range());
```

evm/src/memory/segments.rs
```diff
@@ -54,4 +54,19 @@ impl Segment {
...
+    pub(crate) fn bit_range(&self) -> usize {
+        match self {
+            Segment::Code => 8,
+            Segment::Stack => 256,
+            Segment::MainMemory => 8,
+            Segment::Calldata => 8,
+            Segment::Returndata => 8,
+            Segment::Metadata => 8,
+            Segment::KernelGeneral => 8,
+            Segment::TxnData => 8,
```
dlubarov, line 68: I think these three should be allowed to go up to 256 bits. The rest look good!

dlubarov (approved):
LGTM

evm/src/cpu/kernel/interpreter.rs
wborgeaud: Good point, I've renamed to
```rust
struct InterpreterMemory {
    context_memory: Vec<MemoryContextState>,
}
```

evm/src/cpu/kernel/interpreter.rs
wborgeaud: Good point! Do we need to maintain a pointer to the top of the stack? 
Also, for the code, would the offset become a context+offset? Like right now the kernel code is at the start of the `Code` segment in the first context. So if we want to jump to a kernel function from another context, we would have to jump to the first context? Or would we write the kernel code at every context?

evm/src/memory/segments.rs
wborgeaud: Thanks!

evm/src/cpu/kernel/interpreter.rs
dlubarov: Yeah I think we'll need to add a stack pointer register in the CPU, though our assembly shouldn't need to know about it (unless we wanted stack-based arrays, but doesn't really seem necessary).

I think the kernel code can live in context 0 only. When we jump to kernel code, I think it may be easiest/cheapest to leave the CPU's context register alone, so it still holds the (userland) caller's context. Then when we fetch an instruction, we'd load from context `is_user_mode * context`. Does that sound okay? cc @nbgl who is starting on some related work.

wborgeaud: Yes sounds good!

PR #639, opened 2022-07-25
Inter-row program counter constraints

nbgl:
Enforce correct control flow between rows. An execution trace is forced to occupy a contiguous range of rows. It begins at `program_counter = 0` and continues until the program halts (this is just a flag which we can set if the STOP instruction is executed or the program is killed). The `program_counter` of a row (unless it is the very first row of an execution trace) is set to `next_program_counter` from the previous row.

### DIFF
--- a/evm/src/cpu/columns.rs
+++ b/evm/src/cpu/columns.rs
@@ -17,8 +17,11 @@ pub struct CpuColumnsView<T> {
     pub is_bootstrap_contract: T,
 
     /// Filter. 1 if the row corresponds to a cycle of execution and 0 otherwise.
-    /// Lets us re-use decode columns in non-cycle rows.
+    /// Lets us re-use columns in non-cycle rows.
     pub is_cpu_cycle: T,
+
+    /// If CPU cycle: The program counter for the current instruction.
+    pub program_counter: T,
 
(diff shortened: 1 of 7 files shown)


evm/src/cpu/columns.rs
```diff
@@ -139,6 +142,9 @@ pub struct CpuColumnsView<T> {
     /// If CPU cycle: the opcode, broken up into bits in **big-endian** order.
     pub opcode_bits: [T; 8],
 
+    /// If CPU cycle: The program counter for the next instruction.
+    pub next_program_counter: T,
```
dlubarov, line 146: Is this needed? I would think we could have constraints interact directly with `nv.program_counter`, for logic like jumps etc.

(I see how that could be problematic if the next row is not a CPU cycle, but I think once we finish bootloading and reach CPU cycle rows, all subsequent rows would be CPU cycle rows?)

dlubarov (approved):
LGTM otherwise!

evm/src/cpu/columns.rs
nbgl: I had the same thought. I decided to leave this column in here so this PR can show what the constraints end up being; I plan to remove it in a following PR

PR #640, opened 2022-07-27
RipeMD

typ3c4t:


evm/src/cpu/kernel/asm/ripemd/compression.asm
```diff
@@ -0,0 +1,279 @@
...
+    %jump(rol)
+post_rol:
+    // stack: c, a, b, d, e, F, K, boxes  , rounds, sides, virt
+    SWAP3
+    // stack: d, a, b, c, e, F, K, boxes  , rounds, sides, virt
+    SWAP4
+    // stack: e, a, b, c, d, F, K, boxes  , rounds, sides, virt
+    SWAP7  
+    PUSH 1  
+    SWAP1
```
npwardberkeley, line 249: This would be good as a `%stack` macro I think!

evm/src/cpu/kernel/asm/ripemd/compression.asm
```diff
@@ -0,0 +1,269 @@
...
+    // stack: r0, r1, r2, r3, r4, l0, l1, l2, l3, l4, s0, s1, s2, s3, s4, VR, RD 
+    SWAP10
+    // stack: s0, r1, r2, r3, r4, l0, l1, l2, l3, l4, r0, s1, s2, s3, s4, VR, RD 
+    SWAP1
+    // stack: r1, s0, r2, r3, r4, l0, l1, l2, l3, l4, r0, s1, s2, s3, s4, VR, RD 
+    SWAP6
+    // stack: l1, s0, r2, r3, r4, l0, r1, l2, l3, l4, r0, s1, s2, s3, s4, VR, RD 
+    %add3_u32
+    // stack:         o4, r3, r4, l0, r1, l2, l3, l4, r0, s1, s2, s3, s4, VR, RD 
+    SWAP14
```
npwardberkeley, line 65: These five `SWAP`s could be one `%stack` macro?

evm/src/cpu/kernel/asm/ripemd/compression.asm
```diff
@@ -0,0 +1,269 @@
...
+    DUP12  
+    DUP12  
+    DUP12
+    // stack:  STATE, 0, 0, 16, 5, 1, virt, switch, STATE, virt, retdest 
+    %jump(loop)
+switch:
+    // stack:                                   STATEL, STATE, virt, retdest
+    PUSH mix
+    DUP12 
+    PUSH 0
```
npwardberkeley, line 42: This could be a `%stack` macro

evm/src/cpu/kernel/asm/ripemd/compression.asm
```diff
@@ -0,0 +1,269 @@
...
+    ADD
+    // stack:                a, b, c, d, e, F, K, boxes, rounds, sides, virt
+    %get_box
+    // stack:           box, a, b, c, d, e, F, K, boxes, rounds, sides, virt
+    DUP12
+    DUP2
+    %mload_kernel_code_label(R_data)
+    ADD
+    // stack: virt + r, box, a, b, c, d, e, F, K, boxes, rounds, sides, virt  
+    // %mload_kernel_code_u32_LE(Input_Block)
```
npwardberkeley, line 206: Nit: what is this commented code for?

evm/src/cpu/kernel/asm/ripemd/constants.asm
```diff
@@ -0,0 +1,140 @@
...
+    BYTES 20, 48, 08, 52
+    BYTES 36, 28, 40, 56
+    // Right Round 4
+    BYTES 48, 60, 40, 16
+    BYTES 04, 20, 32, 28
+    BYTES 24, 08, 52, 56
+    BYTES 00, 12, 36, 44
+
+
+global Input_Block:
```
npwardberkeley, line 123: nit: `input_block`

evm/src/cpu/kernel/asm/ripemd/memory.asm
```diff
@@ -0,0 +1,169 @@
...
+    %sub_const(1)
+    SWAP3
+    // stack:               ADDR + 1, rem - 1, length 
+    DUP4
+    %jumpi(store_input)
+    // stack:               ADDR    , 0      , length
+    %pop4
+    // stack:                                  length
+    %jump(ripemd_init)
+
```
npwardberkeley, line 83: nit: spacing (in various places)

evm/src/memory/segments.rs
```diff
@@ -36,10 +36,13 @@ pub(crate) enum Segment {
     /// `StorageTriePointers` with `StorageTrieCheckpointPointers`.
     /// See also `StateTrieCheckpointPointer`.
     StorageTrieCheckpointPointers = 14,
+    /// RipeMD storage
+    ///
```
npwardberkeley, line 40: nit: empty line

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -229,3 +229,70 @@
...
+    // stack: cba, d
+    SWAP1
+    %shl_const(24)
+    OR
+    // stack: dcba
+%endmacro
+
+%macro break
+    %jump(0xdeadbeef)
+%endmacro
```
npwardberkeley, line 298: nit: no newline at end of file

evm/src/cpu/kernel/asm/ripemd/compression.asm
typ3c4t: unfortunately when I had stack macros that went that deep (I think the algo might be exponential in input prefix depth?), my code wouldn't run

evm/src/cpu/kernel/asm/ripemd/memory.asm
typ3c4t: hmm, I guess this is a purposeful style choice---I wanted to right-align the stack and visually align persistent values on the stack at different steps. I found this helped me keep track of what was happening. Has the team established explicit style standards on such things? @dlubarov

npwardberkeley: Ohh I meant the line spacing — having extra lines (or at least inconsistently) between the functions

npwardberkeley: The stack diagram spacing is fine! It's different form how I do it but neither is correct

typ3c4t: oh sorry for the misunderstanding! is the rule that there should always be precisely 1 newline between labels/macros?

evm/src/cpu/kernel/asm/ripemd/compression.asm
npwardberkeley: Ohhh right

npwardberkeley: Yeah that's fine then!

evm/src/cpu/kernel/asm/ripemd/memory.asm
npwardberkeley: Hm not necessarily, just that it should be consistent! I think typically it would be one newline, unless multiple labels are meant to be grouped together (in which case there would be more newlines between the groups)

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -229,3 +229,70 @@
     %select_bool
     // stack: max
 %endmacro
+
+%macro u32
```
dlubarov, line 233: nit: I'd call it `to_u32` or `as_u32` to be more explicit

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -229,3 +229,70 @@
...
+    ADD
+    // stack: x+y, z
+    ADD
+    // stack: x+y+z
+    %u32
+%endmacro
+
+
+// given u32 bytestring abcd return dcba
+%macro reverse_bytes_u32
```
dlubarov, line 256: Did you see the `BYTE` instruction? I think it could lead to a shorter implementation

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -229,3 +229,70 @@
...
+    %shl_const(16)
+    OR
+    // stack: cba, d
+    SWAP1
+    %shl_const(24)
+    OR
+    // stack: dcba
+%endmacro
+
+%macro break
```
dlubarov, line 296: If this will be merged, let's call it something like `%debug` or `%deadbeef`, to make it more clear that it's not something that should be in any production code, and maybe add a note that this causes termination for debugging purposes.

dlubarov: I think for debugging purposes we can also just add a `STOP` opcode? So maybe this isn't useful

evm/src/memory/segments.rs
```diff
@@ -58,6 +61,7 @@ impl Segment {
             Self::StorageTrieAddresses,
             Self::StorageTriePointers,
             Self::StorageTrieCheckpointPointers,
+            Self::RipeMd,
```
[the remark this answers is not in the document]
dlubarov: Just as a general rule of thumb, I'd suggest we should use `KERNEL_GENERAL` by default. But if we have a scenario where function A calls function B and both use `KERNEL_GENERAL`, potentially clobbering each other, it should be fine to add a new segment to avoid clobbering.

evm/src/cpu/kernel/asm/ripemd/compression.asm
```diff
@@ -0,0 +1,263 @@
...
+    DUP5
+    // stack:            e, a, b, c, d, e, F, K, boxes, rounds, sides, virt
+    ADD 
+    %u32    
+    // stack:               a, b, c, d, e, F, K, boxes, rounds, sides, virt
+    SWAP1  
+    SWAP2  
+    PUSH post_rol  
+    SWAP1  
+    PUSH 10
```
dlubarov, line 229: Seems like a good place to use `%stack`

evm/src/cpu/kernel/asm/ripemd/memory.asm
```diff
@@ -0,0 +1,166 @@
...
+    // stack:           xsy, y, 0x100, y
+    SUB
+    DIV
+    SWAP1
+    // stack: y, xs
+    %mstore_ripemd_offset($offset)
+    // stack: xs
+%endmacro 
+
+%macro mstore_ripemd_offset(offset)
```
dlubarov, line 114: We recently added support for macro overloading, so this could potentially share the `mstore_ripemd` name. Fine to keep if you prefer it like this though.

evm/src/cpu/kernel/asm/memory/core.asm
```diff
@@ -62,6 +62,24 @@
...
+%macro mload_kernel_code_label(label)
+    // stack: shift
+    PUSH $label  
+    ADD
+    // stack: label + shift
+    %mload_kernel_code
+    // stack: byte
+%endmacro
+
+%macro mload_kernel_code_label_u32(label)
```
dlubarov, line 74: nit: I think it'd be natural to move the `%mul_const(4)` (which currently the caller does) inside this macro. So this macro would take the index of the u32, rather than its offset in bytes.

evm/src/memory/segments.rs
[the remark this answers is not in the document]
typ3c4t: So do you think in the case of RipeMD, I should keep it as is, giving it its own segment, or do you think it's worth changing it to `KERNEL_GENERAL`?

evm/src/cpu/kernel/asm/util/basic_macros.asm
typ3c4t: I got rid of it!

evm/src/memory/segments.rs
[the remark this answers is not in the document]
dlubarov: Yeah I'd slightly prefer switching to `KERNEL_GENERAL`, just to have one less segment, but it's not very important, we can always do it later.

npwardberkeley (approved):
LGTM :D

PR #641, opened 2022-07-27
Increment program counter on native instructions

nbgl:
When we execute an instruction that takes up one row (is not microcoded), then we increment the program counter by 1.


dlubarov (approved):
LGTM

PR #642, opened 2022-07-28
Small fix for type 0 txns

dlubarov:
Always parse "to" as a scalar. No need for a branch; it's left over from when I was trying to enforce canonical RLP (in which case "to" must be 0 or 20 bytes).

The old code would be wrong if we had multiple txns per proof, as if to=0 we wouldn't write that field to memory, so it could have an old value from a previous txn.


wborgeaud (approved):
👍

PR #644, opened 2022-07-29
Implement PANIC instruction

nbgl:
Make the proof unverifiable whenever the PANIC instruction is executed


dlubarov (approved):
Nice idea 😄

PR #645, opened 2022-07-31
Add a `asm/curve/` directory

dlubarov:
`ecrecover.asm` seems like it could fit in `curve/secp256k1/` or stay where it is (I left it for now); let me know if you have a preference.


evm/src/cpu/kernel/aggregator.rs
```diff
@@ -84,13 +85,12 @@ mod tests {
...
-            .format_timestamp(None)
-            .try_init();
+        env_logger::init_from_env(
+            env_logger::Env::default().filter_or(env_logger::DEFAULT_FILTER_ENV, "debug"),
+        );
 
         // Make sure we can parse and assemble the entire kernel.
         let kernel = combined_kernel();
         debug!("Total kernel size: {} bytes", kernel.code.len());
-        dbg!("Total kernel size: {} bytes", kernel.code.len());
```
dlubarov, line 94: (This syntax wasn't working; was the issue that it wasn't convenient to add `RUST_LOG=debug`? If so the change above should help.)

wborgeaud: Oh sorry, I forgot to remove this. I just do this sometimes because it's less keystrokes to add a `dbg!` than to set the env var, but setting it to "debug" by default also works 👍

wborgeaud (approved):
Looks good to me!

> `ecrecover.asm` seems like it could fit in `curve/secp256k1/` or stay where it is (I left it for now); let me know if you have a preference.

I don't have a strong opinion on this. It should be pretty unambiguous that `ecrecover` uses only the Secp curve, so leaving it in `curve/` is fine I guess.

PR #646, opened 2022-07-31
Allow `%stack` to work with labels

dlubarov:
There's no syntax to distinguish named stack items from labels, so this simply searches the former first. I.e. labels can be shadowed by stack items.


wborgeaud (approved):
Looks good to me!

PR #647, opened 2022-07-31
Store literals as `U256` (or `u8` for `BYTES`)

dlubarov:
Instead of the original strings. Will make optimizations simpler.

Depends on #646.


wborgeaud (approved):
👍

PR #648, opened 2022-07-31
Some simple optimization rules

dlubarov:
Depends on #647.


evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,135 @@
...
+        if let [Push(Label(l)), StandardOp(jump), decl] = window
+            && &jump == "JUMP"
+            && (decl == LocalLabelDeclaration(l.clone()) || decl == GlobalLabelDeclaration(l.clone()))
+        {
+            Some(vec![LocalLabelDeclaration(l)])
+        } else {
+            None
+        }
+    });
+}
```
nbgl, line 75: Maybe this should be a warning instead? I can’t think of a situation where the programmer would actually _intend_ to write this.

dlubarov: I was thinking of this scenario - https://github.com/mir-protocol/plonky2/pull/627#discussion_r929390378

That said, currently it won't remove the jump because of the file inclusion order (and I'm not sure it's worth changing).

Even if it did remove a few jumps like that, it seems fairly negligible performance-wise, so perhaps we should just remove this?

Maybe it could be more relevant if we had more powerful macros that could generate labels, but I didn't have any plans to add that.

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -120,7 +120,9 @@
     // stack: input, ...
     PUSH $c
     // stack: c, input, ...
-    GE // Check it backwards: (input <= c) == (c >= input)
```
dlubarov, line 123: (`GE` and `LE` don't actually exist)

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,135 @@
...
+    });
+
+    // Constant propagation for binary ops: `[PUSH x, PUSH y, BINOP] -> [PUSH BINOP(x, y)]`
+    replace_windows(code, |window| {
+        if let [Push(Literal(x)), Push(Literal(y)), StandardOp(op)] = window {
+            match op.as_str() {
+                "ADD" => Some(vec![Push(Literal(x + y))]),
+                "SUB" => Some(vec![Push(Literal(x - y))]),
+                "MUL" => Some(vec![Push(Literal(x * y))]),
+                "DIV" => Some(vec![Push(Literal(x / y))]),
```
nbgl, line 54: Does this have correct edge-case behavior? Rust generally panics or has UB when an integer operation overflows, whereas the EVM wraps around. `wrapping_add`/`sub`/`mul` may be more appropriate. Similarly, division by `0` should produce `0` and not panic.

dlubarov: Good point! For some reason I thought `U256` implemented "EVM semantics" rather than "Rust semantics", but it seems that's not the case. I made the changes and added a few tests.

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,181 @@
...
+    replace_windows(code, |window| {
+        if let [Push(Literal(x)), StandardOp(op)] = window {
+            match op.as_str() {
+                "ISZERO" => Some(vec![Push(Literal(if x.is_zero() {
+                    U256::one()
+                } else {
+                    U256::zero()
+                }))]),
+                "NOT" => Some(vec![Push(Literal(!x))]),
+                _ => None,
```
wborgeaud, line 40: Could also add `POP` :)

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,181 @@
...
+    replace_windows(code, |window| {
+        if let [Push(Literal(y)), Push(Literal(x)), StandardOp(op)] = window {
+            match op.as_str() {
+                "ADD" => Some(vec![Push(Literal(x.overflowing_add(y).0))]),
+                "SUB" => Some(vec![Push(Literal(x.overflowing_sub(y).0))]),
+                "MUL" => Some(vec![Push(Literal(x.overflowing_mul(y).0))]),
+                "DIV" => Some(vec![Push(Literal(
+                    x.checked_div(y).unwrap_or(U256::zero()),
+                ))]),
+                _ => None,
```
wborgeaud, line 57: Why not also add `EQ, LT, GT`, logic ops and signed ops? Or could add a TODO.

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,181 @@
...
+fn remove_swaps(code: &mut Vec<Item>) {
+    replace_windows(code, |window| {
+        if let [Push(x), Push(y), StandardOp(swap1)] = window
+            && &swap1 == "SWAP1" {
+            Some(vec![Push(y), Push(x)])
+        } else {
+            None
+        }
+    });
+}
```
wborgeaud, line 89: We probably won't hit this case often, but it should be easy to generalize this to `SWAPn`.

evm/src/cpu/kernel/optimizer.rs
dlubarov: Good idea, I added them except for signed operation, which may be slightly more work since `U256` doesn't seem to support them directly. I don't think we use them in the kernel yet so let's revisit if they come up?

evm/src/cpu/kernel/optimizer.rs
dlubarov: Would that be to support things like `[PUSH x, DUP5, PUSH y, SWAP2] -> [PUSH y, DUP5, PUSH x]`? I think it seems a bit tricky, depending on what we want to support between the pushes.

wborgeaud: Oh I was just thinking of supporting a sequence of `n+1` `PUSH` followed by a  `SWAPn`.

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,234 @@
...
+
+/// Remove SWAP1 before a commutative function.
+fn remove_swaps_commutative(code: &mut Vec<Item>) {
+    replace_windows(code, |window| {
+        if let [StandardOp(swap1), StandardOp(f)] = window && &swap1 == "SWAP1" {
+            let commutative = matches!(f.as_str(), "ADD" | "MUL");
```
wborgeaud, line 100: Could also add `EQ, AND, OR, XOR`.

wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/optimizer.rs
dlubarov: Done

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,234 @@
...
+    });
+}
+
+/// Remove push-pop type patterns, such as: `[DUP1, POP]`.
+fn remove_ignored_values(code: &mut Vec<Item>) {
+    replace_windows(code, |[a, b]| {
+        if let StandardOp(pop) = b && &pop == "POP" {
+            match a {
+                Push(_) => Some(vec![]),
+                StandardOp(dup) if dup.starts_with("DUP") => Some(vec![]),
```
nbgl, line 114: You could extend it to other non-side-effecting operations. E.g. if you have `[DUP1, ADD, POP]`, that’s effectively a `POP`.

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,234 @@
...
+                "MUL" => Some(x.overflowing_mul(y).0),
+                "DIV" => Some(x.checked_div(y).unwrap_or(U256::zero())),
+                "SHL" => Some(x << y),
+                "SHR" => Some(x >> y),
+                "AND" => Some(x & y),
+                "OR" => Some(x | y),
+                "XOR" => Some(x ^ y),
+                "LT" => Some(u256_from_bool(x < y)),
+                "GT" => Some(u256_from_bool(x > y)),
+                "EQ" => Some(u256_from_bool(x == y)),
```
nbgl, line 60: Other non-side-effecting operations: `SDIV`, `MOD`, `SMOD`, `EXP`, `SIGNEXTEND`, `SLT`, `SGT`, `BYTE`, `SAR`.

evm/src/cpu/kernel/optimizer.rs
```diff
@@ -0,0 +1,234 @@
...
+fn optimize_asm_once(code: &mut Vec<Item>) {
+    constant_propagation(code);
+    no_op_jumps(code);
+    remove_swapped_pushes(code);
+    remove_swaps_commutative(code);
+    remove_ignored_values(code);
+}
+
+/// Constant propagation.
+fn constant_propagation(code: &mut Vec<Item>) {
```
nbgl, line 31: One thing to keep in mind: constant propagation isn’t _guaranteed_ to be helpful. It depends on how we implement `PUSH` instructions.

In particular, we discussed implementing `PUSHn` as a system call (especially for large `n`). Consider the following code: `[PUSH1 0, NOT]`. It effectively pushes 2^256 − 1 onto the stack, so we could replace it with `[PUSH32 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff]`. But if `PUSH1` is one row while `PUSH32` is a system call, then the original version is actually faster and constant propagation actually does more harm than good.

nbgl (approved):

evm/src/cpu/kernel/optimizer.rs
dlubarov: Good point. Maybe we should treat the optimization rules as suggestions that are passed through a cost estimator. We could suggest the opposite optimization too, `[PUSH x] -> [PUSH !x, NOT]`, which might help with some curve code.

evm/src/cpu/kernel/utils.rs
```diff
@@ -1,16 +1,63 @@
...
+where
+    T: Clone + Debug,
+    F: Fn([T; W]) -> Option<Vec<T>>,
+{
+    let mut start = 0;
+    while start + W <= vec.len() {
+        let range = start..start + W;
+        let window = vec[range.clone()].to_vec().try_into().unwrap();
+        if let Some(replacement) = maybe_replace(window) {
+            vec.splice(range, replacement);
```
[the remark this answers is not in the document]
dlubarov: Yeah true. I think it should be reasonable for now as our kernel is on the order of kilobytes, but if it gets large we may need to switch to a chunked structure, or something tree-based like a rope.

evm/src/cpu/kernel/optimizer.rs
dlubarov: Added `MOD`, `BYTE`, `EXP`. I think the signed ops are less trivial since `U256` doesn't directly implement them, so let's just wait and see if a use case comes up.

evm/src/cpu/kernel/asm/util/basic_macros.asm
```diff
@@ -120,7 +120,9 @@
...
+    %add_const(1) // This will be optimized out.
+    // stack: c + 1, input, ...
+    GT // Check it backwards: (input <= c) == (c + 1 > input)
```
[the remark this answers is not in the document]
dlubarov: Fixed as per our chat

evm/src/cpu/kernel/assembler.rs
```diff
@@ -530,6 +538,6 @@ mod tests {
         constants: HashMap<String, U256>,
     ) -> Kernel {
         let parsed_files = files.iter().map(|f| parse(f)).collect_vec();
-        assemble(parsed_files, constants)
+        assemble(parsed_files, constants, false)
```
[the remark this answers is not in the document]
dlubarov: Done

evm/src/cpu/kernel/optimizer.rs
dlubarov: Good point, added a note for a future improvement.

evm/src/cpu/kernel/optimizer.rs
dlubarov: Ah ok, I added a note for a future improvement.

nbgl (approved):
looks good!

PR #650, opened 2022-08-01
Test for parsing type 0 transactions

dlubarov:
The test didn't catch any bugs; it seems @wborgeaud already found them all for me :)

I made a couple related changes to `Interpreter`:
- Only enforce the `JUMPDEST` rule if we're in kernel mode.
- Flexibility to configure additional halt offsets, to help tests pieces of assembly which don't have the form of functions (i.e. they don't return to an address on the stack, so we can't pass `0xdeadbeef`).
- Convenience methods for dealing with the RLP and txn field segments of memory.

### DIFF
--- a/evm/src/cpu/kernel/tests/transaction_parsing/parse_type_0_txn.rs
+++ b/evm/src/cpu/kernel/tests/transaction_parsing/parse_type_0_txn.rs
@@ -46,6 +46,7 @@ fn process_type_0_txn() -> Result<()> {
     assert_eq!(interpreter.get_txn_field(To), 0.into());
     assert_eq!(interpreter.get_txn_field(Value), 100.into());
     assert_eq!(interpreter.get_txn_field(DataLen), 2.into());
+    assert_eq!(interpreter.get_txn_data(), &[0x42.into(), 0x42.into()]);
     assert_eq!(interpreter.get_txn_field(YParity), 1.into());
     assert_eq!(
         interpreter.get_txn_field(R),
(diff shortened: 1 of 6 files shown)


evm/src/cpu/kernel/tests/transaction_parsing/parse_type_0_txn.rs
```diff
@@ -0,0 +1,64 @@
...
+            "7c5c61ed975ebd286f6b027b8c504842e50a47d318e1e801719dd744fe93e6c6"
+        ))
+    );
+    assert_eq!(
+        interpreter.get_txn_field(S),
+        U256::from_big_endian(&hex!(
+            "1e7b5119b57dd54e175ff2f055c91f3ab1b53eba0b2c184f347cdff0e745aca2"
+        ))
+    );
+
```
wborgeaud, line 62: Can also add
```rust
assert_eq!(interpreter.get_txn_data(), &[0x42.into(), 0x42.into()]);
```
with
```rust
pub(crate) fn get_txn_data(&self) -> &[U256] {
    &self.memory.context_memory[0].segments[Segment::TxnData as usize].content
}
```

wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/tests/transaction_parsing/parse_type_0_txn.rs
dlubarov: Good idea; added.

PR #651, opened 2022-08-01
More constants for kernel ASM

dlubarov:
- `GlobalMetadata` - offsets for global kernel variables in memory
- `ContextMetadata` - offsets for context-specific kernel variables in memory
- `GAS_CONSTANTS`, based on the yellowpaper

Also move constants to a separate module since `aggregator` was getting long.

Depends on #650.

### DIFF
--- a/evm/src/cpu/kernel/context_metadata.rs
+++ b/evm/src/cpu/kernel/context_metadata.rs
@@ -5,17 +5,20 @@ pub(crate) enum ContextMetadata {
     ParentContext = 0,
     /// The program counter to return to when we return to the parent context.
     ParentProgramCounter = 1,
-    CalldataSize = 2,
-    ReturndataSize = 3,
+    /// If we're in a system call, this holds the userspace program counter to return to.
+    UserspaceProgramCounter = 2,
+    CalldataSize = 3,
+    ReturndataSize = 4,
 }
 
 impl ContextMetadata {
-    pub(crate) const COUNT: usize = 4;
+    pub(crate) const COUNT: usize = 5;
 
     pub(crate) fn all() -> [Self; Self::COUNT] {
         [
             Self::ParentContext,
             Self::ParentProgramCounter,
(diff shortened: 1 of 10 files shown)


evm/src/cpu/kernel/global_metadata.rs
```diff
@@ -0,0 +1,23 @@
+/// These metadata fields contain global VM state, stored in the `Segment::Metadata` segment of the
+/// kernel's context (which is zero).
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
+pub(crate) enum GlobalMetadata {
+    /// The larger context ID that has been used so far in this execution. Tracking this allows us
```
nbgl, line 5:
```suggestion
    /// The largest context ID that has been used so far in this execution. Tracking this allows us
```

nbgl (approved):

evm/src/cpu/kernel/context_metadata.rs
```diff
@@ -0,0 +1,37 @@
+/// These metadata fields contain VM state specific to a particular context.
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Ord, PartialOrd, Debug)]
+pub(crate) enum ContextMetadata {
+    /// The ID of the context which created this one.
+    ParentContext = 0,
+    /// The program counter to return to when we return to the parent context.
+    ParentProgramCounter = 1,
+    /// If we're in a system call, this holds the userspace program counter to return to.
+    UserspaceProgramCounter = 2,
```
nbgl, line 9: Oh, in #653 I put put that on the stack

nbgl: This also permits nested syscalls/traps.

dlubarov: Ah right, forgot we decided that. Removed this.

PR #652, opened 2022-08-01
Core transaction processing logic

dlubarov:
With lots of TODOs to fill in afterward; this is just a start.

Depends on #651.

### DIFF
--- a/evm/src/cpu/columns/mod.rs
+++ b/evm/src/cpu/columns/mod.rs
@@ -17,9 +17,6 @@ pub struct CpuColumnsView<T: Copy> {
     /// Filter. 1 if the row is part of bootstrapping the kernel code, 0 otherwise.
     pub is_bootstrap_kernel: T,
 
-    /// Filter. 1 if the row is part of bootstrapping a contract's code, 0 otherwise.
-    pub is_bootstrap_contract: T,
-
     /// Filter. 1 if the row corresponds to a cycle of execution and 0 otherwise.
     /// Lets us re-use columns in non-cycle rows.
     pub is_cpu_cycle: T,
--- a/evm/src/cpu/kernel/asm/core/call.asm
+++ b/evm/src/cpu/kernel/asm/core/call.asm
@@ -1,0 +1,59 @@
+// Handlers for call-like operations, namely CALL, CALLCODE, STATICCALL and DELEGATECALL.
+
+// Creates a new sub context and executes the code of the given account.
+global call:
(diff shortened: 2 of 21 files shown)


evm/src/cpu/columns/mod.rs
```diff
@@ -17,9 +17,6 @@ pub struct CpuColumnsView<T: Copy> {
     /// Filter. 1 if the row is part of bootstrapping the kernel code, 0 otherwise.
     pub is_bootstrap_kernel: T,
 
-    /// Filter. 1 if the row is part of bootstrapping a contract's code, 0 otherwise.
-    pub is_bootstrap_contract: T,
```
dlubarov, line 21: (No longer needed; we're going to bootstrap contracts in kernel assembly now.)

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -0,0 +1,19 @@
...
+global staticcall:
+    // TODO: Set flag to indicate
+    %jump(call_common)
+
+global delegatecall:
+    // TODO:
+    %jump(call_common)
+
+call_common:
+    %create_context
```
nbgl, line 16: I had a look at this and at `create_context` and this is indeed what I was imagining for contract calls.

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -0,0 +1,56 @@
...
+// - the instruction is invalid
+// - there are insufficient stack items
+// - a JUMP/JUMPI destination is invalid
+// - the new stack size would be larger than 1024, or
+// - state modification is attempted during a static call
+global exception:
+    // TODO
+    %jump(terminate_common)
+
+terminate_common:
```
nbgl, line 34: Do we want to charge gas here or has that been done?

dlubarov: Like charge the cost of these terminal instructions? It looks like `STOP`, `RETURN` and `REVERT` are free (in fact the only free ones). The specs do have a "zero gas" constant for them, so we could do `%consume_gas_const(@GAS_ZERO)` and optimize it out if we wanted to match.

Or charge gas for the transaction? I believe that's done at the start, and then leftover gas is refunded upon graceful termination (just not exceptions). I'll add a TODO to refund gas to the parent context upon `STOP`, `RETURN` and `REVERT` or `SELFDESTRUCT`.

PR #653, opened 2022-08-02
Permission levels, jumps, traps

nbgl:
Sorry, this is a really big PR. I will document it a little bit better and make the tests pass before asking for a review (but feel free to comment before that; I only expect to change comments and tests).


evm/src/cpu/decode.rs
```diff
@@ -1,3 +1,9 @@
+// We want to access `All`, `User`, and `Kernel` without having to go through `Availability`. Rust
+// wants constants to be all upper case, but these aren't really constants, just shortcuts. Also,
+// this syntax is consistent with the Rust std (think `None` and `Some`, which are not all upper
+// case).
+#![allow(non_upper_case_globals)]
+
```
wborgeaud, line 6: Totally unimportant but you can get the same behavior by adding the line
```rust
use Availability::{All, Kernel, User};
```

nbgl: Oh, that is much nicer! Thanks :)

evm/src/cpu/columns/general.rs
```diff
@@ -71,18 +93,39 @@ impl<T: Copy> BorrowMut<[T; NUM_SHARED_COLUMNS]> for CpuGeneralColumnsView<T> {
...
 #[derive(Copy, Clone)]
 pub(crate) struct CpuArithmeticView<T: Copy> {
     // TODO: Add "looking" columns for the arithmetic CTL.
     tmp: T, // temporary, to suppress errors
 }
 
+#[derive(Copy, Clone)]
+pub(crate) struct CpuJumpsView<T: Copy> {
+    // Assuming a limb size of 32 bits.
+    pub(crate) input0: [T; 8],
```
dlubarov, line 105: If I understand correctly, these arrays are like copies of `mem_value[0]` etc., and will later be constrained to match. What do you think of placing constraints on `mem_value` directly?

In this case we have columns to spare anyway, but in e.g. a logic row, it seems like we'd want to avoid any redundancy since we wouldn't have columns to spare.

evm/src/cpu/jumps.rs
```diff
@@ -0,0 +1,334 @@
...
+    Lazy::new(|| KERNEL.global_labels["handle_invalid_jump_dst"]);
+
+pub fn eval_packed_exit_kernel<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    nv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    let jumps_lv = lv.general.jumps();
+
+    yield_constr.constraint_transition(
```
dlubarov, line 22: Just to double check, with this design, we can use `PUSH 0` + `EXIT_KERNEL` to begin executing a user contract right? Looks like this code would read `is_kernel_mode=0` as we want.

And I guess if we wanted a system call to call another system call, we could do `PUSH return_label` + `%add_const(0x100000000)` + `%jump(inner_syscall)`, so that `inner_syscall`'s `EXIT_KERNEL` will see `is_kernel_mode=1`? (I don't think it seems like an essential feature, but if we did want to do it.)

evm/src/cpu/columns/general.rs
dlubarov: We chatted and decided to do this in a future change.

evm/src/cpu/columns/general.rs
```diff
@@ -91,5 +113,32 @@ pub(crate) struct CpuLogicView<T: Copy> {
...
+    pub(crate) input1: [T; 8],
+    pub(crate) output: [T; 8],
+
+    pub(crate) input0_upper_sum_inv: T,
+    pub(crate) input0_upper_zero: T,
+
+    pub(crate) dst_valid: T, // TODO: populate this (check for JUMPDEST)
+    pub(crate) dst_valid_or_kernel: T,
+    pub(crate) input0_jumpable: T,
```
dlubarov, line 128: nit: name seems slightly confusing to me, maybe call it `should_jump` or `jump_to_input1`?

dlubarov, line 128: I'm a bit confused about the meaning of this. The logic seems to be `input0_jumpable = dst_valid_or_kernel AND input0 = 0`?

evm/src/cpu/kernel/asm/handlers.asm
```diff
@@ -0,0 +1,4 @@
+global handle_stop:
```
dlubarov, line 1: We can use the recently-added `stop` in place of this `handle_stop`.

I think we can also use `exception` in place of `handle_invalid_jump_dst`. I think an invalid jump would be handled in the same way as other exceptions like stack underflow/overflow.

(And of course we have the `exp` you wrote, though it's currently a "pure" function, so I guess we should add a syscall wrapper around it.)

Feel free to rename them though; maybe `handle_` (or `sys_` or something) is a good way to indicate that they're syscall / trap handlers.

evm/src/cpu/columns/general.rs
nbgl: It’s `input0.bits[32..256] = 0`. If the high bits are nonzero, then it’s not a valid jump address (because there can’t be any code that far into memory), and it also won’t fit in our program counter.

evm/src/cpu/kernel/asm/handlers.asm
nbgl: Will change!

evm/src/cpu/columns/general.rs
```diff
@@ -91,5 +113,67 @@ pub(crate) struct CpuLogicView<T: Copy> {
...
+#[derive(Copy, Clone)]
+pub(crate) struct CpuJumpsView<T: Copy> {
+    // Assuming a limb size of 32 bits.
+    // The top stack value at entry (for jumps, the address; for `EXIT_KERNEL`, the address and new
+    // privilege level).
+    pub(crate) input0: [T; 8],
+    // For `JUMPI`, the second stack value (the predicate). For `JUMP`, 1.
+    pub(crate) input1: [T; 8],
+
+    // The output is only used when we fault due to an invalid address. It contains the usual
```
dlubarov, line 125: I don't think we necessarily need to save that info on the stack, since exceptions like an invalid jump will destroy the context (I don't think there's a way to recover).

Likewise for stack overflow/underflow, out-of-gas, REVERT, etc. - I don't think it's possible to go back to user mode after any exception.

(If there's a parent context (caller) we'll resume that, but its PC is stored separately in `ParentProgramCounter`.)

evm/src/cpu/columns/general.rs
```diff
@@ -91,5 +113,67 @@ pub(crate) struct CpuLogicView<T: Copy> {
     pub(crate) output: [T; 16],
 }
 
+#[derive(Copy, Clone)]
+pub(crate) struct CpuJumpsView<T: Copy> {
+    // Assuming a limb size of 32 bits.
+    // The top stack value at entry (for jumps, the address; for `EXIT_KERNEL`, the address and new
```
dlubarov, line 119: nit: I think `///` is nice for comments like this, even though it's not public. so tools know to associate it with the field. With IntelliJ it enables some nice hover text.

evm/src/cpu/columns/general.rs
```diff
@@ -91,5 +113,67 @@ pub(crate) struct CpuLogicView<T: Copy> {
...
+
+    // Note that the below flags are mutually exclusive.
+    // 1 if the JUMPI falls though (because input1 is 0); else 0.
+    pub(crate) should_continue: T,
+    // 1 if the JUMP/JUMPI does in fact jump to `input0`; else 0. This requires `input0` to be a
+    // valid destination (`input0[0]` is a `JUMPDEST` not in an immediate or we are in kernel mode
+    // and also `input0[1..7]` is 0) and `input1` to be nonzero.
+    pub(crate) should_jump: T,
+    // 1 if the JUMP/JUMPI faults; else 0. This happens when `input0` is not a valid destination
+    // (`input0[0]` is not `JUMPDEST` that is notin an immediate while we are in user mode, or
```
dlubarov, line 163: little typo: notin

evm/src/cpu/jumps.rs
```diff
@@ -0,0 +1,376 @@
...
+    yield_constr.constraint_transition(builder, kernel_constr);
+}
+
+pub fn eval_packed_jump_jumpi<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    nv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    let jumps_lv = lv.general.jumps();
+    let filter = lv.is_jump + lv.is_jumpi; // `JUMP` or `JUMPI`
```
dlubarov, line 61: Do we need to filter by `is_cpu_cycle` as well?

(If the degree increase is problematic I think we could always make `is_jump` dedicated. I think the plan was to share them with kernel bootloading rows, but we could do without.)

evm/src/cpu/jumps.rs
```diff
@@ -0,0 +1,376 @@
...
+
+    // If `JUMP`, re-use the `JUMPI` logic, but setting the second input (the predicate) to be 1.
+    yield_constr.constraint(lv.is_jump * (jumps_lv.input1[0] - P::ONES));
+    for &limb in &jumps_lv.input1[1..] {
+        // Set all limbs (other than the least-signigficant limb) to 0.
+        // NB: Technically, they don't have to be 0, as long as the sum
```
dlubarov, line 67: Don't they need to be zero (as you have it) to avoid the fault condition? So a dishonest prover can't cause a JUMP to fault.

evm/src/cpu/control_flow.rs
```diff
@@ -84,6 +90,8 @@ pub fn eval_packed_generic<P: PackedField>(
     let (halt_pc0, halt_pc1) = get_halt_pcs::<P::Scalar>();
     yield_constr
         .constraint_last_row((lv.program_counter - halt_pc0) * (lv.program_counter - halt_pc1));
+    // Finally, the last row must be in kernel mode.
+    yield_constr.constraint_last_row(lv.is_kernel_mode - P::ONES);
```
dlubarov, line 94: Good catch

evm/src/cpu/syscalls.rs
```diff
@@ -0,0 +1,110 @@
...
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::{CpuColumnsView, COL_MAP};
+use crate::cpu::kernel::aggregator::KERNEL;
+
+const NUM_SYSCALLS: usize = 2;
+
+fn make_syscall_list() -> [(usize, usize); NUM_SYSCALLS] {
+    let kernel = Lazy::force(&KERNEL);
```
dlubarov, line 19: Did you want to make these explicit? Typically we use `KERNEL` etc directly and let its `Deref` do this

dlubarov (approved):
LGTM otherwise!

evm/src/cpu/columns/general.rs
nbgl: That’s fair! I’ll get rid of this

evm/src/cpu/columns/general.rs
nbgl: Sure, I’ll change it!

evm/src/cpu/columns/general.rs
nbgl: Thanks!

evm/src/cpu/jumps.rs
nbgl: Ah yes, this is a documentation bug. I intended to make `is_jump` and `is_jumpi` dedicated due to degree constraints. I’ll make a note of that.

evm/src/cpu/jumps.rs
nbgl: The fault depends on the address (`input0`). This is setting the condition to 1 (`input1 = 1`). In other words, it implements `JUMP(addr)` as `JUMPI(addr=addr, condition=1)`

PR #654, opened 2022-08-02
Make the rest of gates public

Sladuca:
Motivation: Since gates are in a `Vec` of trait objects and we want to allow people to implement their own gates outside the library, they need to be able to construct them in order to deserialize circuit data (see #590).


dlubarov (approved):
I was a little unclear on the suggestion in #590, but at least making the gates public seems okay to me.

PR #657, opened 2022-08-04
Split up memory asm and add more helper functions

dlubarov:


evm/src/cpu/kernel/asm/memory/memcpy.asm
```diff
@@ -0,0 +1,52 @@
+
+// Copies `count` values from
+//     SRC = (src_ctx, src_segment, src_addr)
+// to
+//     DST = (dst_ctx, dst_segment, dst_addr).
+// These tuple definitions are used for brevity in the stack comments below.
+global memcpy:
```
dlubarov, line 7: (This is unchanged, just split off from memory.asm.)

npwardberkeley (approved):
Looks good! Some of the files (say `memcpy.asm`) seem pretty small, but I guess as things get more complex it makes sense to err on the side of splitting things up for clarity?

dlubarov:
Yeah true, I wouldn't usually use files that small. I figured with asm functions it could help make it more obvious where the function begins/ends, e.g. emphasizing that `memcpy_finish` is internal to `memcpy`.

It is a bit annoying adding an `include_str` for each file, but maybe we can eliminate that later.

PR #659, opened 2022-08-04
Support macro-local labels

dlubarov:
Again borrowing syntax from NASM. Example from the test:

    %macro spin
    %%start:
        PUSH %%start
        JUMP
    %endmacro

One thing this lets us do is create "wrapper" macros which call a function, then return to the code immediately following the macro call, such as

    %macro decode_rlp_scalar
        %stack (pos) -> (pos, %%after)
        %jump(decode_rlp_scalar)
    %%after:
    %endmacro

I used this to clean up `type_0.asm`.

However, since such macros need to insert `%%after` beneath any arguments in the stack, using them will be suboptimal in some cases. I wouldn't worry about it generally, but we might want to avoid them in performance-critical code, or functions with many arguments like `memcpy`.


dlubarov:
cc @nbgl, if you have feedback as well

wborgeaud (approved):
Nice 👍

PR #660, opened 2022-08-05
Packing memory operations

dlubarov:
Methods for encoding integers as bytes in memory, as well as the reverse, decoding bytes as integers. All big-endian.

Can be used for `MLOAD` later, but also for things like RLP, where integers get encoded as big-endian bytes.


npwardberkeley (approved):
Looks good!

PR #665, opened 2022-08-13
Trie related segments and metadata

dlubarov:
Summary of the design:
- Tries are stored as immutable, copy-on-write trees.
- All trie data is stored in the `TrieData` segment. Since it's immutable, data is never modified/deleted, new versions are just appended at the next unused offset.
- In order to support reverts, each context stores a pointer to the initial state trie version, plus pointers to the initial versions of each storage trie.

One variation which may be worth considering is storing the whole state trie as one big trie (with sub-tries for storage). Reverts would then be simpler - we'd replace a single pointer. I thought that approach might make hashing the trie a bit more complex, as the node associated with an account would be a special type of node, rather than just another leaf. Either approach seems reasonable though.


wborgeaud (approved):
Looks good to me!

PR #667, opened 2022-08-15
Improved Keccak implementation

dlubarov:
Based on the approach @SyxtonPrime described.

In terms of columns, the changes are:
- Store inputs (`A`) as `u32` limbs, rather than individual bits.
- Remove `C_partial`. It was used to store an intermediate product in a 5-way xor, but we've since realized that we can do a 5-way xor directly.
- Add `C_prime`, an intermediate result used to help verify the relation between `A` and `A'`.


SyxtonPrime (approved):

dlubarov:
cc @npwardberkeley FYI. Angus is working on a writeup so it will make sense later :)

PR #668, opened 2022-08-15
Expand inverse_2exp comment

dlubarov:


field/src/types.rs
```diff
@@ -192,10 +192,17 @@ pub trait Field:
     /// Compute the inverse of 2^exp in this field.
     #[inline]
     fn inverse_2exp(exp: usize) -> Self {
-        // The inverse of 2^exp is p-(p-1)/2^exp when char(F) = p and
-        // exp is at most the t=TWO_ADICITY of the prime field. When
-        // exp exceeds t, we repeatedly multiply by 2^-t and reduce
-        // exp until it's in the right range.
+        // Let p = char(F). Since 2^exp is a scalar, i.e. an element of GF_p,
```
unzvfu, line 195: I think the word 'scalar' is a bit ambiguous in this context. It would be simplest to say
> Since 2^exp is an element of GF_p, ...

For a clearer explicit statement, I think 'prime subfield' is the right terminology:
> Since 2^exp is in the prime subfield, i.e. an element of GF_p, ...

unzvfu (approved):
Just one minor nitpick; otherwise looks good!

field/src/types.rs
dlubarov: Thanks, I made that change.

PR #669, opened 2022-08-16
Keccak memory stark

dlubarov:
The Keccak permutation's inputs/outputs are larger than our CPU's memory bandwidth, so we can't do Keccak-f in a single cycle.

This introduces a Keccak memory table to sit between the CPU and Keccak tables. The CPU table can pass it just a start address, and this table will handle all the necessary memory lookups.

### DIFF
--- a/evm/Cargo.toml
+++ b/evm/Cargo.toml
@@ -23,6 +23,7 @@ rand_chacha = "0.3.1"
 rlp = "0.5.1"
 keccak-rust = { git = "https://github.com/npwardberkeley/keccak-rust" }
 keccak-hash = "0.9.0"
+tiny-keccak = "2.0.2"
 
 [dev-dependencies]
 criterion = "0.3.5"
--- a/evm/src/all_stark.rs
+++ b/evm/src/all_stark.rs
@@ -228,7 +228,7 @@ mod tests {
         keccak_memory_stark: &KeccakMemoryStark<F, D>,
         config: &StarkConfig,
     ) -> Vec<PolynomialValues<F>> {
-        keccak_memory_stark.generate_trace(vec![], config.fri_config.cap_height)
+        keccak_memory_stark.generate_trace(vec![], 1 << config.fri_config.cap_height)
     }
 
     fn make_logic_trace<R: Rng>(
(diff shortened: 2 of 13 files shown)


evm/Cargo.toml
```diff
@@ -23,6 +23,7 @@ rand_chacha = "0.3.1"
 rlp = "0.5.1"
 keccak-rust = { git = "https://github.com/npwardberkeley/keccak-rust" }
 keccak-hash = "0.9.0"
+tiny-keccak = "2.0.2"
```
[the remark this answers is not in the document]
dlubarov: Yeah good point. I was hesitant to add another, but it seemed like keccak-hash doesn't provide access to the Keccak-f permutation (at least it wasn't obvious from the docs), and keccak_rust seems to have very little usage. I'll try to switch other uses to tiny-keccak.

evm/src/all_stark.rs
```diff
@@ -107,16 +136,33 @@ fn ctl_logic<F: Field>() -> CrossTableLookup<F> {
...
 
 fn ctl_memory<F: Field>() -> CrossTableLookup<F> {
+    let cpu_memory_ops = (0..NUM_CHANNELS).map(|channel| {
+        TableWithColumns::new(
+            Table::Cpu,
+            cpu_stark::ctl_data_memory(channel),
+            Some(cpu_stark::ctl_filter_memory(channel)),
+        )
+    });
+    let keccak_memory_reads = (0..200).map(|i| {
```
npwardberkeley, line 146: Should be `KECCAK_WIDTH_BYTES` instead of `200`?

dlubarov: Good idea, done

evm/src/keccak_memory/keccak_memory_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+    let output_cols = (0..50).map(|i| {
+        Column::le_bytes((0..4).map(|j| {
+            let byte_index = i * 4 + j;
+            col_output_byte(byte_index)
+        }))
+    });
+    input_cols.chain(output_cols).collect()
+}
+
+pub(crate) fn ctl_looking_memory_read<F: Field>(i: usize) -> Vec<Column<F>> {
```
npwardberkeley, line 43: Hmm do we need these functions? Could we just call `ctl_looking_memory(i, true)` or `ctl_looking_memory(i, false)` instead? (Not sure which is more idiomatic...)

npwardberkeley (approved):
Looks great!

evm/src/keccak_memory/keccak_memory_stark.rs
dlubarov: Hm yeah not much benefit to the read/write wrappers. I removed them.

PR #670, opened 2022-08-16
`CommonCircuitData` serialization 2

Sladuca:
This is more or less a second attempt at #590.

### Summary of changes
* Implement methods for `CommonCircuitData` innards in `plonky2::util::serialization`.
* Add a new `GateSerialization` trait to be implemented by the user, and make the `write_common_circuit_data` and `read_common_circuit_data` take an implementor as an argument. This trait's implementation handles the details of how to distinguish between different gate types.
* Add the [`as_any` hack](https://bennetthardwick.com/rust/downcast-trait-object/) with `GateRef` by creating a wrapper trait `AnyGate` with a blanket implementation over all implementors of `Gate`. This makes implementing the `write_gate` method realistic
* Add a macro `impl_gate_serializer` that the user can use to auto-implement the `GateSerializer` trait given a list of the gates they wish to use.

### Why so convoluted?

Since the deserializer needs to be able to tell what gates it's reading, the serializer needs to serialize as if the `Arc<dyn Gate>`s were variants of a tagged enum (or something equivalent). We want to avoid something like `GateKind` from #590 since we want to allow the user to add their own gates, so the tagging has to happen on the user's side.

In order for that to happen, the we need some runtime reflection - hence the `as_any` hack. The macro is a convenience so the user doesn't have to implement this themselves.

I'm not sure if there's a way to go about this without the `as_any` hack, since the serializer has to be able to case on the underlying type in order to produce the metadata necessary for deserialization.


Nashtare:
As this approach been abandoned? I was thinking on iterating further from this

Sladuca:
I haven't worked on it since then, so feel free.

dlubarov:
@Nashtare I think Plonky2 can only support a rather limited form of circuit serialization, since it doesn't support dynamic gates whose constraints aren't known until runtime. I'm not sure how useful it would be given that limitation? But if there's a clear use case then we can move forward with this.

Nashtare:
@dlubarov I was thinking of using it in the context of offline preprocessing, similarly to verification keys for SNARKs provided as serialized files, that are then loaded at runtime. This could be used for instance for `AllRecursiveCircuits`, to remove the need for the heavy offline preprocessing when bootstrapping a node.
To a larger extent, any application using `plonky2` could use a similar approach.
I may have missed something obvious that makes this approach infeasible / impractical though.

Nashtare:
Hi @Sladuca @dlubarov. I apologize for the delay, I've been busy with other priorities, but I finally got a PoC for global serialization. I'm getting for the `AllRecursiveCircuits` of the `empty_txn_list` example (degrees 9 to 19), a serialization time of about 26sec, and deserialization time of about 12sec. There are probably some additional tweaks that can be done to speed this up further.
I had to do several design choices that I'm not super happy with to be able to serialize generators, hence if you're interested by this, I'll gladly clean it up and open a draft PR for getting your inputs.

Nashtare:
I haven't pushed anything yet, wanted to clean it up before. I'll do it in the coming days, let me ping you when it's ready!

Nashtare:
@Sladuca You can have a look here: [plonky2/serialization](https://github.com/toposware/plonky2/tree/serialization).

Given that I am keeping in the `empty_txn_list` test the old `all_circuits` to compare with, instead of deserializing from a file, running time may blow up if the test starts using your SWAP. With enough RAM, serialization time for this is about 12 sec, and deserialization takes about 16sec.

I insist there are still several things to cleanup / optimize, so I'll gladly take any comments / suggestions you may have!

Nashtare:
This one has been handled #981 and subsequent PRs.

PR #672, opened 2022-08-17
Delete opcode column

nbgl:
The `opcode` column is just a linear combination of the `opcode_bits` columns, so we can delete it.


dlubarov (approved):
LGTM

PR #674, opened 2022-08-18
Change logic limb size to 32 bits

nbgl:
This was mostly trivial, except it required a rewrite of the `EQ` instruction (the previous method relied on the small limb size).


dlubarov (approved):
LGTM

PR #676, opened 2022-08-19
Change `from_biguint`'s behavior with extension fields

dlubarov:
These appear to be unused for extension fields, so we're free to change the mapping without breaking anything.

As the TODO says, the mapping that's currently implemented doesn't seem natural or useful. It seems more natural to treat the `BigUint` as a base field element, potentially in a non-canonical form.

The new name `from_noncanonical_biguint` is meant to emphasize this behavior.


unzvfu (approved):
lgtm!

Issue #575, opened 2022-06-23
Re-use Rustc build cache between CI invocations


nbgl:
The CI now takes ~15m and it makes me want to tear my hair out.[^1] Compilation is responsible for ~70% of that. We should be able to speed up compilation by storing the Rustc build cache and re-using it in following builds.

[^1]: I know that this much faster than many teams’ CI. My thoughts are with those teams in these trying times.

PR #679, opened 2022-08-21
Added caching for GitHub Workflows

BGluth:
Added caching to the CI workflows.

Under the hood, [rust-cache](https://github.com/Swatinem/rust-cache) calls GitHub's more generic [cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) action with a configuration that makes sense for a Cargo project. The high level idea is this:

- Generate a cache key based on variables that would normally require a full/partial recompile (eg. compiler version, `Cargo.lock`, etc.).
- Check if there is a cache entry for this key.
- If the key is missing, then after the job completes, create a new cache entry for this key containing build artifacts (`~/.cargo` & `target/`).
- If the key exists, then copy the cached directories directly into the workflow container.

So for a cache hit, this should significantly speed up build times since we can skip the large majority of downloading/compiling dependencies. For hopefully the majority of cases, we should only need to recompile whatever project crates were changed by the PR.

A few other notes:
- Since we're using nightly, this means every day our cache entries are going to become invalidated whenever a new nightly version is released.
- By default, caching also includes the workflow job in the key that it's a part of, which in our case is good since `test` and `lints` both use different `RUSTFLAGS`.
- The underlying GitHub cache action ejects cache entries after 10GB total is used or are stale for 7 days.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -23,6 +23,9 @@ jobs:
           toolchain: nightly
           override: true
 
+      - name: rust-cache
+        uses: Swatinem/rust-cache@v2
+
       - name: Run cargo test
         uses: actions-rs/cargo@v1
         with:
@@ -47,6 +50,9 @@ jobs:
           toolchain: nightly
           override: true
           components: rustfmt, clippy
+
+      - name: rust-cache
+        uses: Swatinem/rust-cache@v2
 
       - name: Run cargo fmt
         uses: actions-rs/cargo@v1
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -17,6 +17,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Install nightly toolchain
+        id: rustc-toolchain
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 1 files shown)


nbgl:
Oh nice! This fixes #575.

BGluth:
Not quite ready to merge. While it shaved 5 minutes off the build time (total of ~10), it's still spending ~6 minutes building the sub-crates. Although `rust-cache` does not recommend caching sub-crates, I think in our case it may be very worthwhile.

BGluth:
Ended up swapping out `rust-cache` for the more generic `cache` action. `rust-cache` doesn't offer a lot in terms of configuration (eg. incremental compilation is always disabled).

The best I think we can do is get build times down to close to ~50 seconds, which is still a big improvement. This 50 seconds is coming from all sub-crates being detected as stale regardless if any src files changed or not. The cause of this I think is Git not storing the modified time (`mtime`) of tracked files and always setting `mtime` to the current time on checkout. So when Cargo performs a stale check, it determines any cached local sub-crate artifacts are stale since the src files are always newer. I don't see a good way around this unless we want to start [tracking mtime](https://stackoverflow.com/questions/2179722/checking-out-old-files-with-original-create-modified-timestamps) in the repo, which is definitely not worth doing for this imo. Incremental compilation really brings the time down however, and build times of 50 seconds (total CI time is ~4 min) are still a very good improvement.

Also, probably not worth adding this in unless we realize it will make a big improvement, but if we find that PRs are making significant changes to multiple sub-crates, this may start negatively affecting how much time incremental compilation saves us (since multiple PRs may be sharing a single cache entry). If this becomes a problem, we can use more cache storage to always store a cache entry specific to each PR but always base each new cache entry off of a random existing one.

#### Build Times (Not including tests)

Caching |  incremental  |    Time
------------|-------------------|--------
No          | No                 | ~10 min
Yes         | No                 | ~6 min
Yes         | Yes                | ~50 s

cpubot (approved):
LGTM 👍

PR #680, opened 2022-08-22
Change visibility of `ECDSASecretKeyTarget` and `ECDSAPublicKeyTarget`'s fields

qope:
`ECDSASecretKeyTarget` and `ECDSAPublicKeyTarget`'s fields are private, so you cannot initialize them outside this crate. `ECDSASecretKeyTarget` and `ECDSAPublicKeyTarget` are  parameters of `verify_message_circuit`, so they should be able to be initialized outside this crate.


dlubarov:
Thanks, I think this seems fine, unless @wborgeaud @npwardberkeley had a different API in mind?

wborgeaud (approved):
> Thanks, I think this seems fine, unless @wborgeaud @npwardberkeley had a different API in mind?

No this looks good to me.

PR #682, opened 2022-08-22
Support accessing local row in CTLs

dlubarov:
Previously CTLs could only access the "next" row; this lets them access the "local" row as well.

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -117,10 +117,9 @@ impl<F: Field> Column<F> {
 
     /// Evaluate on an row of a table given in column-major form.
     pub fn eval_table(&self, table: &[PolynomialValues<F>], local_row: usize) -> F {
-        let mut next_row = local_row + 1;
-        if next_row == table[0].len() {
-            next_row = 0;
-        }
+        let degree = table[0].len();
+        debug_assert!(degree.is_power_of_two());
+        let next_row = (local_row + 1) & (degree - 1); // Equivalent to % degree.
 
         self.linear_combination
             .iter()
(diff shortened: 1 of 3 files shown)


evm/src/cross_table_lookup.rs
```diff
@@ -59,54 +74,90 @@ impl<F: Field> Column<F> {
...
             + FE::from_basefield(self.constant)
     }
 
     /// Evaluate on an row of a table given in column-major form.
-    pub fn eval_table(&self, table: &[PolynomialValues<F>], row: usize) -> F {
+    pub fn eval_table(&self, table: &[PolynomialValues<F>], local_row: usize) -> F {
+        let mut next_row = local_row + 1;
+        if next_row == table[0].len() {
+            next_row = 0;
+        }
```
wborgeaud, line 123: nit: Wouldn't `let next_row = (local_row + 1) % table[0].len()` be clearer?

evm/src/cross_table_lookup.rs
```diff
@@ -279,16 +330,22 @@ fn partial_products<F: Field>(
     let mut partial_prod = F::ONE;
     let degree = trace[0].len();
     let mut res = Vec::with_capacity(degree);
-    for i in 0..degree {
+    for next_row in 0..degree {
+        let local_row = if next_row == 0 {
+            degree - 1
+        } else {
+            next_row - 1
+        };
```
wborgeaud, line 338: nit: Could also do `let local_row = (next_row + degree - 1) % degree;` here.

wborgeaud (approved):
Looks good to me!

evm/src/cross_table_lookup.rs
dlubarov: I agree it's clearer, but I think the compiler would resort to the CPU's division instructions, since it wouldn't know that `local_row` is smaller than the table or that it's a power of two. Do you think `& (table[0].len() - 1)` would be any clearer?

wborgeaud: Oh I see. Using `&` could be clearer along with a small comment reminding that `table[0].len()` is a power of 2, but otherwise I think it's totally fine to keep the conditional.

PR #685, opened 2022-08-23
Enumerate `constants_to_targets` in a deterministic order

### LINKED ISSUE #684: Potential Bug: CircuitData is not always built the same for circuits containing Merkle Proofs - breaks usage of merkle proofs within circuits.
When trying to prove knowledge of a leaf in a merkle tree using the `builder.verify_merkle_proof` api the builder generates different CircuitData every time.
This is true for both the `CircuitConfig::standard_recursion_zk_config()` and `CircuitConfig::standard_recursion_config()` configs.

This breaks using merkle proofs inside circuits.
Meaning that when sending the proof to a different device, that device won't be able to always verify the proof since it generated different CircuitData.
For circuits without the merkle proofs this doesn't occur - the CircuitData is always generated to be exactly the same.
Also, about 50% of time both the prover and verify will generate the same CircuitData and then the proof verification does indeed pass.

I created a small repo that reproduces this exactly and makes playing around with this bug easy and clear.
https://gitlab.com/Shalevos/merkle-tree-circuit-data-bug
I tried for a few hours to figure this out myself but would love the help of the devs on this.

Thanks,
Shalev

dlubarov:
Fixes #684.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -2,6 +2,7 @@ use std::cmp::max;
 use std::collections::{BTreeMap, HashMap, HashSet};
 use std::time::Instant;
 
+use itertools::Itertools;
 use log::{debug, info, Level};
 use plonky2_field::cosets::get_unique_coset_shifts;
 use plonky2_field::extension::{Extendable, FieldExtension};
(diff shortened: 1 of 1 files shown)


wborgeaud (approved):
👍

PR #687, opened 2022-08-24
Simpler CPU <-> memory CTL

dlubarov:
Instead of having `NUM_CHANNELS` CTLs between the CPU and memory, we can have a single CTL with `NUM_CHANNELS` lookers, all from the CPU table.

This means the memory table no longer needs to know about channels, so we can simplify it a bit and combine its channel flags into a single flag (to distinguish "real" and "padding" rows).

Depends on #689.


dlubarov:
I think the CTL framework should support this in theory, but at the moment the CTL constraints are failing. I'll debug but please let me know if you spot an issue.

wborgeaud (approved):
Looks good to me!

PR #688, opened 2022-08-24
EVM arithmetic unit: unsigned comparisons

unzvfu:
Use the subtraction operation to implement unsigned less than and greater than operations. A small amount of refactoring was involved.

### DIFF
--- a/evm/src/arithmetic/add.rs
+++ b/evm/src/arithmetic/add.rs
@@ -9,6 +9,21 @@ use crate::arithmetic::columns::*;
 use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
 use crate::range_check_error;
 
+pub(crate) fn u256_add_cc(input0: [u64; N_LIMBS], input1: [u64; N_LIMBS]) -> ([u64; N_LIMBS], u64) {
+    // Input and output have 16-bit limbs
+    let mut output = [0u64; N_LIMBS];
+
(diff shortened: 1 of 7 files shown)


nbgl (approved):
Looks good!

I think this is the correct/straightforward approach, although the way it’s described confused me a little. It is, in fact, equivalent to just performing `input0 - input1` and just returning the borrow bit.

For both subtraction and comparison we have:
- input0,
- input1,
- difference (modulo 2^256),
- borrow bit.

For subtraction, the difference is the output, whereas the borrow bit is provided by the prover. For comparison, the _borrow bit_ is output, and the difference is provided by the prover.

I personally find this explanation more intuitive: they are the same thing, but we’re just outputting a different column.

unzvfu:
> Looks good!
> 
> I think this is the correct/straightforward approach, although the way it’s described confused me a little. It is, in fact, equivalent to just performing `input0 - input1` and just returning the borrow bit.
> 
> For both subtraction and comparison we have:
> 
>     * input0,
> 
>     * input1,
> 
>     * difference (modulo 2^256),
> 
>     * borrow bit.
> 
> 
> For subtraction, the difference is the output, whereas the borrow bit is provided by the prover. For comparison, the _borrow bit_ is output, and the difference is provided by the prover.
> 
> I personally find this explanation more intuitive: they are the same thing, but we’re just outputting a different column.

Thanks for the feedback! I'll update the docs to make that clearer.

PR #689, opened 2022-08-24
Fix for CTL challenges

dlubarov:
See this line -

```rust
challenges: ctl_data.challenges.challenges[i % config.num_challenges],
```

This doesn't work if we have multiple lookers from the same table; then `zs_columns` will contain multiple contiguous entries for the same challenge.

We could fix the index calculation, but it seems a bit error-prone. Seems easier to store the specific challenge as part of `zs_columns`.


wborgeaud (approved):
Good catch! Nice change, I agree it's clearer to store the challenge with the polynomial 👍

PR #691, opened 2022-08-24
Remove keccak_rust in favor of tiny-keccak

dlubarov:
`keccak_rust` doesn't seem to have much usage, and it treats `x` as the major axis of its 5x5 input.  This is not exactly wrong, since Keccak itself doesn't have a notion of axis order. However, there is a convention for mapping bits of the cube to a flat list of bits, which is

> The mapping between the bits of `s` and those of `a` is `s[w(5y + x) + z] = a[x][y][z]`.

Obeying this convention would be awkward with `keccak_rust` - the words in memory would need to be transposed.


npwardberkeley (approved):
Good catch, thanks!

PR #692, opened 2022-08-25
Gate documentation

npwardberkeley:


plonky2/src/gates/random_access.rs
```diff
@@ -24,9 +24,15 @@ use crate::plonk::vars::{
 /// A gate for checking that a particular element of a list matches a given value.
 #[derive(Copy, Clone, Debug)]
 pub struct RandomAccessGate<F: RichField + Extendable<D>, const D: usize> {
+    // Number of bits in the index (log2 of the list size).
```
dlubarov, line 27: nit: IMO best to use `///` for such comments so that tools will associate them with the field.

npwardberkeley: Ahh right, thanks!

npwardberkeley:
Should be ready for review! There isn't much, it's just a few comments, and then reorganizing `interpolation` stuff. After merging this we can add the detailed documentation William started in #546

dlubarov (approved):
LGTM

PR #693, opened 2022-08-25
Have witness generation take a partial trie instead of Merkle proofs

dlubarov:


evm/src/generation/state.rs
```diff
@@ -19,9 +19,6 @@ pub(crate) struct GenerationState<F: Field> {
 
     pub(crate) keccak_inputs: Vec<[u64; keccak::keccak_stark::NUM_INPUTS]>,
     pub(crate) logic_ops: Vec<logic::Operation>,
-
-    /// Non-deterministic inputs provided by the prover.
-    pub(crate) prover_inputs: Vec<U256>,
```
dlubarov, line 24: (This single prover input buffer no longer really makes sense, now that William's `PROVER_INPUT` change lets us tell the prover what kind of input is requested. We might add more specific buffers for e.g. transaction data later though.)

evm/src/generation/partial_trie.rs
```diff
@@ -0,0 +1,24 @@
+use ethereum_types::U256;
+
+/// A partial trie, or a sub-trie thereof.
+pub enum PartialTrie {
```
dlubarov, line 4: cc @BGluth. I changed `Leaf` as per your comment but let me know if you have other feedback on this.

evm/src/generation/partial_trie.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// An empty trie.
+    Empty,
+    /// The digest of trie whose data does not need to be stored.
+    Hash(U256),
+    /// A branch node, which consists of 16 children and an optional value.
+    Branch([Box<PartialTrie>; 16], Option<U256>),
+    /// An extension node, which consists of a list of nibbles and a single child.
+    Extension(Nibbles, Box<PartialTrie>),
+    /// A leaf node, which consists of a list of nibbles and a value.
+    Leaf(Nibbles, Vec<u8>),
```
wborgeaud, line 14: nit: Maybe it would be clearer to use named fields here instead of tuples. Like
```rust
Branch { children: [Box<PartialTrie>; 16], value: Option<U256>},
Extension { nibbles: Nibbles, child: Box<PartialTrie>},
Leaf { nibbles: Nibbles, value: Vec<u8>}
```

wborgeaud (approved):
Looks good to me!

evm/src/generation/partial_trie.rs
dlubarov: Good idea 👍

evm/src/generation/partial_trie.rs
BGluth: Yeah this seems good!

PR #696, opened 2022-08-26
Public memory, part 1

dlubarov:
Sorry for the large PR; a lot of it is just ripping out the existing public input related code. The main new addition is `PublicValues`, which contains trie roots (before and after) and block metadata. These values could be thought of as public inputs which live in memory, but I wanted to avoid "input" to avoid confusion with `GenerationInputs`.

This doesn't yet include the change to the `Z(last)` comparison, where the verifier itself will (in a sense) read the public values, by computing the (reduced) CTL rows for those special read operations and factoring them into the `Z(last)` comparison. I'll add that later.


evm/src/proof.rs
```diff
@@ -42,7 +44,56 @@ pub(crate) struct AllProofChallenges<F: RichField + Extendable<D>, const D: usiz
...
+
+#[derive(Debug, Clone, Default)]
+pub struct TrieRoots {
+    pub state_root: U256,
+    pub transactions_root: U256,
+    pub receipts_root: U256,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct BlockMetadata {
```
wborgeaud, line 67: Do we also need a `block_basefee` here to support the `BASEFEE` opcode?

evm/src/util.rs
```diff
@@ -40,3 +41,29 @@ pub fn trace_rows_to_poly_values<F: Field, const COLUMNS: usize>(
         .map(|column| PolynomialValues::new(column))
         .collect()
 }
+
+/// Returns the 32-bit little-endian limbs of a `U256`.
+pub(crate) fn u256_limbs<F: Field>(u256: U256) -> [F; 8] {
+    u256.0
+        .into_iter()
+        .flat_map(|limb_64| {
+            let lo = (limb_64 & 0xFFFFFFFF) as u32;
```
wborgeaud, line 50: nit: Isn't that the same as `limb_64 as u32`? Or do you think it's clearer that way?

wborgeaud (approved):
Looks good to me!

evm/src/util.rs
dlubarov: Ah, somehow I thought it would fail in debug mode, but looks like that's not the case.

evm/src/proof.rs
dlubarov: I missed that, thanks!

PR #697, opened 2022-08-26
`NUM_TABLES` constant

wborgeaud:
Add a `NUM_TABLES` constant (replacing the current `Table::num_tables()`) and replace all `Vec<T>` indexed by tables with `[T; NUM_TABLES]`.

### DIFF
--- a/evm/src/generation/mod.rs
+++ b/evm/src/generation/mod.rs
@@ -90,7 +90,7 @@ pub(crate) fn generate_traces<F: RichField + Extendable<D>, const D: usize>(
 
     let cpu_trace = trace_rows_to_poly_values(cpu_rows);
     let keccak_trace = all_stark.keccak_stark.generate_trace(keccak_inputs);
-    let keccak_memory_trace = all_stark.keccak_memory_stark.generate_trace(vec![], 16); // TODO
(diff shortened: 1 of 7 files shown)


evm/src/generation/mod.rs
```diff
@@ -90,9 +90,16 @@ pub(crate) fn generate_traces<F: RichField + Extendable<D>, const D: usize>(
 
     let cpu_trace = trace_rows_to_poly_values(cpu_rows);
     let keccak_trace = all_stark.keccak_stark.generate_trace(keccak_inputs);
+    let keccak_memory_trace = all_stark.keccak_memory_stark.generate_trace(vec![], 0); // TODO
```
wborgeaud, line 93: Not sure which parameters to use here. Do we already have this data here?

dlubarov: Sorry I forgot this, it should pass `operations = keccak_memory_inputs` (from `state` which is unpacked a few lines above), and `min_rows = 1 << cap_height` (looks like we don't have access to that config yet so we'd need to add it). I can add it afterward if it's easier.

dlubarov (approved):
Good idea, looks like it already caught one issue!

evm/src/generation/mod.rs
wborgeaud: Thanks! Made the change

PR #698, opened 2022-08-26
cleaner interface for witness extension methods in `u32` and `ecdsa`

Sladuca:
Having to import and directly call a `witness_set_u32_target` function is a bit clunky.

Instead, add the  functions to a `WitnessU32` subtrait of `Witness` and do a blanket `impl WitnessU32 for T where T: Witness` for it.

in other words, we go from this:
```rust
use plonky2_u32::witness_set_u32_target;

let mut pw = PartialWitness::new();
witness_set_u32_target(&mut pw, val);
```
to this:
```rust
use plonky2_32::WitnessU32;

let mut pw = PartialWitness::new();
pw.set_u32_target(val);
```

I Also did the same for `BigUint` in `ecdsa` and something similar for `GeneratedValues`.

### DIFF
--- a/u32/src/witness.rs
+++ b/u32/src/witness.rs
@@ -1,21 +1,32 @@
 use plonky2::iop::generator::GeneratedValues;
 use plonky2::iop::witness::Witness;
-use plonky2_field::types::Field;
+use plonky2_field::types::{Field, PrimeField64};
 
 use crate::gadgets::arithmetic_u32::U32Target;
 
-pub fn generated_values_set_u32_target<F: Field>(
-    buffer: &mut GeneratedValues<F>,
-    target: U32Target,
-    value: u32,
-) {
-    buffer.set_target(target.0, F::from_canonical_u32(value))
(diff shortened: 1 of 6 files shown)


u32/src/witness.rs
```diff
@@ -1,21 +1,32 @@
...
-    witness: &mut W,
-    target: U32Target,
-    value: u32,
-) {
-    witness.set_target(target.0, F::from_canonical_u32(value))
+impl<T: Witness<F>, F: PrimeField64> WitnessU32<F> for T {
+    fn set_u32_target(&mut self, target: U32Target, value: u32) {
+        self.set_target(target.0, F::from_canonical_u32(value));
+    }
+    fn get_u32_target(&self, target: U32Target) -> (u32, u32) {
```
npwardberkeley, line 16: nit: newline here

npwardberkeley (approved):
Looks great! Thanks for the contribution

PR #699, opened 2022-08-26
Make jumps, logic, and syscalls read from/write to memory columns

nbgl:
Removes all the various `input`/`output` columns, and puts constraints directly on memory values

### DIFF
--- a/evm/src/all_stark.rs
+++ b/evm/src/all_stark.rs
@@ -335,16 +335,16 @@ mod tests {
             );
 
             let input0_bit_cols = logic::columns::limb_bit_cols_for_input(logic::columns::INPUT0);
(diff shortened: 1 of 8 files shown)


evm/src/all_stark.rs
```diff
@@ -271,45 +271,18 @@ mod tests {
             cpu_trace_rows.push(row.into());
         }
 
-        for i in 0..num_logic_rows {
+        // Pad to `num_memory_ops` for memory testing.
```
nbgl, line 274: The memory test and logic/syscall/jump tests now both use the memory value columns. Give the memory test its own rows to avoid conflicts (previously it modified existing rows).

dlubarov (approved):
LGTM!

PR #701, opened 2022-08-27
Save columns by verifying invalid opcodes in software

nbgl:
Currently, to execute the invalid opcode exception handler, we must satisfy a bunch of constraints that ensure that the current opcode is in fact invalid.

We can save 20 columns by _not_ performing those checks in the CPU table (any opcode can decode the invalid opcode flag), and instead making the invalid opcode exception handler do them (it `PANIC`s if we trapped incorrectly)


evm/src/cpu/kernel/asm/core/invalid.asm
```diff
@@ -0,0 +1,34 @@
...
+    %mload_current_code
+    // stack: opcode
+    // Python:
+    //   >>> invalid_ranges_inclusive = [(0x0c, 0x0f), (0x1e, 0x1f), (0x21, 0x2f), (0x49, 0x4f),
+    //   ...                             (0x5c, 0x5f), (0xa5, 0xef), (0xf6, 0xf9), (0xfb, 0xfc),
+    //   ...                             (0xfe, 0xfe)]
+    //   >>> inclusive_range_bits = lambda start, end: sum(1 << i for i in range(start, end + 1))
+    //   >>> hex(sum(inclusive_range_bits(start, end) for start, end in invalid_ranges_inclusive))
+    //   '0x5bc0ffffffffffffffffffe00000000000000000f000fe000000fffec000f000'
+    PUSH 0x5bc0ffffffffffffffffffe00000000000000000f000fe000000fffec000f000
```
dlubarov, line 23: Could we compute this in Rust based on the `OPCODES` data, and pass it to this asm as a constant (`fn evm_constants()`)? Would be nice if we don't need to update it manually.

evm/src/cpu/decode.rs
```diff
@@ -28,122 +20,103 @@ use Availability::{All, Kernel, User};
...
+    (0x5f, 0, true, COL_MAP.is_set_receipt_root),
+    (0x60, 5, false, COL_MAP.is_push), // 0x60-0x7f
+    (0x80, 4, false, COL_MAP.is_dup),  // 0x80-0x8f
+    (0x90, 4, false, COL_MAP.is_swap), // 0x90-0x9f
+    (0xa0, 0, false, COL_MAP.is_log0),
+    (0xa1, 0, false, COL_MAP.is_log1),
+    (0xa2, 0, false, COL_MAP.is_log2),
+    (0xa3, 0, false, COL_MAP.is_log3),
+    (0xa4, 0, false, COL_MAP.is_log4),
     // Opcode 0xa5 is PANIC when Kernel. Make the proof unverifiable by giving it no flag to decode to.
```
dlubarov, line 104: Just to double check, does this strategy for `PANIC` still work? It looks like now `handle_invalid` would panic, causing an infinite loop. I suppose it's unsatisfiable since the CPU can never end the loop and reach `halt`?

evm/src/cpu/decode.rs
```diff
@@ -184,15 +157,19 @@ pub fn generate<F: RichField>(lv: &mut CpuColumnsView<F>) {
...
+    for (oc, block_length, kernel_only, col) in OPCODES {
+        let available = !kernel_only || kernel;
         let opcode_match = top_bits[8 - block_length] == oc;
-        lv[col] = F::from_bool(available && opcode_match);
+        let flag = available && opcode_match;
+        lv[col] = F::from_bool(flag);
+        if flag && any_flag_set {
+            panic!("opcode matched multiple flags");
+        }
+        any_flag_set = any_flag_set || flag;
```
dlubarov, line 169: nit: could use `||=`

evm/src/cpu/kernel/asm/core/invalid.asm
```diff
@@ -0,0 +1,34 @@
...
+    // stack: invalid_opcodes_bitmap, opcode
+    SWAP1
+    // stack: opcode, invalid_opcodes_bitmap
+    SHR
+    %and_const(1)
+    // stack: opcode_is_invalid
+    // if the opcode is indeed invalid, then perform an exceptional exit
+    %jumpi(fault_exception)
+    // otherwise, panic because this trap should not have been entered
+panic:
```
dlubarov, line 33: nit: there's already a `global panic`

dlubarov (approved):
LGTM otherwise, nice idea!

evm/src/cpu/decode.rs
nbgl: Yeah, exactly. It’s a bit roundabout, but `PANIC` still works.

evm/src/cpu/decode.rs
nbgl: That was my original attempt, but it turns out that `||=` is not a Rust operator: https://doc.rust-lang.org/reference/expressions/operator-expr.html#compound-assignment-expressions

evm/src/cpu/kernel/asm/core/invalid.asm
nbgl: sure!

evm/src/cpu/kernel/asm/core/invalid.asm
nbgl: Yeah, you’re right, I got a bit lazy there. I’ll think of a nice way of generating this before merging

dlubarov (approved):

PR #702, opened 2022-08-30
Keccak sponge STARK

dlubarov:
It contains a row for each absorb step of the sponge.

I had some mostly-finished assembly for this, but this should be much more efficient.

### DIFF
--- a/evm/src/keccak_sponge/columns.rs
+++ b/evm/src/keccak_sponge/columns.rs
@@ -5,8 +5,7 @@ use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
 
 pub(crate) const KECCAK_WIDTH_BYTES: usize = 200;
 pub(crate) const KECCAK_WIDTH_U32S: usize = KECCAK_WIDTH_BYTES / 4;
-pub(crate) const KECCAK_RATE_BITS: usize = 1088;
-pub(crate) const KECCAK_RATE_BYTES: usize = KECCAK_RATE_BITS / 8;
+pub(crate) const KECCAK_RATE_BYTES: usize = 136;
 pub(crate) const KECCAK_RATE_U32S: usize = KECCAK_RATE_BYTES / 4;
 pub(crate) const KECCAK_CAPACITY_BYTES: usize = 64;
 pub(crate) const KECCAK_CAPACITY_U32S: usize = KECCAK_CAPACITY_BYTES / 4;
@@ -51,8 +50,8 @@ pub(crate) struct KeccakSpongeColumnsView<T: Copy> {
     /// The block being absorbed, which may contain input bytes and/or padding bytes.
     pub block_bytes: [T; KECCAK_RATE_BYTES],
 
-    /// The rate part of the sponge, after the current block is xor'd in, but before the permutation
-    /// is applied.
+    /// The rate part of the sponge, encoded as 32-bit chunks, after the current block is xor'd in,
+    /// but before the permutation is applied.
     pub xored_rate_u32s: [T; KECCAK_RATE_U32S],
 
     /// The entire state (rate + capacity) of the sponge, encoded as 32-bit chunks, after the
--- a/evm/src/keccak_sponge/columns.rs
+++ b/evm/src/keccak_sponge/columns.rs
@@ -1,0 +1,118 @@
+use std::borrow::{Borrow, BorrowMut};
+use std::mem::{size_of, transmute};
+
+use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
+
+pub(crate) const KECCAK_WIDTH_BYTES: usize = 200;
+pub(crate) const KECCAK_WIDTH_U32S: usize = KECCAK_WIDTH_BYTES / 4;
+pub(crate) const KECCAK_RATE_BITS: usize = 1088;
+pub(crate) const KECCAK_RATE_BYTES: usize = KECCAK_RATE_BITS / 8;
+pub(crate) const KECCAK_RATE_U32S: usize = KECCAK_RATE_BYTES / 4;
+pub(crate) const KECCAK_CAPACITY_BYTES: usize = 64;
+pub(crate) const KECCAK_CAPACITY_U32S: usize = KECCAK_CAPACITY_BYTES / 4;
+
+#[repr(C)]
+#[derive(Eq, PartialEq, Debug)]
+pub(crate) struct KeccakSpongeColumnsView<T: Copy> {
+    /// 1 if this row represents a dummy operation (for padding the table); 0 otherwise.
+    pub is_dummy: T,
+
+    /// 1 if this row represents a full input block, i.e. one in which each byte is an input byte,
+    /// not a padding byte; 0 otherwise.
+    pub is_full_input_block: T,
+
+    /// 1 if this row represents the final block of a sponge, in which case some or all of the bytes
+    /// in the block will be padding bytes; 0 otherwise.
+    pub is_final_block: T,
+
+    // The address at which we will read the input block.
+    pub context: T,
+    pub segment: T,
+    pub virt: T,
+
+    /// The timestamp at which inputs should be read from memory.
+    pub timestamp: T,
+
+    /// The length of the original input, in bytes.
+    pub len: T,
+
+    /// The number of input bytes that have already been absorbed prior to this block.
+    pub already_absorbed_bytes: T,
+
+    /// If this row represents a final block row, the `i`th entry should be 1 if the final chunk of
+    /// input has length `i` (in other words if `len - already_absorbed == i`), otherwise 0.
+    ///
+    /// If this row represents a full input block, this should contain all 0s.
+    pub is_final_input_len: [T; KECCAK_RATE_BYTES],
+
+    /// The initial rate bits of the sponge, at the start of this step.
+    pub original_rate_bits: [T; KECCAK_RATE_BITS],
+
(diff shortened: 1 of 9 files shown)


evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+use crate::memory::segments::Segment;
+use crate::stark::Stark;
+use crate::util::{trace_rows_to_poly_values, u32_from_le_bits, u32_to_le_bits, u8_to_le_bits};
+use crate::vars::StarkEvaluationTargets;
+use crate::vars::StarkEvaluationVars;
+
+#[allow(unused)] // TODO: Should be used soon.
+pub(crate) fn ctl_looked_data<F: Field>() -> Vec<Column<F>> {
+    let cols = KECCAK_SPONGE_COL_MAP;
+    let outputs = Column::singles(&cols.updated_state_u32s[..KECCAK_RATE_U32S]);
```
wborgeaud, line 29: Should be `Column::singles(&cols.updated_state_u32s[..8])` to get the 256-bit output.

evm/src/keccak_sponge/columns.rs
```diff
@@ -0,0 +1,118 @@
...
+    /// 1 if this row represents a dummy operation (for padding the table); 0 otherwise.
+    pub is_dummy: T,
+
+    /// 1 if this row represents a full input block, i.e. one in which each byte is an input byte,
+    /// not a padding byte; 0 otherwise.
+    pub is_full_input_block: T,
+
+    /// 1 if this row represents the final block of a sponge, in which case some or all of the bytes
+    /// in the block will be padding bytes; 0 otherwise.
+    pub is_final_block: T,
```
wborgeaud, line 26: 2 out of 3 of these columns are sufficient to deduce the third. So we can remove one of the 3.

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+        cols.timestamp,
+        cols.len,
+    ])
+    .chain(outputs)
+    .collect()
+}
+
+#[allow(unused)] // TODO: Should be used soon.
+pub(crate) fn ctl_looking_keccak<F: Field>() -> Vec<Column<F>> {
+    let input_rate_cols = (0..KECCAK_WIDTH_U32S)
```
wborgeaud, line 43: Should that be `(0..KECCAK_RATE_U32S)` since we want rate columns?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+        .map(|i| Column::le_bits(&KECCAK_SPONGE_COL_MAP.original_rate_bits[i * 32..(i + 1) * 32]));
+    let input_capacity_cols = Column::singles(
+        (0..KECCAK_WIDTH_U32S).map(|i| KECCAK_SPONGE_COL_MAP.original_capacity_u32s[i]),
```
wborgeaud, line 46: Should that be `(0..KECCAK_CAPACITY_U32S)`?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+        (0..KECCAK_WIDTH_U32S).map(|i| KECCAK_SPONGE_COL_MAP.updated_state_u32s[i]),
+    );
+    input_rate_cols
+        .chain(input_capacity_cols)
+        .chain(output_cols)
+        .collect()
+}
+
+#[allow(unused)] // TODO: Should be used soon.
+pub(crate) fn ctl_looking_memory<F: Field>(i: usize) -> Vec<Column<F>> {
```
wborgeaud, line 58: Hm I'm not sure I understand how this is supposed to work. It seems like this is the CTL columns for the `i`th byte in the block, but it always looks up the same address at `virt` regardless of `i`. 
Should the address be `virt + already_absorbed_bytes + i` instead?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+            .map(|x| F::from_canonical_u32(*x))
+            .collect_vec()
+            .try_into()
+            .unwrap();
+
+        let block_u32s = (0..KECCAK_RATE_U32S).map(|i| {
+            u32_from_le_bits(
+                row.block_bits[i * 32..(i + 1) * 32]
+                    .iter()
+                    .map(|f| *f == F::ONE)
```
wborgeaud, line 294: nit: `.map(Field::is_one)` also works.

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+                block.try_into().unwrap(),
+            );
+
+            // xor block into sponge_state's rate elements.
+            let block_u32s = block
+                .chunks(size_of::<u32>())
+                .map(|chunk| u32::from_le_bytes(chunk.try_into().unwrap()));
+            for (state_u32, block_u32) in sponge_state.iter_mut().zip(block_u32s) {
+                *state_u32 ^= block_u32;
+            }
```
wborgeaud, line 185: Hm I'm confused about this. Don't we already XOR the block into the state on line 303?

evm/src/keccak_sponge/columns.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub already_absorbed_bytes: T,
+
+    /// If this row represents a final block row, the `i`th entry should be 1 if the final chunk of
+    /// input has length `i` (in other words if `len - already_absorbed == i`), otherwise 0.
+    ///
+    /// If this row represents a full input block, this should contain all 0s.
+    pub is_final_input_len: [T; KECCAK_RATE_BYTES],
+
+    /// The initial rate bits of the sponge, at the start of this step.
+    pub original_rate_bits: [T; KECCAK_RATE_BITS],
```
wborgeaud, line 49: Could we use a lookup table to compute the XORs and have these be 4/8-bit chunks instead?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -0,0 +1,441 @@
...
+        let mut rows = vec![];
+        for op in operations {
+            rows.extend(self.generate_rows_for_op(op));
+        }
+
+        let num_rows = rows.len().max(min_rows).next_power_of_two();
+        let padding_row = self.generate_padding_row();
+        for _ in rows.len()..num_rows {
+            rows.push(padding_row);
+        }
```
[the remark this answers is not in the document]
dlubarov: Good idea 👍

evm/src/keccak_sponge/keccak_sponge_stark.rs
dlubarov: Ah yes, good catch.

evm/src/keccak_sponge/keccak_sponge_stark.rs
dlubarov: Oh right, forgot to remove this.

evm/src/keccak_sponge/columns.rs
dlubarov: Hm by inner table do you mean prover-provided (instead of preprocessed) table?

I think lookup tables could be better, but perhaps not by a significant margin, depending on our table size. Like with 2 cols for input chunks, 1 for the output chunk, 2 permuted cols, and 3 `Z` cols, we'd be at 8 cols, giving similar costs as having 8 input bit cols.

I like the idea of using the logic table, only slight complication is that it uses 256-bit inputs, which doesn't divide the Keccak rate (1088 bits). We could reduce its inputs to 64 bits, though that'd affect the CPU's CTL costs a bit. Maybe best to keep it at 256 bits and just do 5 lookups here (the last would only use 64 bits; we could optimize it later). Does that sound good?

wborgeaud: > Hm by inner table do you mean prover-provided (instead of preprocessed) table?

Yes, similar to how we do range checks.

> I think lookup tables could be better, but perhaps not by a significant margin, depending on our table size. Like with 2 cols for input chunks, 1 for the output chunk, 2 permuted cols, and 3 `Z` cols, we'd be at 8 cols, giving similar costs as having 8 input bit cols.

Yeah I agree for 4-bit chunks it might not be worth it. For 8-bit chunks it might already be better, but then we'd need to pad the table to have a length of 2^16, which might be wasteful.
 
> I like the idea of using the logic table, only slight complication is that it uses 256-bit inputs, which doesn't divide the Keccak rate (1088 bits). We could reduce its inputs to 64 bits, though that'd affect the CPU's CTL costs a bit. Maybe best to keep it at 256 bits and just do 5 lookups here (the last would only use 64 bits; we could optimize it later). Does that sound good?

Yes sounds good! We can store the rate as `u32`s and then use a `Column::le_u32s` in the 5 CTLs to the logic table.

dlubarov: Ok I made the change. I stored the previous rate part as `u32`s and the block as `u8`s (since they're `u8`s in memory). The logic CTL uses `u32`s, so there's no conversion needed for the previous rate part, but I used `le_bytes` to combine block bytes.

wborgeaud (approved):
Looks good to me!

PR #704, opened 2022-09-02
Added a few derives to `Trie` types

BGluth:
A downstream project needed `Hash` on `Nibbles`, but I also thought it made sense to derive a few other core traits as well.

### DIFF
--- a/evm/src/generation/partial_trie.rs
+++ b/evm/src/generation/partial_trie.rs
@@ -1,5 +1,6 @@
 use ethereum_types::U256;
 
+#[derive(Clone, Debug)]
 /// A partial trie, or a sub-trie thereof. This mimics the structure of an Ethereum trie, except
 /// with an additional `Hash` node type, representing a node whose data is not needed to process
 /// our transaction.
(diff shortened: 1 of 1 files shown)


evm/src/generation/partial_trie.rs
```diff
@@ -22,6 +23,7 @@ pub enum PartialTrie {
     Leaf { nibbles: Nibbles, value: Vec<u8> },
 }
 
+#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
dlubarov, line 26: Hm it's not obvious to me why an ordering would be useful or which one would be most natural, is it just for `BTreeMap`?

dlubarov (approved):
LGTM

evm/src/generation/partial_trie.rs
BGluth: Yeah `PartialOrd` and `Ord` aren't needed downstream. I was just thinking since this is part of the public interface, someone might want to order these for whatever reason. I'm fine taking the ord traits out if it's not really clear what the default impl should be.

PR #706, opened 2022-09-05
Remove `JUMPDEST`s

dlubarov:
We decided not to enforce it in kernel code.


wborgeaud (approved):
There are some left in `exp.asm`. I found them by adding this
```rust
0x5b => {
    assert!(!self.kernel_mode)
} // "JUMPDEST",
```
to `interpreter.rs`. Do you think we should add this check for testing? 

Looks good to me otherwise!

dlubarov:
Good idea, done

PR #707, opened 2022-09-05
Compute answers to FRI queries in parallel

dlubarov:
The speedup was well under a millisecond, so it's rather negligible, but I think the code is a little nicer this way also.


wborgeaud (approved):
Good idea 👍

PR #708, opened 2022-09-05
Per table recursion

wborgeaud:
Recursively verify every Stark proof individually and aggregate these proofs into a single recursive proof.


plonky2/src/plonk/circuit_data.rs
```diff
@@ -139,6 +139,18 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
     {
         compressed_proof_with_pis.verify(&self.verifier_only, &self.common)
     }
+
+    pub fn verifier_data(self) -> VerifierCircuitData<F, C, D> {
```
dlubarov, line 143: If we add this in the public API I think we should have a similar `prover_data` method.

Let's also update `build_prover` and `build_verifier` to use them.

wborgeaud: Good idea, done

plonky2/src/iop/challenger.rs
```diff
@@ -146,6 +146,10 @@ impl<F: RichField, H: Hasher<F>> Challenger<F, H> {
         self.output_buffer
             .extend_from_slice(&self.sponge_state[0..SPONGE_RATE]);
     }
+
+    pub fn state(&self) -> [F; SPONGE_WIDTH] {
```
dlubarov, line 150: Maybe call `duplexing` in here? Seems like we'd never want to get the state without first processing the input buffer.

plonky2/src/iop/challenger.rs
```diff
@@ -176,14 +180,23 @@ impl<F: RichField + Extendable<D>, H: AlgebraicHasher<F>, const D: usize>
         }
     }
 
+    pub fn from_state(sponge_state: [Target; SPONGE_WIDTH]) -> Self {
+        let output_buffer = sponge_state[0..SPONGE_RATE].to_vec();
```
dlubarov, line 184: It seems like this assumes these duplex elements haven't already been consumed from the output buffer (before we saved and restored Challenger's state). That should be true with our current usage, since we call `duplex` before saving the state.

But still, it's not obvious why this is safe, and it could be used in an unsafe way if we're not careful. I'd suggest starting with an empty output buffer so there's no risk of accidentally reusing outputs.

I.e. we could replace `state()` with a `compact()` function which
- Processes any inputs (if none, we can skip `duplexing()`)
- Clears the output buffer
- Returns state

plonky2/src/iop/challenger.rs
```diff
@@ -272,6 +285,32 @@ impl<F: RichField + Extendable<D>, H: AlgebraicHasher<F>, const D: usize>
 
         self.input_buffer.clear();
     }
+
+    pub fn duplexing(&mut self, builder: &mut CircuitBuilder<F, D>) {
```
dlubarov, line 289: If we go with the suggestion above we shouldn't need this. If we do keep it, we could change it to

```rust
if self.input_buffer.is_empty() {
    // Apply the permutation to get a fresh output buffer as from_state expects.
    self.sponge_state = builder.permute::<H>(self.sponge_state);
} else {
    self.absorb_buffered_inputs(builder);
} 
```

plonky2/src/iop/challenger.rs
```diff
@@ -129,7 +129,7 @@ impl<F: RichField, H: Hasher<F>> Challenger<F, H> {
 
     /// Absorb any buffered inputs. After calling this, the input buffer will be empty, and the
     /// output buffer will be full.
-    fn duplexing(&mut self) {
+    pub fn duplexing(&mut self) {
```
dlubarov, line 132: I'd suggest we keep this private and do what we need in `state()` (or with the suggestion below it'd be called `compact()`)

evm/src/cross_table_lookup.rs
```diff
@@ -191,6 +191,21 @@ impl<F: Field> CrossTableLookup<F> {
...
+            ans += ctl
+                .looking_tables
+                .iter()
+                .filter_map(|twc| (twc.table == table).then_some(num_challenges))
+                .sum::<usize>();
+            ans += (ctl.looked_table.table == table)
+                .then_some(num_challenges)
+                .unwrap_or_default();
+        }
+        ans
```
dlubarov, line 207: A little simpler I think
```suggestion
        let mut num_ctls = 0;
        for ctl in ctls {
            let all_tables = iter::once(&ctl.looked_table).chain(&ctl.looking_tables);
            num_ctls += all_tables.filter(|twc| twc.table == table).count();
        }
        num_ctls * num_challenges
```

plonky2/src/iop/challenger.rs
wborgeaud: Hm so if I change the `state()` function to the `compact()` function you suggested, I shouldn't need the modify this function `from_state()`, right?

dlubarov: I was thinking just `output_buffer` would be set to `vec![]` in that case, but it could be the same otherwise

evm/src/cross_table_lookup.rs
```diff
@@ -450,53 +459,57 @@ pub struct CtlCheckVarsTarget<'a, F: Field, const D: usize> {
...
-                    local_z: *looked_z,
-                    next_z: *looked_z_next,
-                    challenges,
-                    columns: &looked_table.columns,
-                    filter_column: &looked_table.filter_column,
-                });
             }
         }
-        ctl_vars_per_table
+        ctl_vars
```
dlubarov, line 512: As a sanity check can we assert `ctl_zs` is now empty?

dlubarov: Similar to line 625

evm/src/proof.rs
```diff
@@ -29,7 +30,7 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize> A
     }
 
     pub fn nums_ctl_zs(&self) -> [usize; NUM_TABLES] {
```
dlubarov, line 32: Seems unused now?

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
 
-pub fn verify_proof_circuit<
+/// Table-wise recursive proofs of an `AllProof`.
+pub struct RecursiveAllProof<
+    F: RichField + Extendable<D>,
+    C: GenericConfig<D, F = F>,
+    const D: usize,
+> {
+    pub recursive_proofs: [ProofWithPublicInputs<F, C, D>; NUM_TABLES],
+    pub cross_table_lookups: Vec<CrossTableLookup<F>>,
```
dlubarov, line 55: Feels a bit odd to me to have this in a "proof" object, since it's part of the instance/config. Maybe it can be an argument to `verify` instead?

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
+    ctl_zs_last: Vec<T>,
+    ctl_challenges: GrandProductChallengeSet<T>,
+    challenger_state_before: [T; SPONGE_WIDTH],
+    challenger_state_after: [T; SPONGE_WIDTH],
+}
+
+impl<T: Copy + Eq + PartialEq + Debug> PublicInputs<T> {
+    fn from_vec(v: &[T], config: &StarkConfig) -> Self {
+        let mut start = 0;
+        let trace_cap = v[start..4 * (1 << config.fri_config.cap_height)]
```
dlubarov, line 69: Looks like we can remove `start` since it's 0, or if we keep it for flexibility, we should also shift the end of the range by `start` right?

dlubarov: Also nit, I think it would be nice to have a `num_cap_elements()` helper, maybe in `FriConfig`

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
+                .collect(),
+        };
+        start += 2 * config.num_challenges;
+        let challenger_state_before = v[start..start + SPONGE_WIDTH].try_into().unwrap();
+        let challenger_state_after = v[start + SPONGE_WIDTH..start + 2 * SPONGE_WIDTH]
+            .try_into()
+            .unwrap();
+
+        start += 2 * SPONGE_WIDTH;
+        let ctl_zs_last = v[start..].to_vec();
```
dlubarov, line 89: I think this would be nicer with an iterator, although it'd be easier if `next_chunk` was stable, but we could use a helper for now -
```suggestion
/// Similar to the unstable `Iterator::next_chunk`. Could be replaced with that when it's stable.
fn next_chunk<T: Debug, const N: usize>(iter: &mut impl Iterator<Item = T>) -> [T; N] {
    (0..N)
        .flat_map(|_| iter.next())
        .collect_vec()
        .try_into()
        .expect("Not enough elements")
}

impl<T: Copy + Eq + PartialEq + Debug> PublicInputs<T> {
    fn from_vec(v: &[T], config: &StarkConfig) -> Self {
        let mut iter = v.iter().copied();
        let trace_cap = (0..1 << config.fri_config.cap_height)
            .map(|_| next_chunk::<_, 4>(&mut iter).to_vec())
            .collect();
        let ctl_challenges = GrandProductChallengeSet {
            challenges: (0..config.num_challenges)
                .map(|_| GrandProductChallenge {
                    beta: iter.next().unwrap(),
                    gamma: iter.next().unwrap(),
                })
                .collect(),
        };
        let challenger_state_before = next_chunk(&mut iter);
        let challenger_state_after = next_chunk(&mut iter);
        let ctl_zs_last = iter.collect();
```

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
+
+        // Verify the proofs.
+        for (proof, verifier_data) in self.recursive_proofs.into_iter().zip(verifier_data) {
+            verifier_data.verify(proof)?;
+        }
+        Ok(())
+    }
+
+    /// Recursively verify every recursive proof.
+    pub fn verify_circuit<W>(
```
dlubarov, line 156: I think we should separate the circuit building and witness generation this is doing, so that e.g. a verify-only user can build the circuit without having to do any witness generation or proving.

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
+        );
+        for (i, (proof_target, inner_data)) in proofs_target.into_iter().enumerate() {
+            builder.verify_proof(proof_target, &inner_data, &verifier_data[i].common);
+        }
+    }
+}
+
+/// Recursively verify a Stark proof.
+/// Outputs the recursive proof and the associated verifier data.
+fn recursively_verify_stark_proof<
```
dlubarov, line 245: Here too, it seems like this method couples preprocessing with witness generation, but I think we should separate them.

evm/src/recursive_verifier.rs
```diff
@@ -1,148 +1,574 @@
...
+    let mut challenger = RecursiveChallenger::<F, C::Hasher, D>::from_state(challenger_state);
+    let challenges = proof_target.get_challenges::<F, C>(
+        &mut builder,
+        &mut challenger,
+        num_permutation_zs > 0,
+        num_permutation_batch_size,
+        inner_config,
+    );
+    challenger.duplexing(&mut builder);
+    let challenger_state = challenger.state();
```
dlubarov, line 406: nit: maybe call them challenger_{before,after}, took me a minute to realize they were distinct vars.

evm/src/recursive_verifier.rs
wborgeaud: Very nice 👍

evm/src/recursive_verifier.rs
wborgeaud: There's `verifier_data_recursive_stark_proof` that generates the verifier data without witness generation, is that what you mean?

evm/src/recursive_verifier.rs
wborgeaud: So right now I've designed this function to be a gadget, since it might be used as part of a larger circuit, so there's no proving involved. Do you think it makes sense to have a copy of this function without the witness generation?

evm/src/recursive_verifier.rs
```diff
@@ -34,118 +49,615 @@ use crate::{
...
+        // Verify the proofs.
+        for (proof, verifier_data) in self.recursive_proofs.into_iter().zip(verifier_data) {
+            verifier_data.verify(proof)?;
+        }
+        Ok(())
+    }
+
+    /// Recursively verify every recursive proof.
+    pub fn verify_circuit(
+        self,
```
dlubarov, line 167: I think this should not take `self` (the list of recursive proofs), it seems unused and requiring it would mean we can't build this circuit without doing some proving work.

evm/src/recursive_verifier.rs
```diff
@@ -34,118 +49,615 @@ use crate::{
...
+        let proofs_target: [_; NUM_TABLES] = std::array::from_fn(|i| {
+            let verifier_data = &verifier_data[i];
+            let proof = &self.recursive_proofs[i];
+            let pt = builder.add_virtual_proof_with_pis(&verifier_data.common);
+            pw.set_proof_with_pis_target(&pt, proof);
+            let inner_data = VerifierCircuitTarget {
+                constants_sigmas_cap: builder
+                    .add_virtual_cap(verifier_data.common.config.fri_config.cap_height),
+            };
+            pw.set_cap_target(
```
dlubarov, line 269: I think the verification key should already be baked into the circuit for soundness reasons; is it not currently?

evm/src/recursive_verifier.rs
```diff
@@ -34,118 +49,615 @@ use crate::{
...
+            builder.verify_proof(
+                recursive_proof,
+                &verifier_data_target,
+                &verifier_data[i].common,
+            );
+        }
+    }
+
+    /// Verifier data to recursively verify every recursive proof.
+    pub fn verifier_data_verify_circuit<W>(
```
dlubarov, line 248: It seems like this function involves both circuit building and witness generation logic, whereas `verify_circuit` contains the same circuit building logic without witness generation? If so, I think we shouldn't need this one; I think it'd be best to call `verify_circuit` once with a list of virtual recursive proof inputs, and then separately call `set_proof_with_pis_target` whenever we're proving.

evm/src/recursive_verifier.rs
```diff
@@ -34,118 +49,615 @@ use crate::{
...
+            challenger_state_after,
+        }
+    }
+}
+
+impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
+    RecursiveAllProof<F, C, D>
+{
+    /// Verify every recursive proof.
+    pub fn verify(
```
dlubarov, line 115: I'm not sure we'll need to natively verify these proofs? Since they get wrapped in the higher-level txn proof (which doesn't have delayed checks so it's more trivial to verify). It's fine to keep for now at least, just something to think about

evm/src/recursive_verifier.rs
dlubarov: So if I understand right, this function does
- create virtual targets, mainly `add_virtual_stark_proof`
- circuit building, same as `verifier_data_recursive_stark_proof`
- witness generation, mainly `set_stark_proof_target`

For our production use case I think we'll want to separate those things, so I guess we'd be calling `verifier_data_recursive_stark_proof`, `set_stark_proof_target`, etc. separately, and wouldn't use this function.

If it's convenient for testing we could keep it, but if so I think we should have it call `verifier_data_recursive_stark_proof` to remove the duplication, and add `cfg(test)` so it's clear it doesn't need to be audited.

evm/src/recursive_verifier.rs
```diff
@@ -404,7 +941,7 @@ pub fn add_virtual_stark_proof<F: RichField + Extendable<D>, S: Stark<F, D>, con
 
 fn add_stark_opening_set<F: RichField + Extendable<D>, S: Stark<F, D>, const D: usize>(
```
dlubarov, line 942: nit: let's add "virtual" to match the other methods

evm/src/recursive_verifier.rs
wborgeaud: Hm not sure I understand, what do you mean by "baked into the circuit"? Btw I removed this function, so maybe it's no longer relevant.

evm/src/recursive_verifier.rs
wborgeaud: Oh of course I don't know why I kept this function...

evm/src/recursive_verifier.rs
wborgeaud: Hm I think it might be a good idea to keep this for now. It might make debugging easier if we run into some bugs and it could help to be able to verify this without having to build a circuit (like if we have decentralized proving and this proof is passed around, nodes might like to verify natively first to make sure it's not spam).

evm/src/recursive_verifier.rs
wborgeaud: Oh yeah good point, I moved them to a `mod tests` and we can remove them when we remove the `all_stark.rs` tests.

evm/src/recursive_verifier.rs
```diff
@@ -328,6 +567,31 @@ pub fn add_virtual_all_proof<F: RichField + Extendable<D>, const D: usize>(
...
+    verifier_data: &[VerifierCircuitData<F, C, D>; NUM_TABLES],
+) -> RecursiveAllProofTargetWithData<D> {
+    let recursive_proofs = std::array::from_fn(|i| {
+        let verifier_data = &verifier_data[i];
+        builder.add_virtual_proof_with_pis(&verifier_data.common)
+    });
+    let verifier_data = std::array::from_fn(|i| {
+        let verifier_data = &verifier_data[i];
+        VerifierCircuitTarget {
+            constants_sigmas_cap: builder
```
dlubarov, line 585: Is `constants_sigmas_cap` currently unconstrained? If so it seems like a security issue; I think the VK should be baked into the circuit, i.e. set to some constants. Maybe like this?

```rust
constants_sigmas_cap: builder.constant_merkle_cap(verifier_data.verifier_only.constants_sigmas_cap)
```

(`constant_merkle_cap` doesn't exist yet)

Might be a big odd to have in this function though, since this part wouldn't be virtual.

evm/src/recursive_verifier.rs
dlubarov: I think there may be a similar concern in other functions, but I made a separate comment so will resolve this one

evm/src/recursive_verifier.rs
```diff
@@ -422,6 +686,27 @@ fn add_stark_opening_set<F: RichField + Extendable<D>, S: Stark<F, D>, const D:
...
+{
+    for i in 0..NUM_TABLES {
+        witness.set_proof_with_pis_target(
+            &recursive_all_proof_target.recursive_proofs[i],
+            &all_proof.recursive_proofs[i],
+        );
+        witness.set_cap_target(
+            &recursive_all_proof_target.verifier_data[i].constants_sigmas_cap,
+            &verifier_data[i].verifier_only.constants_sigmas_cap,
+        );
```
dlubarov, line 707: I think this shouldn't be needed, if we "hardcode" the VK as suggested above

dlubarov (approved):
LGTM! (besides that question about the VK being constrained)

evm/src/recursive_verifier.rs
wborgeaud: Oh yeah of course. I implemented your suggestion and added a comment to the function saying the verifier data is not virtual.

PR #710, opened 2022-09-06
Stack pointer + underflow/overflow checks

nbgl:
Adds a stack pointer and constraints to show that the current instruction will not cause the stack to underflow or overflow.

Note that this does not actually add the kernel routines to handle underflow or overflow. Until this happens, any such error will cause the proof to be unverifiable.

### DIFF
--- a/evm/src/cpu/control_flow.rs
+++ b/evm/src/cpu/control_flow.rs
@@ -68,14 +68,18 @@ pub fn eval_packed_generic<P: PackedField>(
         lv.is_cpu_cycle * is_native_instruction * (lv.is_kernel_mode - nv.is_kernel_mode),
     );
 
-    // If a non-CPU cycle row is followed by a CPU cycle row, then the `program_counter` of the CPU
-    // cycle row is route_txn (the entry point of our kernel) and it is in kernel mode.
+    // If a non-CPU cycle row is followed by a CPU cycle row, then:
+    //  - the `program_counter` of the CPU cycle row is `route_txn` (the entry point of our kernel),
+    //  - execution is in kernel mode, and
+    //  - the stack is empty.
     let pc_diff =
         nv.program_counter - P::Scalar::from_canonical_usize(KERNEL.global_labels["route_txn"]);
     yield_constr.constraint_transition((lv.is_cpu_cycle - P::ONES) * nv.is_cpu_cycle * pc_diff);
     yield_constr.constraint_transition(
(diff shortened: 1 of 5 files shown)


evm/src/cpu/stack_bounds.rs
```diff
@@ -0,0 +1,132 @@
...
+use plonky2::field::types::Field;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::{CpuColumnsView, COL_MAP};
+
+const MAX_STACK_SIZE: u64 = 1024;
+
+const DECREMENTING_FLAGS: [usize; 1] = [COL_MAP.is_pop];
```
dlubarov, line 14: Let's add a comment noting why we don't need other decrementing instructions to check underflow.

evm/src/cpu/stack_bounds.rs
```diff
@@ -0,0 +1,132 @@
...
+
+const INCREMENTING_FLAGS: [usize; 2] = [COL_MAP.is_pc, COL_MAP.is_dup];
```
dlubarov, line 16: There will be more later right? Let's add a TODO

evm/src/cpu/control_flow.rs
```diff
@@ -68,14 +68,18 @@ pub fn eval_packed_generic<P: PackedField>(
...
+    //  - the `program_counter` of the CPU cycle row is `route_txn` (the entry point of our kernel),
+    //  - execution is in kernel mode, and
+    //  - the stack is empty.
     let pc_diff =
         nv.program_counter - P::Scalar::from_canonical_usize(KERNEL.global_labels["route_txn"]);
     yield_constr.constraint_transition((lv.is_cpu_cycle - P::ONES) * nv.is_cpu_cycle * pc_diff);
     yield_constr.constraint_transition(
         (lv.is_cpu_cycle - P::ONES) * nv.is_cpu_cycle * (nv.is_kernel_mode - P::ONES),
     );
+    yield_constr
```
dlubarov, line 81: nit: let's factor out `(lv.is_cpu_cycle - P::ONES) * nv.is_cpu_cycle` now that it's used 3x

dlubarov (approved):
LGTM

PR #678, opened 2022-08-19
Fix starky avx

Sladuca:
Fixes #675

PR #711, opened 2022-09-06
Add `flat_map_iter` to `maybe_rayon`

Sladuca:
In #678 I used `rayon`'s `flat_map_iter` but I forgot to add it to `maybe_rayon`'s `ParallelIteratorMock` trait, which breaks `starky` and `evm` when the `parallel` feature is disabled. This PR fixes that.

This PR also fixes `starky` and `evm` not disabling default features for `plonky2`, which causes `rayon` to be enabled even if it's disabled by `starky` and `evm`, and it also fixes a feature gate for `BytesHash`.

### DIFF
--- a/evm/Cargo.toml
+++ b/evm/Cargo.toml
@@ -5,7 +5,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-plonky2 = { path = "../plonky2" }
+plonky2 = { path = "../plonky2", default-features = false, features = ["rand", "rand_chacha", "timing", "gate_testing"] }
 plonky2_util = { path = "../util" }
 anyhow = "1.0.40"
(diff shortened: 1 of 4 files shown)


evm/Cargo.toml
```diff
@@ -5,7 +5,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-plonky2 = { path = "../plonky2" }
+plonky2 = { path = "../plonky2", default-features = false, features = ["rand", "rand_chacha", "timing", "gate_testing"] }
```
dlubarov, line 8: I don't think we need `rand_chacha` or `gate_testing`, but I can remove them after.

dlubarov (approved):
LGTM

PR #712, opened 2022-09-07
Document batch opening function

wborgeaud:
Give better documentation for the `PolynomialBatch::prove_openings` function as per auditor comments.


dlubarov (approved):
LGTM

PR #714, opened 2022-09-09
blocks in stack manipulation

npwardberkeley:
Cc @wborgeaud @typ3c4t


evm/src/cpu/kernel/stack/stack_manipulation.rs
```diff
@@ -25,25 +25,50 @@ pub(crate) fn expand_stack_manipulation(body: Vec<Item>) -> Vec<Item> {
...
+        .flat_map(|item| match item {
+            StackPlaceholder::Identifier(name) => {
+                stack_names.insert(name.clone());
+                vec![StackItem::NamedItem(name)]
+            }
+            StackPlaceholder::Block(name, n) => {
+                stack_blocks.insert(name.clone(), n);
+                (0..n)
+                    .map(|i| {
+                        let literal_name = format!("block_{}_{}", name, i);
```
dlubarov, line 44: Even though it's unlikely anyone would use a name like `block_{}_{}`, I'd suggest a format that includes a special character like `@{name}.{i}`, so a collision with an `Identifier` name would be impossible. That's what we did for macro-generated labels (following NASM's convention).

evm/src/cpu/kernel/stack/stack_manipulation.rs
```diff
@@ -25,25 +25,50 @@ pub(crate) fn expand_stack_manipulation(body: Vec<Item>) -> Vec<Item> {
...
+    let mut stack_blocks = HashMap::new();
+    let mut stack_names = HashSet::new();
+
     let mut src = names
         .iter()
         .cloned()
-        .map(StackItem::NamedItem)
+        .flat_map(|item| match item {
+            StackPlaceholder::Identifier(name) => {
+                stack_names.insert(name.clone());
```
dlubarov, line 37: Might be slightly simpler if at this point we treat `Identifier` as a block of size 1? I.e. doing `stack_blocks.insert(name.clone(), 1);` here, and removing `stack_names`. It'd also be a bit easier to check for duplicate names if we only have `stack_blocks`.

dlubarov (approved):
LGTM!

evm/src/cpu/kernel/stack/stack_manipulation.rs
npwardberkeley: Ahh good idea

PR #715, opened 2022-09-10
Verify that comparison output is zero or one

nbgl:
Gets rid of a few "not implemented" warnings


unzvfu (approved):
Looks good! Thanks! :)

PR #716, opened 2022-09-12
Change Lagrange polynomial notation 

wborgeaud:
Change `L_1` to `L_0` everywhere to match the conventional notation `L_i(g^i) = 1`.


dlubarov (approved):
LGTM

PR #717, opened 2022-09-12
Memory channel for decode

nbgl:
Add `context` column and connect the decoder to memory

### DIFF
--- a/evm/src/cpu/cpu_stark.rs
+++ b/evm/src/cpu/cpu_stark.rs
@@ -1,4 +1,5 @@
 use std::borrow::{Borrow, BorrowMut};
+use std::iter::repeat;
 use std::marker::PhantomData;
 
 use itertools::Itertools;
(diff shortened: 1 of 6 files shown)


evm/src/cpu/cpu_stark.rs
```diff
@@ -61,16 +62,42 @@ pub fn ctl_filter_logic<F: Field>() -> Column<F> {
...
-    let mut cols: Vec<Column<F>> = Column::singles([
-        channel_map.is_read,
-        channel_map.addr_context,
-        channel_map.addr_segment,
-        channel_map.addr_virtual,
-    ])
-    .collect_vec();
-    cols.extend(Column::singles(channel_map.value));
+    let mut cols = Vec::<Column<F>>::new();
+    if channel == membus::channel_indices::CODE {
```
dlubarov, line 66: I think it may be a little nicer to split this into two functions, for the code channel and `i`th GP channel. Then we wouldn't need to adjust the index, the caller would just enumerate `0..NUM_GP_CHANNELS`.

dlubarov (approved):
LGTM

PR #721, opened 2022-09-14
Remove `num_virtual_targets` from `CommonCircuitData`

wborgeaud:
Virtual targets are only a prover specific thing and the prover already has access to this info via `ProverOnlyCircuitData::representative_map`.


dlubarov (approved):
LGTM

PR #722, opened 2022-09-15
Python prototype of cache-oblivious FFT

nbgl:
Python implementation of the cache-oblivious FFT algorithm. Can be ported to Rust by whoever needs a project.


nbgl:
Not asking for review before merging, because this is throwaway code and totally separate from the rest of the codebase.

PR #724, opened 2022-09-19
Fix `DTH_ROOT` for degree 1 extension

wborgeaud:
As observed by an auditor, `DTH_ROOT=0` is not a  `D`th root of unity. Fixed by setting it to `1`.


unzvfu (approved):
Haha, good point.

PR #725, opened 2022-09-19
Replaced `PartialTrie` definitions with `eth-trie-utils` crate

BGluth:
Since we decided to move `PartialTrie` logic to it's own crate, this PR removes the definitions from `Plonky2` and adds a dependency for `eth-trie-utils`. Also added `Serialize`/`Deserialize` since this is needed by the test parser.


dlubarov (approved):
LGTM!

PR #729, opened 2022-09-21
Added a mapping between code hashes and contract byte code

BGluth:
Added a mapping between an account's `code_hash` field and the actual contract byte code in `GenerationInputs`.

### DIFF
--- a/evm/src/generation/mod.rs
+++ b/evm/src/generation/mod.rs
@@ -1,5 +1,7 @@
+use std::collections::HashMap;
+
 use eth_trie_utils::partial_trie::PartialTrie;
-use ethereum_types::Address;
+use ethereum_types::{Address, H256};
 use plonky2::field::extension::Extendable;
 use plonky2::field::polynomial::PolynomialValues;
 use plonky2::field::types::Field;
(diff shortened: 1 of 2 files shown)


evm/src/generation/mod.rs
```diff
@@ -41,6 +43,10 @@ pub struct GenerationInputs {
     /// storage tries, and nodes therein, that will be accessed by these transactions.
     pub storage_tries: Vec<(Address, PartialTrie)>,
 
+    /// Mapping between smart contract code hashes and the contract byte code.
+    /// All account entries that have `code_hash` filled in will have their corresponding byte code in this map.
```
dlubarov, line 47: I don't think this is strictly necessary - say a certain account receives tokens, then our zkEVM will need the its data `(nonce, balance, codeHash, storageRoot)` to prove that `balance` was updated properly, but we won't need its code

dlubarov (approved):
LGTM, thanks

evm/src/generation/mod.rs
BGluth: Oh true. I'll update the comment.

PR #730, opened 2022-09-22
Plonky2 examples

npwardberkeley:


plonky2/src/plonk/verifier.rs
```diff
@@ -12,7 +12,7 @@ use crate::plonk::validate_shape::validate_proof_with_pis_shape;
 use crate::plonk::vanishing_poly::eval_vanishing_poly;
 use crate::plonk::vars::EvaluationVars;
 
-pub(crate) fn verify<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>(
+pub fn verify<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>(
```
dlubarov, line 15: Usually we go through `CircuitData::verify`, was there an issue with that?

plonky2/examples/fibonacci.rs
```diff
@@ -0,0 +1,43 @@
...
+        let temp = builder.add(prev_target, cur_target);
+        prev_target = cur_target;
+        cur_target = temp;
+    }
+
+    let fib_100 = F::from_canonical_u64(3736710860384812976);
+    let fib_100_target = builder.constant(fib_100);
+    builder.register_public_input(fib_100_target);
+
+    builder.connect(fib_100_target, cur_target);
```
wborgeaud, line 31: I think a more standard design would be to just register `cur_target` as a public input.
That way we keep the same statement "I know the 100th Fibonacci number" without having to hardcode the value in the circuit.

plonky2/src/plonk/verifier.rs
npwardberkeley: Oh oops! Fixed

plonky2/examples/fibonacci.rs
npwardberkeley: Ah yes that makes sense

npwardberkeley: Does it look right now?

npwardberkeley: Also not quite sure how best to do that in `square_root`...

wborgeaud: > Does it look right now?

Yes looks good! You can also remove that part now:
```rust
let fib_100 = F::from_canonical_u64(3736710860384812976);
let fib_100_target = builder.constant(fib_100);
builder.connect(fib_100_target, cur_target);
```

plonky2/examples/factorial.rs
```diff
@@ -0,0 +1,35 @@
...
+    let config = CircuitConfig::standard_recursion_config();
+
+    let pw = PartialWitness::new();
+    let mut builder = CircuitBuilder::<F, D>::new(config);
+
+    let mut cur_target = builder.one();
+    for i in 2..101 {
+        let i_target = builder.constant(F::from_canonical_u32(i));
+        cur_target = builder.mul(cur_target, i_target);
+    }
```
wborgeaud, line 22: Same as for the Fibonacci example, this will only produce `ConstantGate`s.

I'm not sure how to fix this here, one way would be to instead prove "I know the product `n*(n+1)*...(n+k)`" and have `n` be a public input and `k` be a constant (like 100 here).

plonky2/examples/fibonacci.rs
```diff
@@ -0,0 +1,40 @@
...
+    let pw = PartialWitness::new();
+    let mut builder = CircuitBuilder::<F, D>::new(config);
+
+    let mut prev_target = builder.zero();
+    let mut cur_target = builder.one();
+    for _ in 0..99 {
+        let temp = builder.add(prev_target, cur_target);
+        prev_target = cur_target;
+        cur_target = temp;
+    }
```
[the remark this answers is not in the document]
npwardberkeley: Ohh interesting, I see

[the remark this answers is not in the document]
npwardberkeley: So should I create `a` and `b` as virtual targets? And then what would I do to supply their values?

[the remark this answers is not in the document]
wborgeaud: Yeah exactly, then you can use the partial witness to set the values.

plonky2/examples/square_root.rs
```diff
@@ -0,0 +1,35 @@
...
+    let pw = PartialWitness::new();
+    let mut builder = CircuitBuilder::<F, D>::new(config);
+
+    let x = F::rand();
+    let x_squared = x * x;
+    let x_target = builder.constant(x);
+    let x_squared_target = builder.constant(x_squared);
+
+    let x_squared_computed = builder.mul(x_target, x_target);
+    builder.connect(x_squared_target, x_squared_computed);
```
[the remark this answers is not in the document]
npwardberkeley: Ah okay! Does it look right now? 🤔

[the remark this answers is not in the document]
npwardberkeley: Oh oops — I realized that in my code, the generator isn't actually necessary, because I'm just generating multiplication...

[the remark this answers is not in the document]
npwardberkeley: Were you thinking of having a generator to find the square root _from_ the square? Using Tonelli-Shanks or something?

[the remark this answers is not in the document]
wborgeaud: > Were you thinking of having a generator to find the square root _from_ the square? Using Tonelli-Shanks or something?

Yeah that's what I was thinking. But I forgot that computing the square root in the Goldilocks field is a bit involved (since `p=4k+1`). So two options would be
- Implement a square root algorithm. We had one in plonky1 that should still work https://github.com/mir-protocol/plonky/blob/3403436d3df4bfb2b530e2ca2a78c9c2a9b9a8f0/src/field/field.rs#L440.
- Change this example to something like computing the inverse.  This would be somewhat equivalent to the square-root example ("I know a `y` st `y*x=1`" instead of "I know a `y` st `y*y=x`"), and writing the generator would be easier since inversion is already implemented.

plonky2/examples/square_root.rs
```diff
@@ -0,0 +1,84 @@
...
+    const D: usize = 2;
+    type C = PoseidonGoldilocksConfig;
+    type F = <C as GenericConfig<D>>::F;
+
+    let config = CircuitConfig::standard_recursion_config();
+
+    let mut builder = CircuitBuilder::<F, D>::new(config);
+
+    let x = builder.add_virtual_target();
+    let x_squared = builder.mul(x, x);
```
dlubarov, line 50: nit: `builder.square(x)`

plonky2/examples/factorial.rs
```diff
@@ -0,0 +1,40 @@
...
+    builder.register_public_input(initial);
+    builder.register_public_input(cur_target);
+
+    let mut pw = PartialWitness::new();
+    pw.set_target(initial, F::ONE);
+
+    let data = builder.build::<C>();
+    let proof = data.prove(pw)?;
+
+    println!("100 factorial (mod |F|) is: {}", proof.public_inputs[1]);
```
dlubarov, line 37: I think this is out of date? Could call it `100! / (n - 1)!` or just `factorial range`

dlubarov: Oh I guess it's still accurate in a way, since we know n=1 in this code. But I think it would be natural to say something like `println!("Factorial starting at {} is {}!", proof.public_inputs[0], proof.public_inputs[1])`, as if we ignore the private witness data and only consider the proof, that's the statement the proof represents.

plonky2/examples/fibonacci.rs
```diff
@@ -0,0 +1,49 @@
...
+    // Provide initial values.
+    let mut pw = PartialWitness::new();
+    pw.set_target(initial_a, F::ZERO);
+    pw.set_target(initial_b, F::ONE);
+
+    let data = builder.build::<C>();
+    let proof = data.prove(pw)?;
+
+    println!(
+        "100th Fibonacci number (mod |F|) is: {}",
```
dlubarov, line 44: Likewise here, let's include a, b in the print since they're part of the statement we proved

plonky2/examples/square_root.rs
```diff
@@ -0,0 +1,84 @@
...
+    type F = <C as GenericConfig<D>>::F;
+
+    let config = CircuitConfig::standard_recursion_config();
+
+    let mut builder = CircuitBuilder::<F, D>::new(config);
+
+    let x = builder.add_virtual_target();
+    let x_squared = builder.square(x);
+
+    builder.register_public_input(x);
```
dlubarov, line 52: I think this example may be more meaningful if we keep `x` private, so we're proving `I know a square root of n` rather than `m is a square root of n` (which seems less interesting since sending `m, n` would prove the same statement)

dlubarov (approved):
LGTM otherwise

plonky2/examples/factorial.rs
npwardberkeley: Ah good idea!

npwardberkeley: Ah good idea!

plonky2/examples/square_root.rs
npwardberkeley: Ah okay!

PR #734, opened 2022-09-23
stack manipulation: allow empty LHS

npwardberkeley:


typ3c4t:
@dlubarov this allows for stack macros with empty input prefix

typ3c4t (approved):
awesome!

PR #735, opened 2022-09-23
Connect stack to memory

nbgl:
Constraints for memory channels where they correspond to stack operands/results.

Refactors the CPU columns so that all operation flags are in a separate struct. This makes it easier to specify per-operation behavior.


evm/src/cpu/stack.rs
```diff
@@ -0,0 +1,387 @@
...
+use crate::cpu::columns::ops::OpsColumnsView;
+use crate::cpu::columns::CpuColumnsView;
+use crate::cpu::membus::NUM_GP_CHANNELS;
+use crate::memory::segments::Segment;
+
+#[derive(Clone, Copy)]
+struct StackBehavior {
+    num_pops: usize,
+    pushes: bool,
+    disable_other_channels: bool,
```
dlubarov, line 18: This would be false for things like `MLOAD` right?

To be extra safe we could do something like `extra_channels_used: usize`. Otherwise it seems some instructions like MLOAD would need to manually disable the last couple channels?

evm/src/cpu/stack.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+const STACK_BEHAVIORS: OpsColumnsView<Option<StackBehavior>> = OpsColumnsView {
+    stop: None, // TODO
+    add: Some(StackBehavior {
```
dlubarov, line 23: I think it would be nice to have a few shared consts like

```rust
const BINARY_OP: Option<StackBehavior> = Some(StackBehavior {
    num_pops: 2,
    pushes: true,
    disable_other_channels: true,
});
```

to make the `STACK_BEHAVIORS` definition more succinct.

evm/src/cpu/stack.rs
```diff
@@ -0,0 +1,387 @@
...
+            filter * (channel.addr_segment - P::Scalar::from_canonical_u64(Segment::Stack as u64)),
+        );
+        // E.g. if `stack_len == 1` and `i == 0`, we want `add_virtual == 0`.
+        let addr_virtual = lv.stack_len - P::Scalar::from_canonical_usize(i + 1);
+        yield_constr.constraint(filter * (channel.addr_virtual - addr_virtual));
+    }
+
+    // Pushes
+    if stack_behavior.pushes {
+        let channel = lv.mem_channels[NUM_GP_CHANNELS - 1];
```
dlubarov, line 240: So the order is (pops, non-stack ops, pushes)? Feels a tad unusual to me but it's fine if you prefer it.

dlubarov (approved):
LGTM otherwise

evm/src/cpu/stack.rs
nbgl: > This would be false for things like `MLOAD` right?

Exactly!

> To be extra safe we could do something like `extra_channels_used: usize`.

I agree, but I’d like to call it out of the scope of this PR. This is the kind of implementation detail that I wouldn’t want to see in `stack.rs`, but yes we should do this somewhere

PR #740, opened 2022-09-29
Trie roots now use `H256` instead of `U256`

BGluth:
`H256` preserves any leading `0`s, which could be critical in some situations. Also just a slightly more appropriate type for hashes.


dlubarov (approved):
LGTM!

dlubarov (approved):

PR #742, opened 2022-09-30
MSIZE

npwardberkeley:


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -511,6 +511,14 @@ impl<'a> Interpreter<'a> {
         }
     }
 
+    fn run_msize(&mut self) {
+        let num_u256s = self.memory.context_memory[0].segments[Segment::MainMemory as usize]
+            .content
+            .len();
+        let num_bytes = num_u256s * 32;
```
dlubarov, line 518: In our model we aren't packing several memory bytes into one U256, but rather just using one byte of the U256, and leaving the rest as zero. So we don't need `* 32`

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -511,6 +511,14 @@ impl<'a> Interpreter<'a> {
         }
     }
 
+    fn run_msize(&mut self) {
+        let num_u256s = self.memory.context_memory[0].segments[Segment::MainMemory as usize]
```
dlubarov, line 515: Let's use `self.context` instead of `0` to make it more general

dlubarov (approved):
LGTM otherwise

evm/src/cpu/kernel/interpreter.rs
npwardberkeley: Ohhhh right of coruse

PR #743, opened 2022-09-30
Constraints for dup/swap

nbgl:


evm/src/cpu/stack.rs
```diff
@@ -105,29 +105,37 @@ const STACK_BEHAVIORS: OpsColumnsView<Option<StackBehavior>> = OpsColumnsView {
...
+    get_context: None,    // TODO
+    set_context: None,    // TODO
+    consume_gas: None,    // TODO
+    exit_kernel: None,    // TODO
+    staticcall: None,     // TODO
+    mload_general: None,  // TODO
+    mstore_general: None, // TODO
+    revert: None,         // TODO
+    selfdestruct: None,   // TODO
+    invalid: None,        // TODO
```
nbgl, line 138: These are unchanged but `rustfmt` insists on aligning the comments

evm/src/cpu/dup_swap.rs
```diff
@@ -0,0 +1,249 @@
...
+    {
+        let constr = builder.add_extension(channel.addr_virtual, offset);
+        let constr = builder.sub_extension(constr, lv.stack_len);
+        let constr = builder.mul_add_extension(filter, constr, filter);
+        yield_constr.constraint(builder, constr);
+    }
+}
+
+/// Disable unused channels (everything except the first `used_in` and the last `used_out`).
+pub fn disable_unused_channels_packed<P: PackedField>(
```
dlubarov, line 113: This will be in `stack.rs` later right, once `disable_other_channels` is replaced with something more flexible? I think we shouldn't add any more code like this, since it will be replaced soon, we can just leave a TODO in stack.rs.

evm/src/cpu/dup_swap.rs
```diff
@@ -0,0 +1,249 @@
...
+    lv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    yield_constr.constraint(filter * (channel.used - P::ONES));
+    yield_constr.constraint(filter * (channel.is_read - P::Scalar::from_bool(is_read)));
+    yield_constr.constraint(filter * (channel.addr_context - lv.context));
+    yield_constr.constraint(
+        filter * (channel.addr_segment - P::Scalar::from_canonical_u64(Segment::Stack as u64)),
+    );
+    // // Top of the stack is at `addr = lv.stack_len - 1`.
```
dlubarov, line 59: nit: extra `//`

evm/src/cpu/dup_swap.rs
```diff
@@ -0,0 +1,249 @@
...
+use plonky2::iop::ext_target::ExtensionTarget;
+use plonky2::plonk::circuit_builder::CircuitBuilder;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::{CpuColumnsView, MemoryChannelView};
+use crate::cpu::membus::NUM_GP_CHANNELS;
+use crate::memory::segments::Segment;
+
+/// Constrain two channels to have equal values.
+pub fn channels_equal_packed<P: PackedField>(
```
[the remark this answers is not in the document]
nbgl: whoops yes, copy-paste error

evm/src/cpu/dup_swap.rs
```diff
@@ -0,0 +1,249 @@
...
+    let in1_channel = &lv.mem_channels[0];
+    let in2_channel = &lv.mem_channels[1];
+    let out1_channel = &lv.mem_channels[NUM_GP_CHANNELS - 2];
+    let out2_channel = &lv.mem_channels[NUM_GP_CHANNELS - 1];
+
+    channels_equal_packed(filter, in1_channel, out1_channel, yield_constr);
+    channels_equal_packed(filter, in2_channel, out2_channel, yield_constr);
+
+    constrain_channel_packed(true, filter, n, in2_channel, lv, yield_constr);
+    constrain_channel_packed(false, filter, n, out1_channel, lv, yield_constr);
```
dlubarov, line 193: This feels slightly confusing to me since half the logic is here and half in stack.rs; takes a bit of thinking to piece it together and check that each step occurs in the right order / with the right offsets.

How would you feel about using `num_pops: false`, `pushes: false` for SWAP so that the logic can be centralized here, with 4 `constrain_channel_packed` calls?

dlubarov (approved):
LGTM

evm/src/cpu/dup_swap.rs
nbgl: Yeah, fair point!

evm/src/cpu/dup_swap.rs
nbgl: Sure! I’m not exactly clear on how to do that nicely, but we’ll figure something out, and I’ll delete these constraints in the meantime.

PR #755, opened 2022-10-04
Modular operations for the EVM arithmetic unit

unzvfu:
This PR implements the `MOD`, `ADDMOD` and `MULMOD` operations.


field/src/types.rs
```diff
@@ -490,6 +490,19 @@ pub trait Field64: Field {
...
     fn from_noncanonical_u64(n: u64) -> Self;
 
+    /// Returns `n` as an element of this field. Assumes that `|n| < Self::ORDER`.
+    // TODO: Should probably be unsafe.
+    // TODO: Move to `Field`.
+    #[inline]
+    fn from_canonical_i64(n: i64) -> Self {
+        Self::from_canonical_u64(if n < 0 {
+            // Assumption guarantees that -n < ORDER.
+            Self::ORDER - (-n as u64)
```
nbgl, line 500: Subtle bug: this code fails when `n` = -2^63. In that case, `-n` overflows since 2^63 is not in the range of `i64`.

nbgl: One way to do this is:
```suggestion
            Self::ORDER.wrapping_add(n as u64)
```

field/src/types.rs
```diff
@@ -490,6 +490,19 @@ pub trait Field64: Field {
     // TODO: Move to `Field`.
     fn from_noncanonical_u64(n: u64) -> Self;
 
+    /// Returns `n` as an element of this field. Assumes that `|n| < Self::ORDER`.
+    // TODO: Should probably be unsafe.
+    // TODO: Move to `Field`.
+    #[inline]
+    fn from_canonical_i64(n: i64) -> Self {
```
nbgl, line 497: I’m not sure this method name makes sense. We define canonical representation as anything in {0, …, `ORDER`}, so by that definition a negative number can never be canonical.

field/src/types.rs
unzvfu: Good catch!

evm/src/arithmetic/modular.rs
```diff
@@ -0,0 +1,595 @@
...
+
+/// Add constraints for modular operations.
+pub(crate) fn eval_packed_generic<P: PackedField>(
+    lv: &[P; NUM_ARITH_COLUMNS],
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    let filter = lv[columns::IS_ADDMOD] + lv[columns::IS_MULMOD] + lv[columns::IS_MOD];
+    /*
+    // TODO: Do I need to check this here?
+    yield_constr.constraint(filter * filter - filter);
```
nbgl, line 321: No, this is already done by the cross-table lookup :)

evm/src/arithmetic/mul.rs
```diff
@@ -247,23 +230,26 @@ mod tests {
...
+            let mut constraint_consumer = ConstraintConsumer::new(
+                vec![GoldilocksField(2), GoldilocksField(3), GoldilocksField(5)],
+                GoldilocksField::ONE,
+                GoldilocksField::ONE,
+                GoldilocksField::ONE,
+            );
+            eval_packed_generic(&lv, &mut constraint_consumer);
+            for &acc in &constraint_consumer.constraint_accs {
+                assert_eq!(acc, GoldilocksField::ZERO);
+            }
```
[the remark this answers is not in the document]
nbgl: Oh I missed the TODO

evm/src/arithmetic/arithmetic_stark.rs
```diff
@@ -50,6 +51,12 @@ impl<F: RichField, const D: usize> ArithmeticStark<F, D> {
             compare::generate(local_values, columns::IS_LT);
         } else if local_values[columns::IS_GT].is_one() {
             compare::generate(local_values, columns::IS_GT);
+        } else if local_values[columns::IS_ADDMOD].is_one() {
+            modular::generate(local_values, columns::IS_ADDMOD);
+        } else if local_values[columns::IS_MULMOD].is_one() {
+            modular::generate(local_values, columns::IS_MULMOD);
+        } else if local_values[columns::IS_MOD].is_one() {
```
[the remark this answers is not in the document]
unzvfu: That's essentially what I do [here](https://github.com/mir-protocol/plonky2/blob/hamish/evm-arith-unit-modular-ops/evm/src/arithmetic/modular.rs#L332), but perhaps you have in mind a cleaner solution?

evm/src/arithmetic/columns.rs
```diff
@@ -64,7 +64,10 @@ const fn gen_input_cols<const N: usize>(start: usize) -> [usize; N] {
 const GENERAL_INPUT_0: [usize; N_LIMBS] = gen_input_cols::<N_LIMBS>(0);
 const GENERAL_INPUT_1: [usize; N_LIMBS] = gen_input_cols::<N_LIMBS>(N_LIMBS);
 const GENERAL_INPUT_2: [usize; N_LIMBS] = gen_input_cols::<N_LIMBS>(2 * N_LIMBS);
-const AUX_INPUT_0: [usize; N_LIMBS] = gen_input_cols::<N_LIMBS>(3 * N_LIMBS);
+const GENERAL_INPUT_3: [usize; N_LIMBS] = gen_input_cols::<N_LIMBS>(3 * N_LIMBS);
```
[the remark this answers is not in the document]
unzvfu: Good call. I thought my way of using `array::map()` on those column numbers would be faster, but (if I've understood it properly) a [comparison](https://rust.godbolt.org/z/r9GnP3EoY) shows that the range method produces much nicer code!

evm/src/arithmetic/utils.rs
```diff
@@ -17,9 +31,279 @@ pub(crate) fn _range_check_error<const RC_BITS: u32>(file: &str, line: u32, cols
...
+    // involves replacing Field::is_zero() with num::Zero::is_zero()
+    // which is used everywhere. Hence Default::default() it is.
+    [T::default(); N]
+}
+
+pub(crate) fn pol_add_assign<T>(a: &mut [T], b: &[T])
+where
+    T: AddAssign + Copy + Default,
+{
+    debug_assert!(a.len() >= b.len(), "expected {} >= {}", a.len(), b.len());
```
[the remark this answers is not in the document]
unzvfu: Interesting! The iterator code is also cleaner/more idiomatic, so I'll use that.

evm/src/arithmetic/utils.rs
```diff
@@ -17,9 +31,279 @@ pub(crate) fn _range_check_error<const RC_BITS: u32>(file: &str, line: u32, cols
...
+where
+    T: Copy + Default,
+{
+    // NB: This should really be T::zero() from num::Zero, because
+    // default() doesn't guarantee to initialise to zero (though in
+    // our case it always does). However I couldn't work out how to do
+    // that without touching half of the entire crate because it
+    // involves replacing Field::is_zero() with num::Zero::is_zero()
+    // which is used everywhere. Hence Default::default() it is.
+    [T::default(); N]
```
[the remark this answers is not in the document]
unzvfu: Marked as TODO, but I'll fix this properly in an upcoming PR.

evm/src/arithmetic/utils.rs
```diff
@@ -17,9 +31,297 @@ pub(crate) fn _range_check_error<const RC_BITS: u32>(file: &str, line: u32, cols
...
+    a: [ExtensionTarget<D>; 2 * N_LIMBS],
+    root: ExtensionTarget<D>,
+) -> [ExtensionTarget<D>; 2 * N_LIMBS] {
+    let zero = builder.zero_extension();
+    let mut res = [zero; 2 * N_LIMBS];
+    let t = builder.mul_extension(root, a[0]);
+    res[0] = builder.sub_extension(zero, t);
+    for deg in 1..(2 * N_LIMBS - 1) {
+        let t = builder.mul_extension(root, a[deg]);
+        res[deg] = builder.sub_extension(a[deg - 1], t);
```
nbgl, line 286: This could be done with one `arithmetic_extension`. Something like
```suggestion
        res[deg] = builder.arithmetic_extension(F::NEG_ONE, F::ONE, root, a[deg], t);
```

nbgl (approved):
Nice!!

PR #758, opened 2022-10-05
Conditional recursive verifier

wborgeaud:


plonky2/src/plonk/conditional_recursive_verifier.rs
```diff
@@ -0,0 +1,352 @@
...
+                    quotient_polys_cap,
+                    openings,
+                    opening_proof,
+                },
+            public_inputs,
+        } = proof_with_pis0;
+        let ProofWithPublicInputsTarget {
+            proof:
+                ProofTarget {
+                    wires_cap: dummy_wires_cap,
```
dlubarov, line 46: nit: Let's not call it a dummy proof since technically the method is more general. Seems like it could theoreticaly be used to verify either of two nontrivial circuits, in some particular cases where the two circuits happen to have the same size, gates, etc.

plonky2/src/plonk/conditional_recursive_verifier.rs
```diff
@@ -0,0 +1,352 @@
...
+        b: BoolTarget,
+        cap0: MerkleCapTarget,
+        cap1: MerkleCapTarget,
+    ) -> MerkleCapTarget {
+        assert_eq!(cap0.0.len(), cap1.0.len());
+        MerkleCapTarget(
+            cap0.0
+                .into_iter()
+                .zip_eq(cap1.0)
+                .map(|(h0, h1)| HashOutTarget {
```
dlubarov, line 110: nit: may be nice to have a `select_hash` since it comes up in a couple places

dlubarov (approved):
LGTM 👍 

Maybe as a next step we can have a conditional recursion function which only takes one inner circuit, and generates a dummy circuit automatically?

PR #759, opened 2022-10-05
Fix RipeMD padlength issue

typ3c4t:
The problem was downstream of an error in my simplification of the original Python implementation. The original paldength definition was:
```python
x = 64 - length % 64
if x < 9:
    x += 64
return x - 8
```
and in my Python code, I had made the error of computing it as
```py
x = 56 - length % 64
return x + (x < 9)*64
```
where I subtracted `64 - 8 = 56` but still checked `x < 9`, rather than what should have been `x < 1`

Note that, to avoid overflow, in the asm it's now
```py
t = length % 64
return 56 + 64*(t > 55) - t
```
where before the check was `t > 47`

### DIFF
--- a/evm/src/cpu/kernel/tests/hash.rs
+++ b/evm/src/cpu/kernel/tests/hash.rs
@@ -23,27 +23,24 @@ fn ripemd(input: Vec<u8>) -> U256 {
     U256::from(&hasher.finalize()[..])
 }
 
-fn make_random_input() -> (u32, Vec<u8>) {
+fn make_random_input() -> Vec<u8> {
     // Generate a random message, between 0 and 9999 bytes.
     let mut rng = thread_rng();
     let num_bytes = rng.gen_range(0..10000);
-    (num_bytes, (0..num_bytes).map(|_| rng.gen()).collect())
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/tests/hash.rs
```diff
@@ -23,31 +23,62 @@ fn ripemd(input: Vec<u8>) -> U256 {
...
+    (48, vec![
+        86, 124, 206, 245, 74, 57, 250, 43, 60, 30, 254, 43, 143, 144, 242, 215, 13, 103, 237, 61,
+        90, 105, 123, 250, 189, 181, 110, 192, 227, 57, 145, 46, 221, 238, 7, 181, 146, 111, 209,
+        150, 31, 157, 229, 126, 206, 105, 37, 17,
+    ])
+}
 
-    // Load the message onto the stack.
-    let mut initial_stack = vec![U256::from(num_bytes)];
+fn make_input_stack(length: u32, message: Vec<u8>) ->  Vec<U256> {
```
wborgeaud, line 42: nit: Can also remove the `length` parameter and use `message.len()`.

wborgeaud (approved):
Looks good to me!

PR #765, opened 2022-10-07
Fp12 multiplication

typ3c4t:


evm/src/cpu/kernel/asm/fields/Fp6.asm
```diff
@@ -0,0 +1,206 @@
+// cost: 6 loads + 6 offsets + 5 adds = 6*4 + 6*1 + 5*2 = 40
+%macro load_fp6(offset)
```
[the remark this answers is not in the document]
nbgl: This might require some changes to the assembler, but evaluating arithmetic expressions is pretty basic functionality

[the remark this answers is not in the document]
dlubarov: I think we can leave this to the optimizer, its constant propagation is pretty basic but should be able to handle this case.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -238,9 +238,9 @@ impl<'a> Interpreter<'a> {
             0x09 => self.run_mulmod(),                                  // "MULMOD",
             0x0a => self.run_exp(),                                     // "EXP",
             0x0b => todo!(),                                            // "SIGNEXTEND",
-            0x0c => todo!(),                                            // "ADDFP254",
-            0x0d => todo!(),                                            // "MULFP254",
-            0x0e => todo!(),                                            // "SUBFP254",
+            0x0c => self.run_add_p(),                                   // "ADDFP254",
+            0x0d => self.run_mul_p(),                                   // "MULFP254",
+            0x0e => self.run_sub_p(),                                   // "SUBFP254",
```
typ3c4t, line 243: temporary implementations just for testing, and using a hardcoded dummy value of `101` for the prime

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -238,9 +238,9 @@ impl<'a> Interpreter<'a> {
             0x09 => self.run_mulmod(),                                  // "MULMOD",
             0x0a => self.run_exp(),                                     // "EXP",
             0x0b => todo!(),                                            // "SIGNEXTEND",
-            0x0c => todo!(),                                            // "ADDFP254",
-            0x0d => todo!(),                                            // "MULFP254",
-            0x0e => todo!(),                                            // "SUBFP254",
+            0x0c => self.run_addfp254(),                                // "ADDFP254",
+            0x0d => self.run_mulfp254(),                                // "MULFP254",
+            0x0e => self.run_subfp254(),                                // "SUBFP254",
```
typ3c4t, line 243: temporary functions---hardcoding the prime with a dummy value of `101` for the sake of testing

evm/src/cpu/kernel/asm/fields/fp6.asm
```diff
@@ -0,0 +1,440 @@
...
+    SWAP7
+    ADDFP254
+    SWAP6
+    // stack: f0,                     g0, h1, h2, h3, h4, h5
+    ADDFP254
+    // stack:                         h0, h1, h2, h3, h4, h5
+%endmacro
+
+// *backwards order subtraction* cost: 17
+%macro bus_fp6
```
nbgl, line 184: I find the name `bus` a bit confusing. In my experience, this operation is more commonly called `subr`, e.g. in [ARM64](https://developer.arm.com/documentation/ddi0596/2021-12/SVE-Instructions/SUBR--vectors---Reversed-subtract-vectors--predicated--) and [x86](https://www.felixcloutier.com/x86/fsubr:fsubrp:fisubr).

evm/src/cpu/kernel/asm/fields/fp12.asm
```diff
@@ -0,0 +1,103 @@
...
+/// G = g + g'z
+///
+/// H = h + h'z = FG 
+///
+/// h  = fg + sh(f'g')
+/// h' = (f+f')(g+g') - fg - f'g'
+///
+/// Note: each symbol in the stack comments consists of six words
+
+global mul_Fp12:
```
nbgl, line 43: It looks like this function assumes that the inputs are at addresses `0..12` and `12..24`, with the output written to `24..36`. The issue with this approach is that it makes the caller copy a whole bunch of data in order to call this function.

Usually when a function operates on data that has to be read from/written to memory, it would take pointers to the inputs and a pointer to the output buffer. So this function would take three pointers:
- first input
- second input
- the place where the output should be written

evm/src/cpu/kernel/asm/fields/fp6.asm
```diff
@@ -0,0 +1,447 @@
...
+    SUBFP254
+    SWAP6
+    // stack: f0,                     g0, h1, h2, h3, h4, h5
+    SWAP1
+    SUBFP254
+    // stack:                         h0, h1, h2, h3, h4, h5
+%endmacro
+
+// cost: 156
+%macro mul_fp6
```
nbgl, line 212: This is really big, so it should be a function, not a macro.

A macro gets copy-pasted a bunch of times, increasing the kernel size. That’s fine if it’s small, but if it’s big enough, then the overhead of a function call is relatively minimal.

evm/src/cpu/kernel/asm/fields/fp12.asm
```diff
@@ -0,0 +1,93 @@
...
+    %store_fp6(18)
+    PUSH return_on_stack
+    // stack: return_on_stack
+    %jump(mul_Fp12)
+return_on_stack:
+    // stack:
+    %load_fp6(30)
+    %load_fp6(24)
+    // stack: h, h'
+    %jump(0xdeadbeef)
```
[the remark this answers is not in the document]
typ3c4t: sounds good--is there a way to test this without doing this pre/post on-stack processing?

evm/src/cpu/kernel/tests/fields.rs
```diff
@@ -0,0 +1,200 @@
+use anyhow::Result;
+use ethereum_types::U256;
+use rand::{thread_rng, Rng};
+
+use crate::cpu::kernel::aggregator::combined_kernel;
+use crate::cpu::kernel::interpreter::run_with_kernel;
+
+const P254: u32 = 107;
```
nbgl, line 8: Please add a comment to explain why our prime is `107` and when it should be changed and by whom 😛

evm/src/cpu/kernel/asm/fields/fp6_macros.asm
```diff
@@ -0,0 +1,316 @@
+// cost: 6 loads + 6 dup/swaps + 5 adds = 6*4 + 6*1 + 5*2 = 40
+%macro load_fp6
+    // stack: offset
```
nbgl, line 3: I wouldn’t call it an `offset`. Pointer or address are better.

An offset is the difference between two addresses. So `foo + 4` is offset by `4` from `foo`, but `foo + 4` is not itself an offset.

In other words: if it makes sense to add it to a pointer, _then_ it’s an offset.

evm/src/cpu/kernel/asm/fields/fp6_macros.asm
```diff
@@ -0,0 +1,316 @@
...
+    SWAP4
+    // stack: g0, g1, g2, g3, g4, f5, f0, f1, f2, f3, f4, g5
+    SWAP5
+    SWAP11
+    SWAP5
+    // stack: g0, g1, g2, g3, g4, g5, f0, f1, f2, f3, f4, f5
+%endmacro
+
+// cost: 16
+%macro swap_fp6_hole
```
nbgl, line 237: Please give this a comment :)

nbgl (approved):
good work :)

PR #775, opened 2022-10-11
Generate dummy proof with a given shape

wborgeaud:
Generate dummy proof with a given `CommonCircuitData`
Needed to move `circuit_digest` from `CommonCircuitData` to `ProverOnlyCircuitData` and `VerifierOnlyCircuitData` for that.


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -271,6 +271,10 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         );
     }
 
+    pub fn add_gate_to_gate_set(&mut self, gate: GateRef<F, D>) {
```
dlubarov, line 274: Is this public because users will need to call it for cyclic recursion?

plonky2/src/plonk/proof.rs
```diff
@@ -81,9 +81,10 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
 {
     pub fn compress(
         self,
+        circuit_digest: &<<C as GenericConfig<D>>::Hasher as Hasher<C::F>>::Hash,
```
dlubarov, line 84: What do you think of having this (and `decompress`) take `VerifierCircuitData`? To combine the two params into one.

Or we could do it as a larger refactor later. Seems like we have a lot of code where `VerifierOnlyCircuitData`/`CommonCircuitData` are separated, like the `recursive_verifier` test code, when combining them could make things more succinct.

dlubarov: Oh, maybe that wouldn't be ideal since `VerifierCircuitData` owns its data, so it might require some cloning.

plonky2/src/plonk/conditional_recursive_verifier.rs
```diff
@@ -1,20 +1,65 @@
...
+use crate::plonk::config::{AlgebraicHasher, GenericConfig, Hasher};
+use crate::plonk::proof::{
+    OpeningSetTarget, ProofTarget, ProofWithPublicInputs, ProofWithPublicInputsTarget,
+};
 use crate::with_context;
 
+/// Generate a proof having a given `CommonCircuitData`.
+pub fn dummy_proof<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>(
+    common_data: &CommonCircuitData<F, C, D>,
+) -> Result<(ProofWithPublicInputs<F, C, D>, CircuitData<F, C, D>)>
```
dlubarov, line 27: Couple nits,

- Maybe this should be private; I think we should expose a higher-level method, which looks just like `verify_proof` except with a `BoolTarget` flag, and calls `dummy_proof` internally.
- I would return the `VerifierOnlyCircuitData` part, to avoid redundancy between inputs and outputs (and might be slightly more convenient, if it just gets passed to `conditionally_verify_proof`)

dlubarov (approved):
LGTM otherwise

plonky2/src/plonk/circuit_builder.rs
wborgeaud: Hm I don't think users would call this directly but I thought it could potentially be useful to people using `plonky2` as a library.

plonky2/src/plonk/proof.rs
wborgeaud: > Oh, maybe that wouldn't be ideal since `VerifierCircuitData` owns its data, so it might require some cloning.

Yeah that's the issue I ran into here and elsewhere. There's no way currently to go from `CircuitData` to `VerifierCircuitData` without throwing away `ProverOnlyCircuitData` or some cloning. Since the prover should be able to run these functions without throwing away the prover key, I thought it would just be easier to pass the circuit digest, even though it's a bit ugly.

plonky2/src/plonk/conditional_recursive_verifier.rs
wborgeaud: > - Maybe this should be private; I think we should expose a higher-level method, which looks just like `verify_proof` except with a `BoolTarget` flag, and calls `dummy_proof` internally.

Hm I'm not sure I understand what that function look like and how the flag would be use. Can you explain a bit more?

> - I would return the `VerifierOnlyCircuitData` part, to avoid redundancy between inputs and outputs (and might be slightly more convenient, if it just gets passed to `conditionally_verify_proof`)

Sounds good!

dlubarov: Yeah exactly, I was thinking of a `conditionally_verify_proof` function with only one proof/verifier_data. I think it would be useful for handling variable-size inner proofs. E.g. if need to support three possible sizes, we'd call this new `conditionally_verify_proof` three times, and assert one of the three flags is 1, so the prover would have to use one real VK and two dummy VKs.

dlubarov: I agree dummy circuits could be convenient, at least for testing, but I think the typical API for that would look more like
```rust
fn dummy_circuit<...>(config: CircuitConfig, degree_bits: usize) -> CircuitData<...> { ... }
```
I'm not sure if there will be another use case where we already have a `CommonCircuitData` and want to match it? If it's unclear, I'd argue it's best to start private to keep our docs simpler, and since it's easier to change private to public than the reverse.

wborgeaud: Sounds good, I added a `conditionally_verify_proof_or_dummy` function that takes only one proof and I changed `dummy_proof` to `pub(crate)`.

PR #776, opened 2022-10-11
Represent input columns as ranges rather than arrays

unzvfu:
Suggested by Jacqui; seems cleaner.

Some minor changes made so code works with slices instead of arrays where that seemed cleaner and not less efficient.


nbgl (approved):
looks good!!

PR #778, opened 2022-10-12
Updated `eth_trie_utils` to `0.2.1`

BGluth:
Bumped `eth_trie_utils` to `0.2.0`. There were a few public API changes, so I ended up updating the callsites in the tests so that it compiles. I also did a bit of refactoring to reduce code duplication in the tests, so let me know what you think about these changes.

### DIFF
--- a/evm/src/cpu/kernel/tests/mpt/mod.rs
+++ b/evm/src/cpu/kernel/tests/mpt/mod.rs
@@ -9,6 +9,17 @@ mod insert;
 mod load;
 mod read;
 
+/// Helper function to reduce code duplication.
+/// Note that this preserves all nibbles (eg. `0x123` is not interpreted as `0x0123`).
+pub(crate) fn nibbles<T: Into<U256>>(v: T) -> Nibbles {
+    let packed = v.into();
+
+    Nibbles {
+        count: Nibbles::get_num_nibbles_in_key(&packed),
+        packed,
+    }
+}
+
 pub(crate) fn test_account_1() -> AccountRlp {
     AccountRlp {
         nonce: U256::from(1111),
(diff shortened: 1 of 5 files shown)


evm/src/cpu/kernel/tests/mpt/mod.rs
```diff
@@ -9,6 +9,17 @@ mod insert;
...
+/// Note that this preserves all nibbles (eg. `0x123` is not interpreted as `0x0123`).
+pub(crate) fn nibbles<T: Into<U256>>(v: T) -> Nibbles {
+    let packed = v.into();
+
+    Nibbles {
+        count: Nibbles::get_num_nibbles_in_key(&packed),
+        packed,
+    }
+}
+
```
BGluth, line 22: I'm doing something similar in the tests of `eth_trie_utils`. It feels pretty clunky manually constructing `Nibbles` all of the time, so I'm really tempted to add different ways to do so. This is nice, but I really want the user to be aware that it's not padding nibbles to the nearest byte. I could instead make an explicit static method that creates `Nibbles` from a `U256` and make it very clear in the docs about preserving nibbles.

dlubarov: I think this behavior (using the smallest number of nibbles possible) is very natural/expected. I know some callers have RLP bytes that get converted to nibbles, but then in my mind there's no padding, just those callers are starting with bytes already so the nibbles are already even.

So I think it'd be fine to have a function like this in the lib, and the semantics would be fairly clear. If we wanted to be extra clear we could call it something like `Nibbles::from_u256_no_padding`.

A helper like this is also fine for our purposes though, so happy to merge as is.

evm/src/cpu/kernel/tests/mpt/mod.rs
```diff
@@ -38,16 +49,14 @@ pub(crate) fn test_account_2_rlp() -> Vec<u8> {
 /// A `PartialTrie` where an extension node leads to a leaf node containing an account.
 pub(crate) fn extension_to_leaf(value: Vec<u8>) -> PartialTrie {
     PartialTrie::Extension {
-        nibbles: Nibbles {
-            count: 3,
-            packed: 0xABC.into(),
-        },
-        child: Box::new(PartialTrie::Leaf {
+        nibbles: nibbles(0xABC),
```
dlubarov, line 52: Much nicer! Thanks

dlubarov (approved):
LGTM, thanks!

evm/src/cpu/kernel/tests/mpt/mod.rs
BGluth: > I don't think there's a use case where nibbles get padded to the next even number?

Right true, so this is really an non-issue.

> A helper like this is also fine for our purposes though, so happy to merge as is.

Yeah I would need to add another patch bump to add it in, so if this works for now then yeah maybe just stick with the helper until `0.3.*`?

PR #779, opened 2022-10-12
Add Fp254 ops to the CPU table

nbgl:
I will need help from @unzvfu to add `SUBMOD` support to the arithmetic table (but this will be a different PR)


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -238,6 +238,9 @@ impl<'a> Interpreter<'a> {
             0x09 => self.run_mulmod(),                                  // "MULMOD",
             0x0a => self.run_exp(),                                     // "EXP",
             0x0b => todo!(),                                            // "SIGNEXTEND",
+            0x0c => todo!(),                                            // "ADDFP254",
+            0x0d => todo!(),                                            // "MULFP254",
+            0x0e => todo!(),                                            // "SUBFP254",
```
typ3c4t, line 243: wouldn't these just amount to the implementations for `run_{add,sub,mul}mod` but with `N` hardcoded to the p254?

evm/src/cpu/kernel/opcodes.rs
```diff
@@ -20,6 +20,9 @@ pub(crate) fn get_opcode(mnemonic: &str) -> u8 {
         "MULMOD" => 0x09,
         "EXP" => 0x0a,
         "SIGNEXTEND" => 0x0b,
+        "ADDFP254" => 0x0c,
+        "MULFP254" => 0x0d,
+        "SUBFP254" => 0x0e,
```
typ3c4t, line 25: just out of curiosity is there a systematic reason for how these are enumerated (e.g. why had these numbers been skipped)?

evm/src/cpu/modfp254.rs
```diff
@@ -0,0 +1,53 @@
...
+use plonky2::field::packed::PackedField;
+use plonky2::field::types::Field;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::CpuColumnsView;
+
+// Python:
+// >>> P = 21888242871839275222246405745257275088696311157297823662689037894645226208583
```
typ3c4t, line 12: hmm according to https://neuromancer.sk/std/bn/bn254, the prime is 
```py
In [1]: 0x2523648240000001BA344D80000000086121000000000013A700000000000013
Out[1]: 16798108731015832284940804142231733909889187121439069848933715426072753864723
```

typ3c4t (approved):
seems reasonable! just have to make sure the value of the prime is correct

evm/src/cpu/modfp254.rs
wborgeaud: Jacqui's prime is the correct one.
(There are two BN curves that go by the name `bn254`, the one used by Ethereum is [this one](https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bn254), which is different from the curve referenced in neuromancer.sk)

typ3c4t: oh wow! sorry for the remark then :)

evm/src/cpu/kernel/opcodes.rs
nbgl: We have to support all the [EVM opcodes](https://ethereum.org/en/developers/docs/evm/opcodes/) but can fill the gaps with whatever additional operations we want :)

evm/src/cpu/kernel/interpreter.rs
nbgl: You’re right, but I’d just rather do it in a different PR to keep this one small. (Also, I’ll need to write `run_submod` because it doesn’t currently exist.)

PR #780, opened 2022-10-13
EVM shift left operation

unzvfu:
This PR implements the `SHL` instruction. It is almost entirely dedicated to verifying that a given auxiliary value `c` is actually `2^d` for the given `d`. The rest of the work is delegated to `MUL`.

### DIFF
--- a/evm/src/arithmetic/shift.rs
+++ b/evm/src/arithmetic/shift.rs
@@ -143,7 +143,7 @@ pub(crate) fn verify_2exp_packed<P: PackedField>(
     // otherwise. Because of the range checks on exp, it is bounded
     // above by 2^8 + 15 * 2^16.
     // NB: I don't understand why I do lv[exp].into_iter().skip(1).sum::<P>();
-    let sum_tail = |ps: &[P]| ps.into_iter().skip(1).fold(P::ZEROS, |acc, &x| acc + x);
+    let sum_tail = |ps: &[P]| ps.iter().skip(1).fold(P::ZEROS, |acc, &x| acc + x);
     let exp_quo256_is_nonzero = exp_limb0_quo256 + sum_tail(&lv[exp]);
 
     // If exp_quo256_is_nonzero is zero, then exp < 256 and the
@@ -159,7 +159,7 @@ pub(crate) fn verify_2exp_packed<P: PackedField>(
     // Check that the only non-zero limb of pow_exp is the
     // nz_limb_idx-th limb (it doesn't have to be nonzero though).
     let mut limb_sum = P::ZEROS;
-    for (i, &limb_i) in (0..).zip(lv[pow_exp].into_iter()) {
+    for (i, &limb_i) in (0..).zip(lv[pow_exp].iter()) {
         let i = P::Scalar::from_canonical_u64(i);
         yield_constr.constraint(filter * limb_i * (nz_limb_idx - i));
         limb_sum += limb_i;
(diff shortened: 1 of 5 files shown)


nbgl:
@unzvfu I think we should consider another way of doing shifts, which might be simpler.

Using `MUL`/`DIV` for `SHL`/`SHR` makes perfect sense, but computing `1 << input[1]` is the complicated part. Instead of computing it in constraints, we could store them in a lookup table.

We would have a dedicated segment in memory, address `e` stores `2**e`. (This would need to be populated by the kernel at start-up.) The CPU table would read up `lookup_table[e]`, pass `2**e` directly to the arithmetic table, and invoke `MUL` or `DIV`. The arithmetic table would not even need to know that `SHL`/`SHR` exist.

If `e >= 256`, the memory in the lookup table would default to `0`, so `MUL`/`DIV` would yield `0` as desired.

This is your call ofc, but I think this method is worth mentioning :)

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,460 @@
...
+    let exp_mod256_quo16 = lv[exp_mod256_quo16];
+    let exp_mod16_bits: [P; 4] = read_value(lv, exp_mod16_bits);
+
+    // Check that every "bit" of exp_mod16_bits is 0 or 1
+    for b in exp_mod16_bits {
+        yield_constr.constraint(filter * (b * b - b));
+    }
+
+    // Check that exp_mod256_quo16 ∈ [0, 16)
+    // FIXME range_check_error!(SHIFT_INPUT_EXP_MOD256_QUO16, 4);
```
nbgl, line 150: I think you also need to check that `exp_limb0_quo256` is in [0, 256).

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,460 @@
...
+    // is >= 16, and so all limbs of pow_exp are forced to be zero by
+    // the following for loop.
+    let nz_limb_idx = exp_mod256_quo16 + exp_quo256_is_nonzero * P::Scalar::from_canonical_u64(16);
+
+    // Check that the only non-zero limb of pow_exp is the
+    // nz_limb_idx-th limb (it doesn't have to be nonzero though).
+    let mut limb_sum = P::ZEROS;
+    for (i, &limb_i) in (0..).zip(lv[pow_exp].iter()) {
+        let i = P::Scalar::from_canonical_u64(i);
+        yield_constr.constraint(filter * limb_i * (nz_limb_idx - i));
```
nbgl, line 184: clever!

nbgl (approved):

unzvfu:
> @unzvfu I think we should consider another way of doing shifts, which might be simpler.
> 
> Using `MUL`/`DIV` for `SHL`/`SHR` makes perfect sense, but computing `1 << input[1]` is the complicated part. Instead of computing it in constraints, we could store them in a lookup table.
> 
> We would have a dedicated segment in memory, address `e` stores `2**e`. (This would need to be populated by the kernel at start-up.) The CPU table would read up `lookup_table[e]`, pass `2**e` directly to the arithmetic table, and invoke `MUL` or `DIV`. The arithmetic table would not even need to know that `SHL`/`SHR` exist.
> 
> If `e >= 256`, the memory in the lookup table would default to `0`, so `MUL`/`DIV` would yield `0` as desired.
> 
> This is your call ofc, but I think this method is worth mentioning :)

Oh, yeah, that sounds much simpler! I obviously hadn't considered the possibility of using a lookup table from the CPU. I'll need some help to work out how to do that; let's discuss offline.

unzvfu:
Shift left/right has been implemented in #801 along the lines of [Jacqui's suggestion above](https://github.com/mir-protocol/plonky2/pull/780#issuecomment-1278152037), so this PR is superseded.

PR #781, opened 2022-10-13
Redundant `degree_bits`

wborgeaud:
`degree_bits` is a field of `FriParams` and `CommonCircuitData`, but `FriParams` is also a field of `CommonCircuitData`, so there's some (unchecked) redundancy.
Fixed that by removing `degree_bits` from `CommonCircuitData`.


dlubarov (approved):
Good idea

Issue #446, opened 2022-01-24
Support for cyclic recursion


dlubarov:
Support recursion where the inner circuit is the circuit currently being built, as in IVC etc. This is nontrivial since it's circular - we don't know what circuit we're verifying since it hasn't been constructed yet.

I think @ebfull mentioned using public inputs to pass in the verification key (which, in our setting, is basically a Merkle root of preprocessed polynomials). That seems like a good approach.

However, we will need a different workaround to deal with parameters that affect the "shape" of the recursion circuit. Plonky2 has several such parameters:
- degree (determines Merkle proof lengths)
- list of custom gates
- number of public inputs
- sequence of FRI reduction arities
- degree of the quotient polynomial
- number of preprocessed polynomials, and of partial product polynomials

None of these are known until the circuit is complete. We could have the caller specify these params (and raise an error if they don't work), but that would be a bit inconvenient.

Alternatively, we could estimate these params, then add recursion circuitry assuming those params in a clone of the circuit. Then we build the cloned circuit, and if the resulting params don't match our estimates, we update our estimates and try again.

We could generalize this further, allowing the user to define a set of circuits with various (potentially cyclic) recursive dependencies, then searching for a set of parameters that works for the whole set of circuits. That seems a bit difficult though, and I don't see an immediate need for it.

Discussed a bit in Discord; cc @Sladuca.

Sladuca:
> Alternatively, we could estimate these params, then add recursion circuitry assuming those params in a clone of the circuit. Then we build the cloned circuit, and if the resulting params don't match our estimates, we update our estimates and try again.

I think this is the best approach - I'd imagine the number of use cases that need the third approach is very small, and this keeps the details inside the library so we don't end up stuck with it for backwards compatibility if a more efficient way becomes possible down the line.

Sladuca:
Given our discussions on discord, here's an idea how this could work:

1. Add a `Vec` of self-verifiers to be lazily added to the circuit.
2. Add an `add_recursive_self_verifier` method to `CircuitBuilder` that just adds a new entry to the self-verifier `Vec`.
[...]

PR #782, opened 2022-10-14
Cyclic recursion

wborgeaud:
Address #446.


plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+        while self.num_gates() < (common_data.degree() / 2) {
+            self.add_gate(NoopGate, vec![]);
+        }
+        // Make sure we have every gate to match `common_data`.
+        for g in &common_data.gates {
+            self.add_gate_to_gate_set(g.clone());
+        }
+
+        let data = self.build::<C>();
+        ensure!(data.common == common_data, "Common data does not match.");
```
dlubarov, line 206: nit: let's include the two objects and maybe clarify a bit that the `common_data` passed in didn't match the final circuit. I think this will be a common error since users need to guess at `common_data`

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+}
+
+pub struct CyclicPublicInputs<
+    F: RichField + Extendable<D>,
+    C: GenericConfig<D, F = F>,
+    const D: usize,
+> {
+    pub circuit_digest: HashOut<F>,
+    pub constants_sigmas_cap: MerkleCap<F, C::Hasher>,
+    pub base_case: bool,
```
dlubarov, line 46: Thinking about this more, I'm not sure it should be a public input? I think would callers typically have some logic like "either the proof must perform cyclic recursion, or if it doesn't, then X", where X could be something like
- for a hash chain, "the preimage must be zero"
- for a chain of block proofs, "the initial state trie must be empty"
- for our aggregation tree, "a (non-dummy) transaction proof must be recursive verified"

I think it may be best to have `cyclic_recursion` just take any `condition: BoolTarget` (like `conditionally_verify_proof` does), and have the caller handle the rest.

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+        self.register_public_input(base_case.target);
+
+        common_data.num_public_inputs = self.num_public_inputs();
+        // The `conditionally_verify_proof` gadget below takes 2^12 gates, so `degree_bits` cannot be smaller than 13.
+        common_data.fri_params.degree_bits = common_data.fri_params.degree_bits.max(13);
+
+        let proof = self.add_virtual_proof_with_pis(&common_data);
+        let dummy_proof = self.add_virtual_proof_with_pis(&common_data);
+
+        let pis = CyclicPublicInputsTarget::from_slice(&proof.public_inputs, &common_data)?;
```
dlubarov, line 161: Hm what if there are other public inputs before `CyclicPublicInputsTarget` starts?

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+            circuit_digest,
+            constants_sigmas_cap,
+            base_case,
+        })
+    }
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
+    pub fn cyclic_recursion<C: GenericConfig<D, F = F>>(
+        mut self,
```
dlubarov, line 120: In our aggregation circuit we'll want two cyclic recursion gadgets; can we have this take `&mut self` so we can call it twice?

My understanding is this currently consumes `self` so that it can do the `data.common == common_data` check after `build()`. Perhaps we could instead have `CircuitBuilder` hold an `Option<CommonCircuitData<...>>`, tracking any `common_data` passed to `cyclic_recursion`, and then have `build()` check that it matches.

Alternatively we could continue having `cyclic_recursion` consume `self`, but generalize it so it can add multiple instances of a cyclic recursion gadget. Might be a more complicated API though.

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+        };
+
+        // Flag set to true for the base case of the cycle where we verify a dummy proof to bootstrap the cycle. Set to false otherwise.
+        // Unsafe is ok since `base_case` is a public input and its booleaness should be checked in the verifier.
+        let base_case = self.add_virtual_bool_target_unsafe();
+        self.register_public_input(base_case.target);
+
+        common_data.num_public_inputs = self.num_public_inputs();
+        // The `conditionally_verify_proof` gadget below takes 2^12 gates, so `degree_bits` cannot be smaller than 13.
+        common_data.fri_params.degree_bits = common_data.fri_params.degree_bits.max(13);
```
dlubarov, line 156: Hm it seems a little odd to me that the user has to guess these params, but we automatically correct certain mistakes. Might be best to throw an error pointing out the mistakes, to avoid any confusion. Like a user might think they got it to work with 2^12, when really it was bumped to 2^13 (if they don't check the logs for the real size).

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,490 @@
...
+    proof: &'a Option<ProofWithPublicInputs<F, C, D>>,
+    verifier_data: &'a VerifierOnlyCircuitData<C, D>,
+    common_data: &'a CommonCircuitData<F, C, D>,
+}
+
+pub struct CyclicRecursionTarget<const D: usize> {
+    pub proof: ProofWithPublicInputsTarget<D>,
+    pub verifier_data: VerifierCircuitTarget,
+    pub dummy_proof: ProofWithPublicInputsTarget<D>,
+    pub dummy_verifier_data: VerifierCircuitTarget,
```
dlubarov, line 35: Can these be baked into the circuit? Seems like they could be computed without running into a circular dependency.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: `CyclicPublicInputsTarget::from_slice()` only looks at the tail of the slice, so it's not an issue if there are previous public inputs.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: > In our aggregation circuit we'll want two cyclic recursion gadgets; can we have this take `&mut self` so we can call it twice?
> 
> My understanding is this currently consumes `self` so that it can do the `data.common == common_data` check after `build()`. Perhaps we could instead have `CircuitBuilder` hold an `Option<CommonCircuitData<...>>`, tracking any `common_data` passed to `cyclic_recursion`, and then have `build()` check that it matches.

That sounds good to me! Do you think we should do that on the `CircuitBuilder` or add a wrapped version `CyclicCircuitBuilder` and add it there (to avoid cyclic recursion logic in the base circuit builder)?

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: Good point, I just remove it altogether since it's caught in the `CommonCircuitData` equality check.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: Hm what do you mean by "baked into the circuit"? Like as constants? What would be the advantage of that?

plonky2/src/recursion/cyclic_recursion.rs
dlubarov: I think having it in `CircuitBuilder` seems okay to me

plonky2/src/recursion/cyclic_recursion.rs
dlubarov: I don't think we necessarily need to know if the previous proof was a base case? It seems like the first proof can enforce a base-case condition on its own, e.g. for a hash chain we'd have logic like "if there's no inner proof, then our preimage must be zero".

plonky2/src/recursion/cyclic_recursion.rs
dlubarov: Right I mean as constants; was thinking it could be nice if we didn't need to keep track of these targets and populate them later.

I probably should have suggested this for `conditionally_verify_proof_or_dummy` instead. There if we baked in `dummy_proof`, `dummy_verifier_data` as constants, we wouldn't need to return anything. So the API would match `verify_proof` better, and the user wouldn't need to call `set_proof_with_pis_target` or anything to populate dummy proof data.

With `cyclic_recursion` there'd be less advantage; the user would still need to call `set_cyclic_recursion_data_target` since it does other stuff besides the dummy proof.

I suppose the current approach is a bit more efficient currently, since it avoids extra `ConstantGate`s. Though it could change if we added custom select gates later.

Anyway definitely not something we need to change now.

plonky2/src/recursion/cyclic_recursion.rs
dlubarov: Oh I see, makes sense.

I guess if we go with that idea allowing `cyclic_recursion` to be called at any time, that would complicate this a bit, but we could always keep track of where the cyclic recursion public inputs start. (I think we only need one set of public inputs even if doing multiple instances of cyclic recursion)

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: I see. I think adding all these constants could non-trivially increase the circuit size. Another option could be to non-deterministically set these targets using generators. That way we don't have to manually populate the targets and we also don't need constants.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: Hm but then how do you ensure in general that a base case is not following a recursion case? This might not be strictly necessary for certain use cases, but I think it could be useful in general.

plonky2/src/recursion/cyclic_recursion.rs
dlubarov: Good idea, adding generators seems like a nice solution for it

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: As per our discussion, I removed `base_case` from the public inputs. I also removed the `CyclicPublicInputs{,Target}` structs altogether, since they're now just equivalent to `VerifierOnlyCircuitData{,Target}`.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: Added `goal_common_data: Option<CommonCircuitData<F, D>>` to `CircuitBuiler`.

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: I'll implement that in a later PR.

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,456 @@
...
+        Ok(Self {
+            circuit_digest,
+            constants_sigmas_cap,
+        })
+    }
+}
+
+impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
+    /// Add verifier data and register it as public inputs.
+    /// WARNING: Do not register any public input after calling this!
```
dlubarov, line 93: Seems fine for now but is this something we could relax later? Maybe this function could return the start index of cyclic recursion PIs, and that start index could then be passed to `cyclic_recursion` and to `from_slice`.

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,456 @@
...
+    pub fn verifier_data_for_cyclic_recursion<C: GenericConfig<D, F = F>>(
```
dlubarov, line 94: Was the reason for splitting this off that this should be called once while `cyclic_recursion` may be called multiple times? Maybe we could instead give `CircuitBuilder` an `Option<VerifierCircuitTarget>` to remember if we already have these public inputs.

Just in the interest of making this as simple as possible to use, it seems ideal if `cyclic_recursion` is the only method external users need to call.

dlubarov (approved):
Thanks for the changes, it basically looks good to me now. Besides the `goal_common_data` comment, the others are minor things we can tweak later.

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,456 @@
...
+        C::Hasher: AlgebraicHasher<F>,
+        [(); C::Hasher::HASH_SIZE]:,
+    {
+        let dummy_verifier_data = VerifierCircuitTarget {
+            constants_sigmas_cap: self.add_virtual_cap(self.config.fri_config.cap_height),
+            circuit_digest: self.add_virtual_hash(),
+        };
+
+        // Flag set to true for the base case of the cycle where we verify a dummy proof to bootstrap the cycle. Set to false otherwise.
+        let base_case = self.add_virtual_bool_target_safe();
```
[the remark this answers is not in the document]
wborgeaud: Sounds good!

plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -0,0 +1,456 @@
...
+        verifier_data
+    }
+
+    /// Cyclic recursion gadget.
+    /// WARNING: Do not register any public input after calling this!
+    pub fn cyclic_recursion<C: GenericConfig<D, F = F>>(
+        &mut self,
+        previous_virtual_public_inputs: &[Target],
+        verifier_data: &VerifierCircuitTarget, // should be registered as public inputs already
+        common_data: &CommonCircuitData<F, D>,
```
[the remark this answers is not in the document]
wborgeaud: Oops yes of course

plonky2/src/recursion/cyclic_recursion.rs
wborgeaud: Yes it should be easy to relax if needed. I thought it wouldn't be too annoying like that since the recursion gadget could typically be called at the end of the circuit, but if it becomes annoying this can be changed as you suggested.

PR #789, opened 2022-10-18
Implement SUBMOD instruction

unzvfu:
Implementation of the **non-standard** `SUBMOD` instruction in the arithmetic table.

Does the obvious thing: `SUBMOD(a, b, m) := (a - b) % m`, except that the result is always in the range `[0, m)`. (Cf. typical definitions of `%` which produce a negative result if `a-b` is negative.)


nbgl (commented):
Looks good but `eval_ext_circuit` needs to be updated to handle `IS_SUBMOD`.

unzvfu:
> Looks good but `eval_ext_circuit` needs to be updated to handle `IS_SUBMOD`.

Oops, good catch!

nbgl (approved):

PR #791, opened 2022-10-20
Implement MSIZE in the kernel

wborgeaud:


evm/src/cpu/kernel/asm/memory/core.asm
```diff
@@ -393,3 +393,21 @@
...
+%endmacro
+
+%macro mstore_main
+    // stack: offset, value
+    DUP1
+    // stack: offset, offset, value
+    %update_msize
+    // stack: offset, value
+    %mstore_current(@SEGMENT_MAIN_MEMORY)
+%endmacro
```
wborgeaud, line 413: One drawback of this approach is that we need to make sure we always use these macros when touching main memory.

dlubarov: I think it seems fine, we can use these to implement syscalls for MLOAD/MSTORE, and then for the few other opcodes that affect memory size (like Keccak) we can call `update_msize` directly

evm/src/cpu/kernel/asm/memory/metadata.asm
```diff
@@ -45,3 +45,23 @@
 %macro callvalue
     %mload_context_metadata(@CTX_METADATA_CALL_VALUE)
 %endmacro
+
+%macro msize
+    %mload_context_metadata(@CTX_METADATA_MSIZE)
+%endmacro
+
+%macro update_msize
```
dlubarov, line 53: I think we should generalize this a bit, to support `MSTORE8` which uses +1 instead of +32, and a few instructions like `RETURN` expand memory with a variable length.

In particular I think it'd be best to have `update_msize` take the (byte) length as a separate parameter, like that `M(s, f, l)` function from the yellowpaper (page 29), so that we can avoid expanding memory if length=0 as `M(s, f, l)` does.

dlubarov (approved):
LGTM otherwise

evm/src/cpu/kernel/asm/memory/metadata.asm
wborgeaud: So this is actually the +1 version, not the +32. It computes `ceil_div_usize(offset+1, 32) = (offset + 32)/32`. I'm not sure we need the +32 version since we need to mload byte by byte.

For the `M(s, f, l)` function, I'm not sure we need to implement it since we get the same behavior by doing
```
for i in range(f,l):
  %mload_main(f+i)
  ...
```
I assumed all the opcodes touching main memory (and using the `M(s, f, l)` function to update MSIZE) would use such a loop.

dlubarov: Ah I see. I think calling having a "batch update" could be a little faster (so e.g. we only write to `CTX_METADATA_MSIZE` once), but what you mentioned sounds fine too, we can always optimize later

PR #792, opened 2022-10-24
Bumped `eth_trie_utils` to `0.3.0`

BGluth:


BGluth:
It looks like the newest version of `clippy` introduces a lint to convert string formatting that is like this:
```rust
format!("a: {}, b: {}", a, b)
```
Into this format:
```rust
format!("{a}, {b}")
```
Which I didn't even know was an option until now.

dlubarov (approved):
LGTM

PR #795, opened 2022-10-26
Bounded bignum ASM

npwardberkeley:
Sorry, it's pretty messy 😅 Any and all feedback welcomed!

### DIFF
--- a/evm/src/cpu/kernel/asm/bignum/add.asm
+++ b/evm/src/cpu/kernel/asm/bignum/add.asm
@@ -49,39 +49,8 @@ add_loop:
     %jumpi(add_loop)
 add_end:
     // stack: carry_new, i - 1, a_cur_loc + 1, b_cur_loc + 1, retdest
-    %stack (c, i, a, b) -> (c, a)
-    // stack: carry_new, a_cur_loc + 1, retdest
-    // If carry = 0, no need to increment.
-    ISZERO
-    %jumpi(increment_end)
-increment_loop:
-    // stack: cur_loc, retdest
-    DUP1
-    %mload_kernel_general
-    // stack: val, cur_loc, retdest
-    %increment
-    // stack: val+1, cur_loc, retdest
+    %stack (c, i, a, b) -> (c)
+    // stack: carry_new, retdest
     SWAP1
-    DUP2
-    DUP2
-    // stack: cur_loc, val+1, cur_loc, val+1, retdest
-    %mstore_kernel_general
-    // stack: cur_loc, val+1, retdest
-    %increment
-    // stack: cur_loc + 1, val+1, retdest
-    SWAP1
-    // stack: val+1, cur_loc + 1, retdest
-    %eq_const(@BIGNUM_LIMB_BASE)
-    ISZERO
-    %jumpi(increment_end)
-    // stack: cur_loc + 1, retdest
-    PUSH 0
-    DUP2
-    // stack: cur_loc + 1, 0, cur_loc + 1, retdest
-    %mstore_kernel_general
-    %jump(increment_loop)
-increment_end:
-    // cur_loc, retdest
-    POP
-    // retdest
+    // stack: retdest, carry_new
     JUMP
--- a/evm/src/cpu/kernel/asm/bignum/add.asm
+++ b/evm/src/cpu/kernel/asm/bignum/add.asm
@@ -4,53 +4,54 @@
 // Replaces a with a + b, leaving b unchanged.
 global add_bignum:
     // stack: len, a_start_loc, b_start_loc, retdest
-    %stack (l, a, b) -> (0, 0, a, b, l)
-    // stack: carry=0, i=0, a_start_loc, b_start_loc, len, retdest
+    PUSH 0
+    // stack: carry=0, i=len, a_start_loc, b_start_loc, retdest
 add_loop:
-    // stack: carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: carry, i, a_cur_loc, b_cur_loc, retdest
     DUP4
     %mload_kernel_general
-    // stack: b[i], carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: b[cur], carry, i, a_cur_loc, b_cur_loc, retdest
     DUP4
     %mload_kernel_general
-    // stack: a[i], b[i], carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: a[cur], b[cur], carry, i, a_cur_loc, b_cur_loc, retdest
     ADD
     ADD
-    // stack: a[i] + b[i] + carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: a[cur] + b[cur] + carry, i, a_cur_loc, b_cur_loc, retdest
     %stack (val) -> (val, @BIGNUM_LIMB_BASE, @BIGNUM_LIMB_BASE, val)
-    // stack: a[i] + b[i] + carry, 2^128, 2^128, a[i] + b[i] + carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: a[cur] + b[cur] + carry, 2^128, 2^128, a[cur] + b[cur] + carry, i, a_cur_loc, b_cur_loc, retdest
     DIV
-    // stack: (a[i] + b[i] + carry) // 2^128, 2^128, a[i] + b[i] + carry, i, a_i_loc, b_i_loc, len, retdest
+    // stack: (a[cur] + b[cur] + carry) // 2^128, 2^128, a[cur] + b[cur] + carry, i, a_cur_loc, b_cur_loc, retdest
     SWAP2
-    // stack: a[i] + b[i] + carry, 2^128, (a[i] + b[i] + carry) // 2^128, i, a_i_loc, b_i_loc, len, retdest
+    // stack: a[cur] + b[cur] + carry, 2^128, (a[cur] + b[cur] + carry) // 2^128, i, a_cur_loc, b_cur_loc, retdest
     MOD
-    // stack: c[i] = (a[i] + b[i] + carry) % 2^128, carry_new = (a[i] + b[i] + carry) // 2^128, i, a_i_loc, b_i_loc, len, retdest
+    // stack: c[cur] = (a[cur] + b[cur] + carry) % 2^128, carry_new = (a[cur] + b[cur] + carry) // 2^128, i, a_cur_loc, b_cur_loc, retdest
     DUP4
-    // stack: a_i_loc, c[i], carry_new, i, a_i_loc, b_i_loc, len, retdest
+    // stack: a_cur_loc, c[cur], carry_new, i, a_cur_loc, b_cur_loc, retdest
     %mstore_kernel_general
-    // stack: carry_new, i, a_i_loc, b_i_loc, len, retdest
+    // stack: carry_new, i, a_cur_loc, b_cur_loc, retdest
     %stack (c, i, a, b) -> (a, b, c, i)
-    // stack: a_i_loc, b_i_loc, carry_new, i, len, retdest
+    // stack: a_cur_loc, b_cur_loc, carry_new, i, retdest
     %increment
+    // stack: a_cur_loc + 1, b_cur_loc, carry_new, i, retdest
     SWAP1
+    // stack: b_cur_loc, a_cur_loc + 1, carry_new, i, retdest
     %increment
+    // stack: b_cur_loc + 1, a_cur_loc + 1, carry_new, i, retdest
+    %stack (b, a, c, i) -> (i, c, a, b)
+    // stack: i, carry_new, a_cur_loc + 1, b_cur_loc + 1, retdest
+    %decrement
+    // stack: i - 1, carry_new, a_cur_loc + 1, b_cur_loc + 1, retdest
     SWAP1
-    %stack (a, b, c, i) -> (c, i, a, b)
-    // stack: carry_new, i, a_i_loc + 1, b_i_loc + 1, len, retdest
-    SWAP1
-    %increment
-    SWAP1
-    // stack: carry_new, i + 1, a_i_loc + 1, b_i_loc + 1, len, retdest
-    DUP5
-    DUP3
-    // stack: i + 1, len, carry_new, i + 1, a_i_loc + 1, b_i_loc + 1, len, retdest
-    EQ
+    // stack: carry_new, i - 1, a_cur_loc + 1, b_cur_loc + 1, retdest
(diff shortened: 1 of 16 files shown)


evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+increment_loop:
+    // stack: cur_loc, retdest
+    DUP1
+    %mload_kernel_general
+    // stack: val, cur_loc, retdest
+    %increment
+    // stack: val+1, cur_loc, retdest
+    %stack (v, l) -> (l, v, l, v)
+    DUP2
+    // stack: cur_loc, val+1, cur_loc, val+1, retdest
```
nbgl, line 163: This doesn’t make sense. If the stack is `cur_loc, val+1, cur_loc, val+1` and then you do a `DUP2`, then the new stack is `val+1, cur_loc, val+1, cur_loc, val+1`.

nbgl: This edge case should have a unit test.

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    %mstore_kernel_general
+    // stack: cur_loc, val+1, retdest
+    %decrement
+    // stack: cur_loc - 1, val+1, retdest
+    SWAP1
+    // stack: val+1, cur_loc - 1, retdest
+    %eq_const(256)
+    NOT
```
nbgl, line 171: You want a logical not, but `NOT` is bitwise not.

```suggestion
    ISZERO
```

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    // stack: cur_loc, retdest
+    DUP1
+    %mload_kernel_general
+    // stack: val, cur_loc, retdest
+    %increment
+    // stack: val+1, cur_loc, retdest
+    %stack (v, l) -> (l, v, l, v)
+    DUP2
+    // stack: cur_loc, val+1, cur_loc, val+1, retdest
+    %mstore_kernel_general
```
nbgl, line 164: Need to make sure it’s a valid limb (`& 0xff`) before storing. This could be storing `256`, which breaks the 8-bit limb assumption.

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    SWAP1
+    // stack: val+1, cur_loc - 1, retdest
+    %eq_const(256)
+    NOT
+    %jumpi(increment_end)
+    // stack: cur_loc - 1, retdest
+    PUSH 0
+    DUP2
+    // stack: cur_loc - 1, 0, cur_loc - 1, retdest
+    %mstore_kernel_general
```
nbgl, line 177: Is this to make sure that every stored limb is 8 bits? If so, should the 0 be stored to `cur_loc`, not `cur_loc - 1`?

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    %jump(increment_loop)
```
nbgl, line 178: What happens if you run out of space in `a`? For example, if `a_len == 1` and `a == 255` and `b == 1`. Depending on what the memory around `a` is initialized to, this code could never terminate.

evm/src/cpu/kernel/tests/bignum.rs
```diff
@@ -0,0 +1,161 @@
...
+    };
+    while x > max {
+        x = {
+            let limbs = gen_u256_limbs(&mut rng, num_bits);
+            U256(limbs)
+        };
+    }
+    x
+}
+
```
nbgl, line 45: These tests are a good start, but there’s a _lot_ of edge cases that are not tested.

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+
+%macro subtract_limb
+    // stack: a_i, b_i, borrow
+    DUP3
+    DUP2
+    SUB
+    // stack: a_i - borrow, a_i, b_i, borrow
+    DUP3
+    // stack: b_i, a_i - borrow, a_i, b_i, borrow
+    GT
```
nbgl, line 193: This doesn’t work. Consider: `a_i = 0`, `borrow = 1`. A borrow clearly occurs, but `a_i - borrow == 2**256 - 1` due to underflow, so `borrow_new == 0` for all values of `b_i`.

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    // stack: cur_loc - 1, 0, cur_loc - 1, retdest
+    %mstore_kernel_general
+    %jump(increment_loop)
+increment_end:
+    // cur_loc, retdest
+    POP
+    // retdest
+    JUMP
+
+%macro subtract_limb
```
nbgl, line 185: This is simpler:
```
t = a_i - b_i - borrow
c_i = t % 256
borrow_new = t >> 255
```

nbgl (commented):
This is a good start, but I spot two major issues.

Firstly, the testing is very basic, and does not cover a lot of tricky, but important, edge cases.

Secondly, we really cannot do this with 8-bit limbs. Compared to 128-bit limbs, it takes 16x as long to do an addition or subtraction. But it takes 256x as long to do a multiplication (!) because long multiplication is quadratic time. EVM is a 256-bit machine; we need to use that.

evm/src/cpu/kernel/asm/bignum.asm
```diff
@@ -0,0 +1,296 @@
...
+    // stack: a_len-i-1 == 0, b_len-i-1 == 0, a_len-i-1, b_len-i-1, a_i_loc+1, b_i_loc+1, retdest
+    %jumpi(less_at_end)
+    %jumpi(greater_at_end)
+    %jump(eq_loop)
+greater:
+    // stack: a[i], b[i], a_len, a_0_loc, b_len, b_0_loc, retdest
+    %stack (all: 6) -> (1)
+    // stack: 1, retdest
+    SWAP1
+    JUMP
```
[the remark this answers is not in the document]
npwardberkeley: Hm for point 1, would it be better to have an invariant that the highest limbs are not zero, and to ensure that at the end of `sub_bignum` and similar operations?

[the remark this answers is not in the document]
npwardberkeley: Or just to change how this function works?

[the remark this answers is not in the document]
nbgl: I wouldn’t try to enforce such an invariant. It sounds really messy keeping track of it. (For example, what happens if the subtraction creates a bunch of leading zeros? You’d have to make the subtract function return a length.)


Ideally, the function should be correct even if there are leading zeros.

evm/src/cpu/kernel/asm/bignum/ge.asm
```diff
@@ -0,0 +1,80 @@
...
+    // stack: a[i], b[i], i, a_i_loc, b_i_loc, retdest
+    %stack (vals: 2) -> (vals, vals)
+    GT
+    %jumpi(greater)
+    // stack: a[i], b[i], i, a_i_loc, b_i_loc, retdest
+    LT
+    %jumpi(less)
+    // stack: i, a_i_loc, b_i_loc, retdest
+    DUP1
+    %eq_const(0)
```
unzvfu, line 42:
```suggestion
    ISZERO
```
Don't need to save `i` here as it's just thrown away after the jump anyway.

unzvfu (commented):
So I've just had a look through `add.asm`, `ge.asm` and `mul.asm`. Overall it's looking pretty good! I did note a few places could do with some attention; see the inline comments.

I do want to make some general comments that you might want to keep in mind if this code turns out to be important for performance, but you can choose to ignore these for the time being in the interest of getting a working version merged sooner if you like.

The main observation is that good performance really requires 256-bit limbs. This will reduce the number of operations by about half, and will simplify several parts of the code too. The only tricky aspects are (i) using `LT` to get the carry after an addition, and (ii) implementing 256x256 -> 512bit multiplication. For the former you can benefit from implicit overflow of u256's: `sum = a + b + cy_in; cy_out = sum < a;` (note that, because `cy_in` is 0 or 1, we know that the expression for `sum` can't overflow twice).

For the latter, one splits each 256-bit input `a` into `a = a_0 + a_1 * 2^128` where `a_0` and `a_1` are 128 bits, then does the four 128x128 -> 256-bit multiplications and reconstruct the high and low 256 bits of the result. This might sound convoluted, but it's still a bit cheaper than your approach with 128-bit limbs, and it simplifies the surrounding code quite a lot.

Otherwise, note that `DIV` and `MOD` are typically quite expensive (less so in the EVM than a CPU, but for us they still take two rows in the arithmetic table). I suspect there might be an advantage to using e.g. `SHR` for `DIV` and maybe `SHL; SHR` for `MOD` (this latter also requires 256-bit limbs).

Finally, it should be possible (albeit with a bit of a restructure), and probably simpler, to avoid the use of scratch space in `mul_bignum`; instead you can build the result directly in the output memory. This would save most of the scratch space manipulation, including the memcpy call every iteration in `mul_bignum`. It would also simplify the `modmul_bignum` code significantly.

evm/src/cpu/kernel/asm/bignum/add.asm
```diff
@@ -0,0 +1,87 @@
...
+    // stack: val+1, cur_loc + 1, retdest
+    %eq_const(@BIGNUM_LIMB_BASE)
+    ISZERO
+    %jumpi(increment_end)
+    // stack: cur_loc + 1, retdest
+    PUSH 0
+    DUP2
+    // stack: cur_loc + 1, 0, cur_loc + 1, retdest
+    %mstore_kernel_general
+    %jump(increment_loop)
```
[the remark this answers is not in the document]
npwardberkeley: Ah, well the `%eq_const(@BIGNUM_LIMB_BASE)` is what checks for overflow, and in that case it replaces it with zero and then keeps doing the `increment` loop (that is, keeps on carrying). So in theory it can keep on looping, but only if each following limb has the maximum value of 2^128 - 1

I think it's correct, but I might have missed something; but either way, your idea of changing the interface sounds good! I'll discuss with Daniel

evm/src/cpu/kernel/asm/bignum/mul.asm
```diff
@@ -0,0 +1,190 @@
...
+    %decrement
+    // stack: n-1, a_start_loc, bi+1, output_cur+1, len, n, a_start_loc, bi, output_cur, scratch_space, retdest
+    %stack (new: 4, len, old: 4) -> (len, new)
+    // stack: len, n-1, a_start_loc, bi+1, output_cur+1, scratch_space, retdest
+    DUP2
+    // stack: n-1, len, n-1, a_start_loc, bi+1, output_cur+1, scratch_space, retdest
+    ISZERO
+    %jumpi(mul_end)
+    %jump(mul_loop)
+mul_end:
```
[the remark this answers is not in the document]
npwardberkeley: Oh clever!

evm/src/cpu/kernel/asm/bignum/add.asm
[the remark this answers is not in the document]
unzvfu: Hmm, consider the case `a` and `b` have length 1 and `a + b >= 2^128` so the carry is 1. It looks to me that on lines 58--60 `cur_loc = a + 1` which is already an invalid address, right?

Even if it isn't an invalid address, in general the value `val` read on line 60 could be anything up to `2^128 - 1`; if it's the maximum, then the `%increment` on line 62 causes `val` to be 2^128 and this is then stored at `cur_loc` on line 68. But *limbs must be strictly less than 2^128 to be properly normalised.*

This also means the test on line 74 is not correct; it's true that `val + 1 == BIGNUM_LIMB_BASE` indicates overflow, but the overflowing value `val + 1` has already been stored and `cur_loc` has been incremented to point to the next limb.

So I think this code creates invalid (or at least not properly normalised) bignums in the case where the addition overflows. I think it would be much easier to delete the whole `increment_loop` section and just return the carry, but if you want to stick to this approach you'll need to (i) ensure that `a` has space `len + 1` so there's space for the carry (this already creates the problem of different length inputs) and (ii) the increment loop isn't necessary anyway since if `a` and `b` start with the same length then they can at most overflow by a single bit, so the loop should never run more than one iteration (though this isn't true if the inputs have different lengths).

Hope this is clear; let me know if I've misunderstood anything or made a mistake somewhere! :)

evm/src/cpu/kernel/asm/bignum/modmul.asm
```diff
@@ -0,0 +1,168 @@
+// Arithmetic on little-endian integers represented with 128-bit limbs.
+// All integers must be under a given length bound, and are padded with leading zeroes.
+
+// Stores a * b % m in output_loc, leaving a, b, and m unchanged.
+// a, b, and m must have the same length.
+// Both output_loc and scratch_1 must have size length.
+// All of scratch_2, scratch_3, and scratch_4 must have size 2 * length and be initialized with zeroes.
```
unzvfu, line 7: I'm not sure whether this would be an improvement overall, but note that you can remove the need for additional prover inputs by using [Barrett reduction](https://en.wikipedia.org/wiki/Barrett_reduction) which uses a precomputed "pseudo-inverse" of the modulus. The pseudo-inverse can be easily computed (once at the start of the `MODEXP`) using [the trick described here](https://medium.com/wicketh/mathemagic-512-bit-division-in-solidity-afa55870a65).

evm/src/cpu/kernel/asm/bignum/ge.asm
npwardberkeley: Ah well don't I still need to `DUP` it, in case the loop continues?

unzvfu: Oh yes, good catch! :)

evm/src/cpu/kernel/asm/bignum/add.asm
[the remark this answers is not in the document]
npwardberkeley: Ohh I see, thanks! Ah the point of storing the 0 was to overwrite the overflowed 2^128 value, and then increment the next one; but I incorrectly store the 0 in `cur_loc + 1` instead of `cur_loc`. But you're right that the loop isn't necessary; I'll change it to return the carry bit!

npwardberkeley:
Going to close this and remake as two separate PRs!

PR #796, opened 2022-10-27
Implement `EXTCODEHASH, (EXT)CODESIZE, (EXT)CODECOPY`

wborgeaud:


evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -0,0 +1,130 @@
+%macro extcodehash
+    // stack: address
+    %mpt_read_state_trie
+    // stack: account_ptr
```
dlubarov, line 4: Similar to the comment in #797, I think we need to handle `account_ptr == null` separately. I think we can return 0, that would match `EXTCODEHASH` semantics.

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -0,0 +1,130 @@
...
+    // stack: i+1, code_length, codehash, retdest
+    %jump(load_code_loop)
+
+// Check that the hash of the loaded code equals `codehash`.
+load_code_check:
+    // stack: i, code_length, codehash, retdest
+    POP
+    // stack: code_length, codehash, retdest
+    %stack (code_length, codehash, retdest) -> (@SEGMENT_KERNEL_ACCOUNT_CODE, 0, code_length, codehash, retdest, code_length)
+    GET_CONTEXT
```
dlubarov, line 125: nit: I'd suggest using context=0 for a couple very minor reasons
- Seems slightly convenient since we can include 0 in `%stack`
- Arguably it might seem more conventional for the kernel to use "its own" memory, rather than memory "owned by" a userspace call. (Admittedly we already violate this guideline with `ContextMetadata`, but most other kernel stuff uses context=0.)

evm/src/cpu/kernel/asm/account_code.asm
dlubarov: (Unless you plan to check for account existence outside of this? If so let's comment that it assumes the account exists.)

dlubarov (approved):
LGTM

evm/src/cpu/kernel/asm/account_code.asm
wborgeaud: Good points, I changed it

PR #797, opened 2022-10-28
Implement `(SELF)BALANCE`

wborgeaud:


evm/src/cpu/kernel/tests/balance.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+// Stolen from `tests/mpt/insert.rs`
+// Prepare the interpreter by inserting the account in the state trie.
+fn prepare_interpreter(
+    interpreter: &mut Interpreter,
+    address: Address,
+    account: &AccountRlp,
+) -> Result<()> {
```
wborgeaud, line 29: This is the same function as in #796. We can make it public to avoid redundancy.

evm/src/cpu/kernel/asm/balance.asm
```diff
@@ -0,0 +1,21 @@
+%macro balance
+    // stack: address
+    %mpt_read_state_trie
+    // stack: account_ptr
```
dlubarov, line 4: I think we need a special case for `account_ptr == null`, otherwise if a bad prover failed to provide this account's Merkle data, we'll read whatever is at `memory[trie][1]` which could lead to security concerns.

For comparison `%current_account_data` panics if `account_ptr == null`, since the current account should always be in the trie. With `%balance`, I guess it's valid to call it on an account not in the trie, so it might make sense to return 0 rather than panicking?

dlubarov (approved):
LGTM otherwise

evm/src/cpu/kernel/asm/balance.asm
wborgeaud: Good point. I think panicking if `account_ptr == null` makes more sense, since this happens only when the prover is malicious/buggy. Unless we want to optimize calls of `BALANCE` to empty account?

dlubarov: Hm I was assuming the code for our `BALANCE` syscall just be `%balance`; if so I think we'd need to return 0.

Were you thinking the syscall would check whether the account exists before doing `%balance`? That works too.

dlubarov: Also my comment was not quite right, on second thought if a prover fails to provide the relevant trie data, `mpt_read` should panic (upon trying to read from a hash node).

So I guess there's no concern about malicious provers, just `account_ptr` could be null if the account truly doesn't exist.

wborgeaud: Ah ok sounds good, I changed it to return 0 if `accout_ptr == null`.

PR #798, opened 2022-10-31
Use address macro instead of opcode

wborgeaud:
Unless I'm missing something, `ADDRESS` is not a native instruction


dlubarov (approved):
I think `ADDRESS` could also work, it's not quite implemented but I imagined it would lead to a syscall which internally does `%address`. Might be good to use `%address` directly to avoid the syscall overhead 👍

PR #800, opened 2022-10-31
Constrain `mem_channel.used` flag to be binary

nbgl:
This was missing but is required for correctness (it is used as a filter in a CTL).


dlubarov (approved):
LGTM

PR #801, opened 2022-11-02
EVM shift left/right operations

unzvfu:
Code to verify shift left/right instructions on the EVM. Doesn't include generation code, that will come later once some upcoming changes from @nbgl are merged. I think the lack of generation code is probably the cause of the failing test. It is not currently connected to the arithmetic table, but that will be handled in a follow-up PR.

### DIFF
--- a/evm/src/cpu/columns/general.rs
+++ b/evm/src/cpu/columns/general.rs
@@ -9,6 +9,7 @@ pub(crate) union CpuGeneralColumnsView<T: Copy> {
     arithmetic: CpuArithmeticView<T>,
     logic: CpuLogicView<T>,
     jumps: CpuJumpsView<T>,
+    shift: CpuShiftView<T>,
 }
 
 impl<T: Copy> CpuGeneralColumnsView<T> {
@@ -50,6 +51,16 @@ impl<T: Copy> CpuGeneralColumnsView<T> {
     // SAFETY: Each view is a valid interpretation of the underlying array.
     pub(crate) fn jumps_mut(&mut self) -> &mut CpuJumpsView<T> {
         unsafe { &mut self.jumps }
+    }
+
+    // SAFETY: Each view is a valid interpretation of the underlying array.
+    pub(crate) fn shift(&self) -> &CpuShiftView<T> {
+        unsafe { &self.shift }
+    }
+
+    // SAFETY: Each view is a valid interpretation of the underlying array.
+    pub(crate) fn shift_mut(&mut self) -> &mut CpuShiftView<T> {
+        unsafe { &mut self.shift }
     }
 }
 
@@ -142,6 +153,13 @@ pub(crate) struct CpuJumpsView<T: Copy> {
     /// (`input0[0]` is not `JUMPDEST` that is not in an immediate while we are in user mode, or
     /// `input0[1..7]` is nonzero) and `input1` is nonzero.
     pub(crate) should_trap: T,
+}
+
+#[derive(Copy, Clone)]
+pub(crate) struct CpuShiftView<T: Copy> {
+    // 1 if the high limbs of the displacement are all zero, 0 if one
+    // or more of the limbs is nonzero.
+    pub(crate) displacement_high_limbs_are_zero: T,
 }
 
 // `u8` is guaranteed to have a `size_of` of 1.
--- a/evm/src/cpu/columns/general.rs
+++ b/evm/src/cpu/columns/general.rs
@@ -160,6 +160,9 @@ pub(crate) struct CpuShiftView<T: Copy> {
     // 1 if the high limbs of the displacement are all zero, 0 if one
     // or more of the limbs is nonzero.
     pub(crate) displacement_high_limbs_are_zero: T,
+    // For a shift amount of displacement: [T], this is the inverse of
+    // sum(displacement[1..]) or zero if the sum is zero.
+    pub(crate) high_limb_sum_inv: T,
 }
 
 // `u8` is guaranteed to have a `size_of` of 1.
--- a/evm/src/cpu/columns/general.rs
+++ b/evm/src/cpu/columns/general.rs
@@ -157,9 +157,6 @@ pub(crate) struct CpuJumpsView<T: Copy> {
 
 #[derive(Copy, Clone)]
 pub(crate) struct CpuShiftView<T: Copy> {
-    // 1 if the high limbs of the displacement are all zero, 0 if one
-    // or more of the limbs is nonzero.
-    pub(crate) displacement_high_limbs_are_zero: T,
     // For a shift amount of displacement: [T], this is the inverse of
     // sum(displacement[1..]) or zero if the sum is zero.
     pub(crate) high_limb_sum_inv: T,
--- a/evm/src/cpu/kernel/asm/shift.asm
+++ b/evm/src/cpu/kernel/asm/shift.asm
@@ -2,24 +2,24 @@
 ///
 /// Specifically, set SHIFT_TABLE_SEGMENT[i] = 2^i for i = 0..255.
 %macro shift_table_init
-    push 1   // 2^0
-    push 0   // initial offset is zero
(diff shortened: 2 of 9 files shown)


evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,139 @@
...
+    yield_constr.constraint(small_disp_filter * two_exp.addr_context); // kernel mode only
+    yield_constr.constraint(small_disp_filter * (two_exp.addr_segment - shift_table_segment));
+    yield_constr.constraint(small_disp_filter * (two_exp.addr_virtual - displacement.value[0]));
+
+    // When the shift displacement is >= 2^32, constrain two_exp.value
+    // to be zero.
+    let large_disp_filter = is_shift * (P::ONES - high_limbs_are_zero);
+    for &limb in &two_exp.value {
+        yield_constr.constraint(large_disp_filter * limb);
+    }
```
nbgl, line 54: This doesn’t quite work. The `two_exp` memory channel is always used, so if you constrain its `value` to zero, then it might not match the contents of memory (so you can’t satisfy both constraints).

You can fix it by making `two_exp.used` iff `high_limbs_are_zero`:
```rust
yield_constr.constraint(is_shift * (two_exp.used - high_limbs_are_zero))
```

evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,139 @@
...
+    yield_constr.constraint(is_shift * (displacement.used - P::ONES));
+    yield_constr.constraint(is_shift * (displacement.is_read - P::ONES));
+    yield_constr.constraint(is_shift * (output.used - P::ONES));
+    yield_constr.constraint(is_shift * (output.is_read - P::ONES));
+    yield_constr.constraint(is_shift * (two_exp.used - P::ONES));
+    yield_constr.constraint(is_shift * (two_exp.is_read - P::ONES));
+
+    let high_limbs_are_zero = lv.general.shift().displacement_high_limbs_are_zero;
+    yield_constr
+        .constraint(is_shift * (high_limbs_are_zero - high_limbs_are_zero * high_limbs_are_zero));
```
nbgl, line 36: It looks like the CI is failing because `eval_packed` does not fully match `eval_ext_circuit`.

In particular, there’s a sign error: `eval packed` calculates `high_limbs_are_zero - high_limbs_are_zero * high_limbs_are_zero`, whereas `eval_ext_circuit` does `builder.mul_sub_extension(high_limbs_are_zero, high_limbs_are_zero, high_limbs_are_zero)`, which is `high_limbs_are_zero * high_limbs_are_zero - high_limbs_are_zero`.

```suggestion
    yield_constr
        .constraint(is_shift * (high_limbs_are_zero * high_limbs_are_zero - high_limbs_are_zero));
```

evm/src/cpu/kernel/asm/shift.asm
```diff
@@ -0,0 +1,25 @@
...
+        // stack: ost_(i+1), 2^i
+        swap1
+        // stack: 2^i, ost_(i+1)
+        %mul_const(2)
+        // stack: 2^(i+1), ost_(i+1)
+        swap1
+        // stack: ost_(i+1), 2^(i+1)
+    %endrep
+    %pop2
+%endmacro
```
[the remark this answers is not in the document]
unzvfu: Oh, that's neat! I still need to get used to having an almost arbitrarily large stack at my disposal. :)

evm/src/cpu/shift.rs
unzvfu: Oops, good catch!

evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    let high_limbs_sum: P = displacement.value[1..].iter().copied().sum();
+    yield_constr.constraint(is_shift * high_limbs_sum * high_limbs_are_zero);
```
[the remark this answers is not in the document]
unzvfu: Oops, yep, good catch!

evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    let shift_table_segment = P::Scalar::from_canonical_u64(Segment::ShiftTable as u64);
+
+    let high_limbs_are_zero = lv.general.shift().displacement_high_limbs_are_zero;
+    yield_constr
+        .constraint(is_shift * (high_limbs_are_zero * high_limbs_are_zero - high_limbs_are_zero));
+
+    // Only lookup the shifting factor when displacement is < 2^32
+    yield_constr.constraint(is_shift * (two_exp.used - high_limbs_are_zero));
+    yield_constr.constraint(is_shift * (two_exp.is_read - high_limbs_are_zero));
```
nbgl, line 33: Nit: it would be slightly simpler to just `two_exp.is_read` to one regardless of `high_limbs_are_zero`. The value doesn’t matter when `two_exp.used` is zero.
```suggestion
    yield_constr.constraint(is_shift * (two_exp.is_read - P::ONES));
```

evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,123 @@
...
+    yield_constr.constraint(is_shift * t);
+    yield_constr.constraint(is_shift * high_limbs_sum * high_limbs_are_zero);
+
+    // When the shift displacement is < 2^32, constrain the two_exp
+    // mem_channel to be the entry corresponding to `displacement` in
+    // the shift table lookup (will be zero if displacement >= 256).
+    let small_disp_filter = is_shift * high_limbs_are_zero;
+    yield_constr.constraint(small_disp_filter * two_exp.addr_context); // read from kernel memory
+    yield_constr.constraint(small_disp_filter * (two_exp.addr_segment - shift_table_segment));
+    yield_constr.constraint(small_disp_filter * (two_exp.addr_virtual - displacement.value[0]));
```
nbgl, line 49: Same as `two_exp.is_read`: this doesn’t need to be conditioned on `high_limbs_are_zero` because the value is ignored when `two_exp.used == P::ZEROS`.
```suggestion
    yield_constr.constraint(is_shift * two_exp.addr_context); // read from kernel memory
    yield_constr.constraint(is_shift * (two_exp.addr_segment - shift_table_segment));
    yield_constr.constraint(is_shift * (two_exp.addr_virtual - displacement.value[0]));
```

evm/src/cpu/columns/general.rs
```diff
@@ -144,5 +155,15 @@ pub(crate) struct CpuJumpsView<T: Copy> {
     pub(crate) should_trap: T,
 }
 
+#[derive(Copy, Clone)]
+pub(crate) struct CpuShiftView<T: Copy> {
+    // 1 if the high limbs of the displacement are all zero, 0 if one
+    // or more of the limbs is nonzero.
+    pub(crate) displacement_high_limbs_are_zero: T,
```
nbgl, line 162: This is always equal to `two_exp.used`, so you could just use that and not need the equality constraint

evm/src/cpu/shift.rs
```diff
@@ -0,0 +1,123 @@
...
+    // Not needed here; val is the input and we're verifying that output is
+    // val * 2^d (mod 2^256)
+    //let val = lv.mem_channels[0];
+    //let output = lv.mem_channels[NUM_GP_CHANNELS - 1];
+
+    let shift_table_segment = P::Scalar::from_canonical_u64(Segment::ShiftTable as u64);
+
+    let high_limbs_are_zero = lv.general.shift().displacement_high_limbs_are_zero;
+    yield_constr
+        .constraint(is_shift * (high_limbs_are_zero * high_limbs_are_zero - high_limbs_are_zero));
```
nbgl, line 29: `two_exp.used` is already constrained to be a binary flag, so this is not needed (although if you’re going to remove it, then a comment explaining this might help).

nbgl (approved):
Looks good! Just a few nits, but I don’t feel strongly about any of them

PR #802, opened 2022-11-02
Combine all syscalls into one flag

nbgl:
This PR combines all syscalls into one flag, with the target read from one big jumptable.

To create that jumptable, it adds a `JUMPTABLE` directive to the assembler, which emits the address of a label (like `BYTES`, but friendlier to use).


evm/src/cpu/kernel/asm/core/syscall_stubs.asm
```diff
@@ -0,0 +1,54 @@
+// Labels for unimplemented syscalls to make the kernel assemble.
+// Each label should be removed from this file once it is implemented.
+
+global sys_sdiv:
+global sys_smod:
+global sys_signextend:
+global sys_slt:
+global sys_sgt:
+global sys_byte:
```
dlubarov, line 9: I think it would be good to have some native impl of `BYTE`, maybe using the logic table, as the kernel uses it quite a bit. But it's fine for now

nbgl: Sure! `BYTE` should not be difficult to implement. I’ll add a flag for it!

evm/src/cpu/syscalls.rs
```diff
@@ -2,62 +2,81 @@
...
+            cumul * P::Scalar::from_canonical_u64(256) + limb
+        });
+    yield_constr.constraint_transition(filter * (nv.program_counter - target));
+    // Set kernel mode
     yield_constr.constraint_transition(filter * (nv.is_kernel_mode - P::ONES));
+    // Maintain current context
+    yield_constr.constraint_transition(filter * (nv.context - lv.context));
 
+    // This memory channel is constrained in `stack.rs`.
     let output = lv.mem_channels[0].value;
```
dlubarov, line 74: Should the push be changed to be on the last channel now?

dlubarov (approved):

evm/src/cpu/syscalls.rs
nbgl: Yes, it should! Great catch!

PR #803, opened 2022-11-03
`FromTargets` trait

wborgeaud:
Tentative refactor introducing a new `FromTargets` trait with a fn to go from an iterator of `Target`s to `Self`.

Using this trait, functions like `builder.add_virtual_*` or `add_virtual_*(builder,...)` can be removed and replaced with a call to a new `builder.add_virtual()` function that infers the return type. 

As a next step, this can also be used to add functions like `builder.add_constant()` which would create a constant `StructTarget` from `Struct` or `builder.add_with_generators()` which would create a `StructTarget` set non-deterministically in the trace.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -174,10 +174,13 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         (0..n).map(|_i| self.add_virtual_target()).collect()
     }
 
-    pub fn add_virtual<'a, A: FromTargets<'a, F, D, Config = CircuitConfig>>(&mut self) -> A {
-        let len = A::len(&self.config);
+    pub fn add_virtual<'a, Config, A: FromTargets<'a, F, D, Config = Config>>(
+        &mut self,
+        c: &Config,
+    ) -> A {
+        let len = A::len(c);
         let mut targets = self.add_virtual_targets(len).into_iter();
-        A::from_targets(&mut targets, &self.config)
+        A::from_targets(&mut targets, c)
     }
 
     pub fn add_virtual_target_arr<const N: usize>(&mut self) -> [Target; N] {
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -42,6 +42,7 @@ use crate::plonk::permutation_argument::Forest;
 use crate::plonk::plonk_common::PlonkOracle;
 use crate::timed;
 use crate::util::context_tree::ContextTree;
+use crate::util::from_targets::FromTargets;
 use crate::util::partial_products::num_partial_products;
 use crate::util::timing::TimingTree;
 use crate::util::{transpose, transpose_poly_values};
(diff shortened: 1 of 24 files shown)


dlubarov:
This is really neat in a way, I'm just a bit concerned about the effects on readability. E.g. considering `builder.add_virtual_targets(1)` vs `builder.add_virtual(((), 1))`, I think the former is more natural/intuitive - it's more obvious what method is called, the argument has the expected form and name/type (`n: usize` vs `config: ...`), etc.

I also just feel since it adds extra code, like length calculations that we didn't need before, we should have a strong reason to add that extra code.

I think an `add_constant` API might be cleaner in some ways than `add_virtual`, if those conversions don't require a `Config`? So we wouldn't have odd, hard-to-explain args like `((), 1)`. Though it might still increase verbosity a bit, and make it slightly harder to tell what method is being invoked.

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -173,44 +172,13 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
         &mut self,
-        num_coeffs: usize,
-    ) -> PolynomialCoeffsExtTarget<D> {
-        let coeffs = self.add_virtual_extension_targets(num_coeffs);
-        PolynomialCoeffsExtTarget(coeffs)
+        c: Config,
+    ) -> A {
+        let len = A::len(c);
+        let mut targets = self.add_virtual_targets(len).into_iter();
+        A::from_targets(&mut targets, c)
```
dlubarov, line 181: If we do go this direction, should probably assert `targets` is empty, so any error in `len()` is detected

wborgeaud (commented):
>  I think an `add_constant` API might be cleaner in some ways than `add_virtual`, if those conversions don't require a `Config`?

Could you expand on that? Why would `add_constant` be different from `add_virtual`?

Overall I agree with you that the explicit length calculations are annoying. 
I think this API would make `add_constant` and `add_with_generators` easier to implement and would also allow for some nice cleanups. 
Like now we do `builder.add_virtual_*` in `plonky2` and `add_virtual_*(&mut builder,...)` outside `plonky2`. Same thing with `pw.set_*` and `set_*(&mut pw,...)`. With this API we could just standardize this to `builder.add_virtual()` and `pw.set()`. 
But yeah this is mostly aesthetic, so not sure it's worth the large refactor.

dlubarov:
Hmm maybe I didn't fully understand the idea. Could you explain how `constant` would work, like would there be a `ToTargets`, and then the impl is `T::from_targets(constant.to_targets)`?

Just from this first phase, I worry the API might become a little less user-friendly. E.g. considering `add_virtual_cap`, currently its param is named `cap_height`, but that would change to the generic `c`. On `add_virtual_cap` we could add a comment if we needed to clarify the parameter, but seems like we lose that option with a generic `add_virtual`. Users could always find the `FromTargets` impl to get more info, but that seems higher-friction.

It also just feels a little less natural IMO to wrap args in a tuple, especially e.g. `((), 1)`. Not an issue once we're used to it, but it'd be something we need to explain in a tutorial/example. And if users need to extend the API, that seems harder, mainly due to `len()`.

I'm probably not seeing the full picture with all the pros and cons though, since I don't have a good sense of what `add_constant()`/`add_with_generators()` will look like.

wborgeaud:
Yes for constants and generators we would need to add a `ToFieldElements` trait for native types that returns a vector of field elements. Then we'd be able to use it with the `FromTargets` (linking the two using associated types) for constants, generators.
I think that ideally it would be nice to merge the native and target structs (like we do already in some of the Stark code) to make things simpler and remove some redundancy. 

For the config, I agree with you it's a bit less clear with this PR. One way to fix it would be to use config structs, like `CapConfig { height: usize }` instead of just `usize`.

pgebheim:
Hello!

PR #806, opened 2022-11-04
Add no-std Support for Plonky2 and Starky

bhgomes:
This PR adds makes the `plonky2`, `starky`, and associated gadget libraries `no_std` compatible.

## Structure of Changes

For the most part, the changes involve only adding `no_std`-compatible versions of the existing data structures (and importing from `alloc`) and restructuring the feature-selection settings for each crate. However, in some places there are some more specific changes, mostly in the following two areas:

- Serialization in `plonky2`
- Inline Assembly Support in `plonky2_util`

### Serialization

For serialization, we noticed that `Cursor<Vec<u8>>` is used as the canonical reader/writer for `plonky2` objects. However, there is no `no_std` equivalent for `Cursor` without possibly importing a crate. Instead, the serialization methods on the `Buffer` type were generalized to a `Read` and `Write` trait. For writing, a `Vec<u8>` is enough, and we get the added benefit that it is a completely infallible operation so we can remove the error handling in that case. For reading, we still need a `Cursor<Vec<u8>>` so the `Buffer` object still implements `Read` and `Write`. In this case, `Buffer` was delegated to the `std`-only feature set. In future work we may consider finding a canonical writer that is `no-std` compatible.

## Inline Assembly

In `plonky2_util` inline assembly is used to provide a custom compiler hint:

https://github.com/mir-protocol/plonky2/blob/ea7fbed33a8d132ecc4dc92e9ce49639fdfe3753/util/src/lib.rs#L267-L273

To make this `no-std` compatible we add a target selection at compile-time that only selects the targets that support inline assembly as of the authoring of this PR.


dlubarov:
Thank you for the contribution!

Could you explain the motivation a bit? I think it would be interesting to have plonky2 running on targets like wasm, but my understanding is that no_std is not required there, just certain std methods are unimplemented. Similarly for Solana, my understanding is that no_std is not required, although certain things like std::net are missing.

Is there a use case that requires this, or do you see it as more of a best practice, to enforce certain compatibility guarantees?

bhgomes:
It's a bit of both.

Building crates as `no-std` first and then extending by `std` as an optional feature is the best practice (unless the purpose of the crate is to do IO, networking, ... etc.). Even though WASM may work with `std` in some cases, unless you know exactly which APIs are unsupported you will get spurious runtime errors (instead of compile-time ones). In resource-constrained environments, sandboxed environments, or environments that use a subset of WASM like in substrate-based blockchains, very strict adherence to `no-std` is actually required. Also, moving to a `no-std` compatible build makes the `plonky2` crate much friendlier to the rest of the Rust ecosystem by allowing other `no-std` crates to safely import code independent of whatever reasons they may have for being `no-std`.

One other benefit of moving to a minimal dependency implementation (I'm including the standard library as a dependency here) is that we can see exactly where we introduce a dependency and its usage scope.

plonky2/src/hash/hash_types.rs
```diff
@@ -35,24 +37,26 @@ impl<F: Field> HashOut<F> {
         elements[0..elements_in.len()].copy_from_slice(elements_in);
         Self { elements }
     }
+}
 
-    #[cfg(feature = "parallel")]
-    pub fn rand_from_rng<R: rand::Rng>(rng: &mut R) -> Self {
+impl<F> Sample for HashOut<F>
```
dlubarov, line 42: Argh, it would be nice if we could just use the standard `Rng` methods, like

```rust
let elems: [F; 10] = rng.gen();
```

But seems like the design of `rand` makes this impossible if `F` is generic? Since the `Field` trait can't impose a requirement that `Standard` impl `Distribution<Self>`, and we can't have a blanket impl `Distribution<F>`.

dlubarov:
Thanks for the additional context; seems reasonable to me.

For future changes I'd prefer if you could find a way to split things up, ideally with uninteresting changes (like the `alloc` imports) grouped together for ease of reviewing. I've been reading through this and it looks good so far, but still a ways to go. I'll try to finish fairly soon so we can get this in before a lot of conflicts come up.

plonky2/Cargo.toml
```diff
@@ -11,45 +11,46 @@ edition = "2021"
...
-rand_chacha = { version = "0.3.1", optional = true }
-maybe_rayon = { path = "../maybe_rayon" }
-unroll = "0.1.5"
-anyhow = "1.0.40"
-serde = { version = "1.0", features = ["derive"] }
-serde_cbor = "0.11.1"
-keccak-hash = "0.8.0"
-static_assertions = "1.1.0"
+anyhow = { version = "1.0.40", default-features = false }
+derivative = { version = "2.2.0", default-features = false, features = ["use_core"] }
```
dlubarov, line 22: It seems like we're not making that much use of derivative's features/flexibility? We generally like to keep our dependencies lean, so would prefer to avoid this if the benefits are minor.

plonky2/Cargo.toml
```diff
@@ -11,45 +11,46 @@ edition = "2021"
...
+hashbrown = { version = "0.12.3", default-features = false, features = ["ahash", "serde"] }
```
dlubarov, line 23: The lack of DOS resistance seems a bit concerning, but I think it should be okay... I can't think of a case where we have a potentially-huge `HashMap` with untrusted input. The largest `HashMap`s should be in `CircuitBuilder`, thankfully we no longer use `HashMap` in witness generation.

plonky2/src/plonk/proof.rs
```diff
@@ -98,12 +103,13 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
         C::InnerHasher::hash_no_pad(&self.public_inputs)
     }
 
-    pub fn to_bytes(&self) -> anyhow::Result<Vec<u8>> {
-        let mut buffer = Buffer::new(Vec::new());
-        buffer.write_proof_with_public_inputs(self)?;
-        Ok(buffer.bytes())
+    pub fn to_bytes(&self) -> Vec<u8> {
+        let mut buffer = Vec::new();
+        let _ = buffer.write_proof_with_public_inputs(self);
```
dlubarov, line 108: To make it more obvious that this is safe, I'd rather we call unwrap or expect.

Or since we only write to `Vec`s (directly or through a `Cursor`), maybe we should change `Write` to not return any errors? If so it could be a future change though.

plonky2/Cargo.toml
dlubarov: Why don't we enable `ahash-compile-time-rng` just in case? So at least users who compile their own binaries will be protected.

I wonder why it's not the default, maybe some users expect hashbrown's `HashMap` to be deterministic? But we certainly don't since std `HashMap` isn't.

plonky2/src/util/serialization.rs
```diff
@@ -20,63 +23,83 @@ use crate::plonk::proof::{
     CompressedProof, CompressedProofWithPublicInputs, OpeningSet, Proof, ProofWithPublicInputs,
 };
 
-#[derive(Debug)]
-pub struct Buffer(Cursor<Vec<u8>>);
```
dlubarov, line 24: Instead of using `Vec<u8>` as a write-only alternative alongside `Buffer`, why don't we change `Buffer` to not rely on `Cursor`? Seems cleaner IMO to have have a single buffer object, and it'd enable deserialization without std. We could also remove the `Result`s if we're no longer using `Cursor`.

dlubarov:
Looks good otherwise!

plonky2/src/hash/hash_types.rs
bhgomes: `rand` doesn't have the nicest way to do this, but it is possible to do it generically if `rand` were designed a bit differently. We can add this kind of convenience but it requires some significant boilerplate which I don't think is worth it at this point, perhaps in the future.

plonky2/Cargo.toml
bhgomes: Yeah I chose AHash because of its performance given that I don't think that the DOS resistance is that much of a risk in the plonky2 use case. Added `compile-time-rng` though.

plonky2/src/plonk/proof.rs
bhgomes: Technically writing to a `Cursor<Vec<u8>>` (using `std::io`) can error whereas writing to a `Vec` (using its `push` APIs) cannot. I'll add the `expect` to make it more clear, but in general we have a stronger guarantee here than a regular `expect`.

plonky2/src/util/serialization.rs
```diff
@@ -20,63 +23,83 @@ use crate::plonk::proof::{
...
+/// Buffer Position
+pub trait Position {
```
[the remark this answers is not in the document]
bhgomes: They don't need to be, but we can give the user the ability to use their own I/O schemes at no cost to us.

plonky2/src/util/serialization.rs
bhgomes: For this PR I wanted to do the minimal changes to the serialization API that resulted in `no-std` compatibility, in this case, we can get `serialization` for free (but not `deserialization`). We can of course expand on this minimal version to support everything, but then I would recommend using `serde` or `bincode` instead of building your own serialization library. Also, another point to be made is that the implementation of (de)serialization of a particular object should be local to that object instead of local to the (de)serializer. I didn't want to fix this to reduce the scope of this PR, but I think that would also be a great change to make. We can maybe move this to another serialization-themed Issue/PR.

> Instead of using Vec<u8> as a write-only alternative alongside Buffer, why don't we change Buffer to not rely on Cursor? Since Cursor just adds a position it seems easy to DIY.

Are you saying to implement a `no-std` compliant `Cursor` ourselves in `util`?

> Seems cleaner IMO to have have a single buffer object, and it'd enable deserialization without std. We could also remove the Results if we're no longer using Cursor.

I don't think you can ever remove `Result` for deserialization (serialization you can in some circumstances like if you're only doing in-memory serialization). For deserialization you are going from unstructured to structured data, so its not just that the conversion may fail from some I/O or resource error but that the data is actually not structured correctly, like there not being enough bytes for a field element; that's not an I/O error that's an error in the encoding directly.

plonky2/Cargo.toml
bhgomes: I won't go strongly on this one either way except to say that the standard Rust `derive` does not have the correct semantics and that `derivative` is a way to correct this. I didn't replace every `derive` with `derivative` in the crate for minimality reasons, but I think that's the correct thing to do in general. If you want to move to that, I'm happy to do another PR to fix this, but I'm alright removing `derivative`.

bhgomes:
Thanks for the PR comments!

> For future large changes I'd prefer if you could find a way to split things up, ideally with uninteresting changes (like the `alloc` imports) grouped together for ease of reviewing. 

Do you mean in another PR or in my commit structure? Wasn't sure what the policy was for this repo so I just did it all together. Let me know and I'll use whatever standard you want to use for future PRs, happy to do that.

> And we usually avoid rearranging imports, since there isn't really a standard layout (VSCode and IntelliJ defaults differ), although maybe we could enable some rustfmt options to address that.

I had noticed that most imports were using the same declaration patterns but there were some minor anomalies so I just corrected those. And also I found that `imports_granularity = "Module"` matched the majority convention that seemed to be used in the code, so I added that to `rustfmt`.

plonky2/src/util/serialization.rs
dlubarov: Good point, `Result` makes sense at least for deserialization.

We used serde with CBOR in the past, but there were some sub-optimal things we wanted to optimize. I think the main thing was that we wanted to remove length prefixes from `Vec`s whose lengths can be inferred from `CircuitConfig`. With serde I don't think we could access `CircuitConfig` from deserialization code (but correct me if wrong).

My main concern with the code here is that (if std is enabled) we now have two ways to serialize a proof, one with `Buffer` and one with `Vec<u8>`. I guess the easiest solution is to just not `impl Write for Buffer`.

plonky2/Cargo.toml
dlubarov: Hm I'd like to learn more about the concerns with Rust derives, but I'd say let's avoid derivative for now and discuss later.

There are all minor, but I just prefer standard-Rust things by default since
- It's best for build times
- Very little chance of bugs
- No learning curve, since everyone is familiar with them

plonky2/src/util/serialization.rs
dlubarov: It's just a minor thing so we could also leave it for now, and later change `Buffer` to work without Cursor, and at that point remove the `Vec<u8>` writer. WDYT?

plonky2/Cargo.toml
dlubarov: I don't think adding our own `ahash` dependency with `compile-time-rng` affects `hashbrown`; `cargo tree` is showing separate copies of `ahash`. I thought enabling `ahash-compile-time-rng` in hashbrown was the right approach, but it seems like that feature was removed in https://github.com/rust-lang/hashbrown/pull/357.

It seems the suggestion now is to do something like `hashbrown::HashMap::with_hasher(ahash::RandomState::new())` when we build a `HashMap` that needs HashDoS resistance. Since we don't have a clear need for HashDoS resistance currently, I think it's okay to just go with the default deterministic behavior for now.

bhgomes: > I don't think adding our own ahash dependency with compile-time-rng affects hashbrown; cargo tree is showing separate copies of ahash. I thought enabling ahash-compile-time-rng in hashbrown was the right approach, but it seems like that feature was removed in https://github.com/rust-lang/hashbrown/pull/357.

Ahh, sorry I forgot to match the version from the `hashbrown` dependency. Because features are additive as long as you match the same version you can include more features in the dependency (even if it's a sub-dependency). The drawback of this approach is that we have to keep the version synchronized when upgrading the `hashbrown` dependency.

plonky2/Cargo.toml
bhgomes: Yeah I can elaborate in a GitHub issue, but it's non-crucial right now so I removed the dependency.

plonky2/src/util/serialization.rs
bhgomes: > We used serde with CBOR in the past, but there were some sub-optimal things we wanted to optimize. I think the main thing was that we wanted to remove length prefixes from `Vec`s whose lengths can be inferred from `CircuitConfig`. With serde I don't think we could access `CircuitConfig` from deserialization code (but correct me if wrong).

Trying to remove length prefixes is a property of the "(de)serializer" (i.e. encoding format) rather than the "(de)serialized object" so if you wanted this you would create a custom `serde::Serializer` and `serde::Deserializer` implementation that can take a `CircuitConfig` as input (see [`serde_json::ser::Serializer`](https://docs.rs/serde_json/latest/serde_json/ser/struct.Serializer.html) as an example: we can specify the pretty-print by adding a custom [`Formatter`](https://docs.rs/serde_json/latest/serde_json/ser/trait.Formatter.html) implementation). This is very do-able.

> My main concern with the code here is that (if `std` is enabled) we now have two ways to serialize a proof, one with `Buffer` and one with `Vec<u8>`. I guess the easiest solution is to just not `impl Write for Buffer`.

I think this is fine for a few reasons:
1. Abstract interface for writing (if you don't know which writer you want just use the trait)
2. User flexibility if they want serialization and deserialization or just serialization.

I agree it is suboptimal that we don't have deserialization in `no-std` but I don't want to add more complexity to this PR.

> It's just a minor thing so we could also leave it for now, and later change `Buffer` to work without std Cursor, and at that point remove the `Vec<u8>` writer. WDYT?

Yeah, I think that's the best option for engineering complexity right now, we can implement a minimal cursor API (or see if we can get rid of position dependence entirely).

plonky2/Cargo.toml
dlubarov: Ah ok, seems like a pretty good solution then

dlubarov:
Thanks, I think this is good to merge, looks like there's just a few conflicts if you wouldn't mind resolving those?

Re PR flow, ideally we like separate PRs for repetitive diffs, like rustfmt changes or things that touch a lot of imports. Just like to get those merged quickly with minimal conflicts, while taking our time to review and discuss the more interesting changes. I understand it's not always easy to break things up though.

plonky2/src/util/serialization.rs
dlubarov: Sounds good, let's leave it for now and maybe add that Cursor-like functionality later.

Hm I hadn't thought about a custom `Serializer`; I haven't seen a `Serializer` focused on a particular application before. Are you thinking this would handle arbitrary objects, but apply some special optimizations when it encounters a proof object?

Implementing `Serialize`/`Deserialize` seems ideal in terms of user-friendliness, so that proof objects could be embedded in larger objects, which could be serialized with whatever `Serializer` is suitable for the broader application. But it seems like that would prevent accessing `CircuitConfig`.

PR #812, opened 2022-11-10
Remove signed operation code from arithmetic table

unzvfu:
We originally expected signed operations to be implemented in the arithmetic table, but as they are very very infrequently used, we have decided to implement them "in software" to save some arithmetic table columns. This PR removes the code and placeholders that were included prior to the decision to exclude these operations from the arithmetic table.

### DIFF
--- a/evm/src/arithmetic/columns.rs
+++ b/evm/src/arithmetic/columns.rs
@@ -22,25 +22,20 @@ pub const IS_ADD: usize = 0;
 pub const IS_MUL: usize = IS_ADD + 1;
 pub const IS_SUB: usize = IS_MUL + 1;
 pub const IS_DIV: usize = IS_SUB + 1;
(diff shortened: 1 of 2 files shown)


typ3c4t (approved):
though I'm unsure of the details that underlie the change from "in hardware" to "in software", this PR itself looks like a straightforward "config" change, and hence it LGTM---unsure if this causes any downstream effects, but if all checks pass...

unzvfu:
> though I'm unsure of the details that underlie the change from "in hardware" to "in software", this PR itself looks like a straightforward "config" change, and hence it LGTM---unsure if this causes any downstream effects, but if all checks pass...

Thanks Dmitry. I should have been clearer in the PR description, but there won't be any downstream effects; this is basically removing dead code.

PR #813, opened 2022-11-10
Halve number of columns used by arithmetic table

unzvfu:
Split verification of modular operations across two rows so that the arithmetic table is only 72 columns wide instead of 144 columns wide.

[[Fix explained below.]](https://github.com/mir-protocol/plonky2/pull/813#discussion_r1022339721)
~~NB: Test currently fails because of [this line](https://github.com/mir-protocol/plonky2/blob/e0d8fdd2f836ad3767e72b57bf04ee73938fe337/evm/src/arithmetic/modular.rs#L411):~~
```
yield_constr.constraint_last_row(filter);
```
~~I suspect this line is correct, but that I'm not building the `ConstraintConsumer` correctly in the tests. @nbgl Please advise?~~

### DIFF
--- a/evm/src/arithmetic/add.rs
+++ b/evm/src/arithmetic/add.rs
@@ -47,7 +47,7 @@ where
     for (a, b) in larger.zip(smaller) {
         // t should be either 0 or 2^LIMB_BITS
         let t = cy + a - b;
-        yield_constr.constraint(is_op * t * (overflow - t));
+        yield_constr.constraint_transition(is_op * t * (overflow - t));
         // cy <-- 0 or 1
         // NB: this is multiplication by a constant, so doesn't
         // increase the degree of the constraint.
@@ -87,7 +87,7 @@ where
         let t2 = builder.mul_extension(t, t1);
 
         let filtered_limb_constraint = builder.mul_extension(is_op, t2);
-        yield_constr.constraint(builder, filtered_limb_constraint);
+        yield_constr.constraint_transition(builder, filtered_limb_constraint);
 
         cy = builder.mul_const_extension(overflow_inv, t);
     }
--- a/evm/src/arithmetic/add.rs
+++ b/evm/src/arithmetic/add.rs
@@ -35,6 +35,7 @@ pub(crate) fn eval_packed_generic_are_equal<P, I, J>(
     is_op: P,
     larger: I,
     smaller: J,
+    is_two_row_op: bool,
 ) -> P
 where
     P: PackedField,
@@ -47,7 +48,11 @@ where
     for (a, b) in larger.zip(smaller) {
         // t should be either 0 or 2^LIMB_BITS
         let t = cy + a - b;
-        yield_constr.constraint_transition(is_op * t * (overflow - t));
+        if is_two_row_op {
+            yield_constr.constraint_transition(is_op * t * (overflow - t));
+        } else {
+            yield_constr.constraint(is_op * t * (overflow - t));
+        }
         // cy <-- 0 or 1
         // NB: this is multiplication by a constant, so doesn't
         // increase the degree of the constraint.
@@ -62,6 +67,7 @@ pub(crate) fn eval_ext_circuit_are_equal<F, const D: usize, I, J>(
     is_op: ExtensionTarget<D>,
     larger: I,
     smaller: J,
+    is_two_row_op: bool,
 ) -> ExtensionTarget<D>
 where
     F: RichField + Extendable<D>,
@@ -87,7 +93,11 @@ where
         let t2 = builder.mul_extension(t, t1);
 
         let filtered_limb_constraint = builder.mul_extension(is_op, t2);
-        yield_constr.constraint_transition(builder, filtered_limb_constraint);
+        if is_two_row_op {
+            yield_constr.constraint_transition(builder, filtered_limb_constraint);
+        } else {
+            yield_constr.constraint(builder, filtered_limb_constraint);
+        }
 
         cy = builder.mul_const_extension(overflow_inv, t);
     }
@@ -125,6 +135,7 @@ pub fn eval_packed_generic<P: PackedField>(
         is_add,
         output_computed,
         output_limbs.iter().copied(),
+        false,
     );
 }
 
@@ -155,6 +166,7 @@ pub fn eval_ext_circuit<F: RichField + Extendable<D>, const D: usize>(
         is_add,
         output_computed.into_iter(),
         output_limbs.iter().copied(),
+        false,
     );
 }
 
--- a/evm/src/arithmetic/columns.rs
+++ b/evm/src/arithmetic/columns.rs
@@ -52,24 +52,26 @@ pub(crate) const ALL_OPERATIONS: [usize; 17] = [
 /// this cycle.
 ///
 /// Modular arithmetic takes 9 * N_LIMBS columns which is split across
-/// two rows of 4.5 * N_LIMBS columns.
-const NUM_SHARED_COLS: usize = 4 * N_LIMBS + N_LIMBS / 2;
+/// two rows, the first with 5 * N_LIMBS columns and the second with
+/// 4 * N_LIMBS columns. (There are hence N_LIMBS "wasted columns" in
+/// the second row.)
+const NUM_SHARED_COLS: usize = 5 * N_LIMBS;
 
 const GENERAL_INPUT_0: Range<usize> = START_SHARED_COLS..START_SHARED_COLS + N_LIMBS;
 const GENERAL_INPUT_1: Range<usize> = GENERAL_INPUT_0.end..GENERAL_INPUT_0.end + N_LIMBS;
 const GENERAL_INPUT_2: Range<usize> = GENERAL_INPUT_1.end..GENERAL_INPUT_1.end + N_LIMBS;
 const GENERAL_INPUT_3: Range<usize> = GENERAL_INPUT_2.end..GENERAL_INPUT_2.end + N_LIMBS;
+const AUX_INPUT_0_LO: Range<usize> = GENERAL_INPUT_3.end..GENERAL_INPUT_3.end + N_LIMBS;
 
 // The auxiliary input columns overlap the general input columns
 // because they correspond to the values in the second row for modular
 // operations.
-const AUX_INPUT_0: Range<usize> = START_SHARED_COLS..START_SHARED_COLS + 2 * N_LIMBS;
-const AUX_INPUT_1: Range<usize> = AUX_INPUT_0.end..AUX_INPUT_0.end + 2 * N_LIMBS;
+const AUX_INPUT_0_HI: Range<usize> = START_SHARED_COLS..START_SHARED_COLS + N_LIMBS;
+const AUX_INPUT_1: Range<usize> = AUX_INPUT_0_HI.end..AUX_INPUT_0_HI.end + 2 * N_LIMBS;
 // These auxiliary input columns are awkwardly split across two rows,
 // with the first half after the general input columns and the second
 // half after the auxiliary input columns.
-const AUX_INPUT_2_LO: Range<usize> = GENERAL_INPUT_3.end..GENERAL_INPUT_3.end + N_LIMBS / 2;
-const AUX_INPUT_2_HI: Range<usize> = AUX_INPUT_1.end..AUX_INPUT_1.end + N_LIMBS / 2;
+const AUX_INPUT_2: Range<usize> = AUX_INPUT_1.end..AUX_INPUT_1.end + N_LIMBS;
 
 // ADD takes 3 * N_LIMBS = 48 columns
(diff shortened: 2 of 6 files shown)


evm/src/arithmetic/columns.rs
```diff
@@ -45,45 +49,66 @@ pub(crate) const ALL_OPERATIONS: [usize; 17] = [
...
+pub(crate) const MODULAR_OUT_AUX_RED: Range<usize> = GENERAL_INPUT_1;
 pub(crate) const MODULAR_MODULUS: Range<usize> = GENERAL_INPUT_2;
 pub(crate) const MODULAR_OUTPUT: Range<usize> = GENERAL_INPUT_3;
 pub(crate) const MODULAR_QUO_INPUT: Range<usize> = AUX_INPUT_0;
 // NB: Last value is not used in AUX, it is used in MOD_IS_ZERO
 pub(crate) const MODULAR_AUX_INPUT: Range<usize> = AUX_INPUT_1;
 pub(crate) const MODULAR_MOD_IS_ZERO: usize = AUX_INPUT_1.end - 1;
-pub(crate) const MODULAR_OUT_AUX_RED: Range<usize> = AUX_INPUT_2;
+pub(crate) const MODULAR_INPUT_1_LO: Range<usize> = AUX_INPUT_2_LO;
+pub(crate) const MODULAR_INPUT_1_HI: Range<usize> = AUX_INPUT_2_HI;
```
nbgl, line 111: The current CTL code assumes that all the data lives in the same row, and it can’t compare a row in one table against a pair of rows in another table (please correct me if this is wrong, @wborgeaud).

There is an issue here, which is that while most operations need to cross-reference `MODULAR_INPUT_0`, `MODULAR_INPUT_1`, `MODULAR_MODULUS`, and `MODULAR_OUTPUT` against the CPU, the division operation replaces `MODULAR_OUTPUT` with the low half of `MODULAR_QUO_INPUT` (am I understanding this right, @unzvfu?).

That’s a total of 80 limbs, so it won’t fit in one row as it currently stands. Possible solutions:
1. Make the table 80 columns wide (this is probably the easiest, even if not optimal)
2. Split out the code for division so `ADDMOD`, `SUBMOD`, `MULMOD`, and `MOD` can sync `MODULAR_OUTPUT` against the CPU while `DIV` exposes the low part of `MODULAR_QUO_INPUT`.
3. Change the CTL code so we can use `nv` as well as `lv` in the CTL API (@wborgeaud, I don’t know how much work this would be)

evm/src/arithmetic/modular.rs
```diff
@@ -573,15 +643,15 @@ mod tests {
...
 
-                generate(&mut lv, op_filter);
+                generate(&mut lv, &mut nv, op_filter);
 
                 let mut constraint_consumer = ConstraintConsumer::new(
                     vec![GoldilocksField(2), GoldilocksField(3), GoldilocksField(5)],
                     GoldilocksField::ONE,
                     GoldilocksField::ONE,
                     GoldilocksField::ONE,
                 );
```
nbgl, line 653: The arguments of `ConstraintConsumer::new` are `alphas: Vec<P::Scalar>, z_last: P, lagrange_basis_first: P, lagrange_basis_last: P`. My understanding (please confirm, @wborgeaud) is that `z_last` is a filter to enable `constraint_transition`, `lagrange_basis_first` enables `constraint_first_row`, and `lagrange_basis_last` enables `constraint_last_row`.

We have a constraint to ensure that `lv` is not the last row when performing modular operations, so for tests to pass `constraint_last_row` must be disabled (i.e. `lagrange_basis_last == GoldilocksField::ZERO`).

```suggestion
                let mut constraint_consumer = ConstraintConsumer::new(
                    vec![GoldilocksField(2), GoldilocksField(3), GoldilocksField(5)],
                    GoldilocksField::ONE,
                    GoldilocksField::ZERO,
                    GoldilocksField::ZERO,
                );
```

The same applies to the `zero_modulus` test.

evm/src/arithmetic/columns.rs
wborgeaud: > The current CTL code assumes that all the data lives in the same row, and it can’t compare a row in one table against a pair of rows in another table (please correct me if this is wrong, @wborgeaud).

That's correct 👍 

> 3. Change the CTL code so we can use `nv` as well as `lv` in the CTL API (@wborgeaud, I don’t know how much work this would be)

This should be doable without too much effort, let me know if you want to go with this and I'll implement it.

evm/src/arithmetic/modular.rs
wborgeaud: Yes that's correct. If you are trying to emulate a row that is not the last row in the trace, you should set `z_last` to a non-zero value and `lagrange_basis_last` to zero.
Also, small nit but I think it would be better to use random values for the `alphas` here to make the test sounder.

unzvfu: Thanks both! In @nbgl's suggested change, both `lagrange_basis_first` and `lagrange_basis_last` are set to zero, but in the explanations you both provided it seems like only `lagrange_basis_last` should be set to zero. Was `lagrange_basis_first` set to zero by accident in the change or is there something I've missed? (The tests do pass when `lagrange_basis_first` is left equal to one.)

evm/src/arithmetic/columns.rs
unzvfu: > There is an issue here, which is that while most operations need to cross-reference `MODULAR_INPUT_0`, `MODULAR_INPUT_1`, `MODULAR_MODULUS`, and `MODULAR_OUTPUT` against the CPU, the division operation replaces `MODULAR_OUTPUT` with the low half of `MODULAR_QUO_INPUT` (am I understanding this right, @unzvfu?).

@nbgl Yep, spot on.

Regarding the potential solution

>  3. Change the CTL code so we can use `nv` as well as `lv` in the CTL API (@wborgeaud, I don’t know how much work this would be)

this seems like it would be a good idea if it will (or might) be useful for other operations, but otherwise probably not worth complicating the interface just support that one operation. `DIV` is the only operation for which this has come up for me; have either of you two seen other places where adding `nv` to the CTL API would help?

evm/src/arithmetic/modular.rs
```diff
@@ -308,14 +350,16 @@ fn modular_constr_poly<P: PackedField>(
...
+    // This sets is_less_than to 1 unless we get mod_is_zero when
     // doing a DIV; in that case, we need is_less_than=0, since the
     // function checks
     //
     //   output - modulus == out_aux_red + is_less_than*2^256
     //
     // and we were given output = out_aux_red
     let is_less_than = P::ONES - mod_is_zero * lv[IS_DIV];
+    // NB: output, modulus and out_aux_red are all in lv, so can just
+    // call yield_constr.constraint() in eval_ext_circuit_lt here.
```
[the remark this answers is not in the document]
unzvfu: Oops, good catch!

evm/src/arithmetic/columns.rs
wborgeaud: > have either of you two seen other places where adding nv to the CTL API would help?

I don't think I have.

evm/src/arithmetic/modular.rs
wborgeaud: `lagrange_basis_first` should be set to one iff you are emulating the first row of the trace. I'm guessing you don't have any logic depending on whether a row is the first (i.e., you don't use `consumer.constraint_first_row`), so it shouldn't matter if you set `lagrange_basis_first` to zero or one.

evm/src/arithmetic/modular.rs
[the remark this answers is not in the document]
unzvfu: I've modified `eval_packed_generic_lt(...)` and everything it calls to use `yield_constr.constraint_transition(...)` rather than `yield_constr.constraint()`. This should fix the calls from `modular::`, but it does mean it will be calling `.constraint_transition()` in the `compare::` code even though those operations only use a single row. Are there any downsides to this that I might be missing?

evm/src/arithmetic/columns.rs
unzvfu: Updated to use 80 columns instead of 72.

evm/src/arithmetic/add.rs
```diff
@@ -47,7 +47,7 @@ where
     for (a, b) in larger.zip(smaller) {
         // t should be either 0 or 2^LIMB_BITS
         let t = cy + a - b;
-        yield_constr.constraint(is_op * t * (overflow - t));
+        yield_constr.constraint_transition(is_op * t * (overflow - t));
```
nbgl, line 50: The issue with this is that the constraint will be disabled on the very last row. So it works on modulo operations (because `lv` will never be the last row) but it's not sound for e.g. plain addition or multiplication (because they only take up one row, which is permitted to be the last).

Possible solutions:
- Have two versions of the function
- Let the user choose between `constraint` and `constraint_transition` through one of:
  - Accept a closure as an argument
  - Put a flag in the arguments
  - Flag again, but as a `const` generic

evm/src/arithmetic/add.rs
```diff
@@ -87,7 +87,7 @@ where
         let t2 = builder.mul_extension(t, t1);
 
         let filtered_limb_constraint = builder.mul_extension(is_op, t2);
-        yield_constr.constraint(builder, filtered_limb_constraint);
+        yield_constr.constraint_transition(builder, filtered_limb_constraint);
```
nbgl, line 90: Same as above.

evm/src/arithmetic/compare.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) fn eval_packed_generic_lt<P: PackedField>(
     let cy = eval_packed_generic_are_equal(yield_constr, is_op, aux.iter().copied(), lhs_limbs);
     // We don't need to check that cy is 0 or 1, since output has
     // already been checked to be 0 or 1.
-    yield_constr.constraint(is_op * (cy - output));
+    yield_constr.constraint_transition(is_op * (cy - output));
```
nbgl, line 69: Same as `add.rs`.

evm/src/arithmetic/compare.rs
```diff
@@ -134,7 +134,7 @@ pub(crate) fn eval_ext_circuit_lt<F: RichField + Extendable<D>, const D: usize>(
     );
     let good_output = builder.sub_extension(cy, output);
     let filter = builder.mul_extension(is_op, good_output);
-    yield_constr.constraint(builder, filter);
+    yield_constr.constraint_transition(builder, filter);
```
nbgl, line 137: Same as above.

evm/src/arithmetic/columns.rs
```diff
@@ -40,42 +44,65 @@ pub(crate) const ALL_OPERATIONS: [usize; 12] = [
...
+// over two rows).
 pub(crate) const MODULAR_INPUT_0: Range<usize> = GENERAL_INPUT_0;
 pub(crate) const MODULAR_INPUT_1: Range<usize> = GENERAL_INPUT_1;
 pub(crate) const MODULAR_MODULUS: Range<usize> = GENERAL_INPUT_2;
 pub(crate) const MODULAR_OUTPUT: Range<usize> = GENERAL_INPUT_3;
-pub(crate) const MODULAR_QUO_INPUT: Range<usize> = AUX_INPUT_0;
+pub(crate) const MODULAR_QUO_INPUT_LO: Range<usize> = AUX_INPUT_0_LO;
 // NB: Last value is not used in AUX, it is used in MOD_IS_ZERO
+pub(crate) const MODULAR_QUO_INPUT_HI: Range<usize> = AUX_INPUT_0_HI;
 pub(crate) const MODULAR_AUX_INPUT: Range<usize> = AUX_INPUT_1;
```
nbgl, line 106: `MODULAR_AUX_INPUT` doesn't actually use the last limb, so I would exclude it
```suggestion
pub(crate) const MODULAR_AUX_INPUT: Range<usize> = AUX_INPUT_1.start..AUX_INPUT_1.end - 1;
```

evm/src/arithmetic/modular.rs
```diff
@@ -246,21 +270,41 @@ fn generate_modular_op<F: RichField>(
...
+            .iter()
+            .map(|&c| F::from_noncanonical_i64(c)),
+        MODULAR_QUO_INPUT_HI,
+        quot_limbs[N_LIMBS..]
+            .iter()
+            .map(|&c| F::from_noncanonical_i64(c)),
+    ) {
+        lv[i] = lo;
+        nv[j] = hi;
+    }
```
nbgl, line 286: Nit: I think this would be more readable as two separate loops, but granted, it is going to be faster as one loop.

nbgl (commented):
Looks good. The only thing that's blocking is the `constraint_transition` thing (it needs to be fixed for soundness).

evm/src/arithmetic/add.rs
unzvfu: These are all somewhat ugly; I chose to pass a flag in the arguments as it seems the simplest.

nbgl (approved):
looks good!!

PR #815, opened 2022-11-15
Include the FRI prover's PoW witness in the transcript

dlubarov:
We don't think this is required for soundness, but just to remove any doubt.

Old protocol:
```
...
P sends final_poly
V samples random r
P sends pow_witness (not in transcript)
V computes pow_response = H(r, pow_witness)
V asserts pow_response has N leading 0s
...
```

Updated protocol:
```
...
P sends final_poly
P sends pow_witness
V samples random pow_response
V asserts pow_response has N leading 0s
...
```

The updated protocol is arguably simpler, at least for the verifier. For the prover, doing the search efficiently becomes a bit harder, as we need a more compact encoding of the `Challenger` state to clone.

I did something similar to `Challenger::compact()`, but didn't want to use `compact()` itself, as we don't want to do the whole duplexing step until after we observe a particular PoW witness candidate.

### DIFF
--- a/plonky2/src/fri/prover.rs
+++ b/plonky2/src/fri/prover.rs
@@ -1,4 +1,3 @@
-use itertools::Itertools;
 use maybe_rayon::*;
 use plonky2_field::extension::{flatten, unflatten, Extendable};
 use plonky2_field::polynomial::{PolynomialCoeffs, PolynomialValues};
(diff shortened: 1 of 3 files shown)


plonky2/src/fri/prover.rs
```diff
@@ -114,28 +113,56 @@ where
...
 }
 
+/// Performs the proof-of-work (a.k.a. grinding) step of the FRI protocol. Returns the PoW witness.
 fn fri_proof_of_work<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>(
-    current_hash: HashOut<F>,
+    challenger: &mut Challenger<F, C::Hasher>,
     config: &FriConfig,
 ) -> F {
-    (0..=F::NEG_ONE.to_canonical_u64())
+    // let pow_seed = challenger.get_hash();
```
wborgeaud, line 121: Can remove this?

wborgeaud (approved):
Looks good to me!

plonky2/src/fri/prover.rs
dlubarov: Good catch, fixed

PR #816, opened 2022-11-15
WIP: witness generation

nbgl:


evm/src/cpu/simple_logic/eq_iszero.rs
```diff
@@ -7,7 +7,17 @@ use plonky2::iop::ext_target::ExtensionTarget;
 use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
 use crate::cpu::columns::CpuColumnsView;
 
-pub fn generate<F: RichField>(lv: &mut CpuColumnsView<F>) {
+fn limbs(x: U256) -> [u32; 8] {
```
dlubarov, line 10: nit: maybe add "le" in the name to indicate endianness

evm/src/witness/memory.rs
```diff
@@ -0,0 +1,83 @@
...
+    contents: HashMap<MemoryAddress, U256>,
+}
+
+impl MemoryState {
+    pub fn new(kernel_code: &[u8]) -> Self {
+        let mut contents = HashMap::new();
+
+        for (i, &byte) in kernel_code.iter().enumerate() {
+            if byte != 0 {
+                let address = (0, 0, i as u32);
```
dlubarov, line 63: nit: would prefer to do `Segment::Code as usize`, I guess we'll probably never change it from 0 but it's not really guaranteed

evm/src/witness/memory.rs
```diff
@@ -0,0 +1,83 @@
...
+    ) -> Self {
+        let timestamp = (clock * NUM_CHANNELS + channel.index()) as u64;
+        MemoryOp { timestamp, address, op }
+    }
+}
+
+
+#[derive(Clone)]
+pub struct MemoryState {
+    contents: HashMap<MemoryAddress, U256>,
```
dlubarov, line 54: I think a dense representation of memory (like the old `MemorySegmentState`) would probably be more efficient. I wouldn't expect much sparseness based on the cost model, since it charges based on the highest address used.

evm/src/witness/traces.rs
```diff
@@ -0,0 +1,32 @@
+use crate::arithmetic::columns::NUM_ARITH_COLUMNS;
+use crate::cpu::columns::CpuColumnsView;
+use crate::logic;
+use crate::witness::memory::MemoryOp;
+
+type LogicRow<T> = [T; logic::columns::NUM_COLUMNS];
+type ArithmeticRow<T> = [T; NUM_ARITH_COLUMNS];
+
+struct Traces<T: Copy> {
```
dlubarov, line 9: nit: could do `<F: Field>` since we'll only compute traces over the base field; `Copy` might seem like a hint that we'll use it in other ways

PR #818, opened 2022-11-18
Add missing feature to field crate

unzvfu:
Current `main` fails to compile with the the standard "fast compilation" flags: 
```
RUSTFLAGS='-C target-cpu=native' cargo test --release
```
It complains that `#![feature(stdsimd)]` is missing.

Not sure how this got in to main, but it's recent; maybe from the `no-std` PR?


dlubarov (approved):
LGTM, good catch

PR #819, opened 2022-11-18
Implement `CALLVALUE, CALLDATALOAD, CALLDATASIZE, CALLDATACOPY` in interpreter

wborgeaud:


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -537,6 +537,52 @@ impl<'a> Interpreter<'a> {
...
+
+    fn run_calldatacopy(&mut self) {
+        let dest_offset = self.pop().as_usize();
+        let offset = self.pop().as_usize();
+        let size = self.pop().as_usize();
+        for i in 0..size {
+            let calldata_byte = self
+                .memory
+                .mload_general(self.context, Segment::Calldata, offset + i)
+                .byte(0);
```
dlubarov, line 576: nit: we could skip this we convert back to U256 below

dlubarov (approved):
LGTM!

evm/src/cpu/kernel/interpreter.rs
wborgeaud: Good catch, done

PR #820, opened 2022-11-20
Serialization refactor

dlubarov:
Followup to #806.

A few goals here
- Zero dependencies on std.
- Provide a single (memory buffering) impl of `Write` to make sure there's no confusion.
- Move closer to the `std::io` APIs. Hopefully they will be available without std at some point (there have been some discussions...).

More specifically, this
- Changes `Buffer` to not use std's `Cursor`.
- Removes `impl Write` for `Buffer`, since it's implemented for `Vec<u8>`.
- Adds a concrete I/O error type to mimic `std::io`'s.
- Combines `Position` and `Size` into `Remaining`.


dlubarov:
cc @bhgomes @Sladuca

wborgeaud (approved):
Looks good to me!

PR #822, opened 2022-11-21
Domain separator option

dlubarov:
As per our chat with Mirco.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -40,7 +40,7 @@ use crate::plonk::circuit_data::{
     CircuitConfig, CircuitData, CommonCircuitData, ProverCircuitData, ProverOnlyCircuitData,
     VerifierCircuitData, VerifierCircuitTarget, VerifierOnlyCircuitData,
 };
-use crate::plonk::config::{GenericConfig, Hasher};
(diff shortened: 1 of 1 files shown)


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -53,6 +53,11 @@ use crate::util::{log2_ceil, log2_strict, transpose, transpose_poly_values};
 pub struct CircuitBuilder<F: RichField + Extendable<D>, const D: usize> {
     pub config: CircuitConfig,
 
+    /// A domain separator, which is included in the initial Fiat-Shamir seed. This is generally not
+    /// needed, but can be used to ensure that proofs for one application are not valid for another.
+    /// Defaults to zero.
+    domain_separator: Option<F>,
```
wborgeaud, line 59: nit: Could also use `Option<Vec<F>>` here (or just `Vec<F>` defaulting to empty vec). Some other FS implementations (e.g. Merlin) use a byte string for the domain separator, usually containing a description of the circuit, so by using a vector we could also support that.

wborgeaud (approved):
Looks good to me!

plonky2/src/plonk/circuit_builder.rs
dlubarov: Good point, I changed it to `Option<Vec<F>>`, and called `hash_pad` since it's now variable-length.

PR #823, opened 2022-11-21
Make `load_code` a bit more general

dlubarov:
So that it can be used to load code we're going to execute into the code segment of a certain context.


wborgeaud (approved):
Looks good to me!

PR #824, opened 2022-11-23
Cyclic recursion tweaks

dlubarov:
Inverting `base_case` so that the API becomes very similar to `conditionally_verify_proof_or_dummy`, and expanding docs a bit.

### DIFF
--- a/plonky2/src/recursion/cyclic_recursion.rs
+++ b/plonky2/src/recursion/cyclic_recursion.rs
@@ -35,7 +35,7 @@ pub struct CyclicRecursionTarget<const D: usize> {
     pub verifier_data: VerifierCircuitTarget,
     pub dummy_proof: ProofWithPublicInputsTarget<D>,
     pub dummy_verifier_data: VerifierCircuitTarget,
-    pub base_case: BoolTarget,
+    pub condition: BoolTarget,
 }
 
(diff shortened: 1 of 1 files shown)


dlubarov:
I think it could also make sense to split into two variants, mimicking `conditionally_verify_proof` vs `conditionally_verify_proof_or_dummy`, though it's not something we need right now

wborgeaud (approved):
Looks good to me!


> I think it could also make sense to split into two variants, mimicking conditionally_verify_proof vs conditionally_verify_proof_or_dummy, though it's not something we need right now

Could you expand on what the difference would be between the two variants?

dlubarov:
I was thinking we could rename `cyclic_recursion` to something like `conditionally_verify_cyclic_proof_or_dummy`, since it checks either a real proof or a dummy proof, similar to `conditionally_verify_proof_or_dummy`.

Then we could potentially add a generalization which checks one of two proofs, where the first VK is cyclic, and the second VK is passed in by the caller. It would be similar to `conditionally_verify_proof` in that callers could use it to verify one of two real proofs (if they have the same shape), not necessarily dummy proofs.

But maybe we shouldn't do it for now, I'm not sure if it'd be useful in our aggregation circuit. I think an even more general one-of-n recursion API (where one VK is cyclic) could lead to some useful optimizations, though that may get a little complex.

PR #826, opened 2022-11-25
Add Clone support for ProofWithPublicInputsTarget

tikhono:
When I have done some recursive example, I found out the lack for Clone derive for ProofWithPublicInputsTarget.

Here example: https://gist.github.com/tikhono/6b3222ccaa2030280efcbc7038eeb373

Does the usage of clone correct? Or lack of this derive is intendant, so there are some reasonings for this?

### DIFF
--- a/plonky2/src/fri/proof.rs
+++ b/plonky2/src/fri/proof.rs
@@ -112,7 +112,7 @@ pub struct FriProof<F: RichField + Extendable<D>, H: Hasher<F>, const D: usize> 
     pub pow_witness: F,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct FriProofTarget<const D: usize> {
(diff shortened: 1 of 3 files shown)


dlubarov:
Thanks for sending, this sees fine to me if it's needed. That said I think there are a couple other potential fixes,

- It looks like our current recursion examples call `set_proof_with_pis_target` before `verify_proof`. It does seem a bit odd to require that ordering though.
- We could change `verify_proof` to make `proof_with_pis` a reference. I think we usually take references to "target" structs (besides small ones which impl `Copy`); AFAIK there wasn't a particular reason we took a value in this case.

Do you think the reference change would address the issue for you?

tikhono:
Yes, I think reference should work for me.

Thanks for fast response!

dlubarov:
Sounds good! I made #828 to change that, it might just be a bit delayed as our build is broken at the moment (https://github.com/rust-lang/rust/issues/105037)

PR #827, opened 2022-11-28
Cyclic recursion tweaks

dlubarov:
- Have the caller to `cyclic_recursion` create and pass in the virtual proof
- Split `dummy_proof` into preprocessing and proving, so that we don't need to redo the preprocessing work in each `set_cyclic_recursion_data_target` call
- Have the caller update `num_public_inputs` instead of doing it in `cyclic_recursion`. This is a little less convenient but I think it's probably best to leave the caller's config untouched, following the principle of least surprise
- Have `set_cyclic_recursion_data_target` take a sparse set of public inputs. Taking some PIs with the lowest indices didn't seem very general. I still have some reservations about this part of the API - I think it would seem cleaner if PIs of a proof which wasn't selected for verification were simply ignored - but perhaps there are some optimization reasons to keep using them.

### DIFF
--- a/plonky2/src/recursion/cyclic_recursion.rs
+++ b/plonky2/src/recursion/cyclic_recursion.rs
@@ -177,6 +177,8 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
 }
 
 /// Set the targets in a `CyclicRecursionTarget` to their corresponding values in a `CyclicRecursionData`.
+/// The `public_inputs` parameter let the caller specify certain public inputs (identified by their
+/// indices) which should be given specific values. The rest will default to zero.
 pub fn set_cyclic_recursion_data_target<
     F: RichField + Extendable<D>,
     C: GenericConfig<D, F = F>,
--- a/plonky2/src/recursion/cyclic_recursion.rs
+++ b/plonky2/src/recursion/cyclic_recursion.rs
@@ -3,6 +3,7 @@
 use alloc::vec;
 
 use anyhow::{ensure, Result};
+use hashbrown::HashMap;
 use itertools::Itertools;
 
 use crate::field::extension::Extendable;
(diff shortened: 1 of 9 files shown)


plonky2/src/recursion/cyclic_recursion.rs
```diff
@@ -183,10 +183,10 @@ pub fn set_cyclic_recursion_data_target<
     const D: usize,
 >(
     pw: &mut PartialWitness<F>,
-    cyclic_recursion_data_target: &CyclicRecursionTarget<D>,
+    cyclic_recursion_data_target: &CyclicRecursionTarget<F, C, D>,
     cyclic_recursion_data: &CyclicRecursionData<F, C, D>,
     // Public inputs to set in the base case to seed some initial data.
-    public_inputs: &[F],
+    public_inputs: &[(usize, F)],
```
wborgeaud, line 189: nit: Maybe comment on the format of `public_inputs` here (e.g like the comment for `dummy_proof()`) since this is a public function.

plonky2/src/recursion/dummy_circuit.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::plonk::circuit_data::{CircuitData, CommonCircuitData};
+use crate::plonk::config::GenericConfig;
+use crate::plonk::proof::ProofWithPublicInputs;
+
+/// Generate a proof for a dummy circuit. The `public_inputs` parameter let the caller specify
+/// certain public inputs (identified by their indices) which should be given specific values.
+/// The rest will default to zero.
+pub(crate) fn dummy_proof<F, C, const D: usize>(
+    circuit: &CircuitData<F, C, D>,
+    nonzero_public_inputs: &[(usize, F)],
```
wborgeaud, line 19: Is there a reason to not use a HashMap here? To avoid the quadratic search below.

wborgeaud (approved):
Looks good to me, nice cleanup!

plonky2/src/recursion/dummy_circuit.rs
dlubarov: I was thinking these should be pretty small, but yeah maybe best to use `HashMap` just in case

PR #832, opened 2022-12-02
Push constraints

nbgl:


nbgl:
We've decided to do PUSH in a separate table.

PR #839, opened 2022-12-08
Memory load/store constraints

nbgl:
Constraints for `mload_general` and `mstore_general`, as well as corresponding witness generation.

### DIFF
--- a/evm/src/generation/mod.rs
+++ b/evm/src/generation/mod.rs
@@ -125,11 +125,12 @@ fn simulate_cpu<F: RichField + Extendable<D>, const D: usize>(state: &mut Genera
             log::info!("CPU halted after {} cycles", state.traces.clock());
         }
         already_in_halt_loop |= in_halt_loop;
+
+        transition(state);
+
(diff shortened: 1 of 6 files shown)


evm/src/generation/mod.rs
```diff
@@ -125,11 +125,12 @@ fn simulate_cpu<F: RichField + Extendable<D>, const D: usize>(state: &mut Genera
...
         already_in_halt_loop |= in_halt_loop;
+
+        transition(state);
+
         if already_in_halt_loop && state.traces.clock().is_power_of_two() {
             log::info!("CPU trace padded to {} cycles", state.traces.clock());
             break;
         }
-
-        transition(state);
```
nbgl, line 133: This is a minor bugfix. At the start of the loop, `state.registers.program_counter` is the address of the instruction we are about to execute. It's possible for that address to be in the halt loop, even if we haven't executed any instructions in the halt loop yet. We must make sure that we have at least one row corresponding to the halt loop before we can halt.

dlubarov (approved):
LGTM!

PR #843, opened 2022-12-10
Get/set context

nbgl:


evm/src/cpu/contextops.rs
```diff
@@ -0,0 +1,82 @@
...
+fn eval_packed_get<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    let filter = lv.op.get_context;
+    let push_channel = lv.mem_channels[NUM_GP_CHANNELS - 1];
+    yield_constr.constraint(filter * (push_channel.value[0] - lv.context));
+    for &limb in &push_channel.value[1..] {
+        yield_constr.constraint(filter * limb);
+    }
```
dlubarov, line 20: Is this meant as an extra security layer, so that something like a context underflow bug in the kernel wouldn't be exploitable? If the kernel is bug-free it seems like this isn't needed.

dlubarov: Er I was thinking of set. Makes sense for get, ignore my comment :)

PR #845, opened 2022-12-10
Fix to `add_eth`

dlubarov:
It was creating a new account with the hash of an empty storage trie, when really it should be a pointer to an empty storage trie. We can use 0 as this pointer since `@SEGMENT_TRIE_DATA[0] = 0 = @MPT_NODE_EMPTY`.

Also a couple tweaks that helped me debug, like moving the memory value range checks from the interpreter into `MemoryState`, so they're done in actual witness generation as well as interpreter tests.

### DIFF
--- a/evm/tests/transfer_to_new_addr.rs
+++ b/evm/tests/transfer_to_new_addr.rs
@@ -108,7 +108,7 @@ fn test_simple_transfer() -> anyhow::Result<()> {
 
 fn eth_to_wei(eth: U256) -> U256 {
     // 1 ether = 10^18 wei.
-    eth * (U256::from(10u8) << 18)
+    eth * U256::from(10).pow(18.into())
 }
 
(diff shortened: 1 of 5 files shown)


evm/tests/transfer_to_new_addr.rs
```diff
@@ -107,8 +107,8 @@ fn test_simple_transfer() -> anyhow::Result<()> {
 }
 
 fn eth_to_wei(eth: U256) -> U256 {
-    // 1 ether = 2^18 wei.
-    eth << 18
+    // 1 ether = 10^18 wei.
+    eth * (U256::from(10u8) << 18)
```
nbgl, line 111: This is 10 * 2^18, not 10^18 :)

dlubarov: Good catch :)

PR #846, opened 2022-12-11
Simplify `JUMP`/`JUMPI` constraints and finish witness generation

nbgl:
Also re-enables jump constraints, as now they pass the tests

### DIFF
--- a/evm/src/util.rs
+++ b/evm/src/util.rs
@@ -144,11 +144,3 @@ pub(crate) fn biguint_to_u256(x: BigUint) -> U256 {
     let bytes = x.to_bytes_le();
     U256::from_little_endian(&bytes)
 }
-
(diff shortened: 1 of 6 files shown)


evm/src/util.rs
```diff
@@ -144,11 +144,3 @@ pub(crate) fn biguint_to_u256(x: BigUint) -> U256 {
...
     U256::from_little_endian(&bytes)
 }
-
-pub(crate) fn u256_saturating_cast_usize(x: U256) -> usize {
-    if x > usize::MAX.into() {
-        usize::MAX
-    } else {
-        x.as_usize()
-    }
-}
```
nbgl, line 154: Deleting now-dead code (unless there is a reason to keep it?)

PR #849, opened 2022-12-13
Moved some deps to `[dev-dependencies]`

BGluth:
Some deps were moved to `[dev-dependencies]`. Very small amount, and I deliberated a bit if it was even worth a PR.


dlubarov (approved):
LGTM, thanks!

PR #855, opened 2022-12-28
Shrink STARK proofs to a constant degree

dlubarov:
The goal here is to end up with a single "root" circuit representing any EVM proof. I.e. it must verify each STARK, but be general enough to work with any combination of STARK sizes (within some range of sizes that we chose to support). This root circuit can then be plugged into our aggregation circuit.

In particular, for each STARK, and for each initial `degree_bits` (within a range that we choose to support), this adds a "shrinking chain" of circuits. Such a chain shrinks a STARK proof from that initial `degree_bits` down to a constant, `THRESHOLD_DEGREE_BITS`.

The root circuit then combines these shrunk-to-constant proofs for each table. It's similar to `RecursiveAllProof::verify_circuit`; I adapted the code from there and I think we can remove it after. The main difference is that now instead of having one verification key per STARK, we have several possible VKs, one per initial `degree_bits`. We bake the list of possible VKs into the root circuit, and have the prover indicate the index of the VK they're actually using.

This also partially removes the default feature of CTLs. So far we've used filters instead of defaults. Until now it was easy to keep supporting defaults just in case, but here maintaining support would require some more work. E.g. we couldn't use `exp_u64` any more, since the size delta is now dynamic, it can't be hardcoded. If there are no concerns, I'll fully remove the feature after.

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -402,6 +402,9 @@ where
     }
 }
 
+/// Our usual recursion threshold is 2^12 gates, but for these shrinking circuits, we use a few more
+/// gates for a constant inner VK and for public inputs. This pushes us over the threshold to 2^13.
+/// As long as we're at 2^13 gates, we might as well use a narrower witness.
 fn shrinking_config() -> CircuitConfig {
     CircuitConfig {
         num_routed_wires: 40,
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -1,0 +1,410 @@
+use std::collections::BTreeMap;
+use std::ops::Range;
+
+use itertools::Itertools;
+use plonky2::field::extension::Extendable;
+use plonky2::hash::hash_types::RichField;
+use plonky2::hash::hashing::SPONGE_WIDTH;
+use plonky2::iop::challenger::RecursiveChallenger;
+use plonky2::iop::target::Target;
+use plonky2::iop::witness::{PartialWitness, WitnessWrite};
+use plonky2::plonk::circuit_builder::CircuitBuilder;
+use plonky2::plonk::circuit_data::{CircuitConfig, CircuitData};
+use plonky2::plonk::config::{AlgebraicHasher, GenericConfig, Hasher};
+use plonky2::plonk::proof::{ProofWithPublicInputs, ProofWithPublicInputsTarget};
+use plonky2::util::timing::TimingTree;
+
+use crate::all_stark::{all_cross_table_lookups, AllStark, Table, NUM_TABLES};
+use crate::config::StarkConfig;
+use crate::cpu::cpu_stark::CpuStark;
+use crate::cross_table_lookup::{verify_cross_table_lookups_circuit, CrossTableLookup};
+use crate::generation::GenerationInputs;
+use crate::keccak::keccak_stark::KeccakStark;
+use crate::keccak_sponge::keccak_sponge_stark::KeccakSpongeStark;
+use crate::logic::LogicStark;
+use crate::memory::memory_stark::MemoryStark;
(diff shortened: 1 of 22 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -0,0 +1,398 @@
...
+use crate::proof::StarkProofWithMetadata;
+use crate::prover::prove;
+use crate::recursive_verifier::{
+    add_common_recursion_gates, recursive_stark_circuit, PlonkWrapperCircuit, PublicInputs,
+    StarkWrapperCircuit,
+};
+use crate::stark::Stark;
+
+/// The recursion threshold. We end a chain of recursive proofs once we reach this size.
+const THRESHOLD_DEGREE_BITS: usize = 13;
```
dlubarov, line 36: I was having trouble shrinking to 2^12 since these shrinking circuits have a bunch of constants for the verification key, taking a bit of extra witness space. We might be able to make it work, but I think it's not that important, since 2^13 is still small compared to the initial wrapper proofs (which check ~80 queries) or the STARKs themselves.

plonky2/src/gadgets/random_access.rs
```diff
@@ -2,15 +2,15 @@ use alloc::vec::Vec;
...
+use crate::hash::hash_types::{HashOutTarget, MerkleCapTarget, RichField};
 use crate::iop::ext_target::ExtensionTarget;
 use crate::iop::target::Target;
 use crate::plonk::circuit_builder::CircuitBuilder;
+use crate::plonk::circuit_data::VerifierCircuitTarget;
 use crate::util::log2_strict;
 
 impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
-    /// Checks that a `Target` matches a vector at a non-deterministic index.
-    /// Note: `access_index` is not range-checked.
```
dlubarov, line 13: This was outdated, the current impl does effectively range check `access_index` by splitting it into bits (and requiring that the length is a power of two). Not sure if we want to guarantee that though. The old impl was lower-degree and could be useful at some point.

evm/src/recursive_verifier.rs
```diff
@@ -91,7 +98,8 @@ impl<T: Copy + Eq + PartialEq + Debug> PublicInputs<T> {
         };
         let challenger_state_before = next_chunk(&mut iter);
         let challenger_state_after = next_chunk(&mut iter);
-        let ctl_zs_last = iter.collect();
+        let ctl_zs_last: Vec<_> = iter.collect();
+        log::info!("from_vec num Zs: {}", ctl_zs_last.len()); // TODO
```
wborgeaud, line 102: What's the TODO for here?

wborgeaud (approved):
Looks great!

> This also partially removes the default feature of CTLs. So far we've used filters instead of defaults. Until now it was easy to keep supporting defaults just in case, but here maintaining support would require some more work. E.g. we couldn't use exp_u64 any more, since the size delta is now dynamic, it can't be hardcoded. If there are no concerns, I'll fully remove the feature after.

Sounds good, seems like we won't need the default value anywhere.

evm/src/fixed_recursive_verifier.rs
```diff
@@ -0,0 +1,410 @@
...
+        for wrapper_circuit in &self.shrinking_wrappers {
+            proof = wrapper_circuit.prove(&proof)?;
+        }
+        Ok(proof)
+    }
+}
+
+fn shrinking_config() -> CircuitConfig {
+    CircuitConfig {
+        num_routed_wires: 40,
```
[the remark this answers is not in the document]
dlubarov: It's somewhat arbitrary, but I figured if we're at 2^13 anyway we might as well use fewer (preprocessed) polynomials. I'll add a comment.

evm/src/recursive_verifier.rs
dlubarov: Oops I didn't mean to commit it, will remove.

PR #857, opened 2023-01-03
Allow non-tight degree bound

dlubarov:
Reverts the degree adjustment part of #436. As @jimpo pointed out, the adjustment complicates security by allowing rational functions of the form `poly(x) / x`.

In hindsight a tight degree bound shouldn't be necessary. Ultimately we want to check that some witness function `f(x)` exists satisfying (simplified) `c(f(x)) = Z_H(x) q(x)`.  We only need `f(x)` to be low-degree because that allows us to use polynomial identity testing. With PIT we don't care about exact degree bounds; a negligible degree change will have a negligible effect on PIT soundness.

### DIFF
--- a/plonky2/src/fri/oracle.rs
+++ b/plonky2/src/fri/oracle.rs
@@ -3,6 +3,7 @@ use alloc::vec::Vec;
 
 use itertools::Itertools;
 use maybe_rayon::*;
+use plonky2_field::types::Field;
 
 use crate::field::extension::Extendable;
 use crate::field::fft::FftRootTable;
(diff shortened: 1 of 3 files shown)


wborgeaud (approved):
Looks good to me!

ulrich-haboeck:
The whole issue is also discussed in my [summary on FRI](https://eprint.iacr.org/2022/1216). The first construction (Section 4.1.1. therein) is a commitment scheme for polynomials of degree `<= k-1` (where `k` is the rate of the Reed-Solomon code used in the proximity test) and therefore needs a degree correction factor, whereas the "optimized" scheme (Section 4.1.2) does it without that factor and hence is for degree `<= k` polynomials. 

Anyway, I agree with  @dlubarov : We don't need a tight degree bound for polynomials in the circuit satisfiability statement, and this situation does not change when leaving the realm of unique decoding. (It is good enough for DEEP-ALI to work with the simple evaluation quotients - no degree correction factors needed. See the soundness argument in my write-up.)

PR #858, opened 2023-01-03
Remove CTL defaults

dlubarov:
We ended up not needing the feature.

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -633,7 +633,6 @@ pub(crate) mod testutils {
         let CrossTableLookup {
             looking_tables,
             looked_table,
-            default,
         } = ctl;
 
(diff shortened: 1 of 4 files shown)


evm/src/cross_table_lookup.rs
```diff
@@ -580,42 +547,25 @@ pub(crate) fn verify_cross_table_lookups<
...
             let looking_zs_prod = looking_tables
                 .iter()
                 .map(|table| *ctl_zs_openings[table.table as usize].next().unwrap())
                 .product::<F>();
             let looked_z = *ctl_zs_openings[looked_table.table as usize].next().unwrap();
-            let challenge = challenges.challenges[i % config.num_challenges];
-
-            if let Some(default) = default.as_ref() {
-                let combined_default = challenge.combine(default.iter());
-                ensure!(
```
dlubarov, line 613: (I think this was the CTL verifier issue, we were only doing the check if a default was present)

wborgeaud (approved):
Looks good to me!

PR #859, opened 2023-01-03
Remove some older EVM recursion logic

dlubarov:
Some logic was replaced by the constant-degree logic in `fixed_recursive_verifier`.

### DIFF
--- a/evm/src/proof.rs
+++ b/evm/src/proof.rs
@@ -46,11 +46,6 @@ pub(crate) struct AllChallengerState<F: RichField + Extendable<D>, const D: usiz
     pub ctl_challenges: GrandProductChallengeSet<F>,
 }
 
-pub struct AllProofTarget<const D: usize> {
-    pub stark_proofs: [StarkProofTarget<D>; NUM_TABLES],
-    pub public_values: PublicValuesTarget,
-}
-
 /// Memory values which are public.
 #[derive(Debug, Clone, Default)]
 pub struct PublicValues {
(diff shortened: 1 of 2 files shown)


evm/src/proof.rs
```diff
@@ -46,11 +46,6 @@ pub(crate) struct AllChallengerState<F: RichField + Extendable<D>, const D: usiz
     pub ctl_challenges: GrandProductChallengeSet<F>,
 }
 
-pub struct AllProofTarget<const D: usize> {
```
dlubarov, line 49: (We no longer have one circuit that checks all STARK proofs)

evm/src/recursive_verifier.rs
```diff
@@ -59,11 +49,6 @@ pub struct RecursiveAllProof<
     pub recursive_proofs: [ProofWithPublicInputs<F, C, D>; NUM_TABLES],
 }
 
-pub struct RecursiveAllProofTargetWithData<const D: usize> {
```
dlubarov, line 62: (`AllRecursiveCircuits::by_table` now holds this data)

evm/src/recursive_verifier.rs
```diff
@@ -427,85 +334,6 @@ pub(crate) fn add_common_recursion_gates<F: RichField + Extendable<D>, const D:
     )));
 }
 
-/// Returns the recursive Stark circuit verifier data for every Stark in `AllStark`.
-pub fn all_verifier_data_recursive_stark_proof<
```
dlubarov, line 431: (This logic is now in `AllRecursiveCircuits::new` and its helpers)

wborgeaud (approved):
Looks good to me!

PR #860, opened 2023-01-03
Add aggregation circuit

dlubarov:
Which can be used to compress two proofs into one. Each inner proof can be either
- an "EVM root" proof (which typically proves one transaction, though it could be 0 or more)
- another aggregation proof

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -335,8 +335,8 @@ where
         self.root.circuit.prove(root_inputs)
     }
 
-    pub fn verify_root(&self, agg_proof: &ProofWithPublicInputs<F, C, D>) -> anyhow::Result<()> {
-        self.root.circuit.verify(agg_proof.clone())
+    pub fn verify_root(&self, agg_proof: ProofWithPublicInputs<F, C, D>) -> anyhow::Result<()> {
+        self.root.circuit.verify(agg_proof)
     }
 
     pub fn prove_aggregation(
@@ -360,6 +360,7 @@ where
             &self.aggregation.cyclic_vk,
             &self.aggregation.circuit.verifier_only,
         );
+
         self.aggregation.circuit.prove(agg_inputs)
     }
 
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -3,16 +3,19 @@ use std::ops::Range;
 
 use itertools::Itertools;
 use plonky2::field::extension::Extendable;
+use plonky2::gates::noop::NoopGate;
 use plonky2::hash::hash_types::RichField;
 use plonky2::hash::hashing::SPONGE_WIDTH;
 use plonky2::iop::challenger::RecursiveChallenger;
-use plonky2::iop::target::Target;
+use plonky2::iop::target::{BoolTarget, Target};
 use plonky2::iop::witness::{PartialWitness, WitnessWrite};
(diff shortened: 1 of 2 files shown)


dlubarov:
cc @BGluth @cpubot - this will be the last step for the aggregation circuit

BGluth:
I can't comment too much on the internals, but the public interface LGTM!

evm/src/fixed_recursive_verifier.rs
```diff
@@ -246,8 +326,54 @@ where
...
+            &self.root.cyclic_vk,
+            &self.aggregation.circuit.verifier_only,
+        );
+
         self.root.circuit.prove(root_inputs)
     }
+
+    pub fn verify_root(&self, agg_proof: &ProofWithPublicInputs<F, C, D>) -> anyhow::Result<()> {
+        self.root.circuit.verify(agg_proof.clone())
+    }
```
wborgeaud, line 340: nit: Maybe take a value instead of a reference here to make the cloning explicit.

evm/src/fixed_recursive_verifier.rs
```diff
@@ -44,24 +47,47 @@ where
...
 {
-    pub circuit: CircuitData<F, C, D>,
+    circuit: CircuitData<F, C, D>,
     proof_with_pis: [ProofWithPublicInputsTarget<D>; NUM_TABLES],
     /// For each table, various inner circuits may be used depending on the initial table size.
     /// This target holds the index of the circuit (within `final_circuits()`) that was used.
     index_verifier_data: [Target; NUM_TABLES],
+    /// Public inputs used for cyclic verification. These aren't actually used for EVM root
+    /// proofs; the circuit has them just to match the structure of aggregation proofs.
+    cyclic_vk: VerifierCircuitTarget,
```
wborgeaud, line 71: nit: Could we remove this field and just have a generator set these targets to zeros (since they aren't actually used)?

wborgeaud (approved):
Looks great!

evm/src/fixed_recursive_verifier.rs
dlubarov: Done. I was thinking it might be good to have `verify` itself take a reference, since it doesn't really take ownership of any data, but we can consider it later.

evm/src/fixed_recursive_verifier.rs
dlubarov: Hm I like the idea, just a minor complication... currently `conditionally_verify_cyclic_proof` expects its (inner) `cyclic_proof_with_pis`'s VK to match, even if `condition` is false so the cyclic proof wasn't selected for verification.

So in `prove_aggregation`, if an inner proof is an evm proof, we'll have an `agg_proof` target that's mostly ignored, except that its public VK must match. Currently we'll set both proof targets to the evm proof, which happens to have the agg proof's VK in its PIs, since `prove_root` sets it to that.

We could change `prove_aggregation` to always set the agg proof target's public VK, but I think it'd get a tad messy compared to the current code which can simply use `set_proof_with_pis_target`.

Maybe the most natural/elegant solution would be to change `conditionally_verify_cyclic_proof` so that it only enforces that the VKs match if `condition` is true? Though it'd cost a couple extra gates.

dlubarov: I'll leave it for now but I can change this later depending on your feedback.

wborgeaud: Sounds good, definitely easier to keep it like that then 👍

PR #861, opened 2023-01-13
Optimized interpolation gate

jimpo:
Currently there are two implementations of the interpolation gate used in the recursive verifier, one with a gate degree matching the degree of the interpolating polynomial and one with a degree-2 constraint. With the standard recursion config used in the bench_recursion example, it always selects the degree-2 gate because a degree-16 gate to interpolate a degree-16 polynomial would be too high.

This implements a new gate `CosetInterpolationGate` that I believe is the superior choice in all cases. First, it uses a linear time interpolation algorithm instead of the quadratic one used by the current implementations. Interpolation over arbitrary domains requires a quadratic-time computation of barycentric weights and is linear once those weights are computed. Since the domains for the interpolation gate are always a FRI coset, we can normalize the domain by a multiplicative shift so that the interpolation is always over the subgroup. The second trick is that the barycentric evaluation can be computed as a sequence of low-degree computations. This lets the gate be parameterized by a max degree, and the gate implements interpolation with the fewest number of constraints and wires, respecting the degree bound.

Running the bench_recursion benchmark with the standard config, we see

```
[DEBUG plonky2::util::context_tree] | | | 23 gates to evaluate CosetInterpolationGate { subgroup_bits: 4, degree: 6, _phantom: PhantomData<plonky2_field::goldilocks_field::GoldilocksField> }<D=2> constraints
```

with the new gate vs

```
[DEBUG plonky2::util::context_tree] | | | 76 gates to evaluate LowDegreeInterpolationGate { subgroup_bits: 4, _phantom: PhantomData<plonky2_field::goldilocks_field::GoldilocksField> }<D=2> constraints
```

with the old gate.

With an arity-16 reduction and maximum degree of 8, the `CosetInterpolationGate` uses 47 wires and 12 constraints vs 111 wires and 76 constraints for `LowDegreeInterpolationGate`. With an arity-3 reduction and maximum degree of 8, the `CosetInterpolationGate` uses 23 wires and 4 constraints vs 37 wires and 18 constraints for `HighDegreeInterpolationGate`.

It's not a particularly significant optimization, but it does simplify the code by having a single gate implementation so that the existing ones could be removed.


plonky2/src/gates/coset_interpolation.rs
```diff
@@ -0,0 +1,783 @@
...
+    }
+
+    fn barycentric_weights(&self) -> Vec<F> {
+        barycentric_weights(
+            &F::two_adic_subgroup(self.subgroup_bits)
+                .into_iter()
+                .map(|x| (x, F::ZERO))
+                .collect::<Vec<_>>(),
+        )
+    }
```
wborgeaud, line 122: nit: Could we have `barycentric_weights: Vec<F>` be a field of `CosetInterpolationGate` instead? To be sure they're computed only once.

plonky2/src/gates/coset_interpolation.rs
```diff
@@ -0,0 +1,783 @@
...
+        values,
+        barycentric_weights,
+        x,
+        F::Extension::ZERO,
+        F::Extension::ONE,
+    );
+    result
+}
+
+/// Perform a partial interpolation the polynomial defined by its values on an arbitrary domain.
```
wborgeaud, line 470:
```suggestion
/// Perform a partial interpolation of the polynomial defined by its values on an arbitrary domain.
```

wborgeaud (approved):
Looks great, this is a nice improvement!

plonky2/src/gates/coset_interpolation.rs
jimpo: The `InterpolationGate` trait inherits from Copy, so it can't hold a Vec. I'm not sure if that could be removed or how much refactoring it would require.

wborgeaud: Oh right. Then it's fine to leave it like this for now, I can work on refactoring the `InterpolationGate` trait in another PR.

PR #862, opened 2023-01-13
Use error instead of panicking in `FromStr`

wborgeaud:
h/t @npwardberkeley for pointing this out.


npwardberkeley (approved):
Looks great! Thanks!

PR #864, opened 2023-01-15
Block circuit

dlubarov:
I had to make some things public in order to use struct-update syntax for the anticipated `CommonCircuitData`.

cc @BGluth FYI


BGluth:
API looks good from my end!

wborgeaud (approved):
Looks great!

PR #866, opened 2023-01-18
Add range checks to the arithmetic Stark

unzvfu:
This adds range checks to the arithmetic Stark. It involved refactoring and pushing around a bunch of other things, most notably offsetting values that are negative so that the range checks work correctly, and rearranging which registers are used for which inputs so that all inputs and outputs are on the first of two rows (if an operation takes two rows), and so that DIV can be expressed as MOD by just rearranging names. Several bug fixes and comment improvements along the way.

Remaining to do:
- ~~Need to add tests that violate the range limits and verify that the range checks catch them.~~ A future PR will expand testing for all the Starks.
- ~~Refactor the `Operation`, `SimpleBinaryOp`, `ModularBinaryOp`, etc. code to use Daniel's `BinaryOperation` and `TernaryOperation` structure in `mod.rs`. This should be much cleaner and presumably will help with integration.~~ This will be handled in a follow-up PR.


evm/src/arithmetic/columns.rs
```diff
@@ -100,18 +101,28 @@ pub(crate) const MODULAR_INPUT_0: Range<usize> = GENERAL_INPUT_0;
...
+    MODULAR_QUO_INPUT.start..MODULAR_QUO_INPUT.start + N_LIMBS;
+
+// Need one column for the table, then two columns for every value
+// that needs to be range checked in the trace, namely the permutation
+// of the column and the permutation of the range. The two
+// permutations associated to column i will be in columns RC_COLS[2i]
+// and RC_COLS[2i+1].
+pub(crate) const NUM_RANGE_CHECK_COLS: usize = 1 + 2 * NUM_SHARED_COLS;
+pub(crate) const RANGE_COUNTER: usize = START_SHARED_COLS + NUM_SHARED_COLS;
+pub(crate) const RC_COLS: Range<usize> = RANGE_COUNTER..RANGE_COUNTER + 2 * NUM_SHARED_COLS;
```
wborgeaud, line 126: Shouldn't `RC_COLS` be `RANGE_COUNTER + 1..RANGE_COUNTER + 1 + 2 * NUM_SHARED_COLS`?

evm/src/arithmetic/operations.rs
```diff
@@ -0,0 +1,156 @@
...
+
+#[inline]
+fn u64_to_array<F: RichField>(out: &mut [F], x: u64) {
+    debug_assert!(out.len() == 4);
+
+    const MASK: u64 = (1 << 16) - 1;
+    out[0] = F::from_canonical_u64(x & MASK);
+    out[1] = F::from_canonical_u64((x >> 16) % MASK);
+    out[2] = F::from_canonical_u64((x >> 32) % MASK);
+    out[3] = F::from_canonical_u64((x >> 48) % MASK);
```
wborgeaud, line 15: Should these `%` be `&`? You can also get rid of `MASK` and use `F::from_canonical_u16(x as u16)` instead.

evm/src/arithmetic/operations.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+fn u256_to_array<F: RichField>(out: &mut [F], x: U256) {
+    debug_assert!(out.len() == N_LIMBS);
+
+    u64_to_array(&mut out[0..4], x.0[0]);
+    u64_to_array(&mut out[4..8], x.0[1]);
+    u64_to_array(&mut out[8..12], x.0[2]);
+    u64_to_array(&mut out[12..16], x.0[3]);
+}
```
wborgeaud, line 25: Seems like this part of the code depends on `LIMB_BITS=16`. We probably won't change this constant anytime soon, but maybe it's worth adding a couple of `(debug_)assert!`s.

evm/src/arithmetic/operations.rs
```diff
@@ -0,0 +1,156 @@
...
+
+pub trait Operation<F: RichField> {
+    /// Convert operation into one or two rows of the trace.
+    ///
+    /// Morally these types should be [F; NUM_ARITH_COLUMNS], but we
+    /// use vectors because that's what utils::transpose expects.
+    fn to_rows(&self) -> (Vec<F>, Option<Vec<F>>);
+}
```
wborgeaud, line 33: Is the goal the remove this trait at some point? If not, maybe it would be worth renaming to avoid the name collision with `arithmetic::Operation`.

evm/src/arithmetic/operations.rs
```diff
@@ -0,0 +1,156 @@
...
+            .iter()
+            .all(|x| *x == GENERAL_INPUT_0));
+        debug_assert!([ADD_INPUT_1, SUB_INPUT_1, MUL_INPUT_1, CMP_INPUT_1,]
+            .iter()
+            .all(|x| *x == GENERAL_INPUT_1));
+
+        u256_to_array(&mut row[GENERAL_INPUT_0], self.input0);
+        u256_to_array(&mut row[GENERAL_INPUT_1], self.input1);
+
+        // This is ugly, but it avoids the huge amount of boilderplate
```
wborgeaud, line 67:
```suggestion
        // This is ugly, but it avoids the huge amount of boilerplate
```

evm/src/arithmetic/modular.rs
```diff
@@ -191,8 +190,8 @@ fn bigint_to_columns<const N: usize>(num: &BigInt) -> [i64; N] {
 /// NB: `operation` can set the higher order elements in its result to
 /// zero if they are not used.
 fn generate_modular_op<F: RichField>(
-    lv: &mut [F; NUM_ARITH_COLUMNS],
-    nv: &mut [F; NUM_ARITH_COLUMNS],
+    lv: &mut [F],
+    nv: &mut [F],
     filter: usize,
```
wborgeaud, line 195: nit: Seems like some `generate_*` functions use `filter: usize` and some use `op: usize` instead? Is there any difference between the two? If not, maybe we could uniformize.

evm/src/arithmetic/modular.rs
```diff
@@ -263,37 +262,35 @@ fn generate_modular_op<F: RichField>(
     // Higher order terms of the product must be zero for valid quot and modulus:
     debug_assert!(&prod[2 * N_LIMBS..].iter().all(|&x| x == 0i64));
 
+    lv[MODULAR_OUTPUT].copy_from_slice(&output_limbs.map(|c| F::from_canonical_i64(c)));
+    lv[MODULAR_QUO_INPUT].copy_from_slice(&quot_limbs.map(|c| F::from_noncanonical_i64(c)));
```
wborgeaud, line 266: nit: Closures here and in other places below can be removed.

wborgeaud (approved):
Looks good to me!

evm/src/arithmetic/arithmetic_stark.rs
```diff
@@ -16,54 +20,65 @@ pub struct ArithmeticStark<F, const D: usize> {
...
-        } else if local_values[columns::IS_MOD].is_one() {
-            modular::generate(local_values, next_values, columns::IS_MOD);
-        } else if local_values[columns::IS_DIV].is_one() {
-            modular::generate(local_values, next_values, columns::IS_DIV);
-        } else {
-            panic!("the requested operation should not be handled by the arithmetic table");
+        let n_rows = cols[0].len();
+        debug_assert!(cols.iter().all(|col| col.len() == n_rows));
+
+        // TODO: This column is constant; do I really need to set it each time?
```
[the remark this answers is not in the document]
unzvfu: Well, it's called once for each trace that is generated. Not a significant performance issue of course, it's just that morally I think it should be a simple memcpy.

evm/src/arithmetic/arithmetic_stark.rs
```diff
@@ -16,54 +20,65 @@ pub struct ArithmeticStark<F, const D: usize> {
...
+        for i in 0..RANGE_MAX {
+            cols[columns::RANGE_COUNTER][i] = F::from_canonical_usize(i);
+        }
+
+        // For each column c in cols, generate the range-check
+        // permuations and put them in the corresponding range-check
+        // columns rc_c and rc_c+1.
+        for (c, rc_c) in (0..cols.len()).zip(columns::RC_COLS.step_by(2)) {
```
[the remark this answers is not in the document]
unzvfu: Yes! Good catch!

evm/src/arithmetic/arithmetic_stark.rs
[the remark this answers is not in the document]
wborgeaud: What would using a memcpy look like here? Like, would you need to pass around a vector/slice containing `0..RANGE_MAX` in every function indirectly calling `generate_range_checks()`, or could you use some kind of global variable for that?

evm/src/arithmetic/arithmetic_stark.rs
```diff
@@ -104,4 +130,198 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for ArithmeticSta
...
+        };
+
+        let mut rng = ChaCha8Rng::seed_from_u64(0x6feb51b7ec230f25);
+
+        let ops = (0..super::RANGE_MAX)
+            .map(|_| {
+                SimpleBinaryOp::new(
+                    columns::IS_MUL,
+                    U256::from(rng.gen::<u8>()),
+                    U256::from(rng.gen::<u8>()),
```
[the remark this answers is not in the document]
unzvfu: Oh, it's copypasta from somewhere else; updated to be `[u8; 32]`s.

evm/src/arithmetic/arithmetic_stark.rs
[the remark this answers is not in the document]
unzvfu: I was thinking of using a global variable `RANGE_SET` (either const or initialised at startup) that equals `vec![0, 1, 2, ..., RANGE_MAX-1]` and then the loop in question can be replaced with
```rust
cols[columns::RANGE_COUNTER][0..RANGE_MAX].copy_from_slice(&RANGE_SET[..]);
```

evm/src/arithmetic/arithmetic_stark.rs
```diff
@@ -104,4 +130,198 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for ArithmeticSta
...
+                )
+            })
+            .collect::<Vec<_>>();
+
+        // TODO: This is clearly not the right way to build this
+        // vector; I can't work out how to do it using the map above
+        // though, with or without Boxes.
+        let ops = ops.iter().map(|o| o as &dyn Operation<F>).collect();
+        let pols = stark.generate(ops);
```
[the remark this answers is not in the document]
unzvfu: Hmm, well in principle the outputs are already tested in the modules that define the operations (e.g. [here](https://github.com/mir-protocol/plonky2/blob/hamish/arithmetic-range-checks/evm/src/arithmetic/mul.rs#L237) for `MUL`), though those tests are admittedly pretty weak. Still, I think if we need to test the generated output, testing in the corresponding module is the right place for it. Here I feel it should focus on testing the range check functionality and other aspects of building the Stark trace.

evm/src/arithmetic/operations.rs
unzvfu: Yeah, I mentioned in the PR description, but not very clearly, that I think this `Operation` trait should be merged with `arithmetic::Operation` in an upcoming PR (the latter trait appeared unbeknownst to me while I was writing this PR, hence the clash).

evm/src/arithmetic/modular.rs
unzvfu: Uniformised!

evm/src/arithmetic/arithmetic_stark.rs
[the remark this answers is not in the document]
wborgeaud: I see. I think it wouldn't be a bad idea to have an e2e test for the arithmetic Stark generation though. Something like the `basic_trace` test above but with randomized input (ideally sampled from a non-seeded RNG).

PR #872, opened 2023-02-02
Unify generation and verification of ADD/SUB/LT/GT operations

unzvfu:
This PR unifies the generation and verification of the `ADD`, `SUB`, `LT`, and `GT` operations using the observation that they can all be verified using the single equation `x + y = z + cy * 2^256`, where `cy` is 0 or 1, by shuffling around which variables are the inputs and which are the outputs. This reduces the amount of code for generation and verification, and simplifies the calling code.


evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+
+/// Generate row for ADD, SUB, GT and LT operations.
+///
+/// A row consists of four values, GENERAL_REGISTER_[012] and
+/// GENERAL_REGISTER_BIT. The interpretation of these values for each
+/// operation is as follows:
+///
+/// ADD: REGISTER_0 + REGISTER_1, output in REGISTER_2, ignore REGISTER_BIT
+/// SUB: REGISTER_0 - REGISTER_2, output in REGISTER_1, ignore REGISTER_BIT
```
wborgeaud, line 68: Same as above, shouldn't this be `REGISTER_2 - REGISTER_0`?

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+    digits: J,
+    is_two_row_op: bool,
+) -> P
+where
+    P: PackedField,
+    I: Iterator<Item = P>,
+    J: Iterator<Item = P>,
+{
+    let overflow = P::Scalar::from_canonical_u64(1 << LIMB_BITS);
+    let overflow_inv = overflow.inverse();
```
wborgeaud, line 116: nit: There's a `inverse_2exp` that could be used here. But since this is called quite often, maybe it makes sense to just have the inverse as a `const`?

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+        _ => panic!("unexpected operation filter"),
+    };
+}
+
+/// Given a polynomial f represented by its coefficients `pol_coeffs`
+/// and a number n represented by its digits `digits` in base
+/// B=2^LIMB_BITS, with `pol_coeffs` and `digits` being of equal
+/// length as sequences, this method constrains that f(B) == n, but
+/// allowing the possibility of overflow by one bit (which is
+/// returned).
```
wborgeaud, line 102: It's not very clear to me what "overflowing by one bit" means in this context. 
Also it's not obvious to me what the constraints below are doing (it doesn't look like `f(B)==n` at least and it's not very clear what `is_two_row_op` has to do with this). Could you add a small description of the constraints here?

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+        // cy <-- 0 or 1
+        // NB: this is multiplication by a constant, so doesn't
+        // increase the degree of the constraint.
+        cy = t * overflow_inv;
+    }
+    cy
+}
+
+fn eval_ext_circuit_pol_eval_equal<F, const D: usize, I, J>(
+    builder: &mut plonky2::plonk::circuit_builder::CircuitBuilder<F, D>,
```
wborgeaud, line 135: nit: Why use the full path here (and below)?

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+    let expected_cy = eval_packed_generic_pol_eval_equal(
+        yield_constr,
+        filter,
+        pol_sum,
+        z.iter().copied(),
+        is_two_row_op,
+    );
+    // We don't need to check that expected_cy is 0 or 1, since cy has
+    // already been checked to be 0 or 1.
+    if is_two_row_op {
```
wborgeaud, line 218: In which case does `is_two_row_op==true` when this is called? Isn't addition/subtraction/lt/gt a one row op?

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+        // if the operation filters are all zero, then the constraints
+        // should be met even if all values are
+        // garbage.
+        lv[IS_ADD] = F::ZERO;
+        lv[IS_SUB] = F::ZERO;
+        lv[IS_LT] = F::ZERO;
+        lv[IS_GT] = F::ZERO;
+
+        let mut constrant_consumer = ConstraintConsumer::new(
+            vec![GoldilocksField(2), GoldilocksField(3), GoldilocksField(5)],
```
wborgeaud, line 335: nit: These should probably be more random than 2,3,5 to be certain nothing funny is going on.

evm/src/arithmetic/operations.rs
```diff
@@ -64,30 +64,26 @@ impl<F: RichField> Operation<F> for SimpleBinaryOp {
...
-        // required to dispatch directly to each add/sub/etc. operation.
-        match self.op_filter {
-            IS_ADD => add::generate(&mut row),
-            IS_SUB => sub::generate(&mut row),
-            IS_MUL => mul::generate(&mut row),
-            IS_LT | IS_GT => compare::generate(&mut row, self.op_filter),
-            _ => panic!("unrecognised operation"),
+        if self.op_filter == IS_SUB || self.op_filter == IS_GT {
+            u256_to_array(&mut row[GENERAL_REGISTER_0], self.input0);
+            u256_to_array(&mut row[GENERAL_REGISTER_2], self.input1);
```
wborgeaud, line 69: Just want to make sure this is correct. It seems like the code above would output `self.input1 - self.input0`. Is this the expected behavior?

wborgeaud (approved):
Looks good to me!

evm/src/arithmetic/addcc.rs
```diff
@@ -0,0 +1,384 @@
...
+use plonky2::field::packed::PackedField;
+use plonky2::field::types::Field;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::arithmetic::columns::*;
+use crate::arithmetic::utils::read_value_u64_limbs;
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+
+fn u256_add_cc(input0: [u64; N_LIMBS], input1: [u64; N_LIMBS]) -> ([u64; N_LIMBS], u64) {
```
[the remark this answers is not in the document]
unzvfu: Hmm, I should change it to `cy` for "carry" actually. I got into the habit of using `cc` from spending too much time doing Cuda PTX assembly, where [the convention is for an operation that sets the carry flag to have the `.cc` suffix](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#extended-precision-integer-arithmetic-instructions). "cc" stands for the "condition code" register of which one bit holds the carry/borrow flag.

As for `br`, it just stands for subtraction "borrow", the equivalent of carry for subtraction.

evm/src/arithmetic/addcc.rs
unzvfu: It is [used in `modular`](https://github.com/mir-protocol/plonky2/blob/hamish/unify-add-sub-cmp-operations/evm/src/arithmetic/modular.rs#L352) to verify that the result is reduced wrt the modulus. Modular operations are two rows and the inputs for this verification are spread across both rows, hence the need for `.constraint_transition(...)` rather than plain `.constraint(...)`. It's a somewhat ugly solution; happy to consider other options!

wborgeaud: Ah my bad, ignore my comment then.

evm/src/arithmetic/addcc.rs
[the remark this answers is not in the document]
unzvfu: I'll fix this in the next PR.

evm/src/arithmetic/addcc.rs
unzvfu: It's residue from copypasta; fixed.

evm/src/arithmetic/addcc.rs
unzvfu: I'm going to address this in a follow-up PR that does more thorough testing throughout.

PR #873, opened 2023-02-02
Make hash functions generic

Nashtare:
This PR makes the use of hash function (both inner and outer) generic to support user-specific usecases, and make transitions smoother in case a hash function with different parameters (width, rate) is aimed at being used in the future.

The API gets a bit more verbose, but as discussed with @wborgeaud it seems that until `const_generic_exprs` gets more stable, there's no better apparent solution.

EDIT: The move to crates.io messed up the CI when fully rebasing because of the breaking changes, so I added a patch section to the workspace config file. I assume this may come in handy in the future for other breaking changes in between releases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,3 +8,12 @@ opt-level = 3
 
 [profile.bench]
 opt-level = 3
+
+
+[patch.crates-io]
+plonky2_evm = { path = "evm" }
+plonky2_field = { path = "field" }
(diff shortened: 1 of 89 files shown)


Nashtare:
Rebasing with latest main changes also allowed to reduce the new added verbosity on some parts of the code by quite a lot.

dlubarov:
This LGTM, especially with the recent changes, thanks Robin! I might tweak a few things but I'll just do so after merging, to avoid holding this up any longer.

PR #874, opened 2023-02-06
ECC ASM functions renaming

wborgeaud:
No functional changes, just using a more consistent naming schemes for elliptic curve ASM functions.


unzvfu (approved):
Looks good.

PR #876, opened 2023-02-07
Refactor arithmetic operation traits

unzvfu:
This PR refactors the two `Operation` traits by moving the trace generation code into a `Traceable` trait which is now implemented by the `Operation`s. Also took the opportunity to
 
- change the `generate(...)` interfaces for all the operations to use `U256`s instead of reading from the existing trace table (though this is still required to some extent)
- incorporate the `SUBMOD` instruction as an extension to the EVM (artificially limited to kernel mode atm)
- renaming `addcc -> addcy`

Improvement for the future: The `to_rows(...)` methods actually have to calculate the result of an operation in the course of generating the rows of the trace; however the existing `Operation` independently calculates the expected result itself; this duplication is unnecessary and could be fixed by fiddling with the appropriate `Operation` interfaces so that the result and trace rows are returned at the same time.


evm/src/arithmetic/mod.rs
```diff
@@ -137,6 +150,80 @@ impl Operation {
     }
 }
 
+pub trait Traceable<F: PrimeField64> {
```
wborgeaud, line 153: nit: Do we expect this trait to be implemented on something else than `Operation`?

wborgeaud (approved):
Looks good to me!

evm/src/arithmetic/mod.rs
unzvfu: No, good point. I'll just put the function in the `impl Operation` block.

PR #878, opened 2023-02-08
Move SHL and SHR generation to the CPU

unzvfu:
SHL and SHR witness generation should be in the CPU rather than the arithmetic table.


dlubarov (approved):
LGTM

PR #879, opened 2023-02-10
Make le_sum public

0x0ece:
Methods like split_le are public. le_sum is the inverse map so useful to have it pub as well (at the very least, we’d like to access it).


dlubarov (approved):
Sounds good to me 👍

PR #880, opened 2023-02-10
Gas constraints

nbgl:


evm/src/cpu/syscalls.rs
```diff
@@ -69,15 +69,20 @@ pub fn eval_packed<P: PackedField>(
     yield_constr.constraint_transition(filter * (nv.is_kernel_mode - P::ONES));
     // Maintain current context
     yield_constr.constraint_transition(filter * (nv.context - lv.context));
+    // Reset gas counter to zero.
+    yield_constr.constraint_transition(filter * nv.gas);
```
dlubarov, line 73: Is this needed? Seems like it could be unconstrained, since the gas register will get replaced upon kernel_exit

dlubarov (approved):
LGTM!

PR #883, opened 2023-02-13
Introduce crate to seamlessly aggregate proofs of different circuits

nicholas-mainardi:
This PR introduces a crate that provides interfaces and data structures to aggregate an unlimited number of Plonky2 proofs in a single proof, which can be verified with the same verifier data independently from the number of proofs being aggregated.
In order to setup the recursive aggregation, it is only necessary to specify a set of circuits with the same format of public inputs: the available aggregation scheme  will be then capable of aggregating proofs generated with a circuit belonging to the specified set. 
The crate provides also a trait, called `PublicInputAggregation`, that allows to specify the format of the public inputs of the proofs being aggregated, which is necessary in order to properly combine the public inputs of the aggregated proofs in a set of public inputs for the final aggregated proof.

The recursive aggregation provided in this crate has the following features:                                                               
- There is no need to write an ad-hoc recursive circuit for each set of circuits whose proofs needs to be aggregated: the user just needs to specify the set of circuits and how to aggregate their public inputs, then the crate will build the necessary recursive circuits to aggregate an arbitrary number of proofs
- Proofs are aggregated in a tree-like fashion rather than sequentially, which can be more efficient if there are multiple provers that can aggregate simultaneously different chunks of the proof tree  

This crate can be conceived as an additional recursion strategy available in Plonky2, possibly extending the usability and versatility of Plonky2 to a broader set of use cases.

**Note that due to [this bug](https://github.com/rust-lang/rust/issues/104815) in the Rust compiler, the tests can be run only with the following nightly build: `rustc 1.67.0-nightly (e0098a5cc 2022-11-29)`**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Cargo build
-/target
+target/
 Cargo.lock
 
 # Profile-guided optimization
@@ -8,3 +8,5 @@ pgo-data.profdata
 
 # MacOS nuisances
 .DS_Store
+.idea
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["ecdsa", "evm", "field", "insertion", "maybe_rayon", "plonky2", "starky", "system_zero", "u32", "util", "waksman"]
+members = ["ecdsa", "evm", "field", "generic_recursion", "insertion", "maybe_rayon", "plonky2", "starky", "system_zero", "u32", "util", "waksman"]
 
 [profile.release]
 opt-level = 3
--- a/generic_recursion/Cargo.toml
+++ b/generic_recursion/Cargo.toml
@@ -1,0 +1,27 @@
+[package]
+name = "generic_recursion"
+version = "0.1.0"
+edition = "2021"
+authors = [
+    "Daniele Di Benedetto",
+    "Daniele Di Tullio",
+    "Michele d'Amico",
+    "Nicholas Mainardi",
+]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+plonky2 = { version = "0.1.1", default-features = false, features = ["parallel", "std"]}
+plonky2_util = { version = "0.1.0", default-features = false }
+anyhow = "1.0.40"
+log = "0.4.17"
+itertools = "0.10.5"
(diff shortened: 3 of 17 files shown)


dlubarov:
This looks really interesting, thanks for contributing it! I'd like to dig in to understand it more, and see how it compares to what we did [here](https://github.com/mir-protocol/plonky2/blob/main/evm/src/fixed_recursive_verifier.rs). I think there are similarities, though ours is a more narrow solution specifically for our EVM proofs.

How would you feel about making it a separate repo for now, with a link from this repo? I ask partly because it could be a while before we have time to review this in detail, as we're heads down with zkEVM work. We've also just been thinking about splitting things up into smaller repos generally, as arkworks has done.

nicholas-mainardi:
Hi Daniel, thanks for the prompt reply and the interest in our work at HorizenLabs. I understand that it might take a while to review it due to the zkEVM work, also considering the complexity of the PR. 

I took a brief look to your recursive circuits for the EVM proofs before opening the PR, and I think that our approach might allow to reduce the size of the aggregation circuit: indeed, with our approach aggregating any 2 proofs should cost about 2^13 constraints, while in the EVM code I think that you need to pad the size of the circuit to the size of the root circuit, which should amount to 2^15 constraints; on the other hand, our aggregation circuit requires that input proofs are always of constant size (i.e., 2^12 gates), which is done by recursively verifying proofs with a recursive circuit of such constant size, which means that
 a bit more of preprocessing might be required. So, my guess is that a benefit may be observed mostly when either aggregating a root proof with an aggregation proof or aggregating 2 aggregation proofs, while performance should be mostly equivalent when aggregating 2 root proofs.
 
I think it would be ok to move the crate to a separate repo, but it is a bit unclear what you meant to link it from repo: would you like to add the link to our crate in the README of plonky2?

dlubarov:
Good point! I was thinking about adding shrinking proofs on top of the root circuit, but it was getting more complex than I wanted.

Yeah, I was thinking the main readme in this repo could have a list of links to external libraries that can be useful to Plonky2 users. We might move some of our libraries out as well, like the insertion and waksman crates.

pgebheim:
Hello!

pgebheim:
@nicholas-mainardi -- Closing this, but if you respond with a link to your repo we can update the README as specified.

PR #884, opened 2023-02-16
Remove input storage from hash tests

typ3c4t:


evm/src/cpu/kernel/tests/hash.rs
```diff
@@ -38,109 +36,95 @@ fn make_random_input() -> Vec<u8> {
     (0..num_bytes).map(|_| rng.gen()).collect()
 }
 
-fn make_custom_input() -> Vec<u8> {
```
npwardberkeley, line 41: Why are we getting rid of this? Is it because it was only really useful for debugging?

evm/src/cpu/kernel/tests/ripemd.rs
```diff
@@ -1,56 +0,0 @@
...
-
-fn make_input(word: &str) -> Vec<u32> {
-    let mut input: Vec<u32> = vec![word.len().try_into().unwrap()];
-    input.append(&mut word.as_bytes().iter().map(|&x| x as u32).collect_vec());
-    input.push(u32::from_str_radix("deadbeef", 16).unwrap());
-    input
-}
-
-#[test]
-fn test_ripemd_reference() -> Result<()> {
```
npwardberkeley, line 16: Was this also just for debugging?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -65,6 +65,32 @@ pub fn run_interpreter(
...
 }
 
+pub struct InterpreterSetup {
+    pub label: String,
+    pub stack: Vec<U256>,
+    pub segment: Segment,
+    pub memory: Vec<(usize, Vec<U256>)>,
+}
+
+impl InterpreterSetup {
```
npwardberkeley, line 75: Wow this is a really nice interface! Minor nitpick, but since it actually runs the interpreter, the name "interpreter _setup_" doesn't seem to quite capture it; maybe `InterpreterTestWrapper` or something?

evm/src/cpu/kernel/tests/test_utils.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use anyhow::Result;
+use ethereum_types::U256;
+
+use crate::cpu::kernel::aggregator::KERNEL;
+use crate::cpu::kernel::interpreter::Interpreter;
+use crate::memory::segments::Segment;
+use crate::witness::memory::MemoryAddress;
+
+pub struct InterpreterSetup {
```
npwardberkeley, line 11: Is this code duplicated from `kernel/interpreter.rs`? It seems like this one isn't actually imported anywhere?

npwardberkeley (commented):
Looks great!!

evm/src/cpu/kernel/tests/hash.rs
typ3c4t: yeah I did that because William found a bug in my ripeMD impl and this was the bug-causing input

evm/src/cpu/kernel/tests/ripemd.rs
typ3c4t: yes; this was my original test before it got replaced by importing the reference implementation.

evm/src/cpu/kernel/interpreter.rs
typ3c4t: how about this new API (based on a suggestion from Daniel)?

evm/src/cpu/kernel/tests/test_utils.rs
typ3c4t: you're right; this was an accidental vestige of a previous iteration of this PR---just deleted it :)

evm/src/cpu/kernel/interpreter.rs
npwardberkeley: Sorry what are the differences? I can't tell

npwardberkeley: Oh nvm, I can see it. It looks great!

npwardberkeley (approved):

Issue #869, opened 2023-01-27
Faster multiplication by generator


ValarDragon:
Was skimming the code, and couldn't find any sped up routine for multiplying by the primitive root for the goldilocks field.

I believe this is an operation that should be getting done on O(N) points for an N-element proof, due to taking a LDE of the input onto a coset. (Or having the codeword be a coset, though my recollection is thats actually slower)

The goldilocks field's smallest generator is `7`, so should have a faster method of computing it via additions. If theres a faster `mul_by_2` routine on this field, then it could be (roughly):
```
def mul_by_primitive_root(x):
  x2 = x.mul2()
  x3 = x2
  x3 += x
  x6 = x3.mul2()
  x7 = x6
  x7 += x
  return x7
```

Unknown how many times in the code you actually multiply by the primitive root, or how much of a speedup factor this would be in the small field, but thought I'd mention it in case it helps. (In Arkworks for BLS12-381, I at some point thought this would be more than 2x as fast as the general multiplication. But thats a much larger field, and because multiplying by two was twice as fast as normal addition)

Nashtare:
I haven't benchmarked it in the context of plonky2, so my comment is to be taken with a pinch of salt, but with this field size, the ratio between addition and multiplication is much smaller than with large fields (like on BLS12-381 for instance where it is around 4x/5x if I recall correctly?). In [our curve implementation](https://github.com/toposware/cheetah), the ratio is around 1.4, so I doubt such approach would yield any advantage, but maybe my assumptions are wrong with how arithmetic is performed here, especially with assembly instructions..

dlubarov:
Yeah seems like a good idea to special case 7. As Robin said it might be best to still use multiplication, but the reduction should be somewhat faster.

I think a 96-bit reduction would have the same cost as a 67-bit reduction (cc @nbgl, correct me if wrong), so we could implement `mul_by_u32` and it might have other uses.

I think we used to do something like that for our (now deleted) `CrandallField`, and there's still a `from_noncanonical_u96` method in the code, though it's not implemented yet for `GoldilocksField`.

I don't think multiplication by a generator is a significant cost for us so we might not immediately notice a speedup, but it could add up if we also used `mul_by_u32` when evaluating certain constraints etc.

PR #886, opened 2023-02-20
Add FFT-based specification for Poseidon MDS layer on x86 targets

Nashtare:
This PR overrides the default MDS layer implementation of Poseidon on x86 targets to perform an FFT-based matrix-vector multiplication, similarly to what is done on aarch64 targets. Given that the x86_64-specific methods for Poseidon are all disabled, this can yield a significant speed-up on such targets, compared to the naive `mds_row_shf` implementation.

On Ubuntu 22.04, Intel® Core™ i7-9750H CPU @ 2.60GHz × 12, I am getting a 30% speed-up when running the `hashing` benchmark.

### DIFF
--- a/field/src/field_testing.rs
+++ b/field/src/field_testing.rs
@@ -10,8 +10,21 @@ macro_rules! test_field_arithmetic {
 
             use num::bigint::BigUint;
             use rand::rngs::OsRng;
-            use rand::Rng;
+            use rand::{Rng, RngCore};
             use $crate::types::{Field, Sample};
+
+            #[test]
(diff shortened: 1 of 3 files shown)


Nashtare:
I've also added a custom `from_noncanonical_u96()` implementation for Goldilocks, as mentioned in https://github.com/mir-protocol/plonky2/issues/869.

nbgl:
Perfect! It looks good.

Thank you for the contribution :)

nbgl (approved):

PR #888, opened 2023-02-24
Implement logUp

wborgeaud:
Replace the current lookup argument with Ulrich's [logUp protocol](https://ia.cr/2022/1530).
Removed all the consequently unused permutation argument code.


evm/src/cross_table_lookup.rs
```diff
@@ -217,6 +219,93 @@ impl<F: Field> CtlData<F> {
...
+    builder: &mut CircuitBuilder<F, D>,
+    challenger: &mut RecursiveChallenger<F, H, D>,
+    num_challenges: usize,
+) -> GrandProductChallengeSet<Target> {
+    let challenges = (0..num_challenges)
+        .map(|_| get_grand_product_challenge_target(builder, challenger))
+        .collect();
+    GrandProductChallengeSet { challenges }
+}
+
```
wborgeaud, line 308: This is old code from the now deleted `permutation.rs`.

evm/src/lookup.rs
```diff
@@ -1,127 +1,248 @@
...
+        // then one column for the inverse of `table + challenge` and one for the `Z` polynomial.
+        ceil_div_usize(self.columns.len(), constraint_degree - 1) + 2
+    }
 }
 
-/// Given an input column and a table column, generate the permuted input and permuted table columns
-/// used in the Halo2 permutation argument.
-pub fn permuted_cols<F: PrimeField64>(inputs: &[F], table: &[F]) -> (Vec<F>, Vec<F>) {
-    let n = inputs.len();
+/// logUp protocol from https://ia.cr/2022/1530 (TODO link to newer version?)
```
ulrich-haboeck, line 37: The newest version is now on IACR, no new link needed.

evm/src/lookup.rs
```diff
@@ -1,127 +1,248 @@
...
+        constraint_degree, 3,
+        "TODO: Allow other constraint degrees."
+    );
+    let num_helper_columns = lookup.num_helper_columns(constraint_degree);
+    let mut helper_columns: Vec<PolynomialValues<F>> = Vec::with_capacity(num_helper_columns);
+
+    // For each batch of `constraint_degree-1` columns `fi`, compute `sum 1/(f_i+challenge)` and
+    // add it to the helper columns.
+    // TODO: This does one batch inversion per column. It would also be possible to do one batch inversion
+    // for every column, but that would require building a big vector of all the columns concatenated.
```
ulrich-haboeck, line 58: I suppose that you want to say "It would be possible to do one batch inversion for every *group/block* of columns.... "

evm/src/lookup.rs
```diff
@@ -1,127 +1,248 @@
...
+pub struct Lookup {
+    /// Columns whose values should be contained in the lookup table.
+    /// These are the f_i(x) polynomials in the logUp paper.
+    pub(crate) columns: Vec<usize>,
+    /// Column containing the lookup table.
+    /// This is the t(x) polynomial in the paper.
+    pub(crate) table_column: usize,
+    /// Column containing the frequencies of `columns` in `table_column`.
+    /// This is the m(x) polynomial in the paper.
+    pub(crate) frequencies_column: usize,
```
ulrich-haboeck, line 26: Something that popped up when discussing logUp with Daniel. 
Not sure if this is the right place, but we might want to check (or declare) the maximum number of overall trace entries subject to the lookup (i.e. the number of columns, times their length) to satisfy 
```
num_total_logUp_entries < char(F).
```
This will be a practically pointless check for the Goldilocks field, but not necessarily for our `M31`.

evm/src/lookup.rs
```diff
@@ -1,127 +1,248 @@
...
+    let frequencies = &trace_poly_values[lookup.frequencies_column].values;
+    let mut z = Vec::with_capacity(frequencies.len());
+    z.push(F::ZERO);
+    for i in 0..frequencies.len() - 1 {
+        let x = helper_columns[..num_helper_columns - 2]
+            .iter()
+            .map(|col| col.values[i])
+            .sum::<F>()
+            - frequencies[i] * helper_columns[num_helper_columns - 2].values[i];
+        z.push(z[i] + x);
```
ulrich-haboeck, line 107: Maybe a sanity check, that the last value for `z` (which is not computed here) ends up with `F::ZERO`, would be a nice feature?

evm/src/lookup.rs
```diff
@@ -1,127 +1,248 @@
...
+    // add it to the helper columns.
+    // TODO: This does one batch inversion per column. It would also be possible to do one batch inversion
+    // for every column, but that would require building a big vector of all the columns concatenated.
+    // Not sure which approach is better.
+    // Note: these are the h_k(x) polynomials in the paper, with a few differences:
+    //       * Here, the first ratio m_0(x)/phi_0(x) is not included with the columns batched up to create the 
+    //         h_k polynomials; instead there's a separate helper column for it (see below).
+    //       * Here, we use 1 instead of -1 as the numerator (and subtract later).
+    //       * Here, for now, the batch size (l) is always constraint_degree - 1 = 2.
+    for mut col_inds in &lookup.columns.iter().chunks(constraint_degree - 1) {
```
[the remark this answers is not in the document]
npwardberkeley: Correct! https://doc.rust-lang.org/std/slice/struct.Chunks.html

evm/src/arithmetic/columns.rs
```diff
@@ -109,13 +109,7 @@ pub(crate) const DIV_DENOMINATOR: Range<usize> = MODULAR_MODULUS;
...
 
-// Need one column for the table, then two columns for every value
-// that needs to be range checked in the trace, namely the permutation
-// of the column and the permutation of the range. The two
-// permutations associated to column i will be in columns RC_COLS[2i]
-// and RC_COLS[2i+1].
-pub(crate) const NUM_RANGE_CHECK_COLS: usize = 1 + 2 * NUM_SHARED_COLS;
 pub(crate) const RANGE_COUNTER: usize = START_SHARED_COLS + NUM_SHARED_COLS;
-pub(crate) const RC_COLS: Range<usize> = RANGE_COUNTER + 1..RANGE_COUNTER + 1 + 2 * NUM_SHARED_COLS;
+pub(crate) const RC_FREQUENCIES: usize = RANGE_COUNTER + 1;
```
ulrich-haboeck, line 113: Nice, only a single multiplicity column! Whose idea was that! :D

ulrich-haboeck (commented):
Just a few things that I found from the protocol perspective: 

1. A little redundancy in the opening values provided for all `auxiliary_polys`. We query also the helper functions for the reciprocal terms at two points (the requested, and its shift), which is more than we need. 
2. We can re-use the challenge(s) of the cross-table-lookup for every within-table-lookup.

In particular the latter leads to the general question why not choosing a uniform design for both logUp  and CTLs: 
One could implement logUp for vector tables, and choose fractal decompositions for the cross-table permutation argument, and let them share the same challenges. If one wants to optimize columns, one could even reduce number of running sums by batching the sumchecks for each of the arguments (by random linear combinations). Such batching strategy will turn out useful when using smaller fields such as `M31`, where the characteristic bound has practical impact on the number of allowed witness columns. Maybe worth opening a separate issue on that.

Nashtare:
Just a random question @wborgeaud, but is this still in the pipeline? With the increased pressure on the CTLs lately, this may come in handy (either this or logUp + GKR eventually)

PR #889, opened 2023-02-25
Hash asm optimization

npwardberkeley:
Some pretty basic refactoring to cut down opcodes and memory accesses. I'm going to keep poking around, I think there's plenty more low-hanging fruit; but since everything stands on its own, I just wanted to go ahead and get this checked in


evm/src/cpu/kernel/asm/hash/sha2/main.asm
```diff
@@ -33,21 +31,19 @@ global sha2_pad:
     DUP1
     // stack: num_bytes, num_bytes, retdest
     %add_const(8)
-    %div_const(64)
+    %shr_const(6)
```
nbgl, line 34: Nit: `shr_const(6)` is actually very marginally less efficient than `div_const(64)` because it needs an extra memory lookup.

That said, the difference is so small that I'd just go with whatever makes for cleaner code.

npwardberkeley: Ah interesting, thanks! I'll actually change a few the other way then

evm/src/cpu/kernel/asm/hash/sha2/compression.asm
```diff
@@ -8,278 +8,195 @@
...
-    DUP10
-    %add_const(28)
-    %mload_kernel_general_u32
-    // stack: h[0], h[64], a[0]+a[64], b[0]+b[64], c[0]+c[64], d[0]+d[64], e[0]+e[64], f[0]+f[64], g[0]+g[64], num_blocks, scratch_space_addr, message_schedule_addr, i, retdest
+    // stack: g[0]+g[64], h[64], num_blocks, scratch_space_addr, message_schedule_addr, i, a[0]+a[64], b[0]+b[64], c[0]+c[64], d[0]+d[64], e[0]+e[64], f[0]+f[64], 0, h[0], retdest
+    SWAP12
+    // stack: 0, h[64], num_blocks, scratch_space_addr, message_schedule_addr, i, a[0]+a[64], b[0]+b[64], c[0]+c[64], d[0]+d[64], e[0]+e[64], f[0]+f[64], g[0]+g[64], h[0], retdest
+    SWAP13
+    // stack: h[0], h[64], num_blocks, scratch_space_addr, message_schedule_addr, i, a[0]+a[64], b[0]+b[64], c[0]+c[64], d[0]+d[64], e[0]+e[64], f[0]+f[64], g[0]+g[64], 0, retdest
     %add_u32
```
typ3c4t, line 179: could in theory use `%rep` for this repeated trio

evm/src/cpu/kernel/asm/hash/sha2/write_length.asm
```diff
@@ -15,7 +15,7 @@
     %decrement
     SWAP1
     // stack: length, last_addr - 1
-    %shr_const(8)
+    %div_const(256) // equivalent to %shr_const(8)
     // stack: length >> 8, last_addr - 1
     DUP1
```
typ3c4t, line 20: could these also be in a `%rep`?

typ3c4t (approved):
these overall look good!

evm/src/cpu/kernel/asm/hash/sha2/compression.asm
npwardberkeley: Good idea!

PR #904, opened 2023-03-07
Optimize blake2b

npwardberkeley:
Replacing macros with function calls
Reduces kernel size by 50% 🤯

### DIFF
--- a/evm/src/cpu/kernel/asm/hash/blake2b/compression.asm
+++ b/evm/src/cpu/kernel/asm/hash/blake2b/compression.asm
@@ -181,30 +181,15 @@ compression_loop:
     POP
     POP
     // stack: cur_block, retdest
-    %blake2b_internal_state_addr
-    // stack: start, cur_block, retdest
-    PUSH 0
-    // stack: round=0, start, cur_block, retdest
 
     // Run 12 rounds of G functions.
-    %rep 12
-        // stack: round, start, cur_block, retdest
-        %call_blake2b_g_function(0, 4, 8, 12, 0, 1)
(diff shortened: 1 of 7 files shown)


dlubarov:
cc @typ3c4t who might be more familiar with the code

evm/src/cpu/kernel/asm/hash/blake2b/compression.asm
```diff
@@ -181,40 +183,20 @@ compression_loop:
...
     // stack: cur_block, retdest
-    %blake2b_internal_state_addr
-    // stack: start, cur_block, retdest
-    PUSH 0
-    // stack: round=0, start, cur_block, retdest
 
     // Run 12 rounds of G functions.
-    %rep 12
-        // stack: round, start, cur_block, retdest
-        %call_blake2b_g_function(0, 4, 8, 12, 0, 1)
```
dlubarov, line 192: Hm the replacement for this seems a bit messy, and maybe not that much more compact? Since now `run_g_function_round` is doing a significant amount stack manipulation to set up the arguments for each `g` call.

How would you feel about keep this `call_blake2b_g_function` macro? Probably wouldn't be quite as compact (in terms of binary size) as what you have, but at least it's a 12x improvement if we put the 8 `call_blake2b_g_function` calls in a routine. And we could probably make `%call_blake2b_g_function` somewhat more compact, e.g. by making `%blake2b_permutation` a "wrapper" macro that calls a routine.

I think generally replacing macros with routines is good (if they're called several times and contributing significantly to binary size), just in this case your macro seems like a particularly nice fit since there are so many constants we need to pass in.

dlubarov (approved):
LGTM

PR #905, opened 2023-03-08
Cross-table lookup for arithmetic stark

unzvfu:
Create the cross-table lookup mappings between the CPU and Arithmetic tables.

Includes implementing the "modulo BN254 prime" operations in the arithmetic table, as well as some refactoring, renaming, minor bug fixes, and extra addcy tests as bonuses.


evm/src/all_stark.rs
```diff
@@ -89,10 +97,23 @@ impl Table {
...
-    let mut ctls = vec![ctl_keccak_sponge(), ctl_keccak(), ctl_logic(), ctl_memory()];
+    let mut ctls = vec![
+        ctl_add_mul(),
+        ctl_sub(),
+        ctl_lt(),
+        ctl_gt(),
+        ctl_modops(),
+        ctl_bn254ops(),
+        ctl_div(),
+        ctl_mod(),
```
dlubarov, line 108: These are mutually exclusive right? (In the sense that one row of either table won't participate in more than one of these CTLs.)

If so, I think it would be preferable to combine these into one CTL, with a field indicating which operation the CTL entry is for. It should reduce overall columns since each CTL adds a few columns to both tables. This could be left for a future optimization though.

cc @wborgeaud, what do you think, would your branch change the cost model?

unzvfu: > These are mutually exclusive right? (In the sense that one row of either table won't participate in more than one of these CTLs.)

Yes.

> If so, I think it would be preferable to combine these into one CTL, with a field indicating which operation the CTL entry is for. It should reduce overall columns since each CTL adds a few columns to both tables. This could be left for a future optimization though.

Ah I see; I didn't realise that each CTL entry costs some extra columns.

unzvfu: I've made it so the arithmetic table uses a single CTL.

nbgl (approved):

PR #906, opened 2023-03-08
Generalize rust code for field extension arithmetic to be base-field agnostic

typ3c4t:
The main change---and all other changes are downstream---is the replacement of the `bn254_arithmetic` module with the `extension_tower` module. The latter is more or less the former but with the field extension structs abstracted to take a base field as type parameter.


evm/src/extension_tower.rs
```diff
@@ -0,0 +1,1198 @@
+use std::mem::transmute;
+use std::ops::{Add, Div, Mul, Neg, Sub};
+
+use ethereum_types::{U256, U512};
+use rand::distributions::{Distribution, Standard};
+use rand::Rng;
+
+pub trait FieldExt:
+    Sized
+    + Copy
```
nbgl, line 10: Nit: `Copy` already implies `Sized`

evm/src/extension_tower.rs
```diff
@@ -0,0 +1,1198 @@
...
+{
+    type Output = Self;
+
+    fn div(self, rhs: Self) -> Self::Output {
+        self * rhs.inv()
+    }
+}
+
+pub trait Stack {
+    fn on_stack(self) -> Vec<U256>;
```
nbgl, line 1183: Maybe this should return an iterator?

evm/src/extension_tower.rs
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+impl<T: FieldExt> Fp2<T> {
+    /// This function scalar multiplies an Fp2 by an BN254
+    pub fn scale(self, x: T) -> Self {
+        Fp2 {
+            re: x * self.re,
+            im: x * self.im,
+        }
+    }
```
nbgl, line 327: I suggest just implementing `Mul<Rhs = T> for Fp2<T>` and `Mul<Rhs = Fp2<T>> for T`. That way you an use the normal `*` syntax.

nbgl (approved):

PR #907, opened 2023-03-09
Fix plonky2 compilation with wasm32-unknown-unknown target

Nashtare:
Compiling plonky2 for `wasm32-unknown-unknown` currently fails.

Steps to reproduce:

- `rustup update nightly`
- `rustup target add wasm32-unknown-unknown --toolchain nightly`
- `cargo +nightly build --no-default-features --target wasm32-unknown-unknown --manifest-path plonky2/Cargo.toml`

(Note that the whole workspace cannot be compiled to `wasm32-unknown-unknown` because of the hardcoded jemalloc specification in `evm` crate).

The issue is a common one arising when dealing with `rand_core` related crates. Specifying the `getrandom` dependency with the feature "js" enabled when compiling to wasm does the trick.

### DIFF
--- a/plonky2/Cargo.toml
+++ b/plonky2/Cargo.toml
@@ -34,6 +34,9 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 static_assertions = { version = "1.1.0", default-features = false }
 unroll = { version = "0.1.5", default-features = false }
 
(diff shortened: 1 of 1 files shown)


Nashtare:
It may be worth adding an additional CI job checking the correct compilation for such targets.

dlubarov (approved):
LGTM, thanks Robin!

PR #910, opened 2023-03-14
Change OR to ADD when we know bits are disjoint

npwardberkeley:
Replace `OR`s with `ADD`s, which are cheaper, in situations where we know there is no overlap in the bits of numbers being `OR`'d. Most commonly this is when we are combining shifted pieces of numbers


dlubarov (approved):
LGTM

PR #925, opened 2023-03-20
Bignum modexp

npwardberkeley:
`modmul` and `modexp` functions


evm/src/generation/prover_input.rs
```diff
@@ -123,6 +127,93 @@ impl<F: Field> GenerationState<F> {
...
+        let m = &self.memory.contexts[0].segments[Segment::KernelGeneral as usize].content
+            [m_start_loc..m_start_loc + len];
+
+        let a_biguint = mem_vec_to_biguint(a);
+        let b_biguint = mem_vec_to_biguint(b);
+        let m_biguint = mem_vec_to_biguint(m);
+
+        let result_biguint = (a_biguint * b_biguint) / m_biguint;
+        biguint_to_mem_vec(result_biguint)
+    }
```
nbgl, line 216: These should ideally be one function, since they are virtually the same—division and remainder are in some sense the same operation (once you have one, you can get the other one quite cheaply).

I would calculate
```rust
let prod = a_biguint * b_biguint;
let quo = prod / m_biguint;
let rem = prod - quo * m_biguint;
```

I think this would be cleaner code, in addition to being faster (because you don't repeat the same division twice).

(Ideally `num_bigint` would have a `divmod` function, but I couldn't find one…)

evm/src/util.rs
```diff
@@ -179,7 +176,6 @@ pub(crate) fn biguint_to_le_limbs(x: BigUint) -> Vec<u128> {
         .collect()
```
nbgl, line 176: I'd add a note that there's a big optimization opportunity in this function if need be: `digits` should be an iterator, not a vector. This function should ideally perform one malloc, whereas currently it does at least two, maybe three.

evm/src/util.rs
```diff
@@ -161,12 +160,10 @@ pub(crate) fn le_limbs_to_biguint(x: &[u128]) -> BigUint {
     )
 }
 
-#[cfg(test)]
 pub(crate) fn mem_vec_to_biguint(x: &[U256]) -> BigUint {
     le_limbs_to_biguint(&x.iter().map(|&n| n.try_into().unwrap()).collect_vec())
```
nbgl, line 164: Another optimization opportunity here (doesn't need to be done in this PR, or TBH at all, but I'd add a note for whoever's looking at this code next): the iterator is collected into a vector, only for the `Vec` to be iterated over in `le_limbs_to_biguint`.

evm/src/util.rs
```diff
@@ -179,7 +176,6 @@ pub(crate) fn biguint_to_le_limbs(x: BigUint) -> Vec<u128> {
         .collect()
 }
 
-#[cfg(test)]
 pub(crate) fn biguint_to_mem_vec(x: BigUint) -> Vec<U256> {
     biguint_to_le_limbs(x)
         .into_iter()
```
nbgl, line 181: I think this would be cleaner (and faster) with no `Vec`s involved. `biguint_to_le_limbs` `collect`s an iterator into a vector, only for the vector to be read as an iterator. So it shouldn't be a vector in the first place!

evm/src/generation/state.rs
```diff
@@ -39,6 +39,10 @@ pub(crate) struct GenerationState<F: Field> {
     /// useful to see the actual addresses for debugging. Here we store the mapping for all known
     /// addresses.
     pub(crate) state_key_to_address: HashMap<H256, Address>,
+
+    /// Prover inputs containing the result of a MODMUL-related operation, in reverse order so that the next
```
nbgl, line 43: "reverse order" doesn't mean much. I'd write that it's little-endian order, so it becomes big-endian when `pop`ped.

nbgl (commented):
Few comments on the Rust code—I haven't looked at the assembly yet.

One thing to note: you use `num_bigint` for bignum arithmetic. The usual choice for bignum stuff is GMP, which is admittedly not written in Rust, but is highly optimized. [`rug`](https://crates.io/crates/rug) is a Rust wrapper for GMP. [This benchmark](https://crates.io/crates/bigint-benchmark) claims that `rug` is about an order of magnitude faster than `num_bigint`.

evm/src/cpu/kernel/asm/bignum/modexp.asm
```diff
@@ -0,0 +1,164 @@
+// Arithmetic on little-endian integers represented with 128-bit limbs.
+// All integers must be under a given length bound, and are padded with leading zeroes.
+
+// Stores b ^ e % m in output_loc, leaving b, e, and m unchanged.
+// b, e, and m must have the same length.
+// output_loc must have size length and be initialized with zeroes; scratch_1 must have size length.
+// All of scratch_2..scratch_5 must have size 2 * length and be initialized with zeroes.
+global modexp_bignum:
```
nbgl, line 8: Note that this code fails that `length > 0`.

evm/src/util.rs
npwardberkeley: Hm does this change fix it? (Getting rid of the `le_limbs` functions entirely)

evm/src/cpu/kernel/asm/bignum/modexp.asm
```diff
@@ -0,0 +1,164 @@
+// Arithmetic on little-endian integers represented with 128-bit limbs.
+// All integers must be under a given length bound, and are padded with leading zeroes.
+
+// Stores b ^ e % m in output_loc, leaving b, e, and m unchanged.
+// b, e, and m must have the same length.
```
nbgl, line 5: It makes sense for `b` and `m` to have the same length, but I really think `e` should be separate. If `b` and `m` have length _N_ and `e` has length _M_, then this algorithm is _O_(_N_^2 _M_). However, if we were to set _N_ = _M_, then it'd be _O_(_N_^3), which is significantly worse when _N_ ≫ _M_.

As an example, say I have a really big number and I just want to square it. If we know the length of `e` then we can exit the loop way earlier.

evm/src/cpu/kernel/asm/bignum/modexp.asm
```diff
@@ -0,0 +1,164 @@
...
+
+    // Zero out scratch_2..scratch_5.
+    DUP1
+    // stack: length, length, b_start_loc, e_start_loc, m_start_loc, output_loc, scratch_1, scratch_2, scratch_3, scratch_4, scratch_5, retdest
+    %mul_const(8)
+    // stack: 8 * length, length, b_start_loc, e_start_loc, m_start_loc, output_loc, scratch_1, scratch_2, scratch_3, scratch_4, scratch_5, retdest
+    DUP8
+    // stack: scratch_2, 8 * length, length, b_start_loc, e_start_loc, m_start_loc, output_loc, scratch_1, scratch_2, scratch_3, scratch_4, scratch_5, retdest
+    %clear_kernel_general
+    // stack: length, b_start_loc, e_start_loc, m_start_loc, output_loc, scratch_1, scratch_2, scratch_3, scratch_4, scratch_5, retdest
```
nbgl, line 83: This assumes that `scratch_2..scratch_5` are consecutive in memory! The documentation should make this very, very clear.

evm/src/cpu/kernel/asm/bignum/modexp.asm
nbgl: I guess it's fine since you call `iszero_bignum`. But it's a linear-time check, which feels a bit iffy. Either way, we can do these kinds of optimizations later.

evm/src/cpu/kernel/asm/bignum/modmul.asm
```diff
@@ -0,0 +1,192 @@
+// Arithmetic on little-endian integers represented with 128-bit limbs.
+// All integers must be under a given length bound, and are padded with leading zeroes.
+
+// Stores a * b % m in output_loc, leaving a, b, and m unchanged.
+// a, b, and m must have the same length.
+// Both output_loc and scratch_1 must have size length.
+// Both scratch_2 and scratch_3 have size 2 * length and be initialized with zeroes.
+global modmul_bignum:
```
nbgl, line 8: Please describe the algorithm below in a bit more detail. It's a really long function, written in assembly, and it's difficult to tell what's going on.

evm/src/cpu/kernel/asm/bignum/modmul.asm
```diff
@@ -0,0 +1,192 @@
...
+    // Add x into k * m (in scratch_2).
+    PUSH modmul_return_2
+    // stack: modmul_return_2, length, a_start_loc, b_start_loc, output_loc, scratch_2, scratch_3, retdest
+    %stack (return, len, a, b, out, s2) -> (len, s2, out, return, len, a, b, s2)
+    // stack: length, scratch_2, output_loc, modmul_return_2, length, a_start_loc, b_start_loc, scratch_2, scratch_3, retdest
+    %jump(add_bignum)
+modmul_return_2:
+    // stack: carry, length, a_start_loc, b_start_loc, scratch_2, scratch_3, retdest
+    ISZERO
+    %jumpi(no_carry)
```
nbgl, line 91: Wait, can you ever have a carry? `scratch_2` has `2 * 128 * length` bits, so it should fit all values `<= 2^(2 * 128 * length) - 1`. Now, `k`, `m`, and `x` are all `<= 2^(128 * length) - 1`, so  `k * m <= 2^(2 * 128 * length) - 2 * 2^(128 * length + 1) + 1` and `x + k * m <= 2^(2 * 128 * length) - 2^(128 * length)`, which is `<= 2^(128 * length) - 1`. Am I missing something?

nbgl: In any case, if there's a carry, then shouldn't we just PANIC? Because `a * b` definitely fits in `2 * 128 * length` bits, so if `x + k * m` doesn't, then they can't be equal.

evm/src/cpu/kernel/asm/bignum/modmul.asm
```diff
@@ -0,0 +1,192 @@
...
+    %stack (return, len, a, b, s2, s3) -> (len, a, b, s3, return, len, s2, s3)
+    // stack: length, a_start_loc, b_start_loc, scratch_3, modmul_return_3, length, scratch_2, scratch_3, retdest
+    %jump(mul_bignum)
+modmul_return_3:
+    // stack: length, scratch_2, scratch_3, retdest
+
+    // Check that x + k * m = a * b.
+    // Walk through scratch_2 and scratch_3, checking that they are equal.
+    // stack: n=length, i=scratch_2, j=scratch_3, retdest
+modmul_check_loop:
```
nbgl, line 161: One more thing to check: `x < m`. Without it, yes, `x % m == (a * b) % m`, but that doesn't imply that indeed `x == a * b`. (Unless the weaker condition is fine, in which case it should go in a comment ☺️).

nbgl:
Looks good!! A few comments here and there but they're all pretty minor.

I know you're still working on this PR, but it's definitely almost there!

evm/src/cpu/kernel/asm/bignum/modmul.asm
nbgl: I know there's already some comments but they get a bit lost with all the `// stack:` annotations (especially we don't have syntax highlighting for EVM assembly). Maybe surround them with some `***`s for emphasis?

evm/src/cpu/kernel/asm/bignum/modmul.asm
npwardberkeley: Good point!!

npwardberkeley:
Hi Jacqui! Feel free to take another look — I think I've addressed everything except the various more major optimizations you suggested (i.e. allowing `m` to be a different length, and using GMP). I can try to do those now, or just save them for future changes!

evm/src/cpu/kernel/tests/bignum/mod.rs
```diff
@@ -394,3 +467,81 @@ fn test_mul_bignum_all() -> Result<()> {
...
+    let mut modmul_outputs_iter = modmul_outputs.iter();
+    for a in &inputs {
+        for b in &inputs {
+            // For m, skip the first input, which is zero.
+            for m in &inputs[1..] {
+                let output = modmul_outputs_iter.next().unwrap();
+                test_modmul_bignum(a.clone(), b.clone(), m.clone(), output.clone())?;
+            }
+        }
+    }
```
nbgl, line 502:
```suggestion
    let mut modmul_outputs_iter = modmul_outputs.into_iter();
    for a in &inputs {
        for b in &inputs {
            // For m, skip the first input, which is zero.
            for m in &inputs[1..] {
                let output = modmul_outputs_iter.next().unwrap();
                test_modmul_bignum(a.clone(), b.clone(), m.clone(), output)?;
            }
        }
    }
```

evm/src/cpu/kernel/tests/bignum/mod.rs
```diff
@@ -394,3 +467,81 @@ fn test_mul_bignum_all() -> Result<()> {
...
+    // Only test smaller values for exponent.
+    let exp_bit_sizes = vec![2, 100, 127, 128, 129];
+
+    for bit_size in &BIT_SIZES_TO_TEST[3..14] {
+        for exp_bit_size in &exp_bit_sizes {
+            let b = gen_bignum(*bit_size);
+            let e = gen_bignum(*exp_bit_size);
+            let m = gen_bignum(*bit_size);
+            if !m.is_zero() {
+                let output = b.clone().modpow(&e, &m);
```
nbgl, line 518:
```suggestion
                let output = b.modpow(&e, &m);
```

evm/src/cpu/kernel/tests/bignum/mod.rs
```diff
@@ -394,3 +467,81 @@ fn test_mul_bignum_all() -> Result<()> {
...
+                test_modexp_bignum(b, e, m, output)?;
+            }
+
+            let b = max_bignum(*bit_size);
+            let e = max_bignum(*exp_bit_size);
+            let m = max_bignum(*bit_size);
+            if !m.is_zero() {
+                let output = b.clone().modpow(&e, &m);
```
nbgl, line 526:
```suggestion
                let output = b.modpow(&e, &m);
```

evm/src/cpu/kernel/tests/bignum/mod.rs
```diff
@@ -394,3 +467,81 @@ fn test_mul_bignum_all() -> Result<()> {
...
+    for b in &inputs {
+        // Include only smaller exponents, to keep tests from becoming too slow.
+        for e in &inputs[..7] {
+            // For m, skip the first input, which is zero.
+            for m in &inputs[1..] {
+                let output = modexp_outputs_iter.next().unwrap();
+                test_modexp_bignum(b.clone(), e.clone(), m.clone(), output.clone())?;
+            }
+        }
+    }
```
nbgl, line 544:
```suggestion
    let mut modexp_outputs_iter = modexp_outputs.into_iter();
    for b in &inputs {
        // Include only smaller exponents, to keep tests from becoming too slow.
        for e in &inputs[..7] {
            // For m, skip the first input, which is zero.
            for m in &inputs[1..] {
                let output = modexp_outputs_iter.next().unwrap();
                test_modexp_bignum(b.clone(), e.clone(), m.clone(), output)?;
            }
        }
    }
```

evm/src/cpu/kernel/asm/bignum/modexp.asm
```diff
@@ -0,0 +1,139 @@
...
+// end of modexp_loop
+    // stack: len, b_loc, e_loc, m_loc, out_loc, s1, s2, s3, s4, s5, retdest
+    %pop10
+    // stack: retdest
+    JUMP
+len_zero:
+    // stack: len, b_loc, e_loc, m_loc, out_loc, s1, s2, s3, s4, s5, retdest
+    %pop10
+    // stack: retdest
+    JUMP
```
nbgl, line 139: You could merge these:

```suggestion
// end of modexp_loop
len_zero:
    // stack: len, b_loc, e_loc, m_loc, out_loc, s1, s2, s3, s4, s5, retdest
    %pop10
    // stack: retdest
    JUMP
```

evm/src/cpu/kernel/asm/bignum/modmul.asm
```diff
@@ -0,0 +1,167 @@
...
+    // stack: n-1, i+1, j+1, retdest
+    %pop3
+    // stack: retdest
+    JUMP
+
+len_zero:
+    // stack: len, a_loc, b_loc, m_loc, out_loc, s1, s2, s3, retdest
+    %pop8
+    // stack: retdest
+    JUMP
```
nbgl, line 167: Nit: newline at the end of the file

nbgl (approved):
Looks good! Just a few nits :)

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -400,17 +404,18 @@ impl<'a> Interpreter<'a> {
             .debug_offsets
             .contains(&self.generation_state.registers.program_counter)
         {
-            println!("At {}, stack={:?}", self.offset_name(), self.stack());
-        } else if let Some(label) = self.offset_label() {
-            println!("At {label}");
+            // TODO: uncomment
+            // println!("At {}, stack={:?}", self.offset_name(), self.stack());
+        } else if let Some(_label) = self.offset_label() {
+            // println!("At {label}");
```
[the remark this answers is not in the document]
npwardberkeley: Ohh sorry yeah that was just to make testing easier for me

PR #931, opened 2023-03-22
Fake BLS_381 field operations via prover input

typ3c4t:
Use prover input to "fake" having op codes for BLS field arithmetic.


evm/src/generation/prover_input.rs
```diff
@@ -56,6 +57,30 @@ impl<F: Field> GenerationState<F> {
...
     }
 
+    /// Special finite field operations.
+    fn run_sf(&self, input_fn: &ProverInputFn) -> U256 {
+        let field = EvmField::from_str(input_fn.0[1].as_str()).unwrap();
+        let inputs: [U256; 4] = match field {
+            Bls381Base => {
+                let mut inputs: [U256; 4] = [U256::zero(); 4];
+                for i in 0..4 {
+                    inputs[i] = stack_peek(self, i).expect("Empty stack");
```
dlubarov, line 67: nit: `Empty stack` not quite accurate

dlubarov: could also use `std::array::from_fn`

dlubarov (approved):
LGTM

PR #936, opened 2023-03-22
Implement CREATE2 address generation

wborgeaud:


evm/src/cpu/kernel/asm/core/create.asm
```diff
@@ -47,10 +47,10 @@ global sys_create2:
     SWAP4
     %stack (salt) -> (salt, sys_create2_got_address)
     // stack: salt, sys_create2_got_address, value, code_offset, code_len, kexit_info
-    DUP4 // code_len
-    DUP4 // code_offset
+    DUP5 // code_len
+    DUP5 // code_offset
     PUSH @SEGMENT_MAIN_MEMORY
-    PUSH 0 // context
+    GET_CONTEXT
```
wborgeaud, line 53: I'm not sure why this context was set to 0, it should be the current context, right?

dlubarov: Yeah my mistake, current context makes sense

dlubarov (approved):
LGTM

PR #937, opened 2023-03-23
Access lists

wborgeaud:
Implement access lists for addresses and storage keys. Implemented using arrays+linear search for now. Better data structures could be used, will look into it at a later point.

Still TODO:
- Set the addresses access list to the txn's optional access list for type 1/2 txns.
- The current implementation is buggy when a *CALL happens. Need to make sure the access lists are the same throughout the txn and that reverts are handled correctly.


evm/src/cpu/kernel/asm/core/access_lists.asm
```diff
@@ -0,0 +1,104 @@
...
+%%after:
+    // stack: (addr in accessed_address)
+    POP
+%endmacro
+
+/// Inserts the address into the access list if it is not already present.
+/// Return 1 if the address was inserted, 0 if it was already present.
+global insert_accessed_addresses:
+    // stack: addr, retdest
+    PUSH 0 %mload_current(@SEGMENT_ACCESSED_ADDRESSES)
```
dlubarov, line 28: Since this a transaction-level data, I think it should use kernel memory rather than current context memory?

wborgeaud: Yes, though I'm still trying to figure out the details. I think writing everything to kernel memory doesn't quite work, because if a scope reverts, the access list should revert to its state before the scope was entered. So I think we need to also write the full access list in another place (e.g. another context) before any *CALL.

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -1,8 +1,18 @@
 global sys_extcodehash:
     // stack: kexit_info, address
-    // TODO: Charge gas.
-    SWAP1
+    SWAP1 %u256_to_addr
     // stack: address, kexit_info
+    DUP1 %insert_accessed_addresses
+    // stack: address in access_list, address, kexit_info
```
dlubarov, line 6: nit: to me `address in access_list` sounds like the access was warm, but it's the opposite right? I would say something like `cold_access` or `was_inserted`.

wborgeaud: Oh yes good point, this is misleading. Will change to `cold_access`.

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -4,6 +4,7 @@
 global sys_call:
     // stack: kexit_info, gas, address, value, args_offset, args_size, ret_offset, ret_size
     // TODO: Charge gas.
+    DUP3 %u256_to_addr %insert_accessed_addresses POP // TODO: Use return value in gas calculation.
```
dlubarov, line 7: It looks like we were failing to truncate address arguments for a bunch of syscalls, and you've fixed most of it here, but here the original (untruncated) address argument would still be used by some call code further down. Would you mind fixing that also? (Like maybe `SWAP2 %u256_to_addr SWAP2` at the top.)

evm/src/cpu/kernel/asm/core/access_lists.asm
```diff
@@ -0,0 +1,104 @@
...
+/// TODO: Look into using a more efficient data structure for the access lists.
+
+%macro insert_accessed_addresses
+    %stack (addr) -> (addr, %%after)
+    %jump(insert_accessed_addresses)
+%%after:
+    // stack: (addr in accessed_address)
+%endmacro
+
+%macro insert_accessed_addresses_no_return
```
dlubarov, line 16: nit: could use `%insert_accessed_addresses` in this

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -21,6 +21,7 @@ global sys_return:
 global sys_selfdestruct:
     // stack: kexit_info
```
wborgeaud, line 22: This should take an address too, right?

dlubarov: Oh right 👍

dlubarov (approved):
LGTM!

evm/src/cpu/kernel/asm/core/access_lists.asm
dlubarov: Hm seems like you're right. I guess we could use context memory and descend the chain of parent contexts when checking if something is warm.

dlubarov: (But what you have seems good for now, rollbacks in general aren't working yet)

PR #939, opened 2023-03-24
A few fixes for terminal instructions

dlubarov:


evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -23,18 +26,21 @@ global sys_selfdestruct:
     SWAP1 %u256_to_addr
     DUP1 %insert_accessed_addresses_no_return // TODO: Use return value in gas calculation.
     // stack: address, kexit_info
-    SWAP1
-    // TODO: Charge gas.
+    POP // TODO: Transfer balance to address.
+    // stack: kexit_info
     // TODO: Add address to the access list.
-    %consume_gas_const(@GAS_SELFDESTRUCT)
```
dlubarov, line 29: (This is the old macro using `CONSUME_GAS`, which will be going away now that we store gas in `kexit_info`; I can remove it after this PR.)

wborgeaud (approved):
👍

PR #940, opened 2023-03-27
Bumped `eth_trie_utils` to `0.5.0`

BGluth:
Updated the library to use the new version of `eth_trie_utils`.

### DIFF
--- a/evm/src/lib.rs
+++ b/evm/src/lib.rs
@@ -34,6 +34,7 @@ pub mod vars;
 pub mod verifier;
 pub mod witness;
 
+use eth_trie_utils::partial_trie::HashedPartialTrie;
 // Set up Jemalloc
 #[cfg(not(target_env = "msvc"))]
 use jemallocator::Jemalloc;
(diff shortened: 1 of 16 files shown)


evm/src/lib.rs
```diff
@@ -34,10 +34,14 @@ pub mod vars;
...
 // Set up Jemalloc
 #[cfg(not(target_env = "msvc"))]
 use jemallocator::Jemalloc;
 
 #[cfg(not(target_env = "msvc"))]
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
+
+pub type PartialTrie = HashedPartialTrie;
+pub type Node = eth_trie_utils::partial_trie::Node<PartialTrie>;
```
BGluth, line 47: Let me know how you feel about this. I think aliases make using these types a lot more readable, but I wasn't sure about placing this in `lib.rs`. It's needed by both `generation` and `cpu`, so `lib.rs` was the highest up parent module.

dlubarov (approved):

PR #941, opened 2023-03-28
BLS Fp2 Operations and More Systematic Test API

typ3c4t:


evm/src/cpu/kernel/asm/curve/bls381/util.asm
```diff
@@ -44,10 +44,80 @@ global test_add_fp381:
...
+    // stack:                   x_im * y_im, y_re, x_re, z_im, jumpdest
+    %stack (v: 2, y_re: 2, x_re: 2) -> (x_re, y_re, v)
+    // stack:                    x_re , y_re, x_im*y_im, z_im, jumpdest
+    %mul_fp381
+    // stack:                    x_re * y_re, x_im*y_im, z_im, jumpdest
+    %sub_fp381
+    // stack:                                      z_re, z_im, jumpdest
+    %stack (z_re: 2, z_im: 2, jumpdest) -> (jumpdest, z_re, z_im)
+    JUMP
+
```
nbgl, line 123: Nit: newline at the end of the file

evm/src/cpu/kernel/asm/curve/bls381/util.asm
```diff
@@ -44,10 +44,80 @@ global test_add_fp381:
...
+    %add_fp381_2
+    %jump(0xdeadbeef)
+
+global test_sub_fp381_2:
+    %sub_fp381_2
     %jump(0xdeadbeef)
+
+
+global mul_fp381_2:
+    // stack:                          x_re, x_im, y_re, y_im, jumpdest
```
nbgl, line 90: The `stack:` comments are helpful, but you can make them clearer by somehow denoting that `x_re`, ..., `y_im` each take up two slots. Otherwise it looks like `DUP4` is copying `y_im`.

nbgl (approved):
A few minor things, but looks good!

evm/src/cpu/kernel/asm/curve/bls381/util.asm
```diff
@@ -44,10 +44,80 @@ global test_add_fp381:
     %add_fp381
     %jump(0xdeadbeef)
 
+global test_sub_fp381:
+    %sub_fp381
+    %jump(0xdeadbeef)
+
 global test_mul_fp381:
     %mul_fp381
     %jump(0xdeadbeef)
```
nbgl, line 53: This shouldn't be in the kernel, because it's dead code in production. If you want to do tests this way, you can add a conditional compilation feature to the assembler—this shouldn't be difficult.

evm/src/cpu/kernel/asm/curve/bls381/util.asm
```diff
@@ -44,10 +44,80 @@ global test_add_fp381:
...
+    // stack:             z_re, z_im
+%endmacro
+
+global test_add_fp381_2:
+    %add_fp381_2
+    %jump(0xdeadbeef)
+
+global test_sub_fp381_2:
+    %sub_fp381_2
     %jump(0xdeadbeef)
```
nbgl, line 86: Ditto

evm/src/cpu/kernel/asm/curve/bls381/util.asm
typ3c4t: I can just remove these tests for now, since they do work / subsequently more complicated functions which call them pass tests as well.

PR #947, opened 2023-03-31
Selfdestruct gas and set

wborgeaud:


evm/src/cpu/kernel/asm/core/util.asm
```diff
@@ -30,3 +30,41 @@
...
+    DUP1 %mload_trie_data
+    // stack: nonce, account_ptr
+    ISZERO NOT %jumpi(%%false)
+    %increment DUP1 %mload_trie_data
+    // stack: balance, balance_ptr
+    ISZERO NOT %jumpi(%%false)
+    %add_const(2) %mload_trie_data
+    // stack: code_hash
+    PUSH @EMPTY_STRING_HASH
+    EQ
```
dlubarov, line 57: It looks like this falls through to `%%false`, seems like it should jump to the end?

evm/src/cpu/kernel/asm/core/util.asm
```diff
@@ -30,3 +30,41 @@
...
+    POP
+    PUSH 0
+%endmacro
+
+// Returns 1 if the account is dead (i.e., empty or non-existent), 0 otherwise.
+%macro is_dead
+    // stack: addr
+    DUP1 %is_non_existent
+    SWAP1 %is_empty
+    ADD
```
dlubarov, line 69: nit: let's comment like `ADD // OR` for clarity

evm/src/cpu/kernel/asm/core/transfer.asm
```diff
@@ -81,8 +81,8 @@ global add_eth:
     // stack: retdest
     JUMP
 global add_eth_new_account:
-    // TODO: Skip creation if amount == 0?
     // stack: null_account_ptr, addr, amount, retdest
+    DUP3 ISZERO %jumpi(add_eth_new_account_zero)
     POP
```
dlubarov, line 86: nit: could move `POP` up since it's common to the two branches

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -22,17 +22,60 @@ global sys_return:
...
+
+    // Set the balance of the address to 0.
+    // stack: balance, address, recipient, kexit_info
+    PUSH 0
+    // stack: 0, balance, address, recipient, kexit_info
+    DUP3 %mpt_read_state_trie
+    // stack: account_ptr, 0, balance, address, recipient, kexit_info
+    %add_const(1)
+    // stack: balance_ptr, 0, balance, address, recipient, kexit_info
+    %mstore_trie_data
```
dlubarov, line 59: I think we may want this to be a copy-on-write operation, so we can undo it if this context gets reverted. `%make_account_copy` may be useful.

(I forgot to do it in other places like `deduct_eth` which have TODOs for now; feel free to add a TODO here too)

dlubarov (approved):
LGTM!

evm/src/cpu/kernel/asm/core/util.asm
wborgeaud: Oops that's right, fixed.

evm/src/cpu/kernel/asm/core/terminate.asm
wborgeaud: Oh right, I added a TODO.

PR #948, opened 2023-03-31
Self-destruct list instead of set

wborgeaud:
Turns out gas refunds for SELFDESTRUCT were removed in EIP-3529 (and the YP doesn't include updates after Berlin...). So knowing the number of distinct self-destructed addresses isn't actually needed. Therefore the self-destruct set can actually just be a list. Duplicates aren't an issue since the deletion loop can be written like this
```
for addr in selfdestruct_list:
    if is_nonexistent(addr): 
        continue
    delete(addr)
```
Also added a TODO to actually delete the accounts at the end of the tx.


dlubarov (approved):
Nice!

PR #950, opened 2023-03-31
Implement KeccakSpongeStark constraints and fix associated CTL

Nashtare:
This PR implements the constraints of `KeccakSpongeStark` and reintroduces its related CTL.

Regarding the CTL fix, the issue was coming from the way we were storing the hash outputs in memory, which was inconsistent with how the outputs are written in the keccak stark side.
I've adopted a conditional writing (based on the op type) retrieving the `H256` value and storing it in a similar fashion than what's done for the kernel `code_hash` during bootstrapping (changing the way memory values are being written regardless of the operation breaks consistency in other parts).


evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -361,40 +361,242 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for KeccakSpongeS
...
+        yield_constr.constraint(is_final_block * (is_final_block - P::ONES));
 
-        // TODO: If this is the first row, the original sponge state should be 0 and already_absorbed_bytes = 0.
-        // TODO: If this is a final block, the next row's original sponge state should be 0 and already_absorbed_bytes = 0.
+        for &is_final_len in local_values.is_final_input_len.iter() {
+            yield_constr.constraint(is_final_len * (is_final_len - P::ONES));
+        }
+
+        // Ensure that full-input block and final block flags are not set to 1 at the same time.
+        yield_constr.constraint(is_final_block * is_full_input_block);
```
Nashtare, line 384: I think this could be removed, as it would end up being a dummy row which continuity with another dummy row is checked below, but as we define dummy rows as having both flags to 0, I kept this additional check in case.

evm/src/witness/util.rs
```diff
@@ -135,9 +135,18 @@ pub(crate) fn mem_write_gp_log_and_fill<F: Field>(
     channel.addr_context = F::from_canonical_usize(address.context);
     channel.addr_segment = F::from_canonical_usize(address.segment);
     channel.addr_virtual = F::from_canonical_usize(address.virt);
-    for (i, limb) in val_limbs.into_iter().enumerate() {
-        channel.value[2 * i] = F::from_canonical_u32(limb as u32);
-        channel.value[2 * i + 1] = F::from_canonical_u32((limb >> 32) as u32);
+    if row.is_keccak_sponge.is_one() {
```
dlubarov, line 138: Great to see this fixed! I was thinking about changing the endianness of our Keccak STARK, to match EVM which IIUC maps Keccak's byte output to a U256 based on BE ordering. This seems a lot easier though.

Would it be easy to modify the byte order in `generate_keccak_general` instead? That would feel most natural to me since this feels pretty specific to Keccak. But if not I think this is good for now.

Nashtare: I can have a look!

Nashtare: I've updated it to perform the change in `generate_keccak_general` as you suggested.

dlubarov (approved):
Looks great!

In hindsight I'm not sure why I included an explicit `is_final_block` column, seems like we can simplify this a bit if we replace it with the sum of `is_final_input_len`, but I can change that later.

PR #953, opened 2023-04-03
Introduce poseidon2

nicholas-mainardi:
This PR introduces a crate to employ the [recently updated Poseidon2 hash function](https://eprint.iacr.org/2023/323.pdf) in plonky2 as the hash function to generate/verify proofs. The crate provides all necessary trait implementations to seamlessly replace existing code employing Poseidon with Poseidon2.

The crate also provides a Poseidon2 gate and the necessary trait implementations to employ Poseidon2 to compute hashes in plonky2 circuit with minimal modifications to existing code.

Although we decided to keep this implementation in its own crate for simplicity of review, it should be also possible to easily integrate this code in the plonky2 crate, if deemed more appropriate and worthy to be included by maintainers of the plonky2 crate


Nashtare:
Rather than having a specific crate for poseidon2, now that we have the `HashConfig` trait that allows more flexibility on the hash function, I think it may make sense to have a dedicated `plonky2_hash` crate instead, moving poseidon / poseidon2 and whichever other hash functions that would be nice to be supported within. What do you think?

nicholas-mainardi:
Yes, I agree it might be a good strategy, though I think it would require a quite relevant refactoring (e.g., replacing the `Poseidon` trait with a more generic `Hash` trait); but in case you want to introduce it I will be happy to add Poseidon2 to such `plonky2_hash` crate

nicholas-mainardi:
Yes, we run the benchmarks you can find in the crate and we observed about 25-30% performance improvement of Poseidon2 over Poseidon, with small improvements also in the circuit. We speculated that these result should be ok considering that the Poseidon implementation seems to rely heavily on optimized arithmetic for multiplications in Goldilocks, while Poseidon2 should use much less field multiplications (thus it should benefit less from the optimized arithmetic for Goldilocks).

Currently we are not planning to work on asm optimizations

pgebheim:
Hello!

pgebheim:
Closing per last message.

PR #954, opened 2023-04-03
Fix signed syscalls stack

wborgeaud:
Syscalls expect the kernel return info to be at the top of the stack.


unzvfu (approved):
Good catch!

PR #957, opened 2023-04-04
Fix copy opcodes when offset is large

wborgeaud:
The `offset` parameter in the *COPY opcodes can be arbitrarily large, leading to an error when trying to access memory directly. Fixed by adding a check for large offsets.


dlubarov (approved):
LGTM

PR #959, opened 2023-04-05
Implement sys_return and sys_revert

wborgeaud:
Also set returndata size to 0 in sys_stop and sys_selfdestruct.

### DIFF
--- a/evm/src/cpu/kernel/asm/core/terminate.asm
+++ b/evm/src/cpu/kernel/asm/core/terminate.asm
@@ -1,21 +1,10 @@
 // Handlers for operations which terminate the current context, namely STOP,
 // RETURN, SELFDESTRUCT, REVERT, and exceptions such as stack underflow.
 
-// Set parent context's CTX_METADATA_RETURNDATA_SIZE to 0.
-%macro set_parent_returndata_size_zero
-    // stack: (empty)
-    %mload_context_metadata(@CTX_METADATA_PARENT_CONTEXT)
-    %stack (parent_ctx) ->
-        (parent_ctx, @SEGMENT_CONTEXT_METADATA, @CTX_METADATA_RETURNDATA_SIZE, 0)
-    MSTORE_GENERAL
-    // stack: (empty)
-%endmacro
-
-
 global sys_stop:
     // stack: kexit_info
-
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -103,7 +159,7 @@ global fault_exception:
     // stack: (empty)
     PUSH 0 // leftover_gas
     // TODO: Revert state changes.
-    // TODO: Set parent context's CTX_METADATA_RETURNDATA_SIZE to 0.
+    %set_parent_returndata_size_zero
     PUSH 0 // success
     %jump(terminate_common)
```
wborgeaud, line 164: I'm not sure I understand why we don't panic here. Seems like going to `terminate_common` will just continue the execution?

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -1,23 +1,57 @@
 // Handlers for operations which terminate the current context, namely STOP,
 // RETURN, SELFDESTRUCT, REVERT, and exceptions such as stack underflow.
 
+// Set parent context's CTX_METADATA_RETURNDATA_SIZE to 0.
+%macro set_parent_returndata_size_zero
+    // stack: (empty)
```
dlubarov, line 6: I think it would be nice to have a `mstore_parent_context_metadata` macro, since we do this in a few places

evm/src/cpu/kernel/asm/core/terminate.asm
dlubarov: My understanding is that for exceptions of this sort, the block is still valid (so we still need to prove the result), and the txn could potentially still continue if there's a parent context (besides the kernel).

dlubarov (approved):
LGTM

evm/src/cpu/kernel/asm/core/terminate.asm
wborgeaud: Good idea, I added two macros `mstore_parent_context_metadata(field), mstore_parent_context_metadata(field, value)` in `metadata.asm`.

PR #961, opened 2023-04-06
Minor fixes to context creation

wborgeaud:


evm/src/cpu/kernel/asm/core/util.asm
```diff
@@ -1,13 +1,10 @@
 // Return the next context ID, and record the old context ID in the new one's
 // @CTX_METADATA_PARENT_CONTEXT field. Does not actually enter the new context.
 %macro create_context
+    // stack: (empty)
     %next_context_id
-    GET_CONTEXT
-    %stack (ctx, next_ctx)
-       -> (next_ctx, @SEGMENT_NORMALIZED_TXN, @CTX_METADATA_PARENT_CONTEXT,
```
wborgeaud, line 7: Not sure why the segment was `SEGMENT_NORMALIZED_TXN` here (should be `SEGMENT_CONTEXT_METADATA` right?), but this block is exactly the same as `%set_new_ctx_parent_ctx`.

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -18,7 +18,6 @@ global sys_call:
     %address %set_new_ctx_caller
     DUP5 %set_new_ctx_value
     DUP5 DUP5 %address %transfer_eth
-    %set_new_ctx_parent_ctx
```
wborgeaud, line 21: Here and below, `%set_new_ctx_parent_ctx` is already called in `%create_context`.

evm/src/memory/segments.rs
```diff
@@ -132,7 +132,7 @@ impl Segment {
             Segment::KernelGeneral2 => 256,
             Segment::KernelAccountCode => 8,
             Segment::TxnFields => 256,
-            Segment::TxnData => 256,
```
wborgeaud, line 135: Unrelated to the other fixes but I think txn data should be bytes.

dlubarov (approved):
Thanks for fixing!

PR #963, opened 2023-04-07
Fix call logic

wborgeaud:
Only for sys_call for now. Big TODO left is gas calculation for *CALL.
Some call tests like `callcall_00` give the correct updated state (modulo gas calculations).


evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -12,18 +12,21 @@ global sys_call:
...
     %create_context
     // stack: new_ctx, kexit_info, gas, address, value, args_offset, args_size, ret_offset, ret_size
+    // TODO: Consider call depth
 
     // Each line in the block below does not change the stack.
     DUP4 %set_new_ctx_addr
     %address %set_new_ctx_caller
     DUP5 %set_new_ctx_value
-    DUP5 DUP5 %address %transfer_eth
+    DUP5 DUP5 %address %transfer_eth %jumpi(panic) // TODO: Fix this panic.
```
dlubarov, line 21: Ah thanks for fixing that. Looks like the same `%transfer_eth` bug occurs further down also.

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -212,3 +222,34 @@ global after_call_instruction:
     EXIT_KERNEL
     // (Old context) stack: new_ctx
 %endmacro
+
+%macro copy_mem_to_calldata
+    GET_CONTEXT
+    %stack (ctx, new_ctx, args_offset, args_size, new_ctx) ->
```
dlubarov, line 228: Hm why take two copies of `new_ctx`? I actually didn't know `%stack` would accept it, maybe we should change it later to require unique names.

dlubarov (approved):
LGTM

evm/src/cpu/kernel/asm/core/call.asm
wborgeaud: Oh good catch, I cleaned that up. I agree we should probably throw an error when there are duplicate names.

PR #965, opened 2023-04-07
Fix `run_constructor`

dlubarov:
Fixes the crash due to a huge allocation in `returndatacopyPythonBug_Tue_03_48_41-1432`.


evm/src/cpu/kernel/asm/core/create.asm
```diff
@@ -87,7 +87,7 @@ global create_common:
 
 run_constructor:
```
wborgeaud, line 88: Also with #961 you can delete `%set_new_ctx_parent_ctx` on line 96 (it's called in `%create_context`) above.

wborgeaud (approved):
Good catch!

PR #966, opened 2023-04-07
Implement arithmetic opcodes on the interpreter side

Nashtare:
This PR implements the remaining (signed) arithmetic operations on the interpreter's side.

### DIFF
--- a/evm/src/cpu/kernel/interpreter.rs
+++ b/evm/src/cpu/kernel/interpreter.rs
@@ -600,17 +600,17 @@ impl<'a> Interpreter<'a> {
 
             let mut x_bytes = [0u8; 32];
             x.to_big_endian(&mut x_bytes);
-            let mut x_bytes = x_bytes[num_bytes_prepend..].to_vec();
+            let x_bytes = x_bytes[num_bytes_prepend..].to_vec();
             let sign_bit = x_bytes[0] >> 7;
 
-            self.push(if sign_bit == 0 {
-                x_bytes.extend_from_slice(&vec![0; num_bytes_prepend]);
-                U256::from_big_endian(&x_bytes)
+            let mut bytes = if sign_bit == 0 {
+                vec![0; num_bytes_prepend]
             } else {
-                let mut tmp = vec![0xff; num_bytes_prepend];
-                tmp.extend_from_slice(&x_bytes);
-                U256::from_big_endian(&tmp)
-            });
+                vec![0xff; num_bytes_prepend]
+            };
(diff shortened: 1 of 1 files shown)


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -809,6 +918,70 @@ impl<'a> Interpreter<'a> {
...
+
+    match (x_is_pos, y_is_pos) {
+        (true, true) => x.cmp(&y),
+        (true, false) => Ordering::Greater,
+        (false, true) => Ordering::Less,
+        (false, false) => x.cmp(&y).reverse(),
+    }
+}
+
+const MINUS_ZERO: U256 = U256([
```
dlubarov, line 957: Hm the yellowpaper mentions two's complement. If I'm not mistaken, only the all-zero-bits encoding is interpreted as zero, while this value would be interpreted as 2^-255 (which is special in that there's no positive 2^255).

Nashtare: Ah mb. I'll adapt the methods taking into account the two's complements.

Nashtare: I should have adapted all methods with two's complement!

Nashtare: Should be good for review!

Nashtare: I'll also add some tests for all those signed arith. operations in the related `tests` module, for consistency.

dlubarov (approved):
LGTM otherwise 👍

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -513,6 +577,43 @@ impl<'a> Interpreter<'a> {
...
+            let n = n.low_u64() as usize;
+            let num_bytes_prepend = 31 - n;
+
+            let mut x_bytes = [0u8; 32];
+            x.to_big_endian(&mut x_bytes);
+            let mut x_bytes = x_bytes[num_bytes_prepend..].to_vec();
+            let sign_bit = x_bytes[0] >> 7;
+
+            self.push(if sign_bit == 0 {
+                x_bytes.extend_from_slice(&vec![0; num_bytes_prepend]);
```
dlubarov, line 607: It looks like here the sign bits are appended after x, should it be prepended before?

Nashtare: Yes you're right, I've messed up the ordering here. Sorry about that, fixed!

PR #969, opened 2023-04-10
CALL gas

wborgeaud:
Add gas calculation for `sys_call`. Will do the other *call opcodes in another PR.
Makes some call tests pass (e.g. `(Non)ZeroValue_CALL, callcall_00`).

### DIFF
--- a/evm/src/cpu/kernel/asm/core/call.asm
+++ b/evm/src/cpu/kernel/asm/core/call.asm
@@ -1,82 +1,24 @@
 // Handlers for call-like operations, namely CALL, CALLCODE, STATICCALL and DELEGATECALL.
 
-%macro call_charge_gas
-    // stack: cold_access, address, gas, kexit_info, value, args_offset, args_size, ret_offset, ret_size
-    %mul_const(@GAS_COLDACCOUNTACCESS_MINUS_WARMACCESS)
-    %add_const(@GAS_WARMACCESS)
-    // stack: Caaccess, address, gas, kexit_info, value, args_offset, args_size, ret_offset, ret_size
-    DUP5 ISZERO PUSH 1 SUB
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -250,3 +257,78 @@ global after_call_instruction:
     %jump(memcpy)
 %%after:
 %endmacro
+
+// Charge gas for *call opcodes and return the sub-context gas limit.
+// Doesn't include memory expansion costs.
+%macro call_charge_gas
+    // Compute C_aaccess
+    // stack: cold_access, address, gas, kexit_info, value, args_offset, args_size, ret_offset, ret_size
```
dlubarov, line 265: Am I reading it right that `value` is the deepest stack item this uses? If so I think we can remove the others - we usually only comment on args that a macro uses

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -250,3 +257,78 @@ global after_call_instruction:
...
+    %mul_const(@GAS_CALLSTIPEND) ADD
+    %stack (C_callgas, address, gas, kexit_info, value, args_offset, args_size, ret_offset, ret_size) ->
+        (kexit_info, C_callgas, address, value, args_offset, args_size, ret_offset, ret_size)
+%endmacro
+
+// Checked memory expansion.
+%macro checked_mem_expansion
+    // stack: size, offset, kexit_info
+    DUP1 ISZERO %jumpi(%%zero)
+    ADD
```
dlubarov, line 326: Let's add a TODO to use overflow-safe addition?

dlubarov (approved):
LGTM

PR #970, opened 2023-04-10
Env instructions in interpreter

Nashtare:
This PR implements some environment instructions on the interpreter side.

### DIFF
--- a/evm/src/cpu/kernel/interpreter.rs
+++ b/evm/src/cpu/kernel/interpreter.rs
@@ -340,8 +340,8 @@ impl<'a> Interpreter<'a> {
             0x35 => self.run_calldataload(),                            // "CALLDATALOAD",
             0x36 => self.run_calldatasize(),                            // "CALLDATASIZE",
             0x37 => self.run_calldatacopy(),                            // "CALLDATACOPY",
-            0x38 => todo!(),                                            // "CODESIZE",
-            0x39 => todo!(),                                            // "CODECOPY",
+            0x38 => self.run_codesize(),                                // "CODESIZE",
+            0x39 => self.run_codecopy(),                                // "CODECOPY",
             0x3a => self.run_gasprice(),                                // "GASPRICE",
             0x3b => todo!(),                                            // "EXTCODESIZE",
             0x3c => todo!(),                                            // "EXTCODECOPY",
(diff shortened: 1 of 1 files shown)


Nashtare:
I'm not too sure for some of them, still familiarizing myself with this section. I also noticed `SELFBALANCE` was missing from the opcode list, is there any reason for that?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -609,6 +609,32 @@ impl<'a> Interpreter<'a> {
...
+    fn run_address(&mut self) {
+        self.push(
+            self.generation_state.memory.contexts[self.context].segments
+                [Segment::ContextMetadata as usize]
+                .get(ContextMetadata::Address as usize),
+        )
+    }
+
+    fn run_balance(&mut self) {
+        let offset = self.pop().as_usize();
```
dlubarov, line 621: For this one I think we'd need to hash the address (to get the state trie key) and descend the trie based on that key. Maybe put a TODO for now?

(I don't think it's necessarily a requirement for the interpreter to support all opcodes - we use it for more lightweight unit tests, but we could decide to only exercise things like `BALANCE` in integration tests with the "real" zkEVM. So just calling it unsupported could be an option also.)

Nashtare: Probably better to just close this PR for now I think, I believe there's additional tweaks to be done on some others, and it's indeed not a huge priority.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -659,6 +685,103 @@ impl<'a> Interpreter<'a> {
...
+    }
+
+    fn run_extcodecopy(&mut self) {
+        let dest_offset = self.pop().as_usize();
+        let offset = self.pop().as_usize();
+        let size = self.pop().as_usize();
+        for i in 0..size {
+            let extcode_byte = self.generation_state.memory.mload_general(
+                0,
+                Segment::KernelAccountCode,
```
dlubarov, line 740: I think this should be `self.context` and `Segment::Code`

evm/src/cpu/kernel/interpreter.rs
dlubarov: I think the others appear fine to me, except the couple others that call `get_trie_data`

evm/src/cpu/kernel/interpreter.rs
Nashtare: I guess I got confused by [this line](https://github.com/mir-protocol/plonky2/blob/b721236eb3a9e069d79de3f37dd384c2dafd08e9/evm/src/cpu/kernel/asm/account_code.asm#L130) in the kernel

Nashtare: Thanks for the explanation William, and yes indeed I had missed the first pop. Following your recommendation I'll leave it unimplemented.

evm/src/cpu/kernel/interpreter.rs
Nashtare: I've left them as todo as you suggested for now, I may get back to them in a follow-up PR

dlubarov (approved):
LGTM

PR #971, opened 2023-04-10
Remove is_final_block column in KeccakSpongeStark

Nashtare:
Following Daniel's suggestion in #950.

### DIFF
--- a/evm/src/keccak_sponge/keccak_sponge_stark.rs
+++ b/evm/src/keccak_sponge/keccak_sponge_stark.rs
@@ -379,9 +379,6 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for KeccakSpongeS
         // Ensure that full-input block and final block flags are not set to 1 at the same time.
         yield_constr.constraint(is_final_block * is_full_input_block);
 
-        // If this is a full-input block, is_final_input_len should contain all 0s.
-        yield_constr.constraint(is_full_input_block * is_final_block);
-
         // If this is the first row, the original sponge state should be 0 and already_absorbed_bytes = 0.
         let already_absorbed_bytes = local_values.already_absorbed_bytes;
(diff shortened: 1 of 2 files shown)


evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -479,7 +473,13 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for KeccakSpongeS
         );
         yield_constr.constraint(builder, constraint);
 
-        let is_final_block = local_values.is_final_block;
+        let mut is_final_block = builder.add_extension(
```
dlubarov, line 476: I believe starting with an accumulator of zero should give the same performance; addition with zero should be skipped by `arithmetic_extension_special_cases`. There's also an `add_many_extension` method we could use for brevity.

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -382,12 +379,8 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for KeccakSpongeS
         // Ensure that full-input block and final block flags are not set to 1 at the same time.
         yield_constr.constraint(is_final_block * is_full_input_block);
```
dlubarov, line 380: Looks like this is now the same as the next constraint below, we can remove one or the other :)

dlubarov (approved):
LGTM otherwise, nice simplification

evm/src/keccak_sponge/keccak_sponge_stark.rs
Nashtare: I opted for `add_many_extension`, thanks for the tip.

evm/src/keccak_sponge/keccak_sponge_stark.rs
Nashtare: Updated!

PR #972, opened 2023-04-11
Implement rest of *CALL opcodes

wborgeaud:
Finish implementations of `sys_delegatecall, sys_staticcall, sys_callcode`. Mostly copy-pasted from `sys_call`.


dlubarov (approved):
Nice, LGTM! I think we still need some `set_new_ctx_code_size` calls but but I guess that can come later

PR #973, opened 2023-04-11
Check if context is static for state-changing opcodes

wborgeaud:


evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -2,7 +2,15 @@
 
 // Creates a new sub context and executes the code of the given account.
 global sys_call:
+    // Check that the value is zero if the context is static.
     // stack: kexit_info, gas, address, value, args_offset, args_size, ret_offset, ret_size
+    DUP4 ISZERO %not_bit
```
dlubarov, line 7: nit: seems better to do `ISZERO ISZERO` for a is-nonzero check, vs `1 - is_zero`? I guess we could change `not_bit` to use `ISZERO`

dlubarov (approved):
LGTM

evm/src/cpu/kernel/asm/core/call.asm
wborgeaud: Good idea, I changed `%not_bit`.

PR #974, opened 2023-04-11
Set `stack_len_bounds_aux` properly

hratoanina:
Makes the stack_bounds constraints pass.

I'm not really satisfied with putting the computation of `stack_len_bounds_aux` in `try_perform_instruction`. It feels like it should be handled by the individual `generate_[op]` functions, but this would introduce a lot of duplicate code. Maybe in the beginning of `perform_op`?


nbgl (approved):
Looks good. Thank you!

PR #978, opened 2023-04-12
Enable stack constraints

4l0n50:
Makes stack constraints pass by changing the shift left/right behaviors such that `disable_other_channels = false` (those channels are already constrained in `shift.rs`). It also sets the number of pops of `BASIC_TERNARY_OP` to 3.


nbgl (approved):
Thank you!!

Issue #585, opened 2022-06-27
Support deserializing proofs from slices


Sladuca:
The proof serialization [`Buffer`](https://github.com/mir-protocol/plonky2/blob/3346d3f902b6b71eeef58e1dab17e2d5b8c956d6/plonky2/src/util/serialization.rs#L27) seems to only support deserialization from a `Vec<u8>`. In some environments (e.g. Solana BPF) copying into an owned `Vec<u8>` is too expensive.

May I add a read-only `RoBuffer` struct that uses a slice instead like I have [here](https://github.com/proxima-one/plonky2/blob/a356fffd97fad34b74a51c1dd71e189f56f24365/plonky2/src/util/serialization.rs#L618)?

dlubarov:
I think that sounds good. But to avoid duplicating the read methods, I think we should remove the read methods from `Buffer` (and rename to something like `ProofWriter`), so that reads always deal with slices. Don't think there's any performance reason to read from a `Vec` rather than a slice.

cc @wborgeaud who wrote the code, in case you see an issue

Nashtare:
I believe this can be closed?

PR #981, opened 2023-04-12
Serialization support for all circuit components

Nashtare:
This PR introduces serialization support for gates, generators, and circuit component (with primary goal here to support `AllRecursiveCircuits` serialization).

The approach builds over #670, reusing the same approach for gate serialization and a similar one for generator serialization. The `DummyProofGenerator` is a specific one which requires different handling, with associated documentation in the serializer macro description.

It doesn't address the suggestion in #585 yet, hence the specific `to_buffer` / `from_buffer` methods on the evm/ side for the internals of `AllRecursiveCircuits` to remove the copies. I assumed it could be handled separately on a different PR.

The PR also includes an additional serialization check on the `square_root` example, which introduces a new generator outside of the plonky2 crate, to highlight flexibility for projects using plonky2 as backend but with a need for additional gates / generators.

@Sladuca given the work on the gate serialization was originally yours, I took the liberty to mention you as co-author of commit 8b6f75eafb9951743415193a6ad951630a961992, let me know if that's ok for you.


plonky2/src/iop/generator.rs
```diff
@@ -106,6 +110,52 @@ pub trait WitnessGenerator<F: Field>: 'static + Send + Sync + Debug {
...
+impl<F: Field> PartialEq for WitnessGeneratorRef<F> {
+    fn eq(&self, other: &Self) -> bool {
+        let mut buf1 = Vec::new();
+        let mut buf2 = Vec::new();
+        self.0.serialize(&mut buf1).unwrap();
+        other.0.serialize(&mut buf2).unwrap();
+
+        buf1 == buf2
+    }
+}
```
Nashtare, line 149: Definitely not a fan of this. We could alternatively just check that both "id" match, but that would only enforce that we use the same ordering of generators. Maybe that would be acceptable?

plonky2/src/gates/gate.rs
```diff
@@ -191,9 +198,19 @@ pub trait Gate<F: RichField + Extendable<D>, const D: usize>: 'static + Send + S
     }
 }
 
+pub trait AnyGate<F: RichField + Extendable<D>, const D: usize>: Gate<F, D> {
```
npwardberkeley, line 201: This is just to allow gate type checking in `get_gate_tag_impl`, right? I think it'd be nice to have a comment explaining that

plonky2/src/iop/generator.rs
```diff
@@ -106,6 +110,52 @@ pub trait WitnessGenerator<F: Field>: 'static + Send + Sync + Debug {
...
     fn run(&self, witness: &PartitionWitness<F>, out_buffer: &mut GeneratedValues<F>) -> bool;
+
+    fn serialize(&self, dst: &mut Vec<u8>) -> IoResult<()>;
+
+    fn deserialize(src: &mut Buffer) -> IoResult<Self>
+    where
+        Self: Sized;
+}
+
+pub trait AnyWitnessGenerator<F: Field>: WitnessGenerator<F> {
```
npwardberkeley, line 121: Why do we need this trait, instead of having `WitnessGeneratorRef` wrap `Box<dyn WitnessGenerator>`?

Nashtare: That should suffice indeed. Initially I was puzzled by how to deal here the best way, given that it had to handle both WitnessGenerator and SimpleGenerator, but then I fell back to the "easy" hack around of generators ids and Default impl to get to the SimpleGenerator trait.

Nashtare: You're right this one is a leftover from a previous attempt, I can remove it.

plonky2/src/iop/generator.rs
Nashtare: Given that checking equality by "id" was already done for gates, I opted for this as well here.

plonky2/src/iop/generator.rs
Nashtare: I applied your comments, in case we opt to keep going for an approach like this one.

npwardberkeley (approved):

PR #983, opened 2023-04-13
Precompiles interface

wborgeaud:
Implement interface to call the existing precompile assembly.
Still TODO:
- EXPMOD
- SNARKV
- BLAKE2_F
- Add support for EOA -> Precompile txn.


evm/src/cpu/kernel/asm/core/precompiles/sha256.asm
```diff
@@ -0,0 +1,40 @@
...
+    %num_bytes_to_num_words
+    // stack: data_words_len, kexit_info
+    %mul_const(@SHA256_DYNAMIC_GAS)
+    PUSH @SHA256_STATIC_GAS
+    ADD
+    // stack: gas, kexit_info
+    %charge_gas
+
+    // TODO: fix this
+    %zero_out_kernel_general
```
wborgeaud, line 20: @npwardberkeley Tests don't pass without this. Is this expected?  
Also in general, since precompiles are run in their own context, we could just store data in the current context instead of the kernel's context to avoid clobbering.

evm/src/cpu/kernel/asm/core/call.asm
```diff
@@ -29,19 +29,22 @@ global sys_call:
...
     // TODO: Consider call depth
 
+    %stack (new_ctx, kexit_info, callgas, address, value, args_offset, args_size, ret_offset, ret_size) ->
+          (new_ctx, args_offset, args_size, new_ctx, kexit_info, callgas, address, value, args_offset, args_size, ret_offset, ret_size)
+    %copy_mem_to_calldata
+    // stack: new_ctx, kexit_info, callgas, address, value, args_offset, args_size, ret_offset, ret_size
+    DUP5 DUP5 %address %transfer_eth %jumpi(panic) // TODO: Fix this panic.
+    DUP3 %set_new_ctx_gas_limit
+    %set_new_ctx_parent_pc(after_call_instruction)
+    %handle_precompiles
```
dlubarov, line 39: Do you think there's a nice way we could call this with a smaller stack, ideally with only the data a precompile needs?

It feels a little error-prone if this same (kinda large) stack state must appear in each `sys_*call` and each precompile, and might be tricky if we wanted to change the stack layout.

I'm not sure if there's an elegant solution though; if not this is fine.

evm/src/cpu/kernel/asm/core/precompiles/main.asm
```diff
@@ -0,0 +1,133 @@
...
+    PUSH %%after
+    DUP5
+    // stack: address, %%after, new_ctx, kexit_info, gas, address, value, args_offset, args_size, ret_offset, ret_size
+    %jump(handle_precompiles)
+%%after:
+    // stack: new_ctx, kexit_info, callgas, address, value, args_offset, args_size, ret_offset, ret_size
+%endmacro
+
+global handle_precompiles:
+    // stack: addr, retdest
```
dlubarov, line 12: Maybe we should include the full stack state here, or `...` to hint that there's more. I think it's fine to omit stack items that aren't relevant, but in my mind they're relevant/required here since the precompiles expect that stack state.

evm/src/cpu/kernel/asm/core/precompiles/main.asm
```diff
@@ -0,0 +1,133 @@
...
+    DUP1 %eq_const(@ID)     %jumpi(precompile_id)
+    DUP1 %eq_const(@EXPMOD) %jumpi(precompile_expmod)
+    DUP1 %eq_const(@BN_ADD) %jumpi(precompile_bn_add)
+    DUP1 %eq_const(@BN_MUL) %jumpi(precompile_bn_mul)
+    DUP1 %eq_const(@SNARKV) %jumpi(precompile_snarkv)
+    %eq_const(@BLAKE2_F) %jumpi(precompile_blake2_f)
+    // stack: retdest
+    JUMP
+
+global pop_and_return_success:
```
dlubarov, line 25: I think a stack comment here would help, would it be `_unused, kexit_info`?

evm/src/cpu/kernel/asm/core/precompiles/rip160.asm
```diff
@@ -0,0 +1,37 @@
+global precompile_rip160:
+    %stack (address, retdest, new_ctx, kexit_info, gas, address, value, args_offset, args_size, ret_offset, ret_size) ->
+        (new_ctx, kexit_info, ret_offset, ret_size)
+    DUP1
+    SET_CONTEXT
+    // stack: (empty)
+    PUSH 0x100000000 // = 2^32 (is_kernel = true)
```
dlubarov, line 7: If we create a `kexit_info` to use with `%charge_gas`, I think we need to initialize `kexit_info` with the gas that the parent context gave us (from metadata)?

If it's easier we could probably skip the `is_kernel` bit, as it should only be needed if we call a syscall. (In fact I think we can probably get rid of the `is_kernel` bit entirely, as AFAIK we don't have any kernel code that calls a syscall, and doing so may be weird since most syscalls charge gas.)

Alternatively, we could not construct `kexit_info`, but directly calculate `leftover_gas = our_gas_from_metadata - gas_to_charge`. Since that wouldn't involve `%charge_gas`, we'd need to check for out-of-gas ourselves, but it might still be a tad simpler than constructing `kexit_info`?

dlubarov (approved):
LGTM otherwise!

evm/src/cpu/kernel/asm/core/call.asm
wborgeaud: Do you mean like DUPing `address, new_ctx, kexit_info, ret_offset, ret_size` (the info used by the precompiles) in front of the stack before calling `%handle_precompiles`? If so, yes that would work.

evm/src/cpu/kernel/asm/core/precompiles/rip160.asm
wborgeaud: > If we create a kexit_info to use with %charge_gas, I think we need to initialize kexit_info with the gas that the parent context gave us (from metadata)?

Hm I'm confused. I thought `kexit_info>>192` holds the gas used so far in this context (which is 0 here) and that the gas limit is stored in the context metadata. Is that not the case?

> If it's easier we could probably skip the is_kernel bit, as it should only be needed if we call a syscall. (In fact I think we can probably get rid of the is_kernel bit entirely, as AFAIK we don't have any kernel code that calls a syscall, and doing so may be weird since most syscalls charge gas.)

Ah I was wondering about this. Maybe I can keep it like this until we get rid of the `is_kernel` bit, just to make it less confusing.

evm/src/cpu/kernel/asm/core/call.asm
dlubarov: Yeah, and I guess `handle_precompiles` could pop those if no precompile was executed. I don't think it'd save any code (might add a little), but feels a bit nicer to me if precompiles only receive the info they require. It's minor though, can keep it as is you prefer.

evm/src/cpu/kernel/asm/core/precompiles/rip160.asm
dlubarov: Oh you're right, forgot about that. Seems fine as is then.

evm/src/cpu/kernel/asm/core/call.asm
wborgeaud: Yeah I agree it's nicer. I've made the change 👍

evm/src/cpu/kernel/asm/core/precompiles/sha256.asm
npwardberkeley: Oh sorry forgot to respond! Yes I think mu code currently does assume that the memory starts with zeroes — lmk if I should change that!

PR #984, opened 2023-04-14
Fix generic const expressions warning

unzvfu:
Pull request #873 introduced the use of the incomplete `generic_const_exprs` feature ([here](https://github.com/mir-protocol/plonky2/pull/873/files#diff-ab60f6df9f439d7609ceba5ffaa210cfdee8e1c25fae13c3e2c799e0b3c92d04R4)) but did not disable the compiler warning for it, which seems to have been causing CI failures on main and branches since. This PR disables the warning for incomplete features in Plonky2.


dlubarov (approved):
LGTM

PR #985, opened 2023-04-15
Provide methods for serializing Kernel

Nashtare:
Introduces methods for serializing / deserializing the Kernel.
Given the time it takes to bootstrap now with the introduction of the precompile files from #983 (>100sec on my machine), it may come in handy to skip the process when working on other parts of the crate.

### DIFF
--- a/evm/Cargo.toml
+++ b/evm/Cargo.toml
@@ -36,6 +36,7 @@ serde = { version = "1.0.144", features = ["derive"] }
 static_assertions = "1.1.0"
 hashbrown = { version = "0.12.3" }
 tiny-keccak = "2.0.2"
+serde_json = "1.0"
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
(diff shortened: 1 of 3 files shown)


dlubarov:
Sometimes compilation seems to become slow due to a couple particular uses of `%stack`, which searches for the optimal sequence of instructions to rearrange the stack. I think this problem should go away permanently if we abort long searches like [this](https://github.com/mir-protocol/plonky2/pull/987) (unfinished, just the high level idea).

dlubarov:
I think the performance impact of the fallback should be very small, as there are only a few `%stack` uses that take more iterations than that (mainly two which are taking ~99% of the time), and they're in code that isn't performance-sensitive. I think an approximate algorithm can also get pretty close to optimal; our teammate Angus wrote some notes which I've been meaning to revisit.

But I agree there's no harm in just adding `Serialize`/`Deserialize` if it's useful, especially short term before we have that fallback in place.

PR #989, opened 2023-04-17
Unify proof type + link public values

Nashtare:
This PR unifies the proof type in `fixed_recursive_verifier`, so that we deal with the same `EvmProof` at any level of the tree, and includes check on the public values trie roots when merging proofs. A few design questions / assumptions I wasn't sure on how you were envisioning things:

- I assumed two proofs to be aggregated to be representing consecutive chain states
- I assumed root proofs may not represent an entire block, hence the need to check consistency across txn / receipt tries as well in the case where block metadata of two proofs to merge match
- ~~For checking that `block_metadata` match, I only considered `block_timestamp`, although we could extend it to the other fields as well.~~
- ~~The above also stores the latest `block_metadata` when aggregating two proofs. Do we want to keep track of all the block metadatas from past proofs?~~
- Given the function definition of `prove_block()`, the consistency check between trie roots assumes the block proof (if any) is always on the left side.

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -58,10 +58,27 @@ where
     pub public_values: AggregatedPublicValues,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct EvmProofTarget<const D: usize> {
     pub proof: ProofWithPublicInputsTarget<D>,
     pub public_values: AggregatedPublicValuesTarget,
+}
+
+impl<const D: usize> EvmProofTarget<D> {
+    pub fn to_buffer(&self, buffer: &mut Vec<u8>) -> IoResult<()> {
+        buffer.write_target_proof_with_public_inputs(&self.proof)?;
+        self.public_values.to_buffer(buffer)
+    }
+
+    pub fn from_buffer(buffer: &mut Buffer) -> IoResult<Self> {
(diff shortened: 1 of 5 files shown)


Nashtare:
Actually there's an issue in my current block_metadata equality check approach, we'd need at least the leftmost/rightmost pair for an aggregation (whichever the depth) to be able to aggregate further from both sides with another proof.

Nashtare:
Switched to AggregatedPublicValues type to gather info on leftmost / rightmost block_metadata for an aggregation proof to alleviate issue raised in comment above.

dlubarov:
> I assumed root proofs may not represent an entire block

Sorry for the confusing naming, but I think you mean the "aggregation root" proof (not related to `RootCircuitData`)? Our intention was that `RootCircuitData` would typically be for a single txn (although it could be a small chunk of txns in the same EVM proof), while the aggregation root would be for an entire block.

Nashtare:
Ah ok, thanks for the precision!
So we could lighten the current approach on the BlockCircuitData side as we'd always expect "full" aggregation root proofs, i.e. two blocks being merged together, and check only the connection between state tries. For aggreg proofs, we'd still need the conditional check on the txn/receipt tries as we'd perform several aggregations until having a whole block proven.

I was wondering if this approach was something you were considering, or if you were thinking of doing things fundamentally differently?

Nashtare:
@dlubarov I've reverted some changes, using now `PublicValues` as for `AllProof`, assuming aggregations being done only between transactions within a same block, as I think you were envisioning it?

PR #995, opened 2023-04-19
Gas and more for `CREATE(2)`

wborgeaud:


evm/src/cpu/kernel/asm/core/create.asm
```diff
@@ -5,8 +5,13 @@
 // Post stack: address
 global sys_create:
     %check_static
+
+    %stack (kexit_info, value, code_offset, code_len) -> (code_len, code_offset, kexit_info, value, code_offset, code_len)
+    %checked_mem_expansion
     // stack: kexit_info, value, code_offset, code_len
-    // TODO: Charge gas.
+    %charge_gas_const(@GAS_CREATE)
+    // TODO: If using EIP-3860, we should limit and charge gas on `code_len`.
```
wborgeaud, line 13: I'm in general a bit unsure which EIPs we need to implement (this one was added in Shanghai). Does Polygon just use the latest Ethereum fork?

dlubarov: Yeah I've been assuming the latest. I think there are a few things like EIP 1559 that Edge doesn't currently support, but it should be supported at some point, and it would be neat to be able to prove Eth blocks.

dlubarov (approved):
LGTM!

I think we may also need a hook similar to `observe_new_address`, so that the trace generation code can keep track of the newly-created code (keyed by its hash), in case it's later executed in the same trace. Can come later though.

evm/src/cpu/kernel/asm/core/create.asm
wborgeaud: Sounds good! I'll implement this EIP and the others below then 👍

Nashtare: The current supported versions in Edge are these ones I believe:
```
type Forks struct {
	Homestead      *Fork `json:"homestead,omitempty"`
	Byzantium      *Fork `json:"byzantium,omitempty"`
	Constantinople *Fork `json:"constantinople,omitempty"`
	Petersburg     *Fork `json:"petersburg,omitempty"`
	Istanbul       *Fork `json:"istanbul,omitempty"`
	London         *Fork `json:"london,omitempty"`
	EIP150         *Fork `json:"EIP150,omitempty"`
	EIP158         *Fork `json:"EIP158,omitempty"`
	EIP155         *Fork `json:"EIP155,omitempty"`
}
```
(taken from polygon-edge/chain/params.go), with the latest being London. Was there any communication on whether and/or when Shanghai would be supported in Edge?

PR #997, opened 2023-04-19
Restructure BN254 Pairing Precompile to fit EVM Test Spec

typ3c4t:
The API is given by
```
global bn254_pairing:
    // stack: k, inp, out, retdest
```
where 
 - `k` is the number of inputs
 - `out` and `inp` are the `BnPairing` segment offsets for the output and input

Note: each input consists of 6 stack terms---2 for the curve point and 4 for the twisted curve point---and the inputs are placed contiguously

The assembly code currently uses offsets 0-78 for scratch space, after which one can freely place the `out` and `inp`. I recommend placing the latter after the former since it can have arbitrary size. For the sake of clean numbers, perhaps the following two options might make the most sense.
 - `out=88` and `inp=100`
 - `out=100` and `inp=112`


evm/src/cpu/kernel/asm/curve/bn254/curve_arithmetic/pairing.asm
```diff
@@ -0,0 +1,71 @@
...
+    %mul_const(6)
+    ADD
+    // stack:  inp_j=inp+6j, j, k, inp
+    DUP1
+    // stack:  inp_j, inp_j, j, k, inp
+    %load_fp254_2
+    // stack:    P_j, inp_j, j, k, inp
+    %bn_check
+    // stack: valid?, inp_j, j, k, inp
+    %assert_nonzero
```
wborgeaud, line 38: Here we shouldn't panic (which is what `assert_nonzero` does) but return a special value to indicate that the input was malformed. So you could replace this with `ISZERO %jumpi(bn_pairing_bad_input)` and 
```
bn_pairing_bad_input:
     %stack (...) -> (retdest, @U256_MAX)
     JUMP
```

evm/src/cpu/kernel/asm/curve/bn254/curve_arithmetic/pairing.asm
```diff
@@ -0,0 +1,71 @@
...
+///         assert(P.is_valid)
+///         assert(Q.is_valid)
+///     
+///     out = 1
+///     for P, Q in pairs:
+///         out *= miller_loop(P, Q)
+///
+///     return bn254_final_exponent(out)
+
+global bn254_pairing:
```
wborgeaud, line 13: Does this return a bool or the pairing result?

evm/src/cpu/kernel/asm/curve/bn254/curve_arithmetic/pairing.asm
```diff
@@ -0,0 +1,154 @@
...
+///     
+///     out = 1
+///     for P, Q in pairs:
+///         out *= miller_loop(P, Q)
+///
+///     result = bn254_final_exponent(out)
+///     return result == @GENERATOR_PAIRING
+
+global bn254_pairing:
+    // stack:
```
wborgeaud, line 15: So is the expected stack empty when calling this?

typ3c4t: this happened by accident after cutting and pasting events 😂

evm/src/cpu/kernel/tests/bn254.rs
```diff
@@ -308,16 +295,21 @@ fn test_tate() -> Result<()> {
...
         memory: vec![(ptr, inputs)],
     };
     let interpreter = run_interpreter_with_memory(setup).unwrap();
-    let output: Vec<U256> = interpreter.extract_kernel_memory(BnPairing, out..out + 12);
+    // let output: Vec<U256> = interpreter.extract_kernel_memory(BnPairing, out..out + 12);
     let expected = tate(CURVE_GENERATOR, TWISTED_GENERATOR).on_stack();
-
-    assert_eq!(output, expected);
+    println!("{:?}", expected);
+    assert_eq!(interpreter.stack()[0], U256::one());
```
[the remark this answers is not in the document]
typ3c4t: the .asm checks for `== e(P,Q)` and leaves the resultant bool at the top of the stack---what's the point then in checking this with the rust test?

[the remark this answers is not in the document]
wborgeaud: Ah right, I got confused with the `expected` and output `variables`. Maybe we can remove them now that they're unused.

wborgeaud (approved):
Looks good to me!

PR #999, opened 2023-04-20
EIP-3860: Limit and meter initcode

wborgeaud:


evm/src/cpu/kernel/asm/core/intrinsic_gas.asm
dlubarov, line 1: Hm it seems like this extra gas cost is not explicitly defined as contributing to intrinsic gas (by either the EIP or the YP), but it seems to behave the same as if it was part of intrinsic gas? A bit confusing to me, but seems okay to do it like this

dlubarov (approved):
LGTM

evm/src/cpu/kernel/asm/core/intrinsic_gas.asm
wborgeaud: From the EIP https://eips.ethereum.org/EIPS/eip-3860#rules
```
2. For a create transaction, extend the transaction data cost formula to include initcode_cost(initcode). (Note that this is included in transaction intrinsic cost, i.e. transaction with not enough gas to cover initcode cost is invalid.)
```

PR #1000, opened 2023-04-20
EIP-170: Contract code size limit 

wborgeaud:
Also added cost of deploying code in `process_contract_creation_txn_after_constructor`.


dlubarov (approved):
LGTM

PR #1002, opened 2023-04-21
New contract hook

wborgeaud:


evm/src/generation/state.rs
```diff
@@ -93,6 +99,35 @@ impl<F: Field> GenerationState<F> {
...
+
+        let ctx = self.registers.context;
+        let base_address = MemoryAddress::new(ctx, Segment::Returndata, 0);
+        let returndata_size_addr = MemoryAddress::new(
+            ctx,
+            Segment::ContextMetadata,
+            ContextMetadata::ReturndataSize as usize,
+        );
+        let returndata_size = self.memory.get(returndata_size_addr).as_usize();
+        let code = (0..returndata_size)
```
dlubarov, line 117: Could also get the code as

```
self.memory.contexts[ctx]
    .segments[Segment::Returndata as usize]
    .content[..returndata_size]
```

evm/src/generation/state.rs
```diff
@@ -93,6 +99,35 @@ impl<F: Field> GenerationState<F> {
...
+            .map(|i| {
+                let address = MemoryAddress {
+                    virt: base_address.virt.saturating_add(i),
+                    ..base_address
+                };
+                let val = self.memory.get(address);
+                val.as_u32() as u8
+            })
+            .collect::<Vec<_>>();
+
```
dlubarov, line 127: Maybe we should re-hash the code and check that it matches just to be safe? Could be debug-only

dlubarov (approved):

PR #1003, opened 2023-04-21
EIP-3541: Reject new contract code starting with the 0xEF byte

wborgeaud:


evm/src/cpu/kernel/asm/core/process_txn.asm
```diff
@@ -144,6 +144,10 @@ process_contract_creation_txn_after_code_loaded:
 global process_contract_creation_txn_after_constructor:
     // stack: success, leftover_gas, new_ctx, address, retdest
     POP // TODO: Success will go into the receipt when we support that.
+
+    // EIP-3541: Reject new contract code starting with the 0xEF byte
+    PUSH 0 %mload_current(@SEGMENT_RETURNDATA) %eq_const(0xEF) %assert_zero
```
dlubarov, line 149: I think this is meant to be an exception/failure where where the overall block is still valid (unlike malformed txns etc), so we'd still prove the result vs panicking. Maybe add a TODO?

dlubarov (approved):

evm/src/cpu/kernel/asm/core/process_txn.asm
wborgeaud: Ah right! I added a TODO

PR #1006, opened 2023-04-25
Blake2_f function and precompile call

npwardberkeley:
Changes to bring the blake implementation in line with what the yellow paper specifies:
Replaces the blake2b hash function with the blake2_f compression function (which is most of the internal part of the hash function). The rest of blake2b is kept around in case it's useful, but isn't included in the kernel. Adds a new test for blake2_f using test data from the EIP

Also, calling code for blake2_f precompile


evm/src/cpu/kernel/asm/core/precompiles/blake2_f.asm
```diff
@@ -1,3 +1,132 @@
...
+    PUSH blake2_f_contd
+    // stack: blake2_f_contd, kexit_info
+
+    // Load inputs from calldata memory into stack.
+
+    %calldatasize
+    // stack: calldatasize, blake2_f_contd, kexit_info
+    DUP1
+    // stack: calldatasize, calldatasize, blake2_f_contd, kexit_info
+    %assert_eq_const(213)
```
wborgeaud, line 20:
```suggestion
%eq_const(213) ISZERO %jumpi(fault_exception)
```
We don't want to panic here.

evm/src/cpu/kernel/asm/core/precompiles/blake2_f.asm
```diff
@@ -1,3 +1,132 @@
 global precompile_blake2_f:
-    // TODO
-    PANIC
+    // stack: address, retdest, new_ctx, (old stack)
+    %pop2
```
wborgeaud, line 3:
```suggestion
    // stack: retdest, new_ctx, (old stack)
    POP
```
The address isn't on the stack for blake2

evm/src/cpu/kernel/asm/core/precompiles/blake2_f.asm
```diff
@@ -1,3 +1,132 @@
...
+    // stack: calldatasize, blake2_f_contd, kexit_info
+    %decrement
+    // stack: flag_addr=212, blake2_f_contd, kexit_info
+    DUP1
+    // stack: flag_addr, flag_addr, blake2_f_contd, kexit_info
+    PUSH @SEGMENT_CALLDATA
+    GET_CONTEXT
+    // stack: ctx, @SEGMENT_CALLDATA, flag_addr, flag_addr size
+    MLOAD_GENERAL
+    // stack: flag, flag_addr, blake2_f_contd, kexit_info
```
wborgeaud, line 30: We need to check that `flag` is 0 or 1 and go to `fault_exception` otherwise.

evm/src/cpu/kernel/asm/core/precompiles/blake2_f.asm
```diff
@@ -1,3 +1,136 @@
...
+    DUP1
+    // stack: flag_addr, flag_addr, blake2_f_contd, kexit_info
+    PUSH @SEGMENT_CALLDATA
+    GET_CONTEXT
+    // stack: ctx, @SEGMENT_CALLDATA, flag_addr, flag_addr, blake2_f_contd, kexit_info
+    MLOAD_GENERAL
+    // stack: flag, flag_addr, blake2_f_contd, kexit_info
+    DUP1
+    // stack: flag, flag, flag_addr, blake2_f_contd, kexit_info
+    %lt_const(2) ISZERO %jumpi(fault_exception) // Check flag < 2 (flag = 0 or flag = 1)
```
wborgeaud, line 33: nit: Can save the `ISZERO` by doing `%gt_const(1) %jumpi(fault_exception)` instead.

wborgeaud (approved):
Looks good to me!

PR #1009, opened 2023-04-26
expmod precompile calling code

npwardberkeley:


evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,257 @@
...
+    // stack: l_B, len, l_M, l_E, l_B, kexit_info
+    PUSH 96
+    PUSH @SEGMENT_CALLDATA
+    GET_CONTEXT
+    PUSH 0
+    PUSH @SEGMENT_KERNEL_GENERAL
+    PUSH 0
+    // stack: dst=(0, @SEGMENT_KERNEL_GENERAL, b_loc=0), src=(ctx, @SEGMENT_CALLDATA, 96), l_B, len, l_M, l_E, l_B, kexit_info
+    %memcpy
+    // stack: len, l_M, l_E, l_B, kexit_info
```
wborgeaud, line 138: Looking at the debug trace, this stack comment isn't correct. In particular, `kexit_info` is in front of the stack at this point.

npwardberkeley: Might be fixed... but what test do I run to see that?

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,257 @@
...
+    DUP1
+    // stack: len, len, kexit_info
+    %mul_const(3)
+    // stack: out=3*len, len, kexit_info
+    PUSH @SEGMENT_KERNEL_GENERAL
+    PUSH 0
+    PUSH 0
+    PUSH @SEGMENT_RETURNDATA
+    // stack: @SEGMENT_RETURNDATA, 0, 0, @SEGMENT_KERNEL_GENERAL, out, len, kexit_info
+    %mstore_parent_context_metadata(@CTX_METADATA_RETURNDATA_SIZE, 32)
```
wborgeaud, line 249: Shouldn't it be `len` instead of 32?

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
wborgeaud: Yes it's fixed 👍 
I'm running the `stReturnDataTest/modexp_modsize0_returndatasize_0` test.

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
npwardberkeley: Oh right, good catch

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,260 @@
...
+    // stack: m_loc=2*len, len, out, s1, s2, s3, s4, s5, expmod_contd, len, kexit_info
+    SWAP1
+    // stack: len, m_loc, out, s1, s2, s3, s4, s5, expmod_contd, len, kexit_info
+
+    PUSH 0
+    // stack: b_loc=0, e_loc=len, m_loc, out, s1, s2, s3, s4, s5, expmod_contd, len, kexit_info
+    DUP2
+    // stack: len, b_loc, e_loc, m_loc, out, s1, s2, s3, s4, s5, expmod_contd, len, kexit_info
+
+    %jump(modexp_bignum)
```
wborgeaud, line 233: The number on the stack are counting bytes, but they should be counting 128-bit limbs. For example, if the values are 256-bit, currently `len=32` (for 32 bytes) but it should be `len=2` (for 2 128-bit limbs).

npwardberkeley: Ohh okay I fixed that, but a bigger issue is that the calldata is in bytes, whereas my `modexp` code expects its input in limbs — right? So I'll need to pack it before storing in kernel general memory?

npwardberkeley: I added code to pack the inputs and unpack the outputs! It's kinda gnarly, a lot of lines, but I'm not sure how to do it differently. But let me know if you have suggestions to make it cleaner or more readable!

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,376 @@
...
+    // stack: cur_limb, i, cur_address, len, kexit_info
+    DUP2
+    // stack: i, cur_limb, i, cur_address, len, kexit_info
+    %mul_const(16)
+    // stack: offset=16*i, cur_limb, i, cur_address, len, kexit_info
+    %stack (offset, cur_limb) -> (@SEGMENT_RETURNDATA, offset, cur_limb, 16)
+    // stack: @SEGMENT_RETURNDATA, offset, cur_limb, 16, i, cur_address, len, kexit_info
+    %mload_context_metadata(@CTX_METADATA_PARENT_CONTEXT)
+    // stack: parent_ctx, @SEGMENT_RETURNDATA, offset, cur_limb, 16, i, cur_address, len, kexit_info
+    %mstore_unpacking
```
wborgeaud, line 358: `mstore_unpacking` puts a scalar `offset'` on top of the stack which I think you don't account for here.

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,376 @@
...
+    %min_const(16)
+    SWAP1
+    SUB
+    // stack: num_bytes_new = num_bytes - min(16, num_bytes), ctx, segment, offset, retdest, total_num_limbs, len, ..limbs
+    DUP1
+    ISZERO
+    %jumpi(mload_bytes_return)
+    // stack: num_bytes_new, ctx, segment, offset, retdest, total_num_limbs, len, ..limbs
+    %stack (num, addr: 3) -> (addr, num)
+    %jump(mload_bytes_as_limbs)
```
wborgeaud, line 26: Need to update the offset before jumping back to the start of the loop.

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,382 @@
...
+    %jump(modexp_bignum)
+
+expmod_contd:
+    // stack: len, kexit_info
+
+    // Copy the result value from kernel general memory to the parent's return data.
+
+    DUP1
+    // stack: len, len, kexit_info
+    %mstore_parent_context_metadata(@CTX_METADATA_RETURNDATA_SIZE)
```
wborgeaud, line 340: The returndata size should be equal to `l_M` (in bytes).

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,382 @@
...
+    // stack: len, kexit_info
+    DUP1
+    // stack: len, len, kexit_info
+    %mul_const(3)
+    // stack: out=3*len, len, kexit_info
+    PUSH 0
+    // stack: i=0, cur_address=out, len, kexit_info
+expmod_store_loop:
```
wborgeaud, line 348: Should go through the loop in the inverse direction to store the result in big-endian form.

npwardberkeley: Ah right, thanks. I changed it so it goes through in reverse order

Would it be a problem if len/16 > l_M? That should just mean storing extra zeroes in return data (since the actual result of the expmod will necessarily be less than m), going past returndatasize = l_M — I assume that's okay?

wborgeaud: No that's not an issue, as long as `returndatasize = l_M` we can write as many zeros as we want after that.

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,408 @@
...
+    SWAP1
+    // stack: i+1, cur_address-1, end_address, len, kexit_info
+    DUP3
+    DUP2
+    EQ
+    ISZERO
+    %jumpi(expmod_store_loop)
+expmod_store_end:
+    // stack: i, cur_address, len, kexit_info
+    %pop3
```
wborgeaud, line 404: This stack comment is incorrect, it should be `// stack: i, cur_address, end_address, len, kexit_info`. And so we should use `%pop4`.

evm/src/cpu/kernel/asm/core/precompiles/expmod.asm
```diff
@@ -1,3 +1,408 @@
...
+    // stack: kexit_info
+    PUSH 0
+    // stack: dummy=0, kexit_info
+    %jump(pop_and_return_success)
```
wborgeaud, line 408: Could also just return success:
```
// stack: kexit_info
%leftover_gas
// stack: leftover_gas
PUSH 1 // success
%jump(terminate_common)
```

wborgeaud (approved):
Looks good to me!

PR #1013, opened 2023-04-30
Address overflow-related TODOs in ASM code.

LindaGuiga:
This PR addresses the overflow risks mentioned in the discussion [here](https://github.com/mir-protocol/plonky2/pull/930/files/a4ea0965d79561c345e2f77836c07949c7e0bc69#r1143630253).

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/metadata.asm
+++ b/evm/src/cpu/kernel/asm/memory/metadata.asm
@@ -312,3 +312,15 @@ global sys_basefee:
     %mload_context_metadata(@CTX_METADATA_STATIC)
     %jumpi(fault_exception)
 %endmacro
+
+%macro add_or_fault
+    // stack: offset, size, kexit_info, offset, size
+    DUP1
+    %ensure_reasonable_offset
+    // stack: offset, size, kexit_info, offset, size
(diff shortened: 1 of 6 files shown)


evm/src/cpu/kernel/asm/memory/metadata.asm
```diff
@@ -312,3 +312,15 @@ global sys_basefee:
     %mload_context_metadata(@CTX_METADATA_STATIC)
     %jumpi(fault_exception)
 %endmacro
+
+%macro add_or_fault
```
dlubarov, line 316: Hm I was imagining an implementation like

```
sum = x + y
if sum < x
    fault // overflow
return sum
```

I think this implementation is fine too, just less general in a way since `%ensure_reasonable_offset` is about memory offsets/lengths. Maybe we don't need the extra generality, but do you see any downside to the overflow-check approach?

wborgeaud: I also prefer the overflow-check approach since in some cases `offset` can be arbitrarily large if `size` is zero.

dlubarov (approved):
LGTM

PR #1014, opened 2023-05-01
Constrain memory channels at end of bootstrapping

Nashtare:
Fix todo related to constraining memory channels to be disabled at the end of kernel bootstrapping.


dlubarov (approved):
LGTM, thanks

PR #1017, opened 2023-05-02
expmod edge case fix

npwardberkeley:


unzvfu:
Might be worth adding a test to corresponding to this case? Generally good practice to add a test for each bug...

npwardberkeley:
Good idea! But this bug actually was caught by a test case — it’s just that the tests are in a separate branch, `expmod_precompile_tests`. The reason is that I couldn’t test this precompile calling code directly, so for the test cases I duplicated the entire thing with slight changes. I wasn’t sure whether it was worth adding that entire separate file, so William suggested including it in a different branch

PR #1019, opened 2023-05-03
Verify signatures for legacy transactions

Nashtare:
I noticed that signature verification for legacy transactions was implemented but never called, although there may be a reason I'm missing here.


Nashtare:
Nevermind, it is indeed checked already.

PR #1020, opened 2023-05-03
Remove `generic_const_exprs` dependency from field crate

unzvfu:
Like it says on the tin. Turns out the only functions using `generic_const_exprs` were `PackedField::as/from_arr()`, which were only called in the test suite. Removed those functions and replaced with the slice equivalents.


nbgl (approved):
Looks good!!

PR #1021, opened 2023-05-03
Tests on 16-core runners

dlubarov:


dvdplm:
@pgebheim One to close perhaps?

PR #1023, opened 2023-05-08
Fix clippy

Nashtare:
It seems latest nightly is triggering a new error with clippy.


dlubarov (approved):

unzvfu:
Nit: A slightly more elegant fix is
```rust
    let rng_seed = options.seed.unwrap_or_else(|| OsRng.next_u64());
```

PR #1025, opened 2023-05-08
MPT deletion

wborgeaud:
Implement `mpt_delete` with the following assumptions
- Will only be called on keys present in the trie.
- We only want to delete leaf nodes.
- Branch nodes have zero value.

### DIFF
--- a/evm/src/cpu/kernel/asm/mpt/delete/delete.asm
+++ b/evm/src/cpu/kernel/asm/mpt/delete/delete.asm
@@ -11,14 +11,14 @@ global mpt_delete:
     SWAP1 %increment SWAP1
     // stack: node_type, node_payload_ptr, num_nibbles, key, retdest
 
-    DUP1 %eq_const(@MPT_NODE_EMPTY)     %jumpi(panic) // This should never happen.
     DUP1 %eq_const(@MPT_NODE_BRANCH)    %jumpi(mpt_delete_branch)
     DUP1 %eq_const(@MPT_NODE_EXTENSION) %jumpi(mpt_delete_extension)
-         %eq_const(@MPT_NODE_LEAF)      %jumpi(mpt_delete_leaf)
+    DUP1 %eq_const(@MPT_NODE_LEAF)      %jumpi(mpt_delete_leaf)
+         %eq_const(@MPT_NODE_EMPTY)     %jumpi(panic) // This should never happen.
     PANIC
 
 mpt_delete_leaf:
-    // stack: node_payload_ptr, num_nibbles, key, retdest
-    %pop3
+    // stack: node_type, node_payload_ptr, num_nibbles, key, retdest
+    %pop4
     PUSH 0 // empty node ptr
     SWAP1 JUMP
--- a/evm/src/cpu/kernel/asm/mpt/delete/delete.asm
+++ b/evm/src/cpu/kernel/asm/mpt/delete/delete.asm
@@ -1,4 +1,5 @@
 // Return a copy of the given node with the given key deleted.
+// Assumes that the key is in the trie.
 //
 // Pre stack: node_ptr, num_nibbles, key, retdest
 // Post stack: updated_node_ptr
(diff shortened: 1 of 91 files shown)


evm/src/cpu/kernel/asm/mpt/delete/delete.asm
```diff
@@ -1,6 +1,24 @@
...
 global mpt_delete:
-    PANIC // TODO
+    // stack: node_ptr, num_nibbles, key, retdest
+    DUP1 %mload_trie_data
+    // stack: node_type, node_ptr, num_nibbles, key, retdest
+    // Increment node_ptr, so it points to the node payload instead of its type.
+    SWAP1 %increment SWAP1
+    // stack: node_type, node_payload_ptr, num_nibbles, key, retdest
+
+    DUP1 %eq_const(@MPT_NODE_EMPTY)     %jumpi(panic) // This should never happen.
```
hratoanina, line 14: nit: Since it shouldn't happen, this case could be last.

evm/src/cpu/kernel/asm/mpt/delete/delete_branch.asm
```diff
@@ -0,0 +1,151 @@
...
+    DUP4 %add_const(12) %mload_trie_data %append_to_trie_data
+    DUP4 %add_const(13) %mload_trie_data %append_to_trie_data
+    DUP4 %add_const(14) %mload_trie_data %append_to_trie_data
+    DUP4 %add_const(15) %mload_trie_data %append_to_trie_data // Copy child[15]
+    DUP4 %add_const(16) %mload_trie_data %append_to_trie_data // Copy value_ptr
+    // stack: updated_branch_ptr, updated_child_ptr, first_nibble, node_payload_ptr, retdest
+    SWAP1
+    // stack: updated_child_ptr, updated_branch_ptr, first_nibble, node_payload_ptr, retdest
+    DUP2 %increment DUP4 ADD
+    // stack: updated_branch_ptr+first_nibble, updated_child_ptr, updated_branch_ptr, first_nibble, node_payload_ptr, retdest
```
hratoanina, line 58: `updated_branch_ptr+1+first_nibble`

evm/src/cpu/kernel/asm/mpt/delete/delete_branch.asm
```diff
@@ -0,0 +1,151 @@
...
+    // stack: updated_child_ptr, first_nibble, node_payload_ptr, retdest
+    PUSH 0
+// Loop from i=0..16 excluding `first_nibble` and store the number of non-empty children in
+// KernelGeneral[0]. Also store the last non-empty child in KernelGeneral[1].
+loop:
+    // stack: i, updated_child_ptr, first_nibble, node_payload_ptr, retdest
+    DUP1 DUP4 EQ %jumpi(loop_eq_first_nibble)
+    // stack: i, updated_child_ptr, first_nibble, node_payload_ptr, retdest
+    DUP1 %eq_const(16) %jumpi(loop_end)
+    DUP1 DUP5 ADD %mload_trie_data %mload_trie_data ISZERO ISZERO %jumpi(loop_non_empty)
```
hratoanina, line 77: Second `%mload_trie_data` seems redundant: if we have an empty node, it should already be 0.

evm/src/cpu/kernel/asm/mpt/util.asm
```diff
@@ -183,6 +203,17 @@
...
 %endmacro
 
+// Remove the first `k` nibbles from a key part.
+// def merge_nibbles(front_len, front_key, back_len, back_key):
+//     return (front_len + back_len, (front_key<<(back_len*4)) + back_key)
+%macro merge_nibbles
+    // stack: front_len, front_key, back_len, back_key
+    %stack (front_len, front_key, back_len, back_key) -> (back_len, front_key, back_key, back_len, front_len)
+    %mul_const(4) SHL ADD
+    // stack: new_key, back_len, back_key
```
hratoanina, line 213:
```suggestion
    // stack: new_key, back_len, front_len
```

evm/src/cpu/kernel/asm/mpt/delete/delete_extension.asm
```diff
@@ -0,0 +1,96 @@
...
+    %add_const(2) %mload_trie_data
+    %stack (node_child_ptr, key, num_nibbles, node_len, node_key, retdest) ->
+        (node_child_ptr, num_nibbles, key, after_mpt_delete_extension, node_len, node_key, retdest)
+    %jump(mpt_delete)
+
+after_mpt_delete_extension:
+    // stack: updated_child_node_ptr, node_len, node_key, retdest
+    DUP1 %mload_trie_data
+    // stack: child_type, updated_child_node_ptr, node_len, node_key, retdest
+    DUP1 %eq_const(@MPT_NODE_EMPTY)     %jumpi(panic) // This should never happen.
```
hratoanina, line 30: Same nit: could be last case.

evm/src/cpu/kernel/asm/mpt/delete/delete_branch.asm
wborgeaud: This is in case we have an empty node not at the 0 location in the `TrieData`. I don't think it can happen currently, but I think it's better to keep the option open.

Issue #1004, opened 2023-04-23
Endianness discrepancy for KeccakSponge and Memory CTL checks


LindaGuiga:
When trying to fix the Memory CTL checks, we noticed that currently, there is a discrepancy between the way the output of keccak is stored in the KeccakSponge Stark and in Memory Stark. Indeed, the memory currently stores the hashes in Little Endian, whereas KeccakSponge stores them in Big Endian, which results in a failed CTL check. 
This is related to commit 0529fa0, where, in `generate_keccak_general`, KeccakSponge CTL checks were solved by changing the way Cpu row values were written so that they matched KeccakSponge rows, but Memory row values were kept unchanged:
```
    let val_u64s: [u64; 4] =
        core::array::from_fn(|i| u64::from_le_bytes(core::array::from_fn(|j| hash.0[i * 8 + j])));
    let hash_int = U256(val_u64s);

    let mut log_push = stack_push_log_and_fill(state, &mut row, hash_int)?;
    log_push.value = hash.into_uint();
```
The first obvious solution would be to change the endianness in Memory (since it is already changed in Cpu). However, this would result in a failed assertion later on. More specifically:
- The state tries are indexed by a hash of the account address, based on the hash that is read in memory. The key is therefore the hash in Little Endian form.
- When accessing the account, the hash of the account address is computed using the ASM version of Keccak, where the result is in the Big Endian form.
- Searching for the account in the state tries therefore results in an assertion error.

To fix the problem without changing the way Keccak outputs are computed, we thought about adding columns in KeccakSponge to store the bytes of the 8 first outputs (4 bytes per u32 limb, or 32 columns in total). This way, we can compare the values in KeccakSponge, Cpu and Memory thanks to the byte values. Do you have a better solution in mind?

Nashtare:
I guess this can be closed now.

LindaGuiga:
Memory CTL and endianness are handled in [this PR](https://github.com/mir-protocol/plonky2/pull/1026)

Issue #988, opened 2023-04-17
Missing CPU rows for global metadata writes


hratoanina:
One of the reasons why the Memory CTL produces an invalid proof is because there are some memory writes added to the Memory Stark trace, but without corresponding writes in the CPU trace. These writes are about global metadata, and are done right before boostrapping the kernel in the CPU trace:
https://github.com/mir-protocol/plonky2/blob/5dfac7155f4717674a17ec00038f867d4e73bf98/evm/src/generation/mod.rs#L95-L105

The obvious solution is to add CPU rows with write operations somewhere in the CPU trace, but there are issues with this:
- In bootstrapping mode, constraints force all rows to write kernel code in the code segment of the kernel context, and nothing else.
- After bootstrapping, the CPU is forced to start by running the kernel `main` function, and afterwards PC constrains the execution flow.

We're not sure how to solve it, but here's two ideas we had:
- Adding another filter column during bootstrapping solely for these writes.
- Adding global metadata to `PROVER_INPUT` so that the writes could be added in ASM instructions right at the beginning of `main`.

Do you have any thoughts on these approaches, or maybe another one entirely?

hratoanina:
A verifier bus sounds like a more versatile solution (though as far as I know the verifier shouldn't need to interact with the memory any more than these reads?). Using the CPU sounded like the easiest way to implement it, and since the metadata is part of the Stark public values, it feels like there should be an organic way to interact with it in ASM the same way you can interact with (private) input like the MPTs.

We're also not specifically working on this right now (there's also another issue in the Memory CTL check), but we may in the coming days. We'll let you know if we do!

LindaGuiga:
As an update: we are currently working on this issue, and are considering two approaches:
- In the first, the verifier computes the missing values in the CTL products (since the verifier has access to the CTL challenges and the metadata). The verifier can then multiply the obtained product to the rest of the looking evaluations. I believe this resembles what you mentioned.
- In the second, we consider using `PROVER_INPUT` at the very beginning of  `main.asm` to add the metadata, instead of using `apply_metadata_memops`. We believe that in this case, it would suffice to link the prover private input values to the public inputs.

What do you think of these approaches?

dlubarov:
I figure with both approaches, the verifier receives metadata as inputs, and needs a way to convince itself that the given inputs truly correspond to data in memory. With the first approach, that assurance just comes from the fact that the verifier "personally" writes them to memory.

[...]

PR #1026, opened 2023-05-08
Fix the memory CTL and implement the verifier memory bus

hratoanina:
This pull request fixes the memory CTL so that it doesn't have to be disabled anymore.
It also introduces a verifier-only memory bus to check a small set of memory operations.

## `KeccakSponge` endianness
Related issue: https://github.com/mir-protocol/plonky2/issues/1004
In the `KeccakSponge` STARK, the output of the hash is stored in `u32` columns written in big endian format. This introduces a discrepancy with how the columns are written in the `Memory` STARK, and makes the CTL check fail. The easiest solution to implement we found was to add columns in `KeccakSponge` for individual bytes of the output, which can be recombined in little endian format.

## Verifier-only memory bus
Following Daniel's advice [here](https://github.com/mir-protocol/plonky2/issues/988#issuecomment-1512002282), we introduced a set of memory operations which can be directly checked by the verifier by multiplying their contribution to the looked product. For the sake of generality, these extra values can be added to any CTL product though we only need it for memory.
For now, the operations checked by the verifier are the block metadata writes before kernel bootstrapping and the trie hash roots reads at the end of the execution.

Further work:
- To make sure they are consistent, the trie hash roots should be public inputs of the SNARK wrapper. I believe this is not the case yet.
- For the trie hash roots reads, we need the CPU trace length of the underlying CPU STARK. For now it's provided manually, but it should be propagated from the base CPU STARK to its SNARK wrapper, and all the way through the SNARK recursion chain to the root circuit. It could be handled with public inputs, or maybe with another verifier-only read of a global variable keeping track of the number of steps of the CPU?

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -542,21 +542,29 @@ pub(crate) fn eval_cross_table_lookup_checks_circuit<
 pub(crate) fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize>(
     cross_table_lookups: &[CrossTableLookup<F>],
     ctl_zs_lasts: [Vec<F>; NUM_TABLES],
+    ctl_extra_looking_products: Vec<Vec<F>>,
     config: &StarkConfig,
 ) -> Result<()> {
     let mut ctl_zs_openings = ctl_zs_lasts.iter().map(|v| v.iter()).collect::<Vec<_>>();
-    for CrossTableLookup {
-        looking_tables,
-        looked_table,
-    } in cross_table_lookups.iter()
+    for (
+        CrossTableLookup {
+            looking_tables,
+            looked_table,
+        },
+        extra_product_vec,
+    ) in cross_table_lookups
+        .iter()
+        .zip(ctl_extra_looking_products.iter())
     {
-        for _ in 0..config.num_challenges {
-            let looking_zs_prod = looking_tables
+        for c in 0..config.num_challenges {
+            let mut looking_zs_prod = looking_tables
                 .iter()
                 .map(|table| *ctl_zs_openings[table.table as usize].next().unwrap())
                 .product::<F>();
-            let looked_z = *ctl_zs_openings[looked_table.table as usize].next().unwrap();
 
+            looking_zs_prod *= extra_product_vec[c];
+
+            let looked_z = *ctl_zs_openings[looked_table.table as usize].next().unwrap();
             ensure!(
                 looking_zs_prod == looked_z,
                 "Cross-table lookup verification failed."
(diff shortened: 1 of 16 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -452,11 +464,50 @@ where
...
         }
 
+        // Extra products to add to the looked last value
+        let mut extra_looking_products = Vec::new();
+        for _ in 0..NUM_TABLES {
+            extra_looking_products.push(Vec::new());
+        }
+
+        // Arithmetic
+        for _ in 0..stark_config.num_challenges {
```
npwardberkeley, line 474: Can this be done using `extra_looking_products.push(vec![builder.constant(F::ONE); stark_config.num_challenges]);`? Or even combining the four loops into one `vec![vec![...]]` expression?

hratoanina: It could be, the idea is that specific functions for the other CTLs could be added here. I can't see any use case now or in the future but I wanted to leave the option to.

evm/src/cross_table_lookup.rs
```diff
@@ -546,21 +537,29 @@ pub(crate) fn eval_cross_table_lookup_checks_circuit<
...
-        for _ in 0..config.num_challenges {
-            let looking_zs_prod = looking_tables
+        for c in 0..config.num_challenges {
+            let mut looking_zs_prod = looking_tables
                 .iter()
                 .map(|table| *ctl_zs_openings[table.table as usize].next().unwrap())
                 .product::<F>();
-            let looked_z = *ctl_zs_openings[looked_table.table as usize].next().unwrap();
 
+            looking_zs_prod *= extra_product_vec[c];
```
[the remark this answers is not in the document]
hratoanina: Whoops, I misunderstood your comment. You mean doing the multiplication directly with the product to get rid of the mut, right?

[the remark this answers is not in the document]
npwardberkeley: Yes exactly!

hratoanina:
@npwardberkeley Hi Nick, just checking up on this PR. Is there anything you need from us?

evm/src/verifier.rs
```diff
@@ -106,13 +110,163 @@ where
...
+    let public_values = all_proof.public_values;
+
+    // Extra products to add to the looked last value.
+    let mut extra_looking_products = Vec::new();
+    for _ in 0..NUM_TABLES {
+        extra_looking_products.push(Vec::new());
+    }
+
+    // Arithmetic
+    for _ in 0..config.num_challenges {
```
npwardberkeley, line 122: Same with here

evm/src/fixed_recursive_verifier.rs
```diff
@@ -499,6 +550,208 @@ where
...
+            // context
+            builder.connect(row[1], zero);
+            // segment
+            builder.connect(row[2], segment);
+            // virtual
+            let field_target = builder.constant(F::from_canonical_usize(field as usize));
+            builder.connect(row[3], field_target);
+            // values
+            for j in 0..VALUE_LIMBS {
+                builder.connect(row[j + 4], targets[j]);
```
npwardberkeley, line 743: nit: `4 + j` for consistency

npwardberkeley:
Ah sorry! No it looks great, except for a few really nitpicky comments I left!

hratoanina:
@npwardberkeley We realized that with the original design the recursive circuits were not generic and were depending on the public values (because of hardsetting them in `get_memory_extra_looking_products_circuit`). We turned them into public inputs to solve that problem. For convenience we also propagate a `PublicValues` struct from the root proof to the aggregation proof and the block proof, but we can do without if needed.

Note that this PR doesn't connect these newly created public inputs between proofs, since this is already what @Nashtare's PR (https://github.com/mir-protocol/plonky2/pull/989) is about.

PR #1033, opened 2023-05-10
Optimize transpose

0x0ece:
I did some tests and found that this version of transpose is more efficient, significantly reducing the time to run prove().

For example, I have a recursive circuit with 1024x sha256 calls that takes ~110s to prove. With this patch it goes down to 93s!

### DIFF
--- a/plonky2/src/util/mod.rs
+++ b/plonky2/src/util/mod.rs
@@ -1,6 +1,6 @@
-use alloc::vec;
 use alloc::vec::Vec;
 
+use plonky2_maybe_rayon::*;
 #[doc(inline)]
 pub use plonky2_util::*;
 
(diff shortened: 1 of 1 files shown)


Nashtare:
Was there any reason why this hasn't been merged? I have noticed similar improvements both on x86 and ARM.

npwardberkeley:
Ah sorry just to confirm, did you benchmark this on a variety of transpose sizes? If so, we can go ahead and merge it!

Nashtare:
@npwardberkeley  I did some runs against the `transpose` benchmark:

- on arm target, 8 cores: improvement of 50.48% for the smallest size, 76~85% for the others
- on arm target, 32 cores: improvement of 39.83% for the smallest size, ~84~91% for the others
- on x86_64 target, 12 cores: improvement of 36.58% for the smallest size, 72~75% for the others
- on x86_64 target, 192 cores: slowdown of 200% for the first size, improvement of 59~76% for the others

I don't think the observed slowdown for really beefy machines really matters, as it is for instances which we'd most likely never observe in practice (really narrow traces).

npwardberkeley:
Nice, thank you!

PR #1034, opened 2023-05-10
Optimize reduce_polys_base

0x0ece:
Make `reduce_polys_base` parallel.

Unfortunately I had to use two collect(), not sure if there's a better way to do this.
In all practical cases I analyzed, collect() wasn't an issue as the number of polynomials is relatively small (256, 500, up to 800ish). But parallelization improves significantly the perf.

### DIFF
--- a/plonky2/src/util/reducing.rs
+++ b/plonky2/src/util/reducing.rs
@@ -92,7 +92,7 @@ impl<F: Field> ReducingFactor<F> {
             .zip(polys)
             .collect::<Vec<_>>()
             .par_iter()
-            .map(|(base_power, poly)| poly.mul_extension(base_power.clone()))
+            .map(|(base_power, poly)| poly.mul_extension(*base_power))
             .sum()
     }
 
--- a/plonky2/src/util/reducing.rs
+++ b/plonky2/src/util/reducing.rs
@@ -2,6 +2,8 @@ use alloc::vec;
 use alloc::vec::Vec;
 use core::borrow::Borrow;
 
+use plonky2_maybe_rayon::*;
+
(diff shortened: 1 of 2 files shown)


0x0ece:
Mmm... I noticed that collecting self.base.powers() oom many tests. I don't have an easy solution here, prob best to close this, unless anyone has an idea?

plonky2/src/util/reducing.rs
```diff
@@ -82,15 +84,15 @@ impl<F: Field> ReducingFactor<F> {
 
     pub fn reduce_polys_base<BF: Extendable<D, Extension = F>, const D: usize>(
         &mut self,
-        polys: impl IntoIterator<Item = impl Borrow<PolynomialCoeffs<BF>>>,
+        polys: Vec<&PolynomialCoeffs<BF>>,
```
dlubarov, line 87: Could this take something like a `impl MaybeParIter`, to avoid collecting?

0x0ece:
Disregard the comment above, it turns out that I was doing par_iter().zip(x.par_iter()), while instead doing zip().collect().par_iter() seems to work fine.

plonky2/src/util/reducing.rs
0x0ece: I'm not really able to make it compile without a Vec. Tried a bunch of variations...
Anyway the issue with collect is fixed.

Nashtare:
Hey @0x0ece, apologies for never getting back to you!
I think we would still be interested in merging this, if performances are indeed improved significantly. Could you rebase with latest `main`, so that we can do some performance benchmarks between both branches?

PR #1038, opened 2023-05-15
Fix to hash precompile so EVM tests pass

npwardberkeley:
#1029 broke some sha2 and ripemd EVM tests, this fixes them! (I think, I couldn't run the biggest tests on my machine)


wborgeaud (approved):
Looks good to me!

PR #1041, opened 2023-05-16
`TrieInputs` now uses `H256` for storage account addresses

BGluth:
Slightly easier for us if we use `H256` instead of `H160`.


dlubarov:
Do we need to update [generation/mpt.rs#L43](https://github.com/mir-protocol/plonky2/blob/ce6ac9f8886e58b8a8f4eb910e5af9b4b58ea0a5/evm/src/generation/mpt.rs#L43)?

PR #1069, opened 2023-06-02
Fix arithmetic stark padding

Nashtare:
The current padding only applies up to 65536 rows. For some ethereum tests of the test suite (for instance `GeneralStateTests/stSpecialTest/deploymentError`), the ArithmeticStark trace becomes larger and a panic arises.
This handles proper padding in those cases.

### DIFF
--- a/evm/src/arithmetic/arithmetic_stark.rs
+++ b/evm/src/arithmetic/arithmetic_stark.rs
@@ -138,8 +138,10 @@ impl<F: RichField, const D: usize> ArithmeticStark<F, D> {
         }
 
         // Pad the trace with zero rows if it doesn't have enough rows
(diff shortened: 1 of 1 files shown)


unzvfu:
Thanks, this looks good!

To get the CI to pass, could you please pull the latest changes from `main` (specifically 8153dc7825930c3aa92ae0694cab56188322b4ac) which will fix the build failure?

Nashtare:
Sure! Sorry about that!

unzvfu:
> Sure! Sorry about that!

Oh, not at all; I pushed that fix just a few minutes after you opened this PR, so I wouldn't expect you to have spotted it! :)

PR #1071, opened 2023-06-02
Error handling

nbgl:
This PR implements handling for most fatal user errors, namely:
- stack underflow/overflow,
- `JUMP`/`JUMPI` to invalid address, and
- invalid opcode.

The changes are:
1. New `exception` operation that executes a trap into an exception handler. Similarly to `syscall`, the prover can set this flag regardless of the opcode (he kernel is responsible for verifying that the exception handler was entered correctly). This includes constraints and witness generation.
2. Implement exception handlers for the five aforementioned errors. These handlers verify that an exception actually occurred and `PANIC` otherwise. Once this is checked, the handlers jump to the usual exceptional revert routine.
3. Implement `handle_error` in witness generation to create an `exception` row.
4. Witness generation: detect stack overflow and underflow before entering a syscall (and raise an error instead).
5. Ensure that a syscall cannot cause an uncaught stack overflow by adding an extra constraint on `EXIT_KERNEL`.


evm/src/witness/transition.rs
```diff
@@ -250,6 +255,11 @@ fn try_perform_instruction<F: Field>(state: &mut GenerationState<F>) -> Result<(
     row.stack_len = F::from_canonical_usize(state.registers.stack_len);
 
     let opcode = read_code_memory(state, &mut row);
+    (row, opcode)
+}
+
+fn try_perform_instruction<F: Field>(state: &mut GenerationState<F>) -> Result<(), ProgramError> {
+    let (mut row, opcode) = base_row(state);
     let op = decode(state.registers, opcode)?;
```
wborgeaud, line 264: Somewhat related to this PR, is the `assert_eq!(rhs, F::ZERO)` below on line 286 necessary? It makes a bunch of tests fail (e.g. `stStackTests/stackOverflowDUP_0`) and removing it doesn't seem to cause any other issues.

wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/asm/core/exception.asm
```diff
@@ -0,0 +1,299 @@
...
+    BYTES 0  // 0x00, STOP
+    BYTES 2  // 0x01, ADD
+    BYTES 2  // 0x02, MUL
+    BYTES 2  // 0x03, SUB
+    BYTES 2  // 0x04, DIV
+    BYTES 2  // 0x05, SDIV
+    BYTES 2  // 0x06, MOD
+    BYTES 2  // 0x07, SMOD
+    BYTES 2  // 0x08, ADDMOD
+    BYTES 2  // 0x09, MULMOD
```
wborgeaud, line 159: Should be 3.

nbgl: Great catch!!

evm/src/cpu/kernel/asm/core/exception.asm
```diff
@@ -0,0 +1,299 @@
...
+    %endrep
+
+    BYTES 1  // 0x50, POP
+    BYTES 1  // 0x51, MLOAD
+    BYTES 2  // 0x52, MSTORE
+    BYTES 2  // 0x53, MSTORE8
+    BYTES 1  // 0x54, SLOAD
+    BYTES 2  // 0x55, SSTORE
+    BYTES 1  // 0x56, JUMP
+    BYTES 1  // 0x57, JUMPI
```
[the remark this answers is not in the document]
nbgl: Excellent catch also!

evm/src/witness/operation.rs
```diff
@@ -637,3 +673,64 @@ pub(crate) fn generate_mstore_general<F: Field>(
...
 }
+
+pub(crate) fn generate_exception<F: Field>(
+    exc_code: u8,
+    state: &mut GenerationState<F>,
+    mut row: CpuColumnsView<F>,
+) -> Result<(), ProgramError> {
+    if TryInto::<u32>::try_into(state.registers.gas_used).is_err() {
+        panic!();
+    }
```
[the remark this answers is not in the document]
nbgl: I agree. I did it this way solely for consistency with `SYSCALL`.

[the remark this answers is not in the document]
nbgl: might as well fix it for both in this pr though…

evm/src/witness/transition.rs
nbgl: Hmmmm, if that assertion doesn't hold, then the constraints _should_ fail. But since they don't, then there's some kind of bug. I've been meaning to simplify this part a bit anyway, so I'll just rewrite it.

PR #1073, opened 2023-06-05
Prevent shift ops from panicking

Nashtare:
The `U256` type computes Shr/Shl by calling the `as_usize` method, which panics if the value is larger than `usize::max_value()`.


unzvfu (approved):
Good catch!

PR #1082, opened 2023-06-08
Simplify stack bounds constraints

nbgl:
@dlubarov, you were right. I should have done this the simpler way to begin with…

This PR simplifies stack bounds (overflow/underflow) constraints. `POP` is finally constrained.

(Also, a few misc. witness generation panic->exception stuff that's unrelated but I was already touching that code.)

### DIFF
--- a/evm/src/cpu/stack_bounds.rs
+++ b/evm/src/cpu/stack_bounds.rs
@@ -1,14 +1,13 @@
-//! Checks for stack underflow and overflow.
+//! Checks for stack overflow.
 //!
-//! The constraints defined herein validate that stack exceptions (underflow and overflow) do not
-//! occur. For example, if `is_add` is set but an addition would underflow, these constraints would
-//! make the proof unverifiable.
+//! The constraints defined herein validate that stack overflow did not occur. For example, if `dup`
+//! is set but the copy would overflow would underflow, these constraints would make the proof
+//! unverifiable.
 //!
-//! Faults are handled under a separate operation flag, `is_exception` (this is still TODO), which
-//! traps to the kernel. The kernel then handles the exception. However, before it may do so, the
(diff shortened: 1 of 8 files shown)


evm/src/cpu/stack_bounds.rs
```diff
@@ -1,14 +1,13 @@
-//! Checks for stack underflow and overflow.
+//! Checks for stack overflow.
```
wborgeaud, line 1: Just to make sure I'm not missing anything, you're planning to add the underflow checks in another PR, right?

evm/src/cpu/stack_bounds.rs
```diff
@@ -1,14 +1,13 @@
...
 //!
-//! The constraints defined herein validate that stack exceptions (underflow and overflow) do not
-//! occur. For example, if `is_add` is set but an addition would underflow, these constraints would
-//! make the proof unverifiable.
+//! The constraints defined herein validate that stack overflow did not occur. For example, if `dup`
+//! is set but the copy would overflow would underflow, these constraints would make the proof
```
wborgeaud, line 4:
```suggestion
//! is set but the copy would overflow, these constraints would make the proof
```

wborgeaud (approved):
Looks good to me!

evm/src/cpu/stack_bounds.rs
nbgl: Good observation!

The underflow checks are implicit. If we try to read from a negative stack index, the range checks in the memory table will cause the proof to fail.

Because of this, the previous version only had an underflow check for `POP`, saving a memory read. But I removed this special case for simplicity, and now `POP` reads from the stack (implicitly checking for underflow) and just ignores the value.

PR #1090, opened 2023-06-08
Store gas in two limbs to prevent overflow errors

Nashtare:
The gas value stored in kernel exit info can sometimes be greater than the current u32 limit, causing overflows with the existing config. See `stCreate2/returndatacopy_following_create` for a related failing test, where the CALL uses a (too) large amount of gas to send to the subcontext.


Nashtare:
`stCreate2/returndatacopy_following_create` won't be addressed, following offline discussion with William on keeping a single limb for the gas counter, hence closing this PR.

PR #1092, opened 2023-06-12
Move operations to Field

matthiasgoergens:
Implement the TODOs and move `from_noncanonical_u64` and `from_noncanonical_i64` from `Field64` to `Field`.

### DIFF
--- a/field/src/goldilocks_field.rs
+++ b/field/src/goldilocks_field.rs
@@ -119,22 +119,6 @@ impl Field for GoldilocksField {
     }
 
     #[inline]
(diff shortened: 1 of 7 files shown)


field/src/goldilocks_field.rs
```diff
@@ -118,22 +118,6 @@ impl Field for GoldilocksField {
         reduce128(n)
     }
 
-    #[inline]
```
matthiasgoergens, line 121: The diff is misleading here.  Logically, moved `from_noncanonical_u64` and `from_noncanonical_i64`, but github guesses wrong.

npwardberkeley:
Looks good — thanks for adding!

PR #1095, opened 2023-06-13
PUSH0

nbgl:
Per @wborgeaud's request, implement `PUSH0`.


evm/src/cpu/push0.rs
```diff
@@ -0,0 +1,32 @@
...
+use crate::cpu::columns::CpuColumnsView;
+use crate::cpu::membus::NUM_GP_CHANNELS;
+
+pub fn eval_packed<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    let filter = lv.op.push0;
+    let push_value = lv.mem_channels[NUM_GP_CHANNELS - 1].value;
+    for &limb in &push_value[0..] {
```
wborgeaud, line 16: nit: Why not `for &limb in &push_value {` ?

wborgeaud (approved):
Looks good, thanks!

evm/src/cpu/push0.rs
nbgl: lol good catch!! i copy-pasted this code and just thoughtlessly changed `1..` to `0..`…

PR #1102, opened 2023-06-20
Modexp memory context change

npwardberkeley:


wborgeaud (commented):
Also need to update `evm/src/cpu/kernel/asm/core/precompiles/expmod.asm`, but otherwise looks good.

evm/src/cpu/kernel/asm/bignum/util.asm
```diff
@@ -1,13 +1,15 @@
-%macro memcpy_kernel_general
+%macro memcpy_current_general
     // stack: dst, src, len
-    %stack (dst, src, len) -> (0, @SEGMENT_KERNEL_GENERAL, dst, 0, @SEGMENT_KERNEL_GENERAL, src, len, %%after)
+    GET_CONTEXT
+    %stack (context, dst, src, len) -> (context, @SEGMENT_KERNEL_GENERAL, dst, 0, @SEGMENT_KERNEL_GENERAL, src, len, %%after)
```
wborgeaud, line 4: Not entirely sure about how these macros are used, but shouldn't the `0` be a `context`?

evm/src/cpu/kernel/asm/bignum/util.asm
```diff
@@ -1,13 +1,15 @@
...
     %jump(memcpy)
 %%after:
 %endmacro
 
-%macro clear_kernel_general
+%macro clear_current_general
     // stack: dst, len
-    %stack (dst, len) -> (0, @SEGMENT_KERNEL_GENERAL, dst, 0, len, %%after)
+    GET_CONTEXT
+    %stack (context, dst, len) -> (context, @SEGMENT_KERNEL_GENERAL, dst, 0, len, %%after)
```
wborgeaud, line 12: ditto

evm/src/cpu/kernel/asm/bignum/util.asm
npwardberkeley: Ah good catch!

evm/src/cpu/kernel/asm/bignum/util.asm
npwardberkeley: This one is different, though: for tis `clear` macro, `0` is the value we're `memset`ing into the memory

evm/Cargo.toml
```diff
@@ -26,8 +26,8 @@ num-bigint = "0.4.3"
 once_cell = "1.13.0"
 pest = "2.1.3"
 pest_derive = "2.1.0"
-plonky2 = { version = "0.1.2", default-features = false, features = ["timing"] }
-plonky2_util = { version = "0.1.0" }
+plonky2 = { path = "../plonky2", default-features = false, features = ["timing"] }
+plonky2_util = { path = "../util" }
```
wborgeaud, line 30: Hm are these changes relevant to the PR?

npwardberkeley: Oh sorry! That was so I could run the EVM tests, didn't mean to include it in the PR

wborgeaud (approved):
Looks good to me!

PR #1104, opened 2023-06-24
Make serializer work with slices

Nashtare:
Deserializing (especially large objects like `AllRecursiveCircuits`)  can get expensive when copying the bytes into a new `Buffer`.
Also removes the conditional attribute with `std` (I'm not sure why it was to begin with, as `Vec` is imported from `alloc`?).


npwardberkeley (approved):
Looks good, thanks!

PR #1105, opened 2023-06-26
Silence Poseidon warnings for ARM targets

Nashtare:
The MDS layer implementation for x86-64 targets is moved inside a dedicated submodule to silence cargo warnings of dead code (at least until assembly optimizations for x86 targets are re-enabled).


Nashtare:
Any chance this can get merged? It gets quite noisy to compile with `RUSTFLAGS=-Ctarget-cpu=native` on ARM.

nbgl:
@Nashtare on it!

PR #1111, opened 2023-07-06
Lookup serialization

Nashtare:
This PR updates the way lookup tables are being serialized, and more generally gives access to `CommonCircuitData` to gates and generators during serialization/deserialization.

This allows:
- to serialize only once a `lookup_table` in a given circuit
- to remove the need for the special `DummyCircuitGenerator` to be placed first in the GeneratorSerializer object (could have easily been source of errors for external users defining their own gate/generators and needing to implement a custom serializer.

For this, it updates the underlying `F` type of `WitnessGeneratorRef` from `F: Field` to `F: RichField + Extendable<D>`. I am not sure as to why it was set only to `Field` in the first place, as the restriction is natural here.

### DIFF
--- a/plonky2/src/util/serialization/gate_serialization.rs
+++ b/plonky2/src/util/serialization/gate_serialization.rs
@@ -6,12 +6,16 @@ use crate::plonk::circuit_data::CommonCircuitData;
 use crate::util::serialization::{Buffer, IoResult};
 
 pub trait GateSerializer<F: RichField + Extendable<D>, const D: usize> {
-    fn read_gate(&self, buf: &mut Buffer, cd: &CommonCircuitData<F, D>) -> IoResult<GateRef<F, D>>;
+    fn read_gate(
+        &self,
+        buf: &mut Buffer,
+        common_data: &CommonCircuitData<F, D>,
+    ) -> IoResult<GateRef<F, D>>;
     fn write_gate(
         &self,
         buf: &mut Vec<u8>,
         gate: &GateRef<F, D>,
-        cd: &CommonCircuitData<F, D>,
+        common_data: &CommonCircuitData<F, D>,
     ) -> IoResult<()>;
 }
 
--- a/plonky2/src/util/serialization/gate_serialization.rs
+++ b/plonky2/src/util/serialization/gate_serialization.rs
@@ -6,12 +6,13 @@ use crate::plonk::circuit_data::CommonCircuitData;
 use crate::util::serialization::{Buffer, IoResult};
 
 pub trait GateSerializer<F: RichField + Extendable<D>, const D: usize> {
-    fn read_gate(
+    fn read_gate(&self, buf: &mut Buffer, cd: &CommonCircuitData<F, D>) -> IoResult<GateRef<F, D>>;
+    fn write_gate(
         &self,
-        buf: &mut Buffer,
-        common: &CommonCircuitData<F, D>,
-    ) -> IoResult<GateRef<F, D>>;
-    fn write_gate(&self, buf: &mut Vec<u8>, gate: &GateRef<F, D>) -> IoResult<()>;
+        buf: &mut Vec<u8>,
+        gate: &GateRef<F, D>,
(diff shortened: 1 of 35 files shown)


plonky2/src/util/serialization/gate_serialization.rs
```diff
@@ -2,21 +2,31 @@ use plonky2_field::extension::Extendable;
...
+use crate::plonk::circuit_data::CommonCircuitData;
 use crate::util::serialization::{Buffer, IoResult};
 
 pub trait GateSerializer<F: RichField + Extendable<D>, const D: usize> {
-    fn read_gate(&self, buf: &mut Buffer) -> IoResult<GateRef<F, D>>;
-    fn write_gate(&self, buf: &mut Vec<u8>, gate: &GateRef<F, D>) -> IoResult<()>;
+    fn read_gate(
+        &self,
+        buf: &mut Buffer,
+        common: &CommonCircuitData<F, D>,
```
npwardberkeley, line 12: Is there a reason this isn't called `cd` as in the other functions?

plonky2/src/util/serialization/mod.rs
```diff
@@ -1689,12 +1697,14 @@ pub trait Write {
         &mut self,
         gate: &GateRef<F, D>,
         gate_serializer: &dyn GateSerializer<F, D>,
+        common_data: &CommonCircuitData<F, D>,
```
npwardberkeley, line 1700: Same with this

plonky2/src/util/serialization/gate_serialization.rs
Nashtare: No! Will make the arg names uniform!

plonky2/src/util/serialization/mod.rs
Nashtare: common_data is actually being used in quite a lot of places (and consistently across this file). Do you want me to address this particular case still?

plonky2/src/util/serialization/gate_serialization.rs
Nashtare: Addressed, sorry about that!

plonky2/src/util/serialization/mod.rs
npwardberkeley: Hm maybe it should just be called `common_data` everywhere (instead of `cd`)?

plonky2/src/util/serialization/gate_serialization.rs
npwardberkeley: Np, thanks!

plonky2/src/util/serialization/mod.rs
Nashtare: Ok! Addressed in additional commit.

Nashtare:
@npwardberkeley I've added another commit on top of what you already reviewed, addressing a point mentioned in #1114 about circuit building which can take quite some time when we are loaded with lookups.
The commit adds a hash (`keccak` as already present) of the lookup table when defining a new `LookupGate` / `LookupTableGate`, and uses a custom implementation of `Gate::id()` for those two gate types, to ignore the entire table.

For 8192 `LookupGate`s, I am getting a circuit building time of 2.45 sec vs 4.5 sec for the current implementation. Let me know what you think! 🙂

plonky2/src/util/serialization/mod.rs
```diff
@@ -784,7 +783,16 @@ pub trait Read {
             num_lookup_polys,
             num_lookup_selectors,
             luts,
-        })
+        };
+
+        for _ in 0..gates_len {
```
npwardberkeley, line 788: nit: can we do this using a `for_each`?

npwardberkeley (approved):
Looks good!

plonky2/src/util/serialization/mod.rs
Nashtare: I don't think we could, can we? There would be a `?` inside the closure.

PR #1112, opened 2023-07-06
Fix jump operation generation

LindaGuiga:
We noticed that some proof verifications were failing due to a constraint in `jump.rs`: `lv.op.jump * (cond[0] - P::ONES)` (l. 76). 
This is due to the fact that in `generate_jump`, `row.mem_channels[1].value[0]` was only set to 1 if it was a kernel operation. However, the comment in `jump.rs` explains that `row.mem_channels[1].value[0]` is a predicate that is always 1 when the operation is a `JUMP`.
We therefore set the memory value to 1 outside of the `if` condition in `generate_jump`.

### DIFF
--- a/evm/src/witness/operation.rs
+++ b/evm/src/witness/operation.rs
@@ -185,13 +185,14 @@ pub(crate) fn generate_jump<F: Field>(
         state,
         &mut row,
     );
+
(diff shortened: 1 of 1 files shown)


nbgl (approved):
Thank you!!

PR #1114, opened 2023-07-06
Lookup arg optimizations

onsen-egg:
We've been using lookups downstream for a while, with tables of size 2^16 (for u16 range checks) and hundreds of thousands of lookups in a circuit. This PR addresses the main bottlenecks that were at first producing poor building/proving times (around 100x/3x overall vs. baseline, respectively) at this scale.

Changes: 
- using a linear-time pass to count multiplicities vs. the current O(n^2) implementation
- precomputing the RE poly evaluations, which were being re-computed redundantly. This sped up the [`compute quotient polys`](https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/plonk/prover.rs#L229-L243) step of proving by about 10x for our circuits (size 2^21 with width ~200, using the aforementioned size 2^16 tables)

The remaining significant bottleneck we encountered was also at build time, owing to the fact that the id() of `LookupGate` and `LookupTableGate` includes the full table, and this id is hashed every time another gate gets added to the circuit builder. However, our fix for this was hacky so I didn't include it here—a good solution might have some overlap with the serialization stuff mentioned [here](https://github.com/mir-protocol/plonky2/pull/964#issuecomment-1613105117).

Also, for future reference, I identified [here](https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/plonk/prover.rs#L212-L223) and [here](https://github.com/mir-protocol/plonky2/blob/main/plonky2/src/fri/oracle.rs#L188-L192) as routines that may also have some unneeded performance bloat due to lookups—they are probably worth taking a look at for anyone who wants to improve lookup performance further.

Thanks to @Nashtare and the others who implemented this feature, it's been super helpful :)

### DIFF
--- a/plonky2/src/plonk/prover.rs
+++ b/plonky2/src/plonk/prover.rs
@@ -4,6 +4,7 @@ use core::cmp::min;
 use core::mem::swap;
 
 use anyhow::{ensure, Result};
+use hashbrown::HashMap;
 use plonky2_maybe_rayon::*;
 
 use super::circuit_builder::{LookupChallenges, LookupWire};
@@ -58,13 +59,18 @@ pub fn set_lookup_wires<
 
         // Compute multiplicities.
         let mut multiplicities = vec![0; lut_len];
+
+        let table_value_to_idx: HashMap<u16, usize> = common_data.luts[lut_index]
+            .iter()
+            .enumerate()
+            .map(|(i, (inp_target, _))| (inp_target.clone(), i))
+            .collect();
+
         for (inp_target, _) in prover_data.lut_to_lookups[lut_index].iter() {
             let inp_value = pw.get_target(*inp_target);
-            let mut idx = 0;
-            while F::from_canonical_u16(common_data.luts[lut_index][idx].0) != inp_value {
-                idx += 1;
-            }
-            multiplicities[idx] += 1;
+            let idx = table_value_to_idx.get(&u16::try_from(inp_value.to_canonical_u64()).unwrap()).unwrap();
+
+            multiplicities[*idx] += 1;
         }
 
         // Pad the last `LookupGate` with the first entry from the LUT.
--- a/plonky2/src/plonk/prover.rs
+++ b/plonky2/src/plonk/prover.rs
@@ -15,6 +15,7 @@ use crate::field::zero_poly_coset::ZeroPolyOnCoset;
 use crate::fri::oracle::PolynomialBatch;
 use crate::gates::lookup::LookupGate;
 use crate::gates::lookup_table::LookupTableGate;
+use crate::gates::selectors::LookupSelectors;
 use crate::hash::hash_types::RichField;
 use crate::iop::challenger::Challenger;
 use crate::iop::generator::generate_partial_witness;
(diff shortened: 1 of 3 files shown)


Nashtare:
Thanks, that's amazing! 
As for `the fact that the id() of LookupGate and LookupTableGate includes the full table, and this id is hashed every time another gate gets added to the circuit builder.`, I was wondering for a while whether we wanted to have ids of gates and generators be deeply binding to their underlying specification. For instance, the PoseidonGate id is just having the width as extra parameter, while all the parameters defining the permutation can be arbitrary (and in particular the MDS matrix coefficients have been changed from the reference implementation). We could go for something similar, sort of middle-ground identifier that doesn't require having the entire table listed in the id.

onsen-egg:
For the suggestion of hashing the table to use as an ID when adding Lookup-related gates, although I see that it does improve the build time, I don't think it addresses the fundamental problem and we'll still see very poor build times for large tables/many lookups, since you're still paying the cost of hashing the entire table every time you add a LookupGate or LookupTable Gate, (since new_from_table is [called each time](https://github.com/onsen-egg/plonky2/blob/38325103cbeb96f07ac1d5dd83ab9ab9230d2020/plonky2/src/gadgets/lookup.rs#L93) you do this). `I was wondering for a while whether we wanted to have ids of gates and generators be deeply binding to their underlying specification` this is more what is needed here IMO; the gate id just shouldn't have a dependency on the actual contents of the table, which I think is fine if there's a bit of extra bookkeeping. If I were to go for a refactor, I would identify a lookup table using a user-defined string that is provided when a table is created and use that as the id, replacing [this Vec\<LookupTable\> ](https://github.com/onsen-egg/plonky2/blob/38325103cbeb96f07ac1d5dd83ab9ab9230d2020/plonky2/src/plonk/circuit_builder.rs#L135-L136)with a HashMap<String, LookupTable>. In addition to solving the performance problem, it'd also be nice to be able to identify tables by readable strings rather than indices when [adding lookups](https://github.com/onsen-egg/plonky2/blob/38325103cbeb96f07ac1d5dd83ab9ab9230d2020/plonky2/src/gadgets/lookup.rs#L63).

`Also, are you storing the entire table as the current impl does?` yeah, this hasn't been an issue though, even for "large" tables I don't think it's a big deal since we're only storing it once and sharing it through an Arc thanks to your guys' implementation.

onsen-egg:
@npwardberkeley, are these changes in a good state to merge once CI passes? I pushed changes to fix CI but it looks like you need to approve the workflow again, sorry about that.

plonky2/src/plonk/prover.rs
```diff
@@ -58,13 +60,20 @@ pub fn set_lookup_wires<
 
         // Compute multiplicities.
         let mut multiplicities = vec![0; lut_len];
+
+        let mut table_value_to_idx: HashMap<u16, usize> = HashMap::new();
+
+        for (i, (inp_target, _)) in common_data.luts[lut_index].iter().enumerate() {
```
npwardberkeley, line 66: I think this can be done more idiomatically with `.map()` and `.collect()

npwardberkeley (approved):
Looks good apart from one comment!

plonky2/src/plonk/prover.rs
onsen-egg: How does [this](https://github.com/onsen-egg/plonky2/blob/a67cfdcb676b8637110621a8fe35251f0c583ec7/plonky2/src/plonk/prover.rs#L64-L68) look instead?

npwardberkeley:
Looks good!

Nashtare:
> For the suggestion of hashing the table to use as an ID when adding Lookup-related gates, although I see that it does improve the build time, I don't think it addresses the fundamental problem and we'll still see very poor build times for large tables/many lookups, since you're still paying the cost of hashing the entire table every time you add a LookupGate or LookupTable Gate, (since new_from_table is [called each time](https://github.com/onsen-egg/plonky2/blob/38325103cbeb96f07ac1d5dd83ab9ab9230d2020/plonky2/src/gadgets/lookup.rs#L93) you do this).

I definitely agree, but I have taken this approach as gates are currently storing in their IDs all their internal information. If we were to remove it, and have just basic information as we do with generators ID, then your suggestion of having a `HashMap` with table names does indeed make more sense

PR #1115, opened 2023-07-06
Compute `stack_len_bounds_aux` correctly in `generate_exception`

hratoanina:
In `generate_exception`, after reverting the current row, `stack_len_bounds_aux` was computed incorrectly. This made the corresponding constraint in `stack_bounds.rs` fail.

An example is the `JUMP_Bounds2` test from the EVM test suite: the state is computed correctly but the proof isn't valid.


nbgl:
Thank you for this!! Closing in favor of #1119, which has a minor change.

PR #1116, opened 2023-07-06
Remove need for matching start ranges

Nashtare:
Tiny tweak to remove the need for matching start ranges in the `AllRecursiveCircuits` initialization. Allows to scrap out ~4.6GB of useless data when serializing (out of the ~6.5GB for the empty_txn_list example, i.e.  ~72% reduction).
Also fixes some invalid end ranges.

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -6,6 +6,7 @@ use hashbrown::HashMap;
 use itertools::{zip_eq, Itertools};
 use plonky2::field::extension::Extendable;
 use plonky2::fri::FriParams;
+use plonky2::gates::constant::ConstantGate;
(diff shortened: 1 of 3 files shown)


npwardberkeley:
Hi! Can you explain a bit more about how this works please? How does the constant gate remove the need for matching start ranges?

Nashtare:
Hi @npwardberkeley! That actually comes as some magic trick out of my hat. I was puzzled for a bit as to why we needed all ranges to start at 9 (the lowest, necessary for `KeccakSponge`), and upon investigation, I noticed that the `RootCircuit` creation was complaining about CommonCircuitData mismatch, which this missing constant gate. I didn't push the analysis further, noticing it was then working no matter the end ranges.

npwardberkeley:
Oh I see, that makes sense! Thanks!

PR #1117, opened 2023-07-07
Fix generate_set_context

LindaGuiga:
We noticed in some tests with stack overflows that the macro `set_length` caused `context_ops` constraints to fail, due to the `SET_CONTEXT` instruction.
In the case where the context does not change (which is the case for `set_length`), there is a `MemoryOp` to read the stack length, but the memory channels are not set. The failing constraints concerned the said unset memory channel. Thus, this PR sets the memory channels in that case.


nbgl (approved):
Thank you!!

PR #1136, opened 2023-07-16
Fix risk of division by zero

Nashtare:
The modulus in prover_inputs `run_bignum_modmul()` is not being checked to be different than zero, resulting in a panic in several tests from the EVM suite (see for instance `modexpTests` case variants 4, 5, 8, 25, 31, 40, 41, 51, 101.


npwardberkeley (approved):
Good catch, thanks!

PR #1138, opened 2023-07-17
deps(serde): use `rc` feature in `plonky2` crate

bingcicle:
building `plonky2` currently fails:

```console
$ cd plonky2 && cargo build
--> plonky2/src/plonk/circuit_data.rs:348:39
... the trait `Serialize` is not implemented for `Arc<Vec<(u16, u16)>>`
```

We have to enable `rc` feature in `serde`.

Note that this fix is also present in #1114

### DIFF
--- a/plonky2/Cargo.toml
+++ b/plonky2/Cargo.toml
@@ -30,7 +30,7 @@ plonky2_field = { version = "0.1.0", default-features = false }
 plonky2_util = { version = "0.1.0", default-features = false }
 rand = { version = "0.8.4", default-features = false }
 rand_chacha = { version = "0.3.1", optional = true, default-features = false }
-serde = { version = "1.0", default-features = false, features = ["derive"] }
+serde = { version = "1.0", default-features = false, features = ["derive", "rc"] }
 serde_json = "1.0"
 static_assertions = { version = "1.1.0", default-features = false }
 unroll = { version = "0.1.5", default-features = false }


matthiasgoergens:
@dlubarov Please have a look.  Without this fix, it's very hard to use plonky2 as a dependency.

npwardberkeley:
Hm this looks fine but just wondering, are you using a different version of Rust or something? It's been building fine for the rest of the contributors and also on GitHub Actions

bingcicle:
@npwardberkeley To reproduce, you have to cd into `plonky2` and `cargo build` from there - just tested with the latest `nightly`. The build works fine in the workspace level, though I'm not entirely sure why.

The culprit seems to be https://github.com/mir-protocol/plonky2/commit/c0fc349c580ab573872844cddc8df75779f33332, which introduced an `Arc` to the lookup tables, which I guess requires `rc` feature in `serde`:

```rust
for _ in 0..length {
    luts.push(Arc::new(self.read_lut()?));
}
```

npwardberkeley:
Ohh I see — thanks!

Nashtare:
Perhaps it'd be worth adding a CI job for building solely, say the plonky2 and starky crates? That would prevent this kind of issue in the future.

npwardberkeley:
Ah yes that's a great idea, I'll set that up!

PR #1139, opened 2023-07-18
Better document constraints on addcy carries

unzvfu:
The function to which the addcy code delegates does not check that `given_cy` is in fact 0 or 1, and this check was overlooked in the calling code.


unzvfu:
Oops, this was in fact checked properly, but the documentation is unclear.

PR #1140, opened 2023-07-20
Fix negative quotient issue

### LINKED ISSUE #1127: Quotient = -1 causes range check to fail in ArithmeticStark
@unzvfu We noticed that proof verification of `CreatePrecompiles_{6, 7}` fails. This is because, in some cases when computing `SubFp254`, the computed quotient was equal to -1, but the quotient columns are range checked between 0 and 2^16 in `ArithmeticStark`. Since `MODULAR_QUO_INPUT` columns overlap with some `AUX_INPUT_REGISTER` columns, we can't just remove `MODULAR_QUO_INPUT` columns from the range check. Changing modular computations would also affect `divmod` (and, seemingly, other parts of the code as well).

Do you have an idea of how to solve this issue?

unzvfu:
Closes #1127.

The problem was that the `SUBMOD` and `SUBFP254` operations sometimes generate negative quotients values, but these were mistakenly treated as normal unsigned elements, leading to range check errors later on.

The fix is to handle the quotient differently for these subtraction operations; specifically, the quotient is offset so that it is always positive during generation, and the offset is reversed during verification. (This is only possible because the maximum legitimate quotient in a subtraction operation is 16 limbs, so the quotient register has space to store the high and low parts of the resulting 17-bit values.)

### DIFF
--- a/evm/src/arithmetic/columns.rs
+++ b/evm/src/arithmetic/columns.rs
@@ -74,16 +74,6 @@ const AUX_REGISTER_0: Range<usize> = START_SHARED_COLS..START_SHARED_COLS + N_LI
 const AUX_REGISTER_1: Range<usize> = AUX_REGISTER_0.end..AUX_REGISTER_0.end + 2 * N_LIMBS;
 const AUX_REGISTER_2: Range<usize> = AUX_REGISTER_1.end..AUX_REGISTER_1.end + 2 * N_LIMBS - 1;
 
-// Each element c of {MUL,MODULAR}_AUX_REGISTER is -2^20 <= c <= 2^20;
-// this value is used as an offset so that everything is positive in
-// the range checks.
-pub(crate) const AUX_COEFF_ABS_MAX: i64 = 1 << 20;
-
-// Each element c of MODULAR_QUO_INPUT is -2^16 <= c < 2^16;
-// this value is used as an offset so that everything is positive in
-// the range checks.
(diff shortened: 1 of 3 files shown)


unzvfu:
@LindaGuiga @Nashtare Could you please check that this PR fixes #1127 for you (and doesn't break anything else)? Thanks!

LindaGuiga:
> @LindaGuiga @Nashtare Could you please check that this PR fixes #1127 for you (and doesn't break anything else)? Thanks!

It does fix the two `create2callPrecompiles` that were previously failing, thank you!

evm/src/arithmetic/columns.rs
```diff
@@ -79,6 +79,11 @@ const AUX_REGISTER_2: Range<usize> = AUX_REGISTER_1.end..AUX_REGISTER_1.end + 2
 // the range checks.
 pub(crate) const AUX_COEFF_ABS_MAX: i64 = 1 << 20;
 
+// Each element c of MODULAR_QUO_INPUT is -2^16 <= c < 2^16;
+// this value is used as an offset so that everything is positive in
+// the range checks.
+pub(crate) const QUO_INPUT_ABS_MAX: i64 = 1 << 16;
```
nbgl, line 85: Nit: it feels like an odd choice to put this constant in the `columns` file. The convention we've used is that `columns` contains the indices assigned to particular columns, and this value is not an index.

nbgl (approved):
Looks good!!

PR #1146, opened 2023-07-24
Merge syscalls and exceptions constraints.

LindaGuiga:
This PR addresses the comment in exception.rs, which states that there is a lot of overlap between constraints in `exceptions.rs` and in `syscalls.rs`. 
We propose to merge `exceptions.rs` and `syscalls.rs` into one file: `syscalls_exceptions.rs`.
For constraints that are exactly the same, we use `total_filter = lv.op.syscall + lv.op.exception` as the filter. Otherwise, we only consider one of `lv.op.syscall` or `lv.op.exception` as the filter.


nbgl (approved):
Looks great!! Thank you :)

PR #1148, opened 2023-07-27
Lookup check

Nashtare:
@npwardberkeley Another tiny lookup-related PR, just to fix a wrong naming in `LookupTableGate` id, and to remove the second condition in an assertion check.
More specifically, the second condition (`lut_index >= self.get_luts_length()`) in the OR should first be corrected to `lut_index < self.get_luts_length()`. It wasn't caught in unit tests because we were always satisfying the first condition. Secondly, because the `luts` and `lut_to_lookups` are necessarily of the same size, we can't have `lut_index` being greater than the latter (indeed, the only way for it to fail would be for a user to call `add_lookup_from_index()` with an invalid index when designing the circuit, which would be caught prior to reaching `add_all_lookups()`.

Related to this, I added two additional unit tests, to handle the two failing cases mentioned above. I don't know if we want to maintain the assertion of table unused though, as this just prevents inflating needlessly the circuit size, but isn't really an error per se.

### DIFF
--- a/plonky2/src/gadgets/lookup.rs
+++ b/plonky2/src/gadgets/lookup.rs
@@ -77,7 +77,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
     pub fn add_all_lookups(&mut self) {
         for lut_index in 0..self.num_luts() {
             assert!(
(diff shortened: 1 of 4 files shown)


npwardberkeley (approved):
Looks good, thanks for catching those!

PR #1151, opened 2023-07-27
Cut 5 Columns From The Memory Table With This One Weird Trick!

nbgl:
A minor oversight in c38a98f left five unused columns in the memory table. This tiny PR cuts the number of columns in the memory table from 26 to 21 (not including CTL).


dlubarov (approved):
Great catch!

PR #1155, opened 2023-07-31
Generalise transpose

matthiasgoergens:
So that it also works with eg `u64`, not just with field elements.


plonky2/src/util/mod.rs
```diff
@@ -19,7 +19,7 @@ pub(crate) fn transpose_poly_values<F: Field>(polys: Vec<PolynomialValues<F>>) -
     transpose(&poly_values)
 }
 
-pub fn transpose<F: Field>(matrix: &[Vec<F>]) -> Vec<Vec<F>> {
```
Nashtare, line 22: A detail, but I'd rename it to `T`.

Nashtare (approved):

PR #1159, opened 2023-08-01
Error instead of panicking for missing preprocessed circuits

Nashtare:
If the prover is missing a preprocessed circuit when recursively verifying the initial transaction `AllProof` in `prove_root`, it will currently `panic!`. This change allows the prover to gently return early with an `Error`.


wborgeaud (approved):
Looks good to me!

PR #1161, opened 2023-08-01
Fix prove_root()

Nashtare:
This PR fixes an issue introduced by #1026. This wasn't detected as the only usage of `AllRecursiveCircuits` is done with an empty transaction list.
To reproduce it, you can pass the inputs of the `simple_transfer` test to `prove_root()`. You will need to manually alter the `block_gas_limit` field (see below).

Changes:
- fix the endianess of the block beneficiary
- set all other `block_metadata` targets from the first limb of their `U256` counterpart, converted to a u32. This is a safety measure as each should be fitting in 32 bits by design (we're only considering their first limb). Otherwise testing for instance the `simple_transfer` inputs will fail with inconsistent `Target` value.

@npwardberkeley Would you mind reviewing the changes? As you were the one assigned to #1026 you're probably the most familiar with this recent new part. I've also reused the previously unused `set_public_value_targets()` method for clarity.

@BGluth Just FYI, with the changes from #1026 integrated, it is now impossible to handle tests with a gas limit not fitting in 32 bits, which would actually require https://github.com/mir-protocol/evm-tests/pull/30 to be reverted in order to not have new false positives (though, they wouldn't actually return an error per se, the prover would `panic!` when setting the targets with current main / converting to u32 with this PR).

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -46,8 +46,8 @@ use crate::proof::{PublicValues, PublicValuesTarget, StarkProofWithMetadata};
 use crate::prover::prove;
 use crate::recursive_verifier::{
     add_common_recursion_gates, add_virtual_public_values, recursive_stark_circuit,
-    set_block_metadata_target, set_trie_roots_target, PlonkWrapperCircuit, PublicInputs,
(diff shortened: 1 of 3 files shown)


npwardberkeley (approved):
Looks good to me! Thanks for catching!

PR #1163, opened 2023-08-02
Fix trait import in AVX512 tests

unzvfu:
PR #1092 introduced some compile errors in the architecture-specific testing code for fields. Some of these errors were fixed in #1134. The remainder of the errors are fixed in this PR.

### DIFF
--- a/field/src/arch/x86_64/avx512_goldilocks_field.rs
+++ b/field/src/arch/x86_64/avx512_goldilocks_field.rs
@@ -402,7 +402,7 @@ mod tests {
     use crate::goldilocks_field::GoldilocksField;
     use crate::ops::Square;
     use crate::packed::PackedField;
-    use crate::types::Field64;
+    use crate::types::Field;
 
     fn test_vals_a() -> [GoldilocksField; 8] {
         [


npwardberkeley:
@unzvfu Do you think it's worth updating our CI to run a matrix of tests on different architectures?

unzvfu:
> @unzvfu Do you think it's worth updating our CI to run a matrix of tests on different architectures?

@npwardberkeley Oh, yes that would be excellent! Are you able to take care of that?

npwardberkeley:
Yes! Which architectures should it check for? In the code there are references to `aarch64`, `arm`, `riscv32`, `riscv64`, `x86`, `x86_64`, and `wasm32`, so should I cover all those to be safe?

unzvfu:
> Yes! Which architectures should it check for? In the code there are references to `aarch64`, `arm`, `riscv32`, `riscv64`, `x86`, `x86_64`, and `wasm32`, so should I cover all those to be safe?

Great, thanks Nick! I think we need something a bit more specific, since we have three different x86-based implementations: general x86 and vectorised with AVX2 or AVX512. There's also the two aarch64 implementations: general ARM and vectorised with NEON. We don't have any Risc or Wasm assembly, so I don't think those are necessary. Is it possible to add CI targets that have (or don't have) those specific features?

Issue #1150, opened 2023-07-27
Add missing constraints


nbgl:
Operations that still have some missing constraints:

- [x] `KECCAK_GENERAL`: [Memory channels not constrained.](https://github.com/mir-protocol/plonky2/blob/bfd6b98884fe59472841f89ad51cc3e5f4c02b3e/evm/src/cpu/stack.rs#L73)
- [x] `PROVER_INPUT`: Totally unconstrained. [Need to constrain the memory channels](https://github.com/mir-protocol/plonky2/blob/bfd6b98884fe59472841f89ad51cc3e5f4c02b3e/evm/src/cpu/stack.rs#L74) (disabling unused ones) and range-check all the limbs to 32 bits.
- [x] `SHL`/`SHR`: [Missing CTL to the arithmetic table.](https://github.com/mir-protocol/plonky2/blob/bfd6b98884fe59472841f89ad51cc3e5f4c02b3e/evm/src/cpu/cpu_stark.rs#L83)
- [ ] `PUSH1`-`PUSH32`: Unconstrained. (assigned to @Nashtare)
- [x] Syscalls/exceptions: [`output[6]` needs to be range-checked to 32 bits.](https://github.com/mir-protocol/plonky2/blob/b3f00d4ab828354955ed4f31bd2ec8057cdd5bf3/evm/src/cpu/syscalls_exceptions.rs#L112)

In addition, [the memory CTL is currently disabled](https://github.com/mir-protocol/plonky2/blob/bfd6b98884fe59472841f89ad51cc3e5f4c02b3e/evm/src/all_stark.rs#L108).

Nashtare:
By the way, for the memory CTL, there is an attempt at re-enabling it in #1026 which has been pending for quite a bit now!
I think @LindaGuiga had also asked for the range-checks whether these should be over 32 bits or a different range.

nbgl:
Yes, the syscall/exception range checks should be 32 bits. I've updated this issue to specify :) (cc @LindaGuiga)

LindaGuiga:
I will start dealing with the constraints related to syscalls and prover_input!

Nashtare:
@nbgl I can take care of the other two items if they aren't taken on your side!

LindaGuiga:
[...]

PR #1166, opened 2023-08-04
Connect SHL/SHR operations to the Arithmetic table

Nashtare:
This PR addresses a point in https://github.com/mir-protocol/plonky2/issues/1150.

@unzvfu would you mind reviewing this PR, as you implemented the related operations?
Upon difficulties in getting the Arithmetic CTL to work, I realized (unless I missed something obvious) that we were not logging the corresponding MUL/DIV arithmetic operations when performing a shift.
I also updated the comment on the memory mapping, as I believe it is erroneous. We instead have

```
- mem_channel 0: shift
- mem_channel 1: value to be shifted (was referred as channel 0 previously)
- mem_channel 2: 2^shift
- mem_channel 5 (last): result
```

Let me know if I misunderstood anything! 🙏🏼

### DIFF
--- a/evm/src/cpu/cpu_stark.rs
+++ b/evm/src/cpu/cpu_stark.rs
@@ -60,11 +60,14 @@ fn ctl_data_binops<F: Field>(ops: &[usize]) -> Vec<Column<F>> {
 
 /// Create the vector of Columns corresponding to the three inputs and
 /// one output of a ternary operation.
-fn ctl_data_ternops<F: Field>(ops: &[usize]) -> Vec<Column<F>> {
+/// If `is_shift` is `true`, we offset the memory_channels indices used for the inputs
+/// by 1. It will only be valid if the associated filter is one of the two shift flags.
+fn ctl_data_ternops<F: Field>(ops: &[usize], is_shift: bool) -> Vec<Column<F>> {
+    let offset = is_shift as usize;
     let mut res = Column::singles(ops).collect_vec();
-    res.extend(Column::singles(COL_MAP.mem_channels[0].value));
-    res.extend(Column::singles(COL_MAP.mem_channels[1].value));
-    res.extend(Column::singles(COL_MAP.mem_channels[2].value));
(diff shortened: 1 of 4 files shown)


Nashtare:
I double-checked with the stShift folder of [ethrereum/tests](https://github.com/ethereum/tests/tree/develop/GeneralStateTests/stShift), and everything seems to be fine!

evm/src/cpu/cpu_stark.rs
```diff
@@ -60,11 +60,14 @@ fn ctl_data_binops<F: Field>(ops: &[usize]) -> Vec<Column<F>> {
 
 /// Create the vector of Columns corresponding to the three inputs and
 /// one output of a ternary operation.
-fn ctl_data_ternops<F: Field>(ops: &[usize]) -> Vec<Column<F>> {
+/// If `is_shift` is `true`, we offset the memory_channels indices used for the inputs
+/// by 1. It will only be valid if the associated filter is one of the two shift flags.
+fn ctl_data_ternops<F: Field>(ops: &[usize], is_shift: bool) -> Vec<Column<F>> {
```
unzvfu, line 65: Could you explain the need for `is_shift` in a bit more clearly please (both here and in the comment)? I assume it's implementing the mapping described at https://github.com/mir-protocol/plonky2/blob/d51a512caae653e17b2591c988f75414e395e8f8/evm/src/cpu/shift.rs#L53 but it would be helpful to explicitly document which memory channel corresponds to which value when `offset = 1` (including the precomputed value).

unzvfu (approved):
Looks good! Could just benefit from a bit more documentation in the location noted. Thanks Robin!

evm/src/cpu/cpu_stark.rs
Nashtare: We need this offset because the shift operations use the first memory channel to store the shift, which is not used when computing the related arithmetic operation.
Because both MUL/DIV are binary ops, we can just switch all 3 channels used for ternary operations and safely ignore the last one.
I'll update the documentation to make it clearer!

PR #1168, opened 2023-08-04
Add memory checks for prover_input, as well as range_checks for prover_input, syscalls/exceptions

LindaGuiga:
This PR addresses the `prover_input` and syscalls/exceptions points from [this issue](https://github.com/mir-protocol/plonky2/issues/1150). 

- `STACK_BEHAVIORS` is addressed to add stack constraints for `prover_input`.
- CTLs and `ArithmeticStark` are updated to include range checks for `prover_input`, syscalls and exceptions in the `CpuStark`.


LindaGuiga:
The latest push was to rebase on main, since the PR had conflicts with [this merged PR.](https://github.com/mir-protocol/plonky2/pull/1166)

evm/src/arithmetic/mod.rs
```diff
@@ -122,6 +126,9 @@ pub(crate) enum Operation {
         input2: U256,
         result: U256,
     },
+    RangeCheckOperation {
+        values: [u32; NUM_CPU_LIMBS],
+    },
```
nbgl, line 131: Please make it `U256` for consistency. Every `U256` that the arithmetic table deals with is composed of eight `u32` limbs, so it's the same.

evm/src/witness/operation.rs
```diff
@@ -164,6 +164,14 @@ pub(crate) fn generate_prover_input<F: Field>(
...
     let write = stack_push_log_and_fill(state, &mut row, input)?;
 
+    let input_limbs: [u64; 4] = input.0;
+    let mut values = [0_u32; 8];
+    for (i, limb) in input_limbs.into_iter().enumerate() {
+        values[2 * i] = limb as u32;
+        values[2 * i + 1] = (limb >> 32) as u32;
+    }
+    let range_check_op = arithmetic::Operation::range_check(values);
+    state.traces.push_arithmetic(range_check_op);
```
nbgl, line 174: This would be a lot simpler if `arithmetic::Operation::range_check` accepted a `U256`

evm/src/cpu/cpu_stark.rs
```diff
@@ -22,6 +22,12 @@ use crate::memory::{NUM_CHANNELS, VALUE_LIMBS};
 use crate::stark::Stark;
 use crate::vars::{StarkEvaluationTargets, StarkEvaluationVars};
 
+const EXTRA_OPS: [usize; 3] = [
+    COL_MAP.op.syscall,
+    COL_MAP.op.exception,
```
nbgl, line 27: Does this branch need a rebase? I thought these have been combined into one flag.

evm/src/all_stark.rs
```diff
@@ -111,6 +111,8 @@ fn ctl_arithmetic<F: Field>() -> CrossTableLookup<F> {
         vec![
             cpu_stark::ctl_arithmetic_base_rows(),
             cpu_stark::ctl_arithmetic_shift_rows(),
+            cpu_stark::ctl_sys_exc_check_rows(),
+            cpu_stark::ctl_prover_input_check_rows(),
```
nbgl, line 115: Each of these cross-table lookups adds an extra column to the CPU table, so ideally we'd like to have as few of them as possible. Luckily, we can just use the first CTL (`cpu_stark::ctl_arithmetic_base_rows()`) for range checks.

evm/src/cpu/cpu_stark.rs
```diff
@@ -90,6 +128,60 @@ pub fn ctl_filter_logic<F: Field>() -> Column<F> {
...
+    // Create the CPU Table whose columns are the gas value and the rest of the columns are 0.
+    let filter = Some(Column::single(COL_MAP.op.prover_input));
+
+    TableWithColumns::new(
+        Table::Cpu,
+        ctl_data_prover_input_range_check(&OPS, &EXTRA_OPS),
+        filter,
+    )
+}
+
```
nbgl, line 184: These can be deleted: we can reuse the base arithmetic CTL.

evm/src/cpu/cpu_stark.rs
```diff
@@ -114,7 +206,7 @@ pub fn ctl_arithmetic_base_rows<F: Field>() -> TableWithColumns<F> {
     // the third input.
     TableWithColumns::new(
         Table::Cpu,
-        ctl_data_ternops(&OPS, false),
```
nbgl, line 117: We want to add `COL_MAP.op.syscall` + `COL_MAP.op.exception` + `COL_MAP.op.prover_input` to the OPS list. This can be done with the `Column::sum` API, but will require a reworking of `ctl_data_ternops`.

This way we don't have to have a separate cross-table lookup just for range checks.

nbgl (commented):
Looks good, but the cross-table lookup could be made simpler and more efficient :)

evm/src/cpu/cpu_stark.rs
Nashtare: The duplicated constraints have been merged (cf #1146), but the flags haven't been combined yet AFAIK.

evm/src/cpu/cpu_stark.rs
LindaGuiga: @nbgl  The reason why I didn't reuse the base arithmetic CTL here was because I only selected the useful columns in the CPU for `prover_input` and `syscalls/exceptions`. For `prover_input`, the logic would still work with the base arithmetic CTL, but since we only wanted to range check `output[6]` for `syscalls` and `exceptions`, I thought we had to create an additional ctl. Am I missing something?
What I also thought we could do to avoid adding a CPU column is to range check `handler_addr0-2` and `syscall_info/exc_info`, since they're already in the columns selected by the base CTL, instead of only range checking the gas. What do you think?

LindaGuiga:
@nbgl Thank you for your review! 
I replaced the u32 values for range checks by U256 to simplify things, as you suggested.
Regarding the use of `ctl_arithmetic_base_rows` instead of additional CTLs, since the columns required, on the cpu side, for syscalls and exceptions are different from the other operations, I not only range checked the gas but also the handler addresses and syscall/exc_info, since they are also U256s and are already in the "correct" rows. What do you think of this approach?

LindaGuiga:
@nbgl I pushed a new version that does not introduce a new CTL for the range checks but instead adds a column in `ArithmeticStark` containing the opcode if the operation is a range check.
I constrained the new opcode column to be 0 when the operation is anything but a range check.
What do you think of this approach?

LindaGuiga:
@nbgl I pushed a new version that doesn't introduce a new CTL: the idea is that instead of range-checking the last memory channel for arithmetic operations, we check the next row's `mem_channels[0]` which should also contain the result (since it contains the new top of the stack). This enables us to check the new top of the stack for prover_inputs, syscalls and exceptions, which contains the values we want to range check. What do you think of this approach?

pgebheim:
Hey @Nashtare -- are you going to be reviewing this to get it merged?

Nashtare:
@pgebheim  yeah I can, though I didn't as @nbgl  was assigned the review already

nbgl (approved):
Thank you!

PR #1174, opened 2023-08-08
Combine get_context and set_context into one flag

LindaGuiga:
This PR addresses one of the TODOs in ops.rs: `TODO: combine GET_CONTEXT and SET_CONTEXT into one flag`. 
For this, it combines the two flags into one, and uses a boolean (based on the last bit of the opcode) to determine whether the operation is `GET_CONTEXT` or `SET_CONTEXT`. 
The stack-related constraints for `GET_CONTEXT` are added to contextops.rs, as the previously associated `StackBehavior` is now set to `None`.


nbgl:
I notice also that `nv.stack_len` is left unconstrained in `GET_CONTEXT`. But upon checking, this is actually my error and it's unconstrained in `stack.rs`. Oops!

evm/src/witness/transition.rs
```diff
@@ -184,8 +184,8 @@ fn fill_op_flag<F: Field>(op: Operation, row: &mut CpuColumnsView<F>) {
         Operation::Jumpi => &mut flags.jumpi,
         Operation::Pc => &mut flags.pc,
         Operation::Jumpdest => &mut flags.jumpdest,
-        Operation::GetContext => &mut flags.get_context,
-        Operation::SetContext => &mut flags.set_context,
+        Operation::GetContext => &mut flags.context_op,
+        Operation::SetContext => &mut flags.context_op,
```
nbgl, line 188: Nit: For consistency, I'd like to make sure that each flag column appears on the right only once.
```suggestion
        Operation::GetContext | Operation::SetContext => &mut flags.context_op,
```

nbgl (approved):
Looks good! Just one nit :)

PR #1177, opened 2023-08-10
Reduce reallocations

Nashtare:
Replaces some empty Vec / HashMap initializations with `with_capacity` to reduce the number of reallocations. Given than an empty vec will reallocate to a capacity of 4 directly on the first push, some tiny vecs are left untouched.

The number of reallocations can be evaluated with an external crate like [alloc_counter](https://docs.rs/alloc_counter/latest/alloc_counter/). Testing against `bench_recursion` example, i.e. replacing the `prove` calls with the code below

```rust
    let mut timing = TimingTree::new("prove", Level::Debug);
    let ((_, realloc, _), proof) =
        count_alloc(|| prove::<F, C, D>(&data.prover_only, &data.common, pw, &mut timing).unwrap());
    timing.print();

    println!("Reallocations for proving: {:?}", realloc);
````

 for both the initial dummy proof and the two recursive layers, a typical run yields the following:
 
 - main branch:
Reallocations for dummy proof: 119064
Reallocations for 1st rec. proof:  48632
Reallocations for 2nd rec. proof: 51235

 - this branch:
Reallocations for dummy proof: 56476 (-52.6%)
Reallocations for 1st rec. proof:  32910 (-32.3%)
Reallocations for 2nd rec. proof: 35394 (-30.9%)

@npwardberkeley Would you mind taking a look?

### DIFF
--- a/evm/src/keccak_sponge/keccak_sponge_stark.rs
+++ b/evm/src/keccak_sponge/keccak_sponge_stark.rs
@@ -358,15 +358,13 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
         row.updated_state_u32s = sponge_state.map(F::from_canonical_u32);
         let is_final_block = row.is_final_input_len.iter().copied().sum::<F>() == F::ONE;
         if is_final_block {
-            let mut cur_bytes = vec![F::ZERO; 4];
             for (l, &elt) in row.updated_state_u32s[..8].iter().enumerate() {
                 let mut cur_elt = elt;
-                for i in 0..4 {
-                    cur_bytes[i] =
+                (0..4).for_each(|i| {
+                    row.updated_state_bytes[l * 4 + i] =
                         F::from_canonical_u32((cur_elt.to_canonical_u64() & 0xFF) as u32);
                     cur_elt = F::from_canonical_u64(cur_elt.to_canonical_u64() >> 8);
-                    row.updated_state_bytes[l * 4 + i] = cur_bytes[i];
-                }
+                });
 
                 let mut s = row.updated_state_bytes[l * 4].to_canonical_u64();
                 for i in 1..4 {
--- a/evm/src/keccak_sponge/keccak_sponge_stark.rs
+++ b/evm/src/keccak_sponge/keccak_sponge_stark.rs
@@ -211,7 +211,11 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
         operations: Vec<KeccakSpongeOp>,
         min_rows: usize,
     ) -> Vec<[F; NUM_KECCAK_SPONGE_COLUMNS]> {
-        let mut rows = Vec::with_capacity(operations.len().max(min_rows).next_power_of_two());
+        let base_len: usize = operations
+            .iter()
+            .map(|op| op.input.len() / KECCAK_RATE_BYTES + 1)
+            .sum();
+        let mut rows = Vec::with_capacity(base_len.max(min_rows).next_power_of_two());
         for op in operations {
             rows.extend(self.generate_rows_for_op(op));
         }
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

File: evm/src/keccak_sponge/keccak_sponge_stark.rs [hunk 1]
```diff
@@ -358,8 +358,8 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
         row.updated_state_u32s = sponge_state.map(F::from_canonical_u32);
         let is_final_block = row.is_final_input_len.iter().copied().sum::<F>() == F::ONE;
         if is_final_block {
+            let mut cur_bytes = vec![F::ZERO; 4];
```
dlubarov, line 361: Seems like `cur_bytes` isn't really needed, could just inline it

Nashtare: Good point, just addressed it.

dlubarov (approved):
lgtm

dlubarov:
I did a quick skim, @npwardberkeley if you wouldn't mind looking also

field/src/types.rs
```diff
@@ -283,7 +283,7 @@ pub trait Field:
     }
 
     fn cyclic_subgroup_unknown_order(generator: Self) -> Vec<Self> {
-        let mut subgroup = Vec::new();
+        let mut subgroup = Vec::with_capacity(Self::TWO_ADICITY);
```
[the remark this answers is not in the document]
Nashtare: Yes that's true, I had just assumed that in most cases we wouldn't create a group of just a few elements, hence the allocation (for the Goldilocks case this skips four reallocations for instance). But while doing it I also realized this method was actually never being used, is it a leftover? Happy to revert to the default case though.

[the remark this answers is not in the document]
dlubarov: Yeah didn't end up using it, we could also just remove it. But if we keep it I think either the default or some explicit "larger default" like 256 would be good. Like if we added the Baby Bear field, it feels weird to start at 27 (since there's no subgroup of that order), or if we added Mersenne31, might be suboptimal to start at 1

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -211,7 +211,7 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
         operations: Vec<KeccakSpongeOp>,
         min_rows: usize,
     ) -> Vec<[F; NUM_KECCAK_SPONGE_COLUMNS]> {
-        let mut rows = vec![];
+        let mut rows = Vec::with_capacity(operations.len().max(min_rows).next_power_of_two());
```
npwardberkeley, line 214: Doesn't `generate_rows_for_op` typically produce multiple rows per operation? So isn't this value off from the final vector size by that factor?

Nashtare: Hmm good point, this may still cause reallocation. I guess I should change it to something like

```suggestion
        let base_len: usize = operations
            .iter()
            .map(|op| op.input.len() / KECCAK_RATE_BYTES + 1)
            .sum();
        let mut rows = Vec::with_capacity(base_len.max(min_rows).next_power_of_two());
```

npwardberkeley: Yes, that looks good!

field/src/types.rs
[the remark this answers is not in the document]
Nashtare: Given that we don't use it for now, I just reverted the changes on this one. We can always deal with this later on if its being used again.

evm/src/keccak_sponge/keccak_sponge_stark.rs
Nashtare: I've addressed it in follow-up commits, as well as Daniel's comment.

Nashtare: And related to this point, the allocation of the `HashMap` in `get_sigma_map()` at plonky2/src/plonk/permutation_argument.rs is not ideal, as it may still reallocate, but given the usually huge count of distinct keys, it saves quite a bit of reallocations still.

Nashtare: @npwardberkeley Was there anything else I should deal with related to this PR?

npwardberkeley: Oh sorry no, I think it looks good!

PR #1178, opened 2023-08-10
Combine AND and OR flags in CpuStark

Nashtare:
Merge `and` and `or` flag columns together.


nbgl:
These changes look good but there's one thing missing: there's no way for the logic table CTL to distinguish between AND and OR. For example, if we're executing an AND instruction but the logic table actually performs an OR, all the constraints here will be satisfied, but we'll get an incorrect result.

Nashtare:
After discussing with @nbgl, I'll change the approach on dealing with flags merging, hence closing this PR and will reopen when the other approach is ready.

PR #1179, opened 2023-08-10
Combine some arithmetic flags together

Nashtare:
Merge a few arithmetic operation columns together, removing 4 CPU columns.
The handling of the `mul` and `sub` is specific as their gas cost is different.


Nashtare:
After discussing with @nbgl, I'll change the approach on dealing with flags merging, hence closing this PR and will reopen when the other approach is ready.

PR #1184, opened 2023-08-11
Combine jump flags

hratoanina:


evm/src/cpu/jumps.rs
```diff
@@ -7,6 +7,7 @@ use plonky2::iop::ext_target::ExtensionTarget;
 use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
 use crate::cpu::columns::CpuColumnsView;
 use crate::cpu::membus::NUM_GP_CHANNELS;
+use crate::cpu::stack::{eval_ext_circuit_one, eval_packed_one, JUMPI_OP, JUMP_OP};
```
nbgl, line 10: Nit: I would just `use crate::cpu::stack` because it's not clear that `eval_packed_one`, `JUMP_OP`, etc. relate to stack constraints, whereas `stack::eval_packed_one` is a lot more obvious.

nbgl (approved):
Looks good!!

nbgl:
@hratoanina you have write access to the repo, so i'll let you merge this yourself when you're ready :)

PR #1185, opened 2023-08-12
Combine EQ and ISZERO flags

Nashtare:
Those 2 opcodes weren't on the list of TODOs for combining flags, but the process was relatively straightforward so I did it nonetheless.
@nbgl Would you mind reviewing it if you can?


nbgl (approved):
looks good!! thank you :)

PR #1187, opened 2023-08-15
Combine arithmetic flags on the CPU side

### LINKED ISSUE #1182: Missing check for SUBMOD in decode module
It seems the custom opcode `SUBMOD` is missing from the list of opcodes being checked in the decoding module: https://github.com/mir-protocol/plonky2/blob/dc7e0aa79bf051c1e9a01f2c12651253b4fa214d/evm/src/cpu/decode.rs#L25-L62.

Adding a ticket to mention it, though this should be handled by the combination of the arithmetic flags I'm working on.

Nashtare:
This PR combines several arithmetic flags together, removing 12 CPU columns in total.
The combinations are the following:
- basic binary arithmetic ops (ADD, MUL, SUB, DIV, MOD, LT, GT and BYTE)
- ternary arithmetic ops (ADDMOD, MULMOD and SUBMOD)
- FP254 binary arithmetic ops (ADD_FP254, MUL_FP254 and SUB_FP254)
- shift operations (SHL and SHR)

Enforcing that the correct opcode is being used between the Arithmetic and CPU tables is done through the arithmetic table CTL, which maps each operation flag to its associated opcode value, following a trick suggested by @nbgl and already used in #1181. 

For this, I added two additional flags on the arithmetic table side to identify SHL and SHR instructions independently, which unfortunately puts the number of columns above 304, inducing an additional Poseidon permutation for hashing the leaves of this table.

closes #1182


Nashtare:
@nbgl would you mind having a look at it?

evm/src/cpu/columns/ops.rs
```diff
@@ -7,33 +7,17 @@ use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
...
-    // TODO: combine SHL and SHR into one flag
-    pub shl: T,
-    pub shr: T,
+    pub shift: T, // Combines SHL and SHR flags.
     pub keccak_general: T,
     pub prover_input: T,
     pub pop: T,
-    // TODO: combine JUMP and JUMPI into one flag
-    pub jumps: T, // Note: This column must be 0 when is_cpu_cycle = 0.
+    pub jumps: T, // Combines JUMP and JUMPI flags.
```
Nashtare, line 20: This is just a cleanup from a leftover outdated comment.

evm/src/witness/traces.rs
```diff
@@ -31,7 +31,7 @@ pub struct TraceCheckpoint {
 
 #[derive(Debug)]
 pub(crate) struct Traces<T: Copy> {
-    pub(crate) arithmetic_ops: Vec<arithmetic::Operation>,
+    pub(crate) arithmetic_ops: Vec<(arithmetic::Operation, bool)>,
```
Nashtare, line 34: I'm not a fan of this tuple. Maybe it would be cleaner if it was at least put behind an alias or something?

unzvfu: An alias would probably be an improvement. I haven't looked at this PR at all yet, so apologies if this is way off, but maybe the  `bool` should be integrated into the `arithmetic::Operation` type?

Nashtare: I'll think about both approaches and come up with something in the afternoon!

evm/src/arithmetic/mod.rs
```diff
@@ -119,13 +119,19 @@ impl TernaryOperator {
     }
 }
 
+/// An enum representing arithmetic operations that can be either binary or ternary.
+///
+/// Binary operations include a special `is_simulated` flag to differentiate SHL
+/// and SHR operations from MUL and DIV respectively, as the former are simulated
+/// with the latter by scaling their inputs.
```
unzvfu, line 126: I just made a PR https://github.com/topos-protocol/plonky2/pull/3 with a suggestion for a potentially cleaner way to solve this problem. Let me know what you think.

evm/src/cpu/decode.rs
```diff
@@ -22,26 +22,15 @@ use crate::cpu::columns::{CpuColumnsView, COL_MAP};
...
-    (0x0e, 0, true, COL_MAP.op.subfp254),
-    (0x10, 0, false, COL_MAP.op.lt),
-    (0x11, 0, false, COL_MAP.op.gt),
+    // ADD, MUL, SUB, DIV, MOD, LT and GT flags are handled partly manually here, and partly through the Arithmetic table CTL.
+    // ADDMOD, MULMOD and SUBMOD flags are handled partly manually here, and partly through the Arithmetic table CTL.
+    // FP254 operation flags are handled partly manually here, and partly through the Arithmetic table CTL.
     (0x14, 1, false, COL_MAP.op.eq_iszero),
     // AND, OR and XOR flags are handled partly manually here, and partly through the Logic table CTL.
     (0x19, 0, false, COL_MAP.op.not),
-    (0x1a, 0, false, COL_MAP.op.byte),
```
unzvfu, line 42: Hmm, why is `byte` removed from here? Maybe it should be mentioned in one of those comments?

unzvfu (approved):
Looks good! Just one minor comment.

I do have another suggestion for simplifying a section, which I'll send via a PR presently. Up to you whether you think it's worth including.

evm/src/cpu/decode.rs
Nashtare: Ah sorry that's a miss, `byte` is included in the `binary_op` flag which is dealt separately below.
I'll fix the comment.

PR #1189, opened 2023-08-17
Remove is_bootstrap_kernel column

Nashtare:
Now that CPU cycles are identified by the sum of the `op` instruction flags, we can do the same for the kernel bootstrapping flag, which are exactly the rows that are not CPU cycles.


npwardberkeley (approved):
Looks good!

PR #1192, opened 2023-08-18
Misc constraints

Nashtare:
This PR amortizes a few constraints with `mul_add_extension`, and removes unnecessary filtering in over-constrained statements. 
I removed the TODO comment I had previously written in the decode module, as I thought the op flags were being individually binary constrained elsewhere which is not the case, hence the comment was erroneous.


nbgl (approved):
Thank you!!

PR #1193, opened 2023-08-18
Observe PublicValues

Nashtare:
Deals with the TODO on observing public values when generating challenges.
Note that I had to update some tests as we'd have a cast overflow otherwise.

### DIFF
--- a/evm/src/get_challenges.rs
+++ b/evm/src/get_challenges.rs
@@ -1,3 +1,4 @@
+use ethereum_types::{BigEndianHash, H256, U256};
 use plonky2::field::extension::Extendable;
 use plonky2::fri::proof::{FriProof, FriProofTarget};
 use plonky2::hash::hash_types::RichField;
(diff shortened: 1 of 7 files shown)


evm/src/get_challenges.rs
```diff
@@ -12,6 +13,117 @@ use crate::permutation::{
...
+    challenger.observe_element(F::from_canonical_u32(
+        block_metadata.block_gaslimit.as_u32(),
+    ));
+    challenger.observe_element(F::from_canonical_u32(
+        block_metadata.block_chain_id.as_u32(),
+    ));
+    challenger.observe_element(F::from_canonical_u32(
+        block_metadata.block_base_fee.as_u32(),
+    ));
+}
```
wborgeaud, line 79: Not exactly related to this PR, but I think the base fee often is larger than 2^32, which would lead to a panic here and elsewhere. 
For example, at the time of writing this comment, the [latest block](https://etherscan.io/block/17947828) uses `block_base_fee ~= 13.15 Gwei > 2^32 wei`.

wborgeaud (approved):
Looks good to me!

evm/src/get_challenges.rs
Nashtare: Yeah that's true, but by design we only store the Target versions as a single `Target`, hence limited to 32bits for all block metadata. I had written purposely a conversion from `canonical_u32` in `prove_root` to highlight this design limitation. Should we adapt those metadata fields on a case-by-base basis in a follow-up PR?

wborgeaud: > I had written purposely a conversion from `canonical_u32` in `prove_root` to highlight this design limitation.

I'm not sure I understand what you mean by that.

> Should we adapt those metadata fields on a case-by-base basis in a follow-up PR?

Yes that'd be great. In general, I don't think these metadata field should be `U256` if they're required to fit in a `u32`. 

This PR is good to merge, just noticed this issue while reviewing but it's unrelated to the PR's changes.

Nashtare: Ah sorry for the confusion, what I meant is that previously we were converting the block metadata to u64s before setting the target, which was then causing errors in the recursive circuit if they were indeed greater than `u32::MAX`. [This PR](https://github.com/mir-protocol/plonky2/pull/1161) addressed it by converting to u32s to be consistent with the limb size that is expected to fit in a Target.

As for which metadata to have fitting in more than 1 limb, do you see any other field than basefee that would require such a change? The other most likely one seems to be the gaslimit, and as we had discussed this will never be practically relevant.

wborgeaud: > As for which metadata to have fitting in more than 1 limb, do you see any other field than basefee that would require such a change?

Yes I think basefee is the only field that can be larger than 32-bit (at least for the next 83 years, after which the timestamp would become a problem). 

We should probably open an issue for this, but the block difficulty used to be larger than 32-bit but is now 0 after the Merge. The difficulty has been replaced by prevrandao, so maybe we should just store the prevrandao value in `BlockMetadata.difficulty` or rename it to `BlockMetadata.difficulty_or_prevrandao` or `BlockMetadata.prevrandao`.

Nashtare: > Yes I think basefee is the only field that can be larger than 32-bit
Alright, I'll deal with it.

> We should probably open an issue for this, but the block difficulty used to be larger than 32-bit but is now 0 after the Merge.
Are there any plans to deal with Ethereum state pre Merge? Otherwise I don't think that'd be a big issue.

wborgeaud: > Are there any plans to deal with Ethereum state pre Merge? Otherwise I don't think that'd be a big issue.

I don't think so. It's just that if we put the prevrandao value in `BlockMetadata.difficulty`, then we'd need to use 8 limbs for `BlockMetadataTarget.difficulty`.

PR #1194, opened 2023-08-19
Use 2 limbs for BlockBaseFee

Nashtare:
Allows `BlockBaseFee` to be greater than 32 bits, following discussion in #1193.

@wborgeaud: the actual changes are in the first commit. The second one deals with code duplication removal and moves the `get_memory_extra_looking_products_circuit` method to recursive_verifier.rs where I think it should belong (based on where the other "circuit-version" verification methods are implemented).


wborgeaud (approved):
Looks great, thanks!

Issue #1133, opened 2023-07-14
About constraints related to `reg_preimage` 


qope:
Regarding the evm/keccak implementation, it appears that the constraints related to `reg_preimage` are only the following one.

https://github.com/mir-protocol/plonky2/blob/00579850be0fc9caeae31ec21815b923d985187c/evm/src/keccak/keccak_stark.rs#L267-L273

This means that `reg_preimage` can be chosen independently of `reg_a` and others.

I think this would cause a soundness issue when `reg_preimage` is subject to cross-table lookups.

Am I overlooking something? I appreciate your feedback. Thank you.

shuklaayush:
Shouldn't this also be constraining the higher bits?

dlubarov:
Thanks for reporting, @npwardberkeley could you take a look later?

shuklaayush:
I'd like to create a PR for this if that's fine

npwardberkeley:
Oh yes, go for it!

shuklaayush:
I created a PR to constrain the higher bits of the `preimage`. Should I also add a constraint that `reg_a` and `reg_preimage` should be the same at the start of each input?

Nashtare:
Closing as #1201 and #1202 have been merged.

PR #1201, opened 2023-08-27
fix: constrain higher bits of reg_preimage

shuklaayush:
Adds a missing constraint that ensures the upper limb of the preimage is same across rows

Related to https://github.com/mir-protocol/plonky2/issues/1133#issuecomment-1690366448


npwardberkeley (approved):
Looks good!

PR #1203, opened 2023-08-28
Constrain next row's stack length

hratoanina:
Constrain `nv.stack_len` in `cpu/stack.rs`.


hratoanina:
This addresses a comment from @nbgl here: https://github.com/mir-protocol/plonky2/pull/1174#issuecomment-1674069063

nbgl (approved):
Looks great!! Thank you.

PR #1205, opened 2023-08-30
update CI YML for self-hosted runners

npwardberkeley:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -10,9 +10,9 @@ on:
     branches:
       - "**"
 
-jobs:  
+jobs:
   test:
-    name: Test Suite
+    name: Test on GitHub Runners
```
Nashtare, line 15: Random question as I may be missing something, but why is this kept? Is it because the self-hosted runners aren't on linux?

npwardberkeley: Hm no they are on Linux... I guess it just felt more complete to keep it, in case something goes wrong with the self-hosted runners? But I'm fine with axing it if you think that makes more sense

Nashtare: I'm not really opinionated on the matter, mostly asked out of curiosity. Removing it would also allow for much faster CI jobs completion in general, so quite convenient, but your point is valid.

npwardberkeley: Good point... @dlubarov thoughts?

dlubarov: I don't have the full context but if these are beefier VMs, I agree it would be nice to remove the old config for the speedup.

Nashtare:
Is this still on the pipeline? (can't get my eyes off a faster CI 👀)

npwardberkeley:
Ah sorry yes it is — I just have to cut back a little on the AWS machines. It's on pause for a bit because I went a little crazy and already ran up a big bill for this month 😅 But I'll try to get it all working next week!

dvdplm:
@npwardberkeley You ok if I close this for the time being? When the time comes to revive this you can always re-open the PR yeah?

PR #1206, opened 2023-08-31
Add missing links between public values

LindaGuiga:
This PR aims at implementing the missing public value links, especially for public values introduced by the receipts and logs [PR](https://github.com/mir-protocol/plonky2/pull/1097):
- It applies reads for the transaction numbers, the cumulative gas used and block bloom filters (before and after the current transaction).
- It initializes block bloom filters (it creates a segment for the block bloom filters instead of using `GlobalMetadata`, to avoid code duplication, since each bloom filter is comprised of eight 32-bits words).
- It adds links in aggregation and block proofs for `cum_gas_used`, `txn_number` and `block_filters`.
- It also links state trie roots between block proofs and ensures that transaction and receipt tries are empty at the beginning of a block.

Please let me know if there's anything I missed, or any suggestions/corrections!

### DIFF
--- a/evm/src/cpu/kernel/constants/global_metadata.rs
+++ b/evm/src/cpu/kernel/constants/global_metadata.rs
@@ -42,43 +42,49 @@ pub(crate) enum GlobalMetadata {
     BlockGasLimit = 18,
     BlockChainId = 19,
     BlockBaseFee = 20,
-
+    BlockGasUsed = 21,
+    /// Before current transactions block values.
+    BlockGasUsedBefore = 22,
+    /// After current transactions block values.
+    BlockGasUsedAfter = 23,
     /// Gas to refund at the end of the transaction.
-    RefundCounter = 21,
+    RefundCounter = 24,
     /// Length of the addresses access list.
-    AccessedAddressesLen = 22,
+    AccessedAddressesLen = 25,
     /// Length of the storage keys access list.
-    AccessedStorageKeysLen = 23,
+    AccessedStorageKeysLen = 26,
     /// Length of the self-destruct list.
-    SelfDestructListLen = 24,
+    SelfDestructListLen = 27,
     /// Length of the bloom entry buffer.
-    BloomEntryLen = 25,
+    BloomEntryLen = 28,
 
     /// Length of the journal.
-    JournalLen = 26,
+    JournalLen = 29,
     /// Length of the `JournalData` segment.
-    JournalDataLen = 27,
+    JournalDataLen = 30,
     /// Current checkpoint.
-    CurrentCheckpoint = 28,
-    TouchedAddressesLen = 29,
+    CurrentCheckpoint = 31,
+    TouchedAddressesLen = 32,
     // Gas cost for the access list in type-1 txns. See EIP-2930.
-    AccessListDataCost = 30,
+    AccessListDataCost = 33,
     // Start of the access list in the RLP for type-1 txns.
-    AccessListRlpStart = 31,
+    AccessListRlpStart = 34,
     // Length of the access list in the RLP for type-1 txns.
-    AccessListRlpLen = 32,
+    AccessListRlpLen = 35,
     // Boolean flag indicating if the txn is a contract creation txn.
-    ContractCreation = 33,
-    IsPrecompileFromEoa = 34,
-    CallStackDepth = 35,
-    /// Transaction logs list length
-    LogsLen = 36,
-    LogsDataLen = 37,
-    LogsPayloadLen = 38,
+    ContractCreation = 36,
+    IsPrecompileFromEoa = 37,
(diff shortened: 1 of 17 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -674,6 +724,88 @@ where
...
+        // The initial number of transactions is 0.
+        builder.connect(x.extra_block_data.txn_number_before, zero);
+        // The initial gas used is 0
+        builder.connect(x.extra_block_data.gas_used_before, zero);
+
+        // The initial bloom filter is all zeroes.
+        let initial_bloom = builder.constants(&[F::ZERO; 64]);
+        for i in 0..x.extra_block_data.block_bloom_before.len() {
+            builder.connect(x.extra_block_data.block_bloom_before[i], initial_bloom[i]);
+        }
```
wborgeaud, line 789: nit: Why not just something like
```rust
for t in x.extra_block_data.block_bloom_before {
   builder.connect(t, zero);
}
```
?

evm/src/fixed_recursive_verifier.rs
```diff
@@ -674,6 +724,88 @@ where
...
+            .iter()
+            .zip(rhs.trie_roots_before.state_root)
+        {
+            builder.connect(limb0, limb1);
+        }
+
+        // Connect block numbers.
+        let one = builder.one();
+        let prev_block_nb = builder.sub(rhs.block_metadata.block_number, one);
+        builder.connect(lhs.block_metadata.block_number, prev_block_nb);
```
wborgeaud, line 745: Should we also add a constraint of the form `(1-has_parent_block) * agg_root_proof.block_number = 0` somewhere to ensure that the genesis block has `block_number=0` (and thus to ensure there's no underflow here except for the genesis block)?

Also applies to checking that the genesis block has a predetermined state trie root.

evm/src/fixed_recursive_verifier.rs
```diff
@@ -804,12 +940,26 @@ where
             block_inputs
                 .set_proof_with_pis_target(&self.block.parent_block_proof, parent_block_proof);
         } else {
+            // Initialize state_root_after and the block number for correct connection between blocks.
+            let state_trie_root_keys = 24..32;
+            let block_number_key = TrieRootsTarget::SIZE * 2 + 6;
+            let mut nonzero_pis = HashMap::new();
+            for (key, &value) in state_trie_root_keys.zip_eq(&u256_limbs::<F>(
+                public_values.trie_roots_before.state_root.into_uint(),
```
wborgeaud, line 948: nit: Can use `h256_limbs` instead of `u256_limbs(hash.into_uint())`.

evm/src/fixed_recursive_verifier.rs
```diff
@@ -804,12 +940,26 @@ where
...
+            )) {
+                nonzero_pis.insert(key, value);
+            }
+            nonzero_pis.insert(
+                block_number_key,
+                F::from_canonical_usize(public_values.block_metadata.block_number.as_usize())
```
wborgeaud, line 954: Just to make sure I'm not missing anything, these public values will only be verified for the initial block proof of the block chain. So in this case `block_number` will be zero, right? Therefore, this should just be `F::NEG_ONE`.

evm/src/cpu/kernel/constants/global_metadata.rs
```diff
@@ -103,6 +109,9 @@ impl GlobalMetadata {
             Self::BlockGasLimit,
             Self::BlockChainId,
             Self::BlockBaseFee,
+            Self::BlockGasUsed,
```
wborgeaud, line 112: What does `BlockGasUsed` represent? Looking in this PR, it doesn't seem to be used except for the check
```
builder.connect(
            x.block_metadata.block_gas_used,
            x.extra_block_data.gas_used_after,
);
```
which I don't really understand.

wborgeaud (approved):
Looks great!

evm/src/cpu/kernel/constants/global_metadata.rs
wborgeaud: Ah right, looks good then 👍

evm/src/fixed_recursive_verifier.rs
LindaGuiga: For the genesis block's state trie, I added a check that the state trie was empty.

wborgeaud: We can leave it as a TODO for now, but I think it could be useful to support non-empty genesis state tries e.g. for premines.

LindaGuiga: It makes sense, thank you! I removed the state trie check for genesis blocks and added a TODO instead.

Issue #1204, opened 2023-08-30
Implement PREVRANDAO instruction


pgebheim:
Discussion today was that this could be added with the value as an external input.

@wborgeaud could you elaborate?

wborgeaud:
So currently our implementation of PREVRANDAO (0x44) returns the `difficulty` field of the block header (which is always 0 post-merge). 
As explained in [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399), we should instead return the `mixHash` field of the block header. So to fix this issue we just have to add `mixHash` to the block metadata and return it in `sys_prevrandao`. 
I think we can just have `mixHash` be unconstrained in Plonky2 and assume it's been verified upstream in the consensus layer.

PR #1207, opened 2023-08-31
Add `random` value to block metadata and fix `sys_prevrandao`

wborgeaud:
See #1204 for details.


evm/src/proof.rs
```diff
@@ -68,6 +68,7 @@ pub struct BlockMetadata {
     pub block_timestamp: U256,
     pub block_number: U256,
     pub block_difficulty: U256,
+    pub block_random: U256,
```
Nashtare, line 71: Perhaps an overlook, but shouldn't this rather be an `H256` out of consistency with `mix_hash` output?

evm/src/cpu/kernel/asm/core/syscall_stubs.asm
LindaGuiga, line 1: Should this be removed from `syscall_stubs.asm` and put in (for example) `metadata.asm` instead?

evm/src/proof.rs
```diff
@@ -336,6 +346,9 @@ impl BlockMetadataTarget {
         builder.connect(bm0.block_timestamp, bm1.block_timestamp);
         builder.connect(bm0.block_number, bm1.block_number);
         builder.connect(bm0.block_difficulty, bm1.block_difficulty);
+        for i in 0..8 {
+            builder.connect(bm0.block_random[i], bm1.block_random[i]);
```
LindaGuiga, line 350: `block_random` is connected here, but it isn't set in `set_block_metadata_target`. This leads to a `generators not run` error in a recursive proof.

evm/src/cpu/kernel/asm/core/syscall_stubs.asm
```diff
@@ -10,11 +10,9 @@ global sys_blockhash:
     EXIT_KERNEL
 
 // This is a temporary version that returns the block difficulty (i.e. the old version of this opcode).
```
LindaGuiga, line 12: Shouldn't this comment be deleted or updated?

LindaGuiga (commented):
Thanks for implementing this! Looks great, but there's just one missing thing: `block_random` is not set in `set_block_metadata_target`.

Nashtare (approved):
Looks good!

PR #1208, opened 2023-08-31
Add blockhash opcode

### LINKED ISSUE #1173: Implement BLOCKHASH opcode
The BLOCKHASH opcode returns the hash of one of the 256 most recent complete blocks (not including the current one).

Here are some ideas on how to implement it:

- Output a value provided by the prover using `PROVER_INPUT`. (Not very sound)
- Add the last 256 block hashes to the public values and check for consistency in the block aggregation proof.
- Write the block hashes somewhere in the state trie (or the txs/receipts tries), preferably at a slot that cannot otherwise be accessed (e.g. the all zeros slot). This would slightly impair Ethereum equivalence. (Can be used to query older blocks if needed)
- Use an extra vector commitment (e.g. Merkle tree) that is updated at every block. (Can be used to query older blocks if needed)

LindaGuiga:
This PR implements the BLOCKHASH opcode, following @wborgeaud's second suggestion [in this issue](https://github.com/mir-protocol/plonky2/issues/1173). It includes links between proofs, a couple of unit tests and the challenger observing the additional public values (the previous and current block hashes).

closes #1173

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/metadata.asm
+++ b/evm/src/cpu/kernel/asm/memory/metadata.asm
@@ -249,7 +249,7 @@ global blockhash:
     // stack: cur_block_number, block_number, retdest
     DUP1 DUP3 %increment GT %jumpi(zero_hash) // if block_number >= cur_block_number
     // stack: cur_block_number, block_number, retdest
-    DUP2 PUSH 256 ADD 
+    DUP2 PUSH 256 %add_or_fault
     // stack: block_number+256, cur_block_number, block_number, retdest
     DUP2 GT %jumpi(zero_hash) // if cur_block_number > block_number + 256
     // If we are here, the provided block number is correct
--- a/evm/src/cpu/kernel/asm/memory/metadata.asm
+++ b/evm/src/cpu/kernel/asm/memory/metadata.asm
@@ -247,9 +247,12 @@ global blockhash:
     // stack: block_number, retdest
     %mload_global_metadata(@GLOBAL_METADATA_BLOCK_NUMBER)
     // stack: cur_block_number, block_number, retdest
+    // Check for an overflow, since we're incrementing `block_number` afterwards.
+    DUP2 %eq_const(@U256_MAX) %jumpi(zero_hash)
+    // stack: cur_block_number, block_number, retdest
     DUP1 DUP3 %increment GT %jumpi(zero_hash) // if block_number >= cur_block_number
     // stack: cur_block_number, block_number, retdest
-    DUP2 PUSH 256 %add_or_fault
+    DUP2 PUSH 256 ADD
     // stack: block_number+256, cur_block_number, block_number, retdest
     DUP2 GT %jumpi(zero_hash) // if cur_block_number > block_number + 256
     // If we are here, the provided block number is correct
--- a/evm/src/cpu/kernel/asm/memory/metadata.asm
+++ b/evm/src/cpu/kernel/asm/memory/metadata.asm
@@ -235,6 +235,45 @@ global sys_basefee:
     SWAP1
     EXIT_KERNEL
 
+global sys_blockhash:
+    // stack: kexit_info, block_number
+    %charge_gas_const(@GAS_BLOCKHASH)
+    SWAP1
+    // stack: block_number, kexit_info
+    %blockhash
+    EXIT_KERNEL
+
+global blockhash:
+    // stack: block_number, retdest
+    %mload_global_metadata(@GLOBAL_METADATA_BLOCK_NUMBER)
+    // stack: cur_block_number, block_number, retdest
+    DUP1 DUP3 %increment GT %jumpi(zero_hash) // if block_number >= cur_block_number
+    // stack: cur_block_number, block_number, retdest
+    DUP2 PUSH 256 ADD 
+    // stack: block_number+256, cur_block_number, block_number, retdest
+    DUP2 GT %jumpi(zero_hash) // if cur_block_number > block_number + 256
+    // If we are here, the provided block number is correct
+    SUB
+    // stack: cur_block_number - block_number, retdest
+    PUSH 256 SUB
+    // stack: block_hash_number, retdest
+    %mload_kernel(@SEGMENT_BLOCK_HASHES)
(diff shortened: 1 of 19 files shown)


evm/src/cpu/kernel/asm/memory/metadata.asm
```diff
@@ -235,6 +235,45 @@ global sys_basefee:
...
+    %blockhash
+    EXIT_KERNEL
+
+global blockhash:
+    // stack: block_number, retdest
+    %mload_global_metadata(@GLOBAL_METADATA_BLOCK_NUMBER)
+    // stack: cur_block_number, block_number, retdest
+    DUP1 DUP3 %increment GT %jumpi(zero_hash) // if block_number >= cur_block_number
+    // stack: cur_block_number, block_number, retdest
+    DUP2 PUSH 256 ADD
```
wborgeaud, line 252: This should be an addition with an overflow check.

evm/src/cpu/kernel/tests/block_hash.rs
```diff
@@ -0,0 +1,97 @@
...
+fn test_small_index_block_hash() -> Result<()> {
+    let mut rng = rand::thread_rng();
+
+    let blockhash_label = KERNEL.global_labels["blockhash"];
+    let retdest = 0xDEADBEEFu32.into();
+    let cur_block_number = 512;
+    let block_number = rng.gen::<u8>() as usize;
+    let initial_stack = vec![retdest, block_number.into()];
+
+    let hashes: Vec<U256> = (20..277).map(|elt| elt.into()).collect();
```
wborgeaud, line 79: nit: Not that it matters, but this is a bit weird. Why not random hashes like in the other tests?

evm/src/get_challenges.rs
```diff
@@ -100,6 +100,37 @@ fn observe_block_metadata_target<
...
+    challenger: &mut Challenger<F, C::Hasher>,
+    block_hashes: &BlockHashes,
+) {
+    for i in 0..256 {
+        challenger.observe_elements(
+            &u256_limbs::<F>(U256::from_big_endian(&block_hashes.prev_hashes[i].0))[0..8],
+        );
+    }
+    challenger
+        .observe_elements(&u256_limbs::<F>(U256::from_big_endian(&block_hashes.cur_hash.0))[0..8])
```
wborgeaud, line 117: nit: Can use `h256_limbs` instead

evm/src/cpu/kernel/tests/block_hash.rs
```diff
@@ -0,0 +1,97 @@
...
+    assert_eq!(
+        result[0],
+        0.into(),
+        "Resulting block hash {:?} different from expected hash {:?}",
+        result[0],
+        0
+    );
+
+    Ok(())
+}
```
wborgeaud, line 97: Could be nice to have a test for the overflow, e.g. using `cur_block_number=0` and `block_number =  U256::MAX`.

evm/src/recursive_verifier.rs
```diff
@@ -900,3 +976,28 @@ pub(crate) fn set_block_metadata_target<F, W, const D: usize>(
...
+                .unwrap();
+        witness.set_target_arr(
+            &block_hashes_target.prev_hashes[8 * i..8 * (i + 1)],
+            &block_hash_limbs,
+        );
+    }
+    let cur_block_hash_limbs: [F; 8] =
+        u256_limbs::<F>(U256::from_big_endian(&block_hashes.cur_hash.0))[..8]
+            .try_into()
+            .unwrap();
```
wborgeaud, line 1001: nit: Can use `h256_limbs`.

wborgeaud (approved):
Looks great!

evm/src/cpu/kernel/tests/block_hash.rs
LindaGuiga: When I added the overflow test, the interpreter failed with 'Integer overflow when casting to usize' in `recert_checkpoint`. So I'll add the test with `#[should_panic]` on top.

evm/src/cpu/kernel/asm/memory/metadata.asm
```diff
@@ -235,6 +235,45 @@ global sys_basefee:
...
+    %blockhash
+    EXIT_KERNEL
+
+global blockhash:
+    // stack: block_number, retdest
+    %mload_global_metadata(@GLOBAL_METADATA_BLOCK_NUMBER)
+    // stack: cur_block_number, block_number, retdest
+    DUP1 DUP3 %increment GT %jumpi(zero_hash) // if block_number >= cur_block_number
+    // stack: cur_block_number, block_number, retdest
+    DUP2 PUSH 256 %add_or_fault
```
wborgeaud, line 252: I don't think we should use `add_or_fault` to check for an overflow here, we should return 0 in the case of an overflow, not an error.

But actually, I think we can go back to doing `%add_const(256)` since we've check that `block_number < cur_block_number` and we can safely assume `cur_block_number` is less than `U256::MAX -  256`.
We should however check for an overflow in the `%increment`, but for that we can just add
```
// stack: block_number, retdest
DUP1 %eq_const(@U256_MAX) %jumpi(zero_hash)
```
above.

evm/src/cpu/kernel/tests/block_hash.rs
wborgeaud: See my other comment, if we remove `%add_or_fault` this shouldn't panic.

evm/src/get_challenges.rs
wborgeaud: Looks good thanks! Also nit, but I think we can remove the `[0..8]`.

evm/src/cpu/kernel/asm/memory/metadata.asm
LindaGuiga: As mentioned in the following comments, I left this as is and added the overflow check on `block_number` directly.

evm/src/cpu/kernel/asm/memory/metadata.asm
LindaGuiga: I added the check on the following line so that `zero_hash` has the expected stack.

Issue #1120, opened 2023-07-10
Reduce the size of the memory table


nbgl:
Few ideas:
1. Native instruction for reading 32 bytes (in separate memory locations) into one register. This will require a separate table, but this table is needed to sanely do PUSH anyway.
2. Native instruction for _writing_ 32 bytes (again, in separate memory locations) from one register. This is very similar to 1. but will require range checks to show that the register was broken up correctly.
3. Keep the top of the stack in the CPU table registers.

Another idea, but I'm not sure how realistic it is, would be to only have one read row per instruction. We do not write to the code segment (? or at least it's frozen once we do write to it), so we don't have to worry about it changing. But I'm not sure how to do multiplicity in the CTL.

Nashtare:
@nbgl Now that I have spent some time on the byte packing approach, I am not sure this will yield a significant enough improvement to justify the new table. 
Available implementation is there if it matters: [plonky2/packing_table](https://github.com/topos-protocol/plonky2/tree/packing_table).

Let's consider a typical example of a full `mload_packing` call, reading 32 bytes. We currently do:

```asm
  - 31 ADD (3 memory rows each)
  -  3 DUP3 (2 memory rows each)
  - 93 DUP4 (2 memory rows each)
  - 31 DUP5 (2 memory rows each)
  - 32 MLOAD_GENERAL (5 memory rows each)
  - 31 %add_const (4 memory rows each)
  - 31 %eq_const (4 memory rows each)
  - 31 %shl_const (4 memory rows each)
  -  1 %stack which corresponds to 2 SWAP and 4 POP (4 and 1 memory rows each)
  -  4 POP (1 memory row each)
  - 31 JUMPI (2 memory rows)

→ total = 957 memory rows
```

With the `BytePackingStark` table, we can reduce it to:

```asm
  - 1 MLOAD_32BYTES (5 + 32 memory rows)
  - 1 SWAP (4 memory rows)
  - 1 JUMP (1 memory row)

→ total 42 memory rows
```

For a global reduction of 915 memory rows for a full `mload_packing` call. However, many places only read a single byte, removing only 19 rows. For portions actually using a full `mload_packing`, like the curve precompiles, the gain is negligible compared to the overhead induced by the rest of the operations. Below are a few trace sizes obtained from specific tests from the EVM suite:

-  `pairingTest_0`

| Branch | Arith |BP | CPU |Keccak | KeccakSp |Logic | Memory |
[...]

PR #1212, opened 2023-09-06
Move byte packing / unpacking to a distinct table

Nashtare:
Following suggestion for points 1 and 2 in #1120, this PR introduces a new `BytePacking` table to reduce the CPU/Memory overhead induced by the `%mload_packing` and `%mstore_unpacking` operations in `packing.asm`.

More specifically (thanks @nbgl for the explanations!), there are now two new instructions, `MLOAD_32BYTES` and `MSTORE_32BYTES` which can handle sequence of bytes of length up to 32. This allows to move the heavy packing / unpacking logic from software to constraints.

In terms of trace lengths reduction, see [the tables in this comment](https://github.com/mir-protocol/plonky2/issues/1120#issuecomment-1707429453) for a comparison between the current main branch and this one, against a few EVM test cases. It does incur a light overhead for native token transfer in terms of proving time but significantly reduces the number of CPU ops and memory accesses for precompiles.

Note that because both packing/unpacking are done within the same table, the range check necessary for writing is also (needlessly) applied on the read bytes. 

~In terms of # of columns, I'd ideally remove an extra one, so that we would remove one Poseidon permutation per leaf hashing, but I assume this can be left as a follow-up PR.~
EDIT: I actually removed the FILTER column, effectively getting 1 less Poseidon permutation per leaf hashing. We still may remove other columns but this becomes less important now.


Nashtare:
@unzvfu Would you mind taking a look at this?

unzvfu:
> @unzvfu Would you mind taking a look at this?

Sure, maybe tomorrow with any luck.

Noticed that you're force-pushing to the branch. This is almost never necessary and should absolutely be avoided on public repos since it breaks the repo on others' machines. Much better to just have the extra commits rather than rewrite the history. Perhaps check your config in case it's being enabled automatically?

Nashtare:
> Sure, maybe tomorrow with any luck.

Take your time, and thanks again!

> Noticed that you're force-pushing to the branch. This is almost never necessary and should absolutely be avoided on public repos since it breaks the repo on others' machines.

I agree, though some were necessary because of breaking changes in latest main. I've dealt with them before requesting for a review though to not have you or anyone looking get history changes in the process.

Nashtare:
Note that we may also be able to combine the two new flags on the CPU side, though I haven't put any thoughts on this yet. This can probably be left as a follow-up PR.

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -0,0 +1,553 @@
...
+        yield_constr.constraint_transition(
+            current_sequence_end * next_filter * (next_sequence_start - one),
+        );
+
+        // The remaining length of a byte sequence must decrease by one or be zero.
+        let current_sequence_length = vars.local_values[SEQUENCE_LEN];
+        let current_remaining_length = current_sequence_length
+            - (0..NUM_BYTES)
+                .map(|i| vars.local_values[index_bytes(i)] * P::Scalar::from_canonical_usize(i + 1))
+                .sum::<P>();
```
unzvfu, line 292: Nit: Might be clearer to assign this complicated expression to an intermediate variable, like `current_position` or similar? For example,
```suggestion
        // There is only one `i` for which `vars.local_values[index_bytes(i)]` is non-zero,
        // and `i+1` is the current position: 
        let current_position = (0..NUM_BYTES)
            .map(|i| vars.local_values[index_bytes(i)] * P::Scalar::from_canonical_usize(i + 1))
            .sum::<P>();
        let current_remaining_length = current_sequence_length - current_position;
```

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -0,0 +1,553 @@
...
+        let next_sequence_length = vars.next_values[SEQUENCE_LEN];
+        let next_remaining_length = next_sequence_length
+            - (0..NUM_BYTES)
+                .map(|i| vars.next_values[index_bytes(i)] * P::Scalar::from_canonical_usize(i + 1))
+                .sum::<P>();
```
unzvfu, line 297: Nit: Same as for `current_remaining_length` above; might be worth pulling that long expression out into a function?

unzvfu (approved):
Thanks Robin, this is excellent! Could just do with a bit more documentation in places.

evm/src/byte_packing/columns.rs
```diff
@@ -0,0 +1,46 @@
...
+pub(crate) const ADDR_VIRTUAL: usize = ADDR_SEGMENT + 1;
+pub(crate) const TIMESTAMP: usize = ADDR_VIRTUAL + 1;
+
+/// The total length of a sequence of bytes.
+/// Cannot be greater than 32.
+pub(crate) const SEQUENCE_LEN: usize = TIMESTAMP + 1;
+
+// 32 byte limbs hold a total of 256 bits.
+const BYTES_VALUES_START: usize = SEQUENCE_LEN + 1;
+pub(crate) const fn value_bytes(i: usize) -> usize {
```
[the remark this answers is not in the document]
Nashtare: I originally had those for both `index_bytes` and `value_bytes`, however when removing the u32 limb columns and reconstructed the values directly within the CTL, I was getting an assertion error and ended up removing them.
For clarity as you suggest I'll add them back and switch the CTL range from `(value_bytes(i * 4)..value_bytes((i + 1) * 4))` to `(value_bytes(i * 4)..value_bytes(i * 4) + 4)`.

Nashtare:
Thanks again for the review @unzvfu! 🙏🏼 
I still have to deal with the extensive documentation, but I've addressed most of the technical comments. 

I'll ping you again once I've added the documentation.

unzvfu:
> @unzvfu I've pushed a first attempt at a more descriptive module explanation. I'm not sure how extensive you'd like it to be (and as I wrote the code I'm not sure where the dark spots are) though I tried addressing all the points you mentioned.

That's great, thanks @Nashtare.

PR #1216, opened 2023-09-06
Display actual trace lengths instead of number of ops

Nashtare:
The current `TraceCheckpoint` used for logging info only displays the number of operations per STARK module.

### DIFF
--- a/evm/src/witness/traces.rs
+++ b/evm/src/witness/traces.rs
@@ -8,13 +8,14 @@ use plonky2::timed;
 use plonky2::util::timing::TimingTree;
 
 use crate::all_stark::{AllStark, NUM_TABLES};
+use crate::arithmetic::{BinaryOperator, Operation};
 use crate::config::StarkConfig;
 use crate::cpu::columns::CpuColumnsView;
 use crate::keccak_sponge::columns::KECCAK_WIDTH_BYTES;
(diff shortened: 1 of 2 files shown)


Nashtare:
@npwardberkeley Would you mind having a look when you have time? It's fairly short!

evm/src/witness/traces.rs
```diff
@@ -48,6 +49,47 @@ impl<T: Copy> Traces<T> {
...
+    /// Returns the actual trace lengths for each STARK module.
+    //  Uses a `TraceCheckPoint` as return object for convenience.
+    pub fn get_lengths(&self) -> TraceCheckpoint {
+        TraceCheckpoint {
+            arithmetic_len: self
+                .arithmetic_ops
+                .iter()
+                .map(|op| match op {
+                    Operation::TernaryOperation {
+                        operator: _,
```
npwardberkeley, line 61: I think these unused struct fields in the match can be replaced with `..`

npwardberkeley (approved):
Looks good! (Apart from one comment)

evm/src/witness/traces.rs
Nashtare: Learning new things every day 😄 Thanks! Definitely much cleaner that way!

PR #1195, opened 2023-08-19
chore: add rust-toolchain

jonathanpwang:
Resolves https://github.com/mir-protocol/plonky2/issues/1186

PR #1220, opened 2023-09-07
Latest nightly fix

### LINKED ISSUE #1186: Compile issue: the destructor for this type cannot be evaluated in constant functions
I am attempting to run tests in the `plonky2_evm` package and receive a compilation error:

```bash
$ RUSTFLAGS="-C target-cpu=native" cargo +nightly test --release --package plonky2_evm
   Compiling plonky2_evm v0.1.1 (/home/jimpo/Code/github.com/mir-protocol/plonky2/evm)
error[E0493]: destructor of `[std::ops::RangeInclusive<u8>; N]` cannot be evaluated at compile-time
  --> evm/src/cpu/kernel/constants/exc_bitfields.rs:7:53
   |
7  | const fn u256_from_set_index_ranges<const N: usize>(ranges: [RangeInclusive<u8>; N]) -> U256 {
   |                                                     ^^^^^^ the destructor for this type cannot be evaluated in constant functions
...
29 | }
   | - value is dropped here

For more information about this error, try `rustc --explain E0493`.
error: could not compile `plonky2_evm` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `plonky2_evm` (lib test) due to previous error
```

My rustc version is:

```bash
$ rustc +nightly --version
rustc 1.73.0-nightly (a6f8aa5a0 2023-08-11)
$ cargo +nightly --version
cargo 1.73.0-nightly (d78bbf4bd 2023-08-03)
```

BGluth:
There's one compiler error that's preventing us from building on the latest nightly, and I managed to resolve it by making a paremeter into a reference. I tried a few other things before resorting to using a reference (like `#![feature(const_precise_live_drops)]`), but I couldn't get it to compile. I *think* using a reference here should have no performance changes (since it's still a `const fn` at the end of the day), but let me know if there's a better way to fix this error.

For reference, the error was this:
```sh
error[E0493]: destructor of `[std::ops::RangeInclusive<u8>; N]` cannot be evaluated at compile-time
  --> evm/src/cpu/kernel/constants/exc_bitfields.rs:7:53
   |
7  | const fn u256_from_set_index_ranges<const N: usize>(ranges: [RangeInclusive<u8>; N]) -> U256 {
   |                                                     ^^^^^^ the destructor for this type cannot be evaluated in constant functions
...
29 | }
   | - value is dropped here
```

Resolves #1186.

### DIFF
--- a/evm/src/cpu/kernel/constants/exc_bitfields.rs
+++ b/evm/src/cpu/kernel/constants/exc_bitfields.rs
@@ -4,7 +4,7 @@ use ethereum_types::U256;
 
 /// Create a U256, where the bits at indices inside the specified ranges are set to 1, and all other
 /// bits are set to 0.
-const fn u256_from_set_index_ranges<const N: usize>(ranges: [RangeInclusive<u8>; N]) -> U256 {
+const fn u256_from_set_index_ranges<const N: usize>(ranges: &[RangeInclusive<u8>; N]) -> U256 {
     let mut j = 0;
     let mut res_limbs = [0u64; 4];
     while j < ranges.len() {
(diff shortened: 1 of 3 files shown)


BGluth:
I don't know if we want to do this, but we could also update the CI to use the latest nightly (or keep it on `2023-06-30` if we want something that we know will keep working).

npwardberkeley:
That sounds good to me!! It'd be nice to be on the latest version everywhere, including in the CI. Regarding #1195, can we set the `rust-toolchain` to just `nightly`? Or does it always need to be a particular one that we'd need to keep updating?

BGluth:
We can definitely just have it point to the latest `nightly`. It just might mean that CI will randomly fail (which maybe is a good thing anyways) and we'll have to fix it.

PR #1221, opened 2023-09-07
Remove `LookupGate` and `LookupTableGate` for build optimization

LindaGuiga:
When using lookups with a larger lookup table (of size 2^16), we noticed that `add_all_lookups` incurred a very high overhead in build time, due to `find_slot`. 
This PR addresses this issue as follows:
- We remove `last_lut_row` from `LookupTableGate`.
- Since `LookupTableGate` doesn't have any generators, the previous point enables us to remove the gates from the circuit.
- We also remove `LookupGates`, and only add their generators in `add_all_lookups`.
We still keep the two gate types, since they allow us to use their methods (such as `get_wire_ith_input`).

With these changes, when testing with 500,000 lookups on a lookup table of size 2^16, we go, from ~3100s execution time for `add_all_lookups` to ~0.27s. The tests were carried out on a MacBook Pro with processor: 2,3 GHz Dual-Core Intel Core i5 (using 4 threads on 4 cores).

Please let me know what you think!

### DIFF
--- a/plonky2/src/gadgets/lookup.rs
+++ b/plonky2/src/gadgets/lookup.rs
@@ -1,8 +1,11 @@
+use plonky2_util::ceil_div_usize;
+
 use crate::field::extension::Extendable;
-use crate::gates::lookup::LookupGate;
+use crate::gates::lookup::{LookupGate, LookupGenerator};
 use crate::gates::lookup_table::{LookupTable, LookupTableGate};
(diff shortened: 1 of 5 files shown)


Nashtare:
@onsen-egg Friendly ping as this may be of interest to you 🙂

wborgeaud:
The performance issue has been solved with an alternative approach in #1258.

Issue #1149, opened 2023-07-27
Store the top of the stack in the CPU table


nbgl:
Currently the top of the stack is stored in memory. Keeping it directly in the CPU table could reduce the number of memory accesses in a trace (issue #1120).

Consider the `ADD` instruction. Its semantics are:
```
    x := mem_stack[stack_len - 1]
    y := mem_stack[stack_len - 2]
    z := x + y
    stack_len := stack_len - 1
    stack[stack_len - 1] := z
```
It requires two memory reads and one write. Together with the opcode, this means that the `ADD` instruction currently requires four memory accesses.

Suppose now that we have a special register called `stack_top`, which contains the top of the stack, replacing `stack[stack_len - 1]`. The `ADD` instruction becomes:
```
    x := stack_top
    y := mem_stack[stack_len - 2]
    z := x + y
    stack_len := stack_len - 1
    stack_top := z
```
Instead of four memory accesses, we only need two.

Not every instruction will benefit. Any instruction that pops the top _n_ ≥ 1 elements and performs a push will have its number of memory accesses reduced by two. `SWAP*` will require three memory accesses instead of five. `DUP1` will require two accesses instead of three, although `DUPn` does not benefit for _n_ ≥ 2. `MLOAD_GENERAL` will access memory four times from six, although `MSTORE_GENERAL` does not benefit as it does not store anything to the stack.

Instructions that purely push, such as `PUSH*` or `GET_CONTEXT` do not benefit either, as they have to write the old `stack_top` to memory.

`POP` is the only instruction whose memory burden increases. Since it needs to read the new top of the stack into the CPU register, it has to perform two memory interactions instead of one. Indeed, any instruction that pops without pushing a new element onto the stack has to read in the new top of the stack. (Other such instructions do not benefit from this change but are not harmed by it either; `POP` is unique in that it does not _use_ the top of the stack, so traditionally it hasn't needed to read it.)

### `pow7`
Consider the function:
```
global pow7:
    // stack: x, retdest
    DUP1
    DUP1
    DUP1
    DUP1
    // stack: x, x, x, x, retdest
    MUL
    // stack: x ** 2, x, x, retdest
    MUL
    // stack: x ** 3, x, retdest
    DUP1
    MUL
    // stack: x ** 6, x, retdest
    MUL
    // stack: x ** 7, retdest
    SWAP1
    // stack: retdest, x ** 7
    JUMP
```

The memory op counts per operation are:
[...]

PR #1229, opened 2023-09-12
Make next row available to CTLs

hratoanina:
Cf. the Related work section of https://github.com/mir-protocol/plonky2/issues/1149.

When declaring a CTL, we can now access both `lv` and `nv`. For ease of implementation, all `nv` columns are declared after the `lv` columns. This PR does not bring anything concrete since no CTL needs it, but it paves the way to future optimizations.

In term of implementation details, the only thing of note is that the CTL Z polynomials are upside down, with the full product in the first row and the initial value in the last row. This allows us to compute the transition constraint while only using values from `lv` and `nv` (previously, we would have needed values from the row after `nv`).


npwardberkeley (approved):
Looks great! Thanks!

PR #1230, opened 2023-09-12
Swap ordering in stack macro

Nashtare:
Slight overlook from #1212, but swapping `offset` and `len` in the stack macro of `mstore_unpacking` removes one `SWAP` from the set of operations it expands to (5 `SWAP` + 2 `DUP` originally).

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/packing.asm
+++ b/evm/src/cpu/kernel/asm/memory/packing.asm
@@ -43,7 +43,7 @@ global mload_packing_u64_LE:
 global mstore_unpacking:
     // stack: context, segment, offset, value, len, retdest
     %stack(context, segment, offset, value, len, retdest) -> (context, segment, offset, value, len, offset, len, retdest)
-    // stack: context, segment, offset, value, len, len, offset, retdest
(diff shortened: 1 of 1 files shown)


evm/src/cpu/kernel/asm/memory/packing.asm
```diff
@@ -42,10 +42,10 @@ global mload_packing_u64_LE:
 // Post stack: offset'
 global mstore_unpacking:
     // stack: context, segment, offset, value, len, retdest
-    %stack(context, segment, offset, value, len, retdest) -> (context, segment, offset, value, len, len, offset, retdest)
+    %stack(context, segment, offset, value, len, retdest) -> (context, segment, offset, value, len, offset, len, retdest)
     // stack: context, segment, offset, value, len, len, offset, retdest
```
unzvfu, line 46: Nit: This comment now has the wrong order too. Up to you whether you adjust the comment or just remove it entirely (the `%stack` macro is communicating the same information in any case).

unzvfu (approved):
Looks good, just one minor fix.

PR #1231, opened 2023-09-12
Remove risks of panic

Nashtare:
The `as_u32` / `as_u64` methods of `U256` will panic if the integer is actually greater, which is not ideal.
This PR handles soft conversions which error instead of panicking. 

Also prevents panicking from the interpreter methods and tests.


evm/src/util.rs
```diff
@@ -45,6 +47,29 @@ pub fn trace_rows_to_poly_values<F: Field, const COLUMNS: usize>(
         .collect()
 }
 
+/// Returns the lowest LE 32-bit limb of a `U256` as a field element,
+/// and errors if the integer is actually greater.
+pub(crate) fn u256_lowest_limb<F: Field>(u256: U256) -> Result<F, ProgramError> {
```
npwardberkeley, line 52: I feel like this function name seems to imply that it would take the lowest limb of any U256, not just U256s that entirely fit in one limb?

npwardberkeley: I think `u256_to_u32` or something similar would be clearer

evm/src/fixed_recursive_verifier.rs
```diff
@@ -1022,32 +1007,10 @@ where
         block_inputs
             .set_verifier_data_target(&self.block.cyclic_vk, &self.block.circuit.verifier_only);
 
-        set_block_hashes_target(
```
npwardberkeley, line 1025: Are these changes supposed to be part of this PR? I'm confused how these many function calls being replaced by one call to `set_public_value_targets`, when there aren't any changes to `set_public_value_targets` itself

evm/src/util.rs
Nashtare: It's true the correct naming is confusing, I'll address that. Same for the u64 version.

evm/src/fixed_recursive_verifier.rs
Nashtare: Indeed that's unrelated, it's just that the public values method for setting target includes all these internally, so while addressing the return values to output a ’Result’ I also cleaned this up. I can revert and address it into a subsequent PR if you prefer!

npwardberkeley: Oh no I think that's fine, since it's a small change! I was just checking

npwardberkeley (approved):
Lgtm!

PR #1232, opened 2023-09-13
Transactions

4l0n50:
This PR updates the transaction trie with the processed transactions.

The transaction trie update is done as follows:

- After reading and decoding each transaction from the prover inputs we copy the txn rlp and its size to the trie data segment. The pointer to this data is then used as payload for inserting a node with key RLP(txn_counter) in the transaction trie.

- For hashing the transaction trie we implemented the function encode_txn. This function simply copies the rlp encoded transaction to the address received as input


evm/src/cpu/kernel/tests/receipt.rs
```diff
@@ -119,7 +127,7 @@ fn test_receipt_encoding() -> Result<()> {
     // Get the expected RLP encoding.
     let expected_rlp = rlp::encode(&rlp::encode(&receipt_1));
 
-    let initial_stack = vec![retdest, 0.into(), 0.into()];
+    let initial_stack: Vec<U256> = vec![retdest, 0.into(), 0.into()];
```
Nashtare, line 130: Those changes seem unnecessary, but they silenced the clippy warning of `needless vec` (which is actually needed because of the `Interpreter::new_with_kernel` expected parameter types).

evm/src/cpu/kernel/asm/transactions/router.asm
```diff
@@ -34,3 +37,30 @@ read_txn_from_memory:
     // At this point, since it's not a type 1 or 2 transaction,
     // it must be a legacy (aka type 0) transaction.
     %jump(process_type_0_txn)
+
+global update_txn_trie:
+    // stack: txn_rlp_len, txn_counter, num_nibbles, retdest
+    // We need to copy the transaction rlp since otherwise
+    // it will be eventually overwritten when hashing the
+    // state trie. We will write in the trie data segment.
```
wborgeaud, line 45: I'm a bit confused by this comment. Isn't copying to `SEGMENT_TRIE_DATA` necessary in any case since `mpt_insert` expects data there?

evm/src/cpu/kernel/tests/mpt/load.rs
```diff
@@ -237,3 +242,41 @@ fn load_all_mpts_ext_to_leaf() -> Result<()> {
...
+
+    let mut expected_trie_data = vec![
+        0.into(),
+        U256::from(PartialTrieType::Leaf as u32),
+        2.into(),
+        128.into(), // Nibble
+        5.into(),   // value_ptr
+        txn.len().into(),
+    ];
+    expected_trie_data.extend::<Vec<U256>>(txn.into_iter().map(|b| b.into()).collect());
```
wborgeaud, line 276: nit: Can also do `expected_trie_data.extend(txn.into_iter().map(|b| U256:.from(b)))` to avoid the collect.

evm/src/generation/mpt.rs
```diff
@@ -114,7 +114,9 @@ pub(crate) fn all_mpt_prover_inputs(trie_inputs: &TrieInputs) -> Vec<U256> {
     );
 
     mpt_prover_inputs(&trie_inputs.transactions_trie, &mut prover_inputs, &|rlp| {
-        rlp::decode_list(rlp)
+        let mut parsed_txn = vec![U256::from(rlp.len())];
+        parsed_txn.extend::<Vec<U256>>(rlp.iter().map(|&b| b.into()).collect());
```
wborgeaud, line 118: Same as above, can avoid the collect.

wborgeaud (approved):
Looks good to me!

Nashtare:
@wborgeaud Sorry for the force-push, the history got completely messed up during a rebase, and induced an error in one of the tests which was easier to just revert to the state of your review (commit [fa63f90](https://github.com/mir-protocol/plonky2/pull/1232/commits/fa63f9009380ff4470466c90d9487193a3c3c036)), and then do a clean merge followed by applying your comments.
Let me know if it still looks good for you!

wborgeaud:
No worries, still looks good 👍

PR #1233, opened 2023-09-13
Remove redundant Keccak sponge cols

Nashtare:
The current trace layout for `KeccakSpongeStark` contains the `updated_state` in u32s, as well as the `final_digest` in bytes. Those correspond to the first 8 limbs of the `updated_state` and can therefore be used to recompute the 32-bit limbs in the CTLs.

### DIFF
--- a/evm/src/keccak_sponge/columns.rs
+++ b/evm/src/keccak_sponge/columns.rs
@@ -5,10 +5,13 @@ use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
 
 pub(crate) const KECCAK_WIDTH_BYTES: usize = 200;
 pub(crate) const KECCAK_WIDTH_U32S: usize = KECCAK_WIDTH_BYTES / 4;
+pub(crate) const KECCAK_WIDTH_MINUS_DIGEST_U32S: usize =
+    (KECCAK_WIDTH_BYTES - KECCAK_DIGEST_BYTES) / 4;
 pub(crate) const KECCAK_RATE_BYTES: usize = 136;
 pub(crate) const KECCAK_RATE_U32S: usize = KECCAK_RATE_BYTES / 4;
 pub(crate) const KECCAK_CAPACITY_BYTES: usize = 64;
 pub(crate) const KECCAK_CAPACITY_U32S: usize = KECCAK_CAPACITY_BYTES / 4;
+pub(crate) const KECCAK_DIGEST_U32S: usize = 8;
 pub(crate) const KECCAK_DIGEST_BYTES: usize = 32;
 
 #[repr(C)]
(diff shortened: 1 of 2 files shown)


evm/src/keccak_sponge/columns.rs
```diff
@@ -5,10 +5,13 @@ use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
 
 pub(crate) const KECCAK_WIDTH_BYTES: usize = 200;
 pub(crate) const KECCAK_WIDTH_U32S: usize = KECCAK_WIDTH_BYTES / 4;
+pub(crate) const KECCAK_WIDTH_MINUS_DIGEST_U32S: usize =
+    (KECCAK_WIDTH_BYTES - KECCAK_DIGEST_BYTES) / 4;
 pub(crate) const KECCAK_RATE_BYTES: usize = 136;
 pub(crate) const KECCAK_RATE_U32S: usize = KECCAK_RATE_BYTES / 4;
 pub(crate) const KECCAK_CAPACITY_BYTES: usize = 64;
 pub(crate) const KECCAK_CAPACITY_U32S: usize = KECCAK_CAPACITY_BYTES / 4;
+pub(crate) const KECCAK_DIGEST_U32S: usize = 8;
```
npwardberkeley, line 14: nit: should this be calculated as `KECCAK_DIGEST_BYTES / 4`?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -49,15 +49,34 @@ pub(crate) fn ctl_looked_data<F: Field>() -> Vec<Column<F>> {
...
+        .map(|c| {
+            Column::linear_combination(
+                c.iter()
+                    .enumerate()
+                    .map(|(i, &b)| (b, F::from_canonical_usize(1 << (8 * i)))),
+            )
+        })
+        .collect::<Vec<_>>();
+
+    res.extend(digest_u32s);
```
npwardberkeley, line 75: For both of these cases, I think we can call `extend` without `collect`ing first?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -357,23 +390,28 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
...
+        row.partial_updated_state_u32s.copy_from_slice(
+            &sponge_state[KECCAK_DIGEST_U32S..]
+                .iter()
+                .copied()
+                .map(|i| F::from_canonical_u32(i))
+                .collect::<Vec<_>>(),
+        );
+        for (l, &elt) in sponge_state[..KECCAK_DIGEST_U32S].iter().enumerate() {
+            let mut cur_elt = elt;
+            (0..4).for_each(|i| {
```
npwardberkeley, line 404: nit: why a `for_each` and then a `for`?

npwardberkeley (approved):

evm/src/keccak_sponge/keccak_sponge_stark.rs
Nashtare: That's a good question 😄

PR #1234, opened 2023-09-13
Change padding rule for CPU

Nashtare:
I assumed it would be simpler to just open a PR for it, so that you can share your comments directly on the code @nbgl. Feel free to override the PR if you have a better/simpler/more efficient approach!


evm/src/cpu/control_flow.rs
```diff
@@ -66,20 +61,30 @@ pub fn eval_packed_generic<P: PackedField>(
...
+    yield_constr.constraint_transition(lv.null * (nv.null - P::ONES));
+
     let is_cpu_cycle: P = COL_MAP.op.iter().map(|&col_i| lv[col_i]).sum();
     let is_cpu_cycle_next: P = COL_MAP.op.iter().map(|&col_i| nv[col_i]).sum();
-    // Once we start executing instructions, then we continue until the end of the table.
-    yield_constr.constraint_transition(is_cpu_cycle * (is_cpu_cycle_next - P::ONES));
+    // Once we start executing instructions, then we continue until the end of the table
+    // or we reach dummy padding rows.
+    yield_constr
+        .constraint_transition(is_cpu_cycle * (is_cpu_cycle_next - P::ONES) * (nv.null - P::ONES));
```
nbgl, line 74: Nit: I would write it as 
```suggestion
    yield_constr
        .constraint_transition(is_cpu_cycle * (is_cpu_cycle_next + nv.null - P::ONES));
```

evm/src/cpu/control_flow.rs
```diff
@@ -66,20 +61,30 @@ pub fn eval_packed_generic<P: PackedField>(
...
     // counter. Also, the next row has the same kernel flag.
     let is_native_instruction: P = NATIVE_INSTRUCTIONS.iter().map(|&col_i| lv[col_i]).sum();
     yield_constr.constraint_transition(
-        is_native_instruction * (lv.program_counter - nv.program_counter + P::ONES),
+        (nv.null - P::ONES)
+            * is_native_instruction
+            * (lv.program_counter - nv.program_counter + P::ONES),
+    );
+    yield_constr.constraint_transition(
+        (nv.null - P::ONES) * is_native_instruction * (lv.is_kernel_mode - nv.is_kernel_mode),
```
nbgl, line 86: I would not change this. That way we can verify that by the time we get to the null rows, execution actually _is_ done. In other words, we need to make sure that the prover can't just start inserting null rows before the entire program has correctly finished.

evm/src/cpu/control_flow.rs
```diff
@@ -91,13 +96,11 @@ pub fn eval_packed_generic<P: PackedField>(
...
 
-    // The last row must be a CPU cycle row.
-    yield_constr.constraint_last_row(is_cpu_cycle - P::ONES);
-    // Also, the last row's `program_counter` must be inside the `halt` infinite loop. Note that
-    // that loop consists of two instructions, so we must check for `halt` and `halt_inner` labels.
-    let (halt_pc0, halt_pc1) = get_halt_pcs::<P::Scalar>();
-    yield_constr
-        .constraint_last_row((lv.program_counter - halt_pc0) * (lv.program_counter - halt_pc1));
+    // The last row must be a CPU cycle row or a dummy padding row.
+    yield_constr.constraint_last_row((is_cpu_cycle - P::ONES) * (lv.null - P::ONES));
```
nbgl, line 100: For simplicity I'd just require that the last row is a padding row:
```suggestion
    yield_constr.constraint_last_row(lv.null - P::ONES);
```

evm/src/cpu/control_flow.rs
```diff
@@ -91,13 +96,11 @@ pub fn eval_packed_generic<P: PackedField>(
...
+    // Also, the last row's `program_counter` must be inside the `halt` infinite loop.
+    let halt_pc = get_halt_pc::<P::Scalar>();
+    yield_constr.constraint_last_row(lv.program_counter - halt_pc);
```
nbgl, line 103: This doesn't make sense anymore. I would make sure that if lv.null is set, then program counter is at the `halt` instructions.
```suggestion
    // Also, a padding row's `program_counter` must be at the `halt` label.
    // In particular, it ensures that the first padding row may only be added
    // after we jumped to the `halt` function. Subsequent padding rows may set
    // the `program_counter` to arbitrary values (there's no transition
    // constraints) so we can place this requirement on them too.
    let halt_pc = get_halt_pc::<P::Scalar>();
    yield_constr.constraint(lv.null * (lv.program_counter - halt_pc));
```

evm/src/cpu/kernel/asm/halt.asm
```diff
@@ -1,6 +1 @@
 global halt:
```
nbgl, line 1: We must put a PANIC instruction below `halt` to ensure that the prover cannot choose to _not_ halt at this address (in other words, it could execute whatever comes after `halt` in memory).

```suggestion
global halt:
    PANIC
```

nbgl (commented):
This approach looks good—it's basically the same as mine. A few details can be simplified; I've noted those in comments.

Two missing constraints:
1. Make sure that `lv.null` is mutually exclusive with the op flags.
2. We must ensure that the prover can only start emitting `null` rows _after_ we've jumped to `halt`. The prover should not be able to stop execution midway through.

evm/src/cpu/stack.rs
```diff
@@ -206,6 +206,13 @@ pub fn eval_packed<P: PackedField>(
             eval_packed_one(lv, op, stack_behavior, yield_constr);
         }
     }
+
+    // Disable memory channels for padding rows
+    let is_null = lv.null;
+    for i in 0..NUM_GP_CHANNELS {
+        let channel = lv.mem_channels[i];
+        yield_constr.constraint(is_null * channel.used);
+    }
```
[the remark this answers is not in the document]
Nashtare: I had put it there because the constraints on the unused memory channels are there, but it's true that's not ideal. I'll move it.

evm/src/cpu/control_flow.rs
Nashtare: Well _technically_ it doesn't have to be, but I agree that we can ignore this possibility 😄

evm/src/cpu/kernel/asm/halt.asm
Nashtare: Oh right! good catch

evm/src/cpu/control_flow.rs
Nashtare: The first constraint would fail if unchanged though, because at the last actual CPU row, the PC won't be increased (as kept at `halt_pc` for all subsequent padding rows).

Nashtare: Actually nevermind! As with `PANIC` below the `halt` label we need to adapt slightly the padding to directly write dummy rows when reaching the `halt_pc` the former constraint becomes valid again!

Nashtare:
Thanks for the review @nbgl! I've applied your comments. I just have a few questions regarding your main comment:

> Make sure that lv.null is mutually exclusive with the op flags.

That was indeed an overlook on my side. I've added this constraint implicitely through constraining the first row to NOT be a padding row. On `bootstrap_kernel` side, we are enforcing then that the op flags must be zero on the first row.
We then have 
`yield_constr
        .constraint_transition(is_cpu_cycle * (is_cpu_cycle_next + nv.halt_state - P::ONES));` 
 which guarantees us that subsequent rows never have both op flags and the halt flag activated.
 
Regarding your comment on adding `PANIC` under the `halt` label, this forces to handle the padding slightly differently (i.e. we don't execute the last instruction (as it would hence panic)) and instead directly process dummy rows. Is that correct?

nbgl:
> Regarding your comment on adding `PANIC` under the `halt` label, this forces to handle the padding slightly differently (i.e. we don't execute the last instruction (as it would hence panic)) and instead directly process dummy rows. Is that correct?

Exactly. There's no point continuing to process instructions once we've jumped to `halt`. It's simpler not to give the prover the choice.

Nashtare:
great! Let me know if there is anything else bugging you.
Regarding my previous message, I'll add a comment in the code explaining how the constraints relate between modules for clarity.

Nashtare:
The CI error is unrelated and happened on other branches as well, opening a PR to fix clippy.

evm/src/cpu/bootstrap_kernel.rs
```diff
@@ -64,8 +64,10 @@ pub(crate) fn eval_bootstrap_kernel<F: Field, P: PackedField<Scalar = F>>(
     let next_values: &CpuColumnsView<_> = vars.next_values.borrow();
 
     // IS_BOOTSTRAP_KERNEL must have an init value of 1, a final value of 0, and a delta in {0, -1}.
-    let local_is_bootstrap = P::ONES - local_values.op.into_iter().sum::<P>();
-    let next_is_bootstrap = P::ONES - next_values.op.into_iter().sum::<P>();
+    let local_is_bootstrap =
+        P::ONES - local_values.op.into_iter().sum::<P>() - local_values.halt_state;
+    let next_is_bootstrap =
+        P::ONES - next_values.op.into_iter().sum::<P>() - next_values.halt_state;
```
nbgl, line 70: Organization: I think it would be neater to have `halt_state` be the special, implicit flag (that's set when all other flags are 0). Intuitively, `is_bootstrap` does _something_ whereas `halt_state` does nothing. Not a dealbreaker though.

evm/src/cpu/control_flow.rs
```diff
@@ -55,10 +51,23 @@ pub fn eval_packed_generic<P: PackedField>(
     nv: &CpuColumnsView<P>,
     yield_constr: &mut ConstraintConsumer<P>,
 ) {
+    // The null special instruction must be boolean.
+    yield_constr.constraint(lv.halt_state * (lv.halt_state - P::ONES));
+    // The first row cannot be a null row. In parallel, the `bootstrap_kernel` module
+    // enforces that the first row is a kernel bootstrapping row, implicitely enforcing
+    // that the operation flags are also turned off on that row.
+    yield_constr.constraint_first_row(lv.halt_state);
```
nbgl, line 59: Reduntant constraint: Since this is already enforced in `bootstrap_kernel`, I would delete it here. There's no need for duplication.

Also, in some sense, it's okay for the first row to be `halt_state` (I think). It just means that the proof does nothing whatsoever.

evm/src/cpu/control_flow.rs
```diff
@@ -80,13 +89,23 @@ pub fn eval_packed_generic<P: PackedField>(
...
+    yield_constr.constraint_last_row(lv.halt_state - P::ONES);
+
+    // Also, a padding row's `program_counter` must be at the `halt` label.
+    // In particular, it ensures that the first padding row may only be added
+    // after we jumped to the `halt` function. Subsequent padding rows may set
+    // the `program_counter` to arbitrary values (there's no transition
+    // constraints) so we can place this requirement on them too.
+    let halt_pc = get_halt_pc::<P::Scalar>();
+    yield_constr.constraint(lv.halt_state * (lv.program_counter - halt_pc));
+
```
nbgl, line 108: Organization: I still think this is a kinda awkward place to put these constraints. I would probably start a new `halt.rs` file. (Also not a dealbreaker.)

evm/src/cpu/control_flow.rs
```diff
@@ -80,13 +89,23 @@ pub fn eval_packed_generic<P: PackedField>(
...
     // Finally, the last row must be in kernel mode.
     yield_constr.constraint_last_row(lv.is_kernel_mode - P::ONES);
```
nbgl, line 110: Redundant constraint: `halt_state` actually makes this redundant, so it should be deleted.

evm/src/generation/mod.rs
```diff
@@ -278,26 +279,34 @@ pub fn generate_traces<F: RichField + Extendable<D>, const D: usize>(
...
+            row.program_counter = F::from_canonical_usize(pc);
+            row.is_kernel_mode = F::ONE;
+            row.gas = F::from_canonical_u64(state.registers.gas_used);
+            row.stack_len = F::from_canonical_usize(state.registers.stack_len);
+            row.halt_state = F::ONE;
+
+            while !state.traces.clock().is_power_of_two() {
+                state.traces.push_cpu(row);
+                row.clock += F::ONE;
+            }
```
nbgl, line 304: Bug: We've have a constraint that the last row must be a `halt_state` row. That means that we must push at least one padding row.

The particular edge case where that matters is if the trace is a power of 2 _without_ padding rows. We can't have zero padding rows, so we must pad with another _trace length_ rows.

```suggestion
            loop {
                state.traces.push_cpu(row);
                row.clock += F::ONE;
                if state.traces.clock().is_power_of_two() {
                    break;
                }
            }
```

nbgl (commented):
Looks good!!

A few comments:
- One small, subtle bug,
- Two redundant constraints,
- Two organization notes (not very important, consider them optional/it's your call to make).

evm/src/cpu/control_flow.rs
Nashtare: Actually you're right, I had somehow misread the way `bootstrap_kernel` was constrained on the first row..

Nashtare:
Apologies for the back and forth. Now I think all your comments have been addressed.

nbgl (approved):
Thank you!! Looks great :)

PR #1236, opened 2023-09-15
Update clippy in CI

Nashtare:
Clippy seems to have a weird bug, known upstream, for which a temporary fix seems to be deactivating incremental compilation. It happened on all the recent branches that have been pushed, probably triggered by the nightly update overnight.

cc @BGluth as you could encounter it in what you're doing if the toolchain isn't fixed.


npwardberkeley:
Looks good, thanks!

PR #1237, opened 2023-09-15
Lookup next row

npwardberkeley:


hratoanina:
Hi! Looks like you also decided to tackle this.
There's a pending PR for that: https://github.com/mir-protocol/plonky2/pull/1229

We're mostly taking the same approach, with, from what I see, two differences:
- You can access the next row at the `Column` level, while we declare a new set of `Column`s for the next row, separate from the local row's `Column`s. Your approach is much more generic and looks better.
- You had to add a commitment for the previous row for each evaluation point. There's a way to avoid that by turning the CTL Z polynomial upside down: the first row contains the full product, the last row contains the initializing value, and the transition constraint now only involves the local row (and next row, with this PR) evaluations.

npwardberkeley:
Ohh nice thank you! Wow I missed your PR. Yes that looks exactly right!

Since your PR is much more complete, do you want to just add in my `Column` approach, and then I can review it?

hratoanina:
> Ohh nice thank you! Wow I missed your PR. Yes that looks exactly right!
> 
> Since your PR is much more complete, do you want to just add in my `Column` approach, and then I can review it?

Sure, I'll do that!

PR #1241, opened 2023-09-19
Remove SEQUENCE_LEN in BytePackingStark

Nashtare:
This PR removes the `SEQUENCE_LEN` column from the `BytePackingStark` module, reusing the `get_active_position()` helper method to retrieve the sequence length (i.e. final position) when `SEQUENCE_END` is activated.

credits to @hratoanina for noticing this!


unzvfu (approved):
Looks good!

PR #1246, opened 2023-09-20
Remove `generic_const_exprs` feature from EVM crate

Nashtare:
This PR removes the unstable `generic_const_exprs` feature from the EVM crate. This alleviates both the risk of weird and unpredictable behavior (cf. what happened on #1235 for instance) induced by the high instability of the feature, and the growing compilation time incurred by type-checker getting stuck. This is notably visible when trying to include an eight STARK module, with compilation time completely blowing-up (7min55 vs 1min45 with 7 modules, i.e. +350% increase, in debug mode).

The approach somewhat follows the one taken in #1024.

It yields a compilation time improvement of around 20% on my machine, for no noticeable performance downgrade (proving time varies between -3% and +2% across several runs between latest `main` and this branch), although I'd like to extensively ensure this across different test cases and architectures.

In summary, this adds two associated types to the `Stark` trait, i.e. `EvaluationFrame` and `EvaluationFrameTarget` which allow to access the local view of a frame (i.e. local and next rows), replacing the previous `StarkEvaluationVars` and `StarkEvaluationTargets` structs.

*Side Note: I haven't dealt with `starky` crate at all, although the same approach could be applied, but I wanted to get validation on the approach before generalizing.*


evm/src/stark.rs
```diff
@@ -22,7 +22,16 @@ const QUOTIENT_ORACLE_INDEX: usize = 2;
 /// Represents a STARK system.
 pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
     /// The total number of columns in the trace.
-    const COLUMNS: usize;
+    const COLUMNS: usize = Self::EvaluationFrameTarget::COLUMNS;
```
Nashtare, line 25: Note that keeping this const is not _technically_ needed, but I felt it was feeling cleaner to use it for the `fri` methods of the trait, rather than rely on `Self::EvaluationFrameTarget::COLUMNS` directly there.

Nashtare:
I also tested it on ARM targets, both light and beefy machines, and didn't get any noticeable difference in proving time ($|\delta| < 0.8$%).

evm/src/stark.rs
Nashtare: We could also leave it as it was so that it's implementer-specified (with some comment emphasizing that both associated consts must be consistent within `Stark`, `Stark::EvaluationFrame` and `Stark::EvaluationFrameTarget`.

evm/src/lib.rs
```diff
@@ -4,7 +4,6 @@
 #![allow(clippy::type_complexity)]
 #![allow(clippy::field_reassign_with_default)]
 #![feature(let_chains)]
-#![feature(generic_const_exprs)]
```
unzvfu, line 7: :partying_face:

unzvfu (approved):
This is great; solves the problem with a minimum of fuss! Thanks heaps!

It would be nice to have a solution that avoids all those `.try_into().unwrap()` type conversions (which is a code smell imho), but I can't think of a way to do that without more extensive structural changes, and those type conversions are pretty harmless here anyway.

Nashtare:
@unzvfu I'm terribly sorry, I thought my commit addressing your comment had been pushed, will open a follow up PR for cleanup.

Issue #1247, opened 2023-09-21
Panic handling


Nashtare:
There are several places where the prover can currently panic on invalid inputs, either through `unwrap()` / `expect()` or specific methods like `U256::as_u32()` which implicitely perform an unwrapping.

Ideally we'd like to have all those removed by returning an `Error` instead, though this will imply some important refactoring in several pieces of the evm crate. Is there any preferred approach for tackling this? 

@npwardberkeley: for context, I've been trying to remove also the unchecked calls to `U256::as_usize()` which are not part of #1231 that you reviewed, which made me realize that some modules like the `prover_input` one are filled with panic risks.

dvdplm:
I would suggest adopting a hard but simple rule on panics: not allowed outside tests. A total ban on `unwrap()`s. All `expect()` calls should carry a full explanation as to why the panic will never happen. Example:

```rust
let a = Some(true);
// moar code
let b = a.expect("Initialized as Some above and never touched after; qed");
```

Code processing user input and/or data off the network should never assume the data is valid.

Nashtare:
Handled by #1231 and #1250

PR #1250, opened 2023-09-22
Handle additional panics

Nashtare:
This PR partly addresses #1247. More specifically, it handles (on the EVM crate):

- all risky `as_usize()` calls
- all risky `expect()` calls
- `U256` methods with implicit internal panic (`from_big_endian` / `from_big_endian` / `div_mod`).

I haven't changed (unless required from cascade changes) the kernel / tests + interpreter files as those don't matter as much (panic in testing is fine, and the kernel risks of panic would arise during the offline preprocessing phase which would be caught ahead).

### DIFF
--- a/evm/src/generation/mod.rs
+++ b/evm/src/generation/mod.rs
@@ -218,7 +218,7 @@ pub fn generate_traces<F: RichField + Extendable<D>, const D: usize>(
     GenerationOutputs,
 )> {
     let mut state = GenerationState::<F>::new(inputs.clone(), &KERNEL.code)
-        .map_err(|_| anyhow!("Failed to parse all the initial prover inputs"))?;
+        .map_err(|err| anyhow!("Failed to parse all the initial prover inputs: {:?}", err))?;
 
     apply_metadata_and_tries_memops(&mut state, &inputs);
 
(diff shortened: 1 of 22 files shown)


evm/src/generation/mod.rs
```diff
@@ -216,7 +217,8 @@ pub fn generate_traces<F: RichField + Extendable<D>, const D: usize>(
     PublicValues,
     GenerationOutputs,
 )> {
-    let mut state = GenerationState::<F>::new(inputs.clone(), &KERNEL.code);
+    let mut state = GenerationState::<F>::new(inputs.clone(), &KERNEL.code)
+        .map_err(|_| anyhow!("Failed to parse all the initial prover inputs"))?;
```
wborgeaud, line 221: nit: Could also do
```rust
.map_err(|err| anyhow!("Failed to parse all the initial prover inputs: {:?}", err))?;
```
to get more info on which ProgramError was thrown. Also applies elsewhere in this PR.

evm/src/util.rs
```diff
@@ -70,6 +70,15 @@ pub(crate) fn u256_to_u64<F: Field>(u256: U256) -> Result<(F, F), ProgramError>
     ))
 }
 
+/// Safe alternative to `U256::as_usize()`, which errors in case of overflow instead of panicking.
+pub(crate) fn u256_to_usize(u256: U256) -> Result<usize, ProgramError> {
+    if TryInto::<usize>::try_into(u256).is_err() {
+        return Err(ProgramError::IntegerTooLarge);
+    }
+
+    Ok(u256.as_usize())
```
wborgeaud, line 79: nit
```suggestion
    u256.try_into().map_err(|_| ProgramError::IntegerTooLarge)
```

wborgeaud (approved):
Looks good to me!

PR #1251, opened 2023-09-25
Fix observe_block_metadata

LindaGuiga:
We noticed the recursive proofs were failing because the challenges were observed in the a different order in the recursive and non-recursive versions. This PR fixes the issue.


Nashtare (approved):
We definitely need a non-dummy aggregation test running in the CI 😅

PR #1252, opened 2023-09-25
Combine mload_32bytes and mstore_32bytes flags

LindaGuiga:
This PR combines the two flags: `mload_32bytes` and `mstore_32bytes`, as well as the two associated CTLs.
This required changing the order of the stack for `MSTORE_32BYTES`.


nbgl (approved):
Looks good, just a few nits. Thank you :)

evm/src/cpu/cpu_stark.rs
```diff
@@ -125,38 +125,12 @@ pub fn ctl_arithmetic_shift_rows<F: Field>() -> TableWithColumns<F> {
     TableWithColumns::new(Table::Cpu, columns, Some(Column::single(COL_MAP.op.shift)))
 }
 
-pub fn ctl_data_byte_packing<F: Field>() -> Vec<Column<F>> {
+pub fn ctl_data_byte32<F: Field>() -> Vec<Column<F>> {
     ctl_data_keccak_sponge()
 }
```
Nashtare, line 130: nit: as we're re-using `ctl_data_keccak_sponge` both for `MLOAD_32BYTES` and `MSTORE_32BYTES`, perhaps a tiny comment explaining the distinction may be useful? (as the doc there is valid for MLOAD but the last memory channel is being read in the case of MSTORE, not pushed to).

LindaGuiga:
Given all the changes introduced recently, it seems impossible (for now at least) to merge the two flags given the current state of things. Indeed, removing one memory channel implied changing the CTL for `mstore_32bytes`, and makes it different from `mload_32bytes` (in `mstore_32bytes`, we set the length in the CTL by computing next_value - current_offset, and we can't do that in `mload_32bytes` since we just want to check that the loaded value is correct). I am therefore closing this PR.

PR #1256, opened 2023-09-26
Combine PUSH0 and PC flags.

LindaGuiga:
This PR combines the PUSH0 and GETPC flags. Even though the two opcodes are very different, this combination allows us to reduce the number of columns by 1 in the Cpu STARK.


nbgl (approved):
Looks good!! thank you :)

PR #1258, opened 2023-09-27
Optimize lookup builder

nicholas-mainardi:
This PR optimizes the instantiation of `LookupGate` and `LookupTableGate` instances when building a circuit. The optimization seems to be particularly effective on big lookup tables (e.g., with 2^16 elements).
To this extent, I have added a couple of tests in `lookup_test.rs` that perform lookups over a big lookup table. The speed up achieved by the optimization can be observed by running such tests before and after applying the commit introducing the optimization (that is, the "optimize lookup builder" commit with identifier 8c842a2048a0da933cd313c57ff671911438c9c6). 

This optimization might allow to significantly reduce the circuit building time when doing larger range checks or using hash functions with larger lookup tables. For example, while using 8-bit lookups by default, Monolith can also be used with 16-bit lookup tables in order to reduce the number of gates that need to be instantiated for each permutation.

### DIFF
--- a/plonky2/src/gadgets/lookup.rs
+++ b/plonky2/src/gadgets/lookup.rs
@@ -128,14 +128,14 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
                 let last_lut_gate = self.num_gates();
                 let num_lut_entries = LookupTableGate::num_slots(&self.config);
                 let num_lut_rows = (self.get_luts_idx_length(lut_index) - 1) / num_lut_entries + 1;
+                let gate =
+                    LookupTableGate::new_from_table(&self.config, lut.clone(), last_lut_gate);
                 // Also for `LookupTableGate` we replace the `num_lut_entries` calls to `find_slot`
                 // with a single call to `add_gate`; note that in this case there is no need to
                 // separately handle the last chunk of LUT entries that cannot fill all the slots of
                 // a `LookupTableGate`, as the generator already handles empty slots
                 for _ in 0..num_lut_rows {
-                    let gate =
-                        LookupTableGate::new_from_table(&self.config, lut.clone(), last_lut_gate);
-                    self.add_gate(gate, vec![]);
+                    self.add_gate(gate.clone(), vec![]);
                 }
 
                 let first_lut_gate = self.num_gates() - 1;
--- a/plonky2/src/gadgets/lookup.rs
+++ b/plonky2/src/gadgets/lookup.rs
@@ -89,25 +89,53 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
 
                 let lookups = self.get_lut_lookups(lut_index).to_owned();
 
-                for (looking_in, looking_out) in lookups {
-                    let gate = LookupGate::new_from_table(&self.config, lut.clone());
+                let gate = LookupGate::new_from_table(&self.config, lut.clone());
+                let num_slots = LookupGate::num_slots(&self.config);
+
+                /* Given the number of lookups and the number of slots for each gate, it is possible
+                 to compute the number of gates that will employ all their slots; for such gates,
+                 the `num_slots` calls to the `find_slot` function can be replaced with a single
+                 `add_gate` call.
+                */
+                // lookup_iter will iterate over the lookups that can be placed in fully utilized
+                // gates, splitting them in chunks that can be placed in the same `LookupGate`
+                let lookup_iter = lookups.chunks_exact(num_slots);
+                // `last_chunk` will contain the remainder of lookups, which cannot fill all the
+                // slots of a `LookupGate`; this last chunk will be processed by incrementally
+                // filling slots, to avoid that the `LookupGenerator` is run on unused slots
+                let last_chunk = lookup_iter.remainder();
+                // handle chunks that can fill all the slots of a `LookupGate`
+                lookup_iter.for_each(|chunk| {
+                    let row = self.add_gate(gate.clone(), vec![]);
+                    for (i,(looking_in, looking_out)) in chunk.iter().enumerate() {
+                        let gate_in = Target::wire(row, LookupGate::wire_ith_looking_inp(i));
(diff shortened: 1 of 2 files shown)


Nashtare:
How do this compare with https://github.com/mir-protocol/plonky2/pull/1221?

plonky2/src/gadgets/lookup.rs
```diff
@@ -89,25 +89,53 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
-                for (looking_in, looking_out) in lookups {
-                    let gate = LookupGate::new_from_table(&self.config, lut.clone());
+                let gate = LookupGate::new_from_table(&self.config, lut.clone());
+                let num_slots = LookupGate::num_slots(&self.config);
+
+                /* Given the number of lookups and the number of slots for each gate, it is possible
+                 to compute the number of gates that will employ all their slots; for such gates,
+                 the `num_slots` calls to the `find_slot` function can be replaced with a single
+                 `add_gate` call.
+                */
```
wborgeaud, line 99: nit: We don't really use `/* */` for comments elsewhere and usually use `//` instead.

plonky2/src/gadgets/lookup.rs
```diff
@@ -89,25 +89,53 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+                    let row = self.add_gate(gate.clone(), vec![]);
+                    for (i, (looking_in, looking_out)) in chunk.iter().enumerate() {
+                        let gate_in = Target::wire(row, LookupGate::wire_ith_looking_inp(i));
+                        let gate_out = Target::wire(row, LookupGate::wire_ith_looking_out(i));
+                        self.connect(gate_in, *looking_in);
+                        self.connect(gate_out, *looking_out);
+                    }
+                });
+                // deal with the last chunk
+                for (_, (looking_in, looking_out)) in last_chunk.iter().enumerate() {
```
wborgeaud, line 118: The `enumerate` isn't needed, right?

plonky2/src/gadgets/lookup.rs
```diff
@@ -89,25 +89,53 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
                 // Create LUT gates. Nothing is connected to them.
                 let last_lut_gate = self.num_gates();
                 let num_lut_entries = LookupTableGate::num_slots(&self.config);
                 let num_lut_rows = (self.get_luts_idx_length(lut_index) - 1) / num_lut_entries + 1;
-                let num_lut_cells = num_lut_entries * num_lut_rows;
-                for _ in 0..num_lut_cells {
+                // Also for `LookupTableGate` we replace the `num_lut_entries` calls to `find_slot`
+                // with a single call to `add_gate`; note that in this case there is no need to
+                // separately handle the last chunk of LUT entries that cannot fill all the slots of
+                // a `LookupTableGate`, as the generator already handles empty slots
```
wborgeaud, line 134: nit: This comment makes sense in the context of this PR, but not in general.

wborgeaud (commented):
I would also be interested in a performance comparison with #1221. 

On paper I would have a slight preference for the approach taken here. Conceptually it's the same, but in #1221 the lookup gates aren't really gates anymore and are replaced with noop gates which aren't technically doing a noop. What do you think @Nashtare @LindaGuiga ?

nicholas-mainardi:
Thanks for the quick response. I haven't seen earlier #1221, we employ a similar logic to remove the unnecessary calls to `find_slot`, but for this PR we didn't delve too much in the lookup gates logic, so I think that the main difference is that we didn't replace them with Noop gates. 
In terms of build time, I have run the 2 tests with big lookup tables that I added in this PR also on #1221, and honestly the latter seems to be definitely faster, likely due to the replacement of lookup gates with Noop ones. 
This is the comparison on my Ubuntu machine, equipped with 4 cores 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz  (8 threads) and 32 GB RAM:

- `test_big_lut`: 0.47s build time for #1221, 5.9s for this PR
- `test_many_lookups_on_big_lut`: 0.99s build time for #1221, 6.47s for this PR

The performance gap is definitely not negligible, so performance-wise I would understand if you prefer to integrate #1221 instead; otherwise, if you prefer the approach taken in this PR as it might be less contrived, I will be happy to proceed with the PR and address comments. I will wait for your feedback on how to proceed, thanks

LindaGuiga:
@nicholas-mainardi  @wborgeaud I agree that this approach is less contrived! But before opening #1221, I had also tried a similar approach (keeping the lookup gates), and for 500,000 lookups with LUT size 2^16, the gap was even larger for me in that case: ~6s with the `Noop` gates VS 69s with the lookup gates. I don't know if the gap is also this large for your implementation though, @nicholas-mainardi!

nicholas-mainardi:
Actually I noticed that I create a new instance of `LookupTableGate` in each iteration rather than cloning an existing one; by cloning, the performance between the 2 implementations become much closer:

- `test_big_lut`: 0.42s for #1221, 0.44s for this PR
- `test_many_lookups_on_big_lut` (65536 lookups): 0.91s for #1221, 0.76s for this PR
- `test_many_lookups_on_big_lut` (500k lookups): 2.09s for #1221, 1.6s for this PR

So after the last commit I think that now the 2 solutions should be quite comparable in terms of performance.

wborgeaud:
Oh that's great! I think we can go forward with this approach then.

plonky2/src/gadgets/lookup.rs
nicholas-mainardi: Yes, you are right, I will remove it

plonky2/src/gadgets/lookup.rs
wborgeaud: I just meant that the comment made sense when looking at the code you replaced, but not so much to an external reader not looking at the history of the file.
But the new comment looks good 👍

wborgeaud (approved):
Looks good, thanks for the PR!

PR #1259, opened 2023-09-27
Combine JUMPDEST and KECCAK_GENERAL flags.

LindaGuiga:
This combines the JUMPDEST and KECCAK_GENERAL flags. The two opcodes are unrelated, but it still enables us to gain one column on the Cpu STARK side.


evm/src/cpu/gas.rs
```diff
@@ -99,6 +98,15 @@ fn eval_packed_accumulate<P: PackedField>(
     let ternary_op_cost = P::Scalar::from_canonical_u32(G_MID.unwrap())
         - lv.opcode_bits[1] * P::Scalar::from_canonical_u32(G_MID.unwrap());
     yield_constr.constraint_transition(lv.op.ternary_op * (nv.gas - lv.gas - ternary_op_cost));
+
+    // For JUMPDEST and KECCAK_GENERAL.
+    // JUMPDEST is differentiated from KECCAK_GENERAL by its second bit set to 1.
+    let jumpdest_keccak_general_gas_cost = lv.opcode_bits[1]
+        * P::Scalar::from_canonical_u32(G_JUMPDEST.unwrap())
```
dvdplm, line 105: These `unwrap`s are not great. I know they are not new in this PR but it seems to me like using `Option` in `SIMPLE_OPCODES` has had some unfortunate consequences. Perhaps it should be refactored to use an enum so we get `OpsColumnsView<ManualOrCode>`, with `enum ManualOrCode { Manual, Code(u32) }`?
Alternatively, have two collections, one for manually managed opcodes and the other for simple opcodes and go through each separately?

Nashtare: They need to be wrapped in an `Option`, because we have a distinction here between `Some(0)` and `None`, though they are pretty harmless as we only `unwrap` hardcoded constants so it doesn't really bother me to rely on unwrap here.

dvdplm: To me having to resort to using `unwrap` is a code smell. Would having two hard-coded lists (manual vs "simple") work?

evm/src/cpu/decode.rs
```diff
@@ -147,6 +147,19 @@ pub fn eval_packed_generic<P: PackedField>(
...
+    // Manually check lv.op.jumpdest_keccak_general.
+    // KECCAK_GENERAL is a kernel-only instruction, but not JUMPDEST.
+    // JUMPDEST is differentiated from KECCAK_GENERAL by its second bit set to 1.
+    yield_constr.constraint(
+        (P::ONES - kernel_mode) * lv.op.jumpdest_keccak_general * (P::ONES - lv.opcode_bits[1]),
+    );
+
+    // Check the JUMPDEST and KERNEL_GENERAL opcodes.
+    let jumpdest_keccak_general_constr = (opcode - P::Scalar::from_canonical_usize(0x21_usize))
+        * (opcode - P::Scalar::from_canonical_usize(0x5b_usize))
```
Nashtare, line 160: nit: we don't need to specify `_usize` here, the compiler will infer the type.
Additionally for clarity, I think it could be worth defining the two opcodes similarly to how it's done below in the circuit version, i.e.:

```rust
    let jumpdest_opcode = P::Scalar::from_canonical_usize(0x5b_usize);
    let keccak_general_opcode = P::Scalar::from_canonical_usize(0x21_usize);
    let jumpdest_keccak_general_constr = (opcode - keccak_general_opcode)
        * (opcode - jumpdest_opcode) * lv.op.jumpdest_keccak_general;
    yield_constr.constraint(jumpdest_keccak_general_constr);
```

evm/src/cpu/stack.rs
```diff
@@ -53,6 +53,20 @@ pub(crate) const MLOAD_GENERAL_OP: Option<StackBehavior> = Some(StackBehavior {
...
+    disable_other_channels: true,
+});
+
+pub(crate) const JUMPDEST_OP: Option<StackBehavior> = Some(StackBehavior {
+    num_pops: 0,
+    pushes: false,
+    new_top_stack_channel: None,
+    disable_other_channels: true,
+});
+
```
Nashtare, line 69: We never exploit that those are wrapped inside an `Option`, as we unwrap them directly, so I'd just change their type to simply `StackBehavior`.
We could also apply this to the others, either in this PR or in a follow-up one.

Nashtare (approved):
LGTM!

PR #1261, opened 2023-09-28
Make gas fit in 2 limbs

Nashtare:
Following discussion in https://github.com/mir-protocol/evm-tests/issues/35, handle gas (limit/used) in two limbs.
The primary goal is for testing, as currently more than 7000 tests can't pass the check due to the recent zkEVM updates.

The additional CPU column is I think acceptable, given that we've already reduced the width by quite a bit (116 columns with this PR), and there are a few PRs opened or incoming which should bring it close to 100. In the worst case, we could always revert this PR once we get close to production-readiness?

If that's considered a too large overhead for what it brings, we could instead keep it up to date whenever there are consequent changes on `main`, and bind the test runner to this branch? Or we could go for the partial solution of just augmenting to 2 limbs gas limit.

### DIFF
--- a/evm/src/cpu/gas.rs
+++ b/evm/src/cpu/gas.rs
@@ -70,20 +70,22 @@ fn eval_packed_accumulate<P: PackedField>(
         })
         .sum();
 
-    let constr = nv.gas - (lv.gas + gas_used);
+    let gas_diff = nv.gas[1] * P::Scalar::from_canonical_u64(1 << 32) + nv.gas[0]
+        - (lv.gas[1] * P::Scalar::from_canonical_u64(1 << 32) + lv.gas[0]);
+    let constr = gas_diff - gas_used;
     yield_constr.constraint_transition(filter * constr);
 
     for (maybe_cost, op_flag) in izip!(SIMPLE_OPCODES.into_iter(), lv.op.into_iter()) {
         if let Some(cost) = maybe_cost {
             let cost = P::Scalar::from_canonical_u32(cost);
-            yield_constr.constraint_transition(op_flag * (nv.gas - lv.gas - cost));
+            yield_constr.constraint_transition(op_flag * (gas_diff - cost));
         }
     }
 
     // For jumps.
     let jump_gas_cost = P::Scalar::from_canonical_u32(G_MID.unwrap())
         + lv.opcode_bits[0] * P::Scalar::from_canonical_u32(G_HIGH.unwrap() - G_MID.unwrap());
-    yield_constr.constraint_transition(lv.op.jumps * (nv.gas - lv.gas - jump_gas_cost));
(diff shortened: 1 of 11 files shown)


evm/src/cpu/gas.rs
```diff
@@ -70,20 +70,22 @@ fn eval_packed_accumulate<P: PackedField>(
         })
         .sum();
 
-    let constr = nv.gas - (lv.gas + gas_used);
+    let gas_diff = nv.gas[1] * P::Scalar::from_canonical_u64(1 << 32) + nv.gas[0]
+        - (lv.gas[1] * P::Scalar::from_canonical_u64(1 << 32) + lv.gas[0]);
+    let constr = gas_diff - gas_used;
```
wborgeaud, line 75: Could overflowing the field size be an issue here?

evm/src/proof.rs
```diff
@@ -101,14 +101,15 @@ pub struct BlockMetadata {
     pub block_number: U256,
     /// The difficulty (before PoS transition) of this block.
     pub block_difficulty: U256,
+    /// The `mix_hash` value of this block.
     pub block_random: H256,
-    /// The gas limit of this block. It must fit in a `u32`.
+    /// The gas limit of this block.
```
wborgeaud, line 106:
```suggestion
    /// The gas limit of this block. It must fit in a `u64`.
```

evm/src/proof.rs
```diff
@@ -101,14 +101,15 @@ pub struct BlockMetadata {
...
     pub block_gaslimit: U256,
     /// The chain id of this block.
     pub block_chain_id: U256,
     /// The base fee of this block.
     pub block_base_fee: U256,
-    /// The total gas used in this block. It must fit in a `u32`.
+    /// The total gas used in this block.
```
wborgeaud, line 112:
```suggestion
    /// The total gas used in this block. It must fit in a `u64`.
```

wborgeaud (approved):
Looks good to me!

I think this is fine to merge and, as you said, maybe we can revert it later on for a production release.

evm/src/cpu/gas.rs
Nashtare: Hmm could be. The largest gas_used in the EVM test suite is 53 bits long, so we'd be fine even in this extreme case.
Couldn't we just add a specific TODO here saying that before prod, this should be reverted to only 1 limb?

evm/src/proof.rs
Nashtare: I hadn't mentioned it because other fields also implicitely have to fit in a u64, but will add a comment for all for clarity.

evm/src/cpu/gas.rs
wborgeaud: > Couldn't we just add a specific TODO here saying that before prod, this should be reverted to only 1 limb?

Yes sounds good 👍

PR #1263, opened 2023-10-01
Fix rlp encoding of access_lists

Nashtare:
Also derive `Clone` to be reused on the test-runner side.

### DIFF
--- a/evm/src/generation/mpt.rs
+++ b/evm/src/generation/mpt.rs
@@ -33,7 +33,7 @@ impl Default for AccountRlp {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub enum AccessListInner {
     List(Vec<AccessListItemRlp>),
     Item(AccessListItemRlp),
@@ -64,13 +64,13 @@ impl Decodable for AccessListInner {
     }
 }
 
-#[derive(RlpEncodable, RlpDecodable, Debug)]
+#[derive(Clone, RlpEncodable, RlpDecodable, Debug)]
 pub struct AccessListItemRlp {
     pub address: Address,
     pub storage_keys: Vec<U256>,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct AddressOption(pub Option<Address>);
 
 impl Encodable for AddressOption {
@@ -98,7 +98,7 @@ impl Decodable for AddressOption {
     }
 }
 
-#[derive(RlpEncodable, RlpDecodable, Debug)]
+#[derive(Clone, RlpEncodable, RlpDecodable, Debug)]
 pub struct LegacyTransactionRlp {
     pub nonce: U256,
     pub gas_price: U256,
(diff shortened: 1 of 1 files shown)


evm/src/generation/mpt.rs
```diff
@@ -89,13 +120,13 @@ pub struct AccessListTransactionRlp {
     pub to: AddressOption,
     pub value: U256,
     pub data: Bytes,
-    pub access_list: Vec<AccessListItemRlp>,
+    pub access_list: Vec<AccessListInner>,
```
wborgeaud, line 123: I'm a bit confused by this. EIP-2930 says that the format for `access_list` is `[[{20 bytes}, [{32 bytes}...]]...]`, i.e., `Vec<AccessListItemRlp>`. But this change seems to indicate that it could also be something like `Vec<Vec<AccessListItemRlp>>`.
Are there txns with this access list format?

dvdplm: I agree, this doesn't look right: like it stands this would be accepted: `vec![AccessListInner::List(vec![AccessListItemRlp{address: …, storage_keys: …}, AccessListItemRlp{address: …, storage_keys: …}])]`, which seems odd.

evm/src/generation/mpt.rs
```diff
@@ -33,13 +33,44 @@ impl Default for AccountRlp {
...
+        match self {
+            AccessListInner::List(list) => s.append_list(list),
+            AccessListInner::Item(item) => s.append(item),
+        };
+    }
+}
+
+impl Decodable for AccessListInner {
+    fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
+        if rlp.is_list() {
```
dvdplm, line 53: If this is in the hot loop it can probably be optimized a bit by checking `is_empty()` once and then switch based on byte 0 instead, like:

```rust
if rlp.is_empty() { return DecoderError:RlpExpectedToBeData }
// Can only be list or data now and byte 0 is != 0xc0
if rlp.as_raw[0] > 0xc0 { 
  // do list stuff 
} else {
  // do data stuff
}
```

dvdplm (changes requested):
A unit test or two would be good.

evm/src/generation/mpt.rs
Nashtare: I agree that this isn't the regular encoding, but the EVM test suite has some weird block RLP encodings on some of its tests, which either need to be ignored when parsing or require this specific handling.
The other weird encoding is at the transactions section of the block, I've seen so far three versions:

- a list of txns
- a single txn
- a list of strings (ie doubly encoded)

Nashtare: What I could do then, if just port these changes directly to evm-tests instead. I need some logic there anyway to handle the non-RLP prefix for those transactions, so it wouldn't add too much, and at least here we'd respect the specs.

wborgeaud: Yeah I think that's probably a better idea. These are public structs so I feel like we should have them follow the specs and handle the test suite's oddities in the `evm-tests` repo.

wborgeaud: Good idea 👍

PR #1264, opened 2023-10-02
Derive clone for txn RLP structs

Nashtare:
Also move them behind a transaction_testing module for clarity on their expected usage.
@wborgeaud Would you mind having another look?

### DIFF
--- a/evm/src/generation/mpt.rs
+++ b/evm/src/generation/mpt.rs
@@ -33,88 +33,6 @@ impl Default for AccountRlp {
     }
 }
 
-pub mod transaction_testing {
-    use super::*;
-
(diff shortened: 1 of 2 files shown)


evm/src/generation/mpt.rs
```diff
@@ -33,92 +33,96 @@ impl Default for AccountRlp {
     }
 }
 
-#[derive(RlpEncodable, RlpDecodable, Debug)]
-pub struct AccessListItemRlp {
-    pub address: Address,
-    pub storage_keys: Vec<U256>,
-}
+pub mod transaction_testing {
```
wborgeaud, line 36: tiny nit: Having this module in the middle of the file looks a bit weird. Maybe we can move it to the end of the file?

PR #1270, opened 2023-10-02
Remove extra SHL/SHR CTL.

LindaGuiga:
This PR changes the way SHL and SHR are generated so that all the CPU CTLs looking into `ArithmeticStark` have the same form, and we can save one CTL.

Here, shift operations are generated on the arithmetic side rather on the CPU side. This leads to specific constraints for the shift operations, since now, the first input is `shift` rather than `1 << shift`. 
The three input registers are used in the Arithmetic STARK: the first is `shift`, the second is the input, and the third is `1 << shift`.
This way, we can have the same CTL shape for all arithmetic operations, and it removes the special shift CTL.

@unzvfu Would mind looking at this if you have the time?

### DIFF
--- a/evm/src/arithmetic/mod.rs
+++ b/evm/src/arithmetic/mod.rs
@@ -38,14 +38,11 @@ impl BinaryOperator {
             BinaryOperator::Add => input0.overflowing_add(input1).0,
             BinaryOperator::Mul => input0.overflowing_mul(input1).0,
             BinaryOperator::Shl => {
-                // Compute the shifted displacement, so we can turn the left
-                // left into a multiplication.
-                let shifted_input1 = if input1.bits() <= 32 {
-                    U256::one() << input1
+                if input0 < U256::from(256usize) {
+                    input1 << input0
                 } else {
                     U256::zero()
-                };
-                input0.overflowing_mul(shifted_input1).0
+                }
             }
             BinaryOperator::Sub => input0.overflowing_sub(input1).0,
             BinaryOperator::Div => {
@@ -56,17 +53,10 @@ impl BinaryOperator {
                 }
             }
             BinaryOperator::Shr => {
-                // Compute the shifted displacement, so we can turn the
-                // right shift into a multiplication.
-                let shifted_input1 = if input1.bits() <= 32 {
-                    U256::one() << input1
+                if input0 < U256::from(256usize) {
+                    input1 >> input0
                 } else {
                     U256::zero()
-                };
-                if shifted_input1.is_zero() {
-                    U256::zero()
-                } else {
-                    input0 / shifted_input1
                 }
             }
             BinaryOperator::Mod => {
--- a/evm/src/arithmetic/mod.rs
+++ b/evm/src/arithmetic/mod.rs
@@ -9,6 +9,7 @@ mod byte;
 mod divmod;
 mod modular;
 mod mul;
+mod shift;
 mod utils;
 
 pub mod arithmetic_stark;
@@ -35,13 +36,37 @@ impl BinaryOperator {
     pub(crate) fn result(&self, input0: U256, input1: U256) -> U256 {
         match self {
             BinaryOperator::Add => input0.overflowing_add(input1).0,
-            BinaryOperator::Mul | BinaryOperator::Shl => input0.overflowing_mul(input1).0,
+            BinaryOperator::Mul => input0.overflowing_mul(input1).0,
+            BinaryOperator::Shl => {
+                // Compute the shifted displacement, so we can turn the left
+                // left into a multiplication.
+                let shifted_input1 = if input1.bits() <= 32 {
+                    U256::one() << input1
+                } else {
+                    U256::zero()
+                };
+                input0.overflowing_mul(shifted_input1).0
+            }
             BinaryOperator::Sub => input0.overflowing_sub(input1).0,
-            BinaryOperator::Div | BinaryOperator::Shr => {
+            BinaryOperator::Div => {
                 if input1.is_zero() {
                     U256::zero()
                 } else {
                     input0 / input1
+                }
+            }
+            BinaryOperator::Shr => {
+                // Compute the shifted displacement, so we can turn the
+                // right shift into a multiplication.
+                let shifted_input1 = if input1.bits() <= 32 {
+                    U256::one() << input1
+                } else {
+                    U256::zero()
+                };
+                if shifted_input1.is_zero() {
+                    U256::zero()
+                } else {
+                    input0 / shifted_input1
                 }
             }
             BinaryOperator::Mod => {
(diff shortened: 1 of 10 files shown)


evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+    result: U256,
+) {
+    // TODO: It would probably be clearer/cleaner to read the U256
+    // into an [i64;N] and then copy that to the lv table.
+    // The first input is the shift we need to apply.
+    u256_to_array(&mut lv[INPUT_REGISTER_0], shift);
+    // The second register holds the input which needs shifting.
+    u256_to_array(&mut lv[INPUT_REGISTER_1], input);
+    u256_to_array(&mut lv[OUTPUT_REGISTER], result);
+    // If `shift > 2^32`, the shifted displacement is set to 0.
```
unzvfu, line 55:
```suggestion
    // If `shift >= 256`, the shifted displacement is set to 0.
```

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+    } else {
+        U256::one() << shift
+    };
+
+    u256_to_array(&mut lv[INPUT_REGISTER_2], shifted_displacement);
+
+    let input0 = read_value_i64_limbs(lv, INPUT_REGISTER_1); // input
+    let input1 = read_value_i64_limbs(lv, INPUT_REGISTER_2); // 1 << shift
+
+    if is_shl {
```
unzvfu, line 68: I think the code in the two parts of this if statement can easily be refactored with the code in `mul.rs` and `divmod.rs` that they were copied from.

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+
+        lv[AUX_INPUT_REGISTER_0].copy_from_slice(&out);
+    }
+}
+
+/// Evaluates the constraints for an SHL opcode.
+/// The logic is very similar to the one for MUL. The only difference is that
+/// the inputs are in `INPUT_REGISTER_1`  and `INPUT_REGISTER_2` instead of
+/// `INPUT_REGISTER_0` and `INPUT_REGISTER_1`.
+pub fn eval_packed_shl<P: PackedField>(
```
unzvfu, line 131: Again, I think it's preferable and easy for the code to be shared between here and `mul.rs` rather than copy-pasting it. Same applies to all the other functions from here on.

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+        let mut rng = ChaCha8Rng::seed_from_u64(0x6feb51b7ec230f25);
+        let mut lv = [F::default(); NUM_ARITH_COLUMNS].map(|_| F::sample(&mut rng));
+        let mut nv = [F::default(); NUM_ARITH_COLUMNS].map(|_| F::sample(&mut rng));
+
+        // set `IS_MUL == 1` and ensure all constraints are satisfied.
+        lv[IS_SHL] = F::ONE;
+        lv[IS_SHR] = F::ZERO;
+
+        for _i in 0..N_RND_TESTS {
+            let shift = U256::from(rng.gen::<usize>());
```
unzvfu, line 362: If `shift` is a random `usize`-sized value, then the probability that it is less than 256 is about 2^-56. :)
```suggestion
            let shift = U256::from(rng.gen::<u8>());
```
Probably better to have a separate test for big shifts where the expected result is zero.

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+                GoldilocksField::ONE,
+                GoldilocksField::ONE,
+                GoldilocksField::ONE,
+            );
+            eval_packed_generic(&lv, &nv, &mut constraint_consumer);
+            for &acc in &constraint_consumer.constraint_accs {
+                assert_eq!(acc, GoldilocksField::ZERO);
+            }
+        }
+    }
```
unzvfu, line 393: A similar set of tests for right shift would also be good.

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,394 @@
...
+            u256_to_array(&mut lv[INPUT_REGISTER_2], shifted);
+            let mut full_input = U256::from(0);
+            // set inputs to random values
+            for ai in INPUT_REGISTER_1 {
+                lv[ai] = F::from_canonical_u16(rng.gen());
+                full_input =
+                    U256::from(lv[ai].to_canonical_u64()) + full_input * U256::from(1 << 16);
+            }
+
+            let output = full_input * shifted;
```
unzvfu, line 378: Note that making the suggested change of `usize` to `u8` above triggers an overflow exception on this line. Probably easiest to just do
```suggestion
            let output = full_input << shift;
```

unzvfu (changes requested):
Overall this is a nice improvement, thanks! There are several things that will need fixing or tidying up though; see my comments.

evm/src/arithmetic/mod.rs
```diff
@@ -35,15 +36,39 @@ impl BinaryOperator {
...
+            BinaryOperator::Mul => input0.overflowing_mul(input1).0,
+            BinaryOperator::Shl => {
+                // Compute the shifted displacement, so we can turn the left
+                // left into a multiplication.
+                let shifted_input1 = if input1.bits() <= 32 {
+                    U256::one() << input1
+                } else {
+                    U256::zero()
+                };
+                input0.overflowing_mul(shifted_input1).0
```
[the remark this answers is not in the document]
LindaGuiga: The argument order for the generation on the Arithmetic side was indeed reversed: I fixed this.
I limited `input1` to 2^32 bits (mistakenly) basing my code on this piece of code in operation.rs:
```
if input0.bits() <= 32 {
        let (_, read) = mem_read_gp_with_log_and_fill(LOOKUP_CHANNEL, lookup_addr, state, &mut row);
        state.traces.push_memory(read);
    } else {
        // The shift constraints still expect the address to be set, even though no read will occur.
        let channel = &mut row.mem_channels[LOOKUP_CHANNEL];
        channel.addr_context = F::from_canonical_usize(lookup_addr.context);
        channel.addr_segment = F::from_canonical_usize(lookup_addr.segment);
        channel.addr_virtual = F::from_canonical_usize(lookup_addr.virt);
    }
```
where we carry out a memory read if the number of bits is less than 32. I changed the limitation to 16 bits in the new version, but I was wondering why there was a difference between the memory reads and the limit in `shift`?

LindaGuiga:
@unzvfu Thank you for your review!

As suggested in one of the comments, I added separate tests for `shift > 256`, and I realized a mistake had been introduced (I think in PR #1166, which introduced `IS_SHR` and `IS_SHL`) for that case: the modulus was not set to 0 but to 1 in that case (and some constraints involved only `IS_DIV` when they should have included both `IS_DIV` and `IS_SHR`).
I fixed this issue in the latest PR, where I also introduced the tests for `shift > 256`. If you think this should go into another PR, I can always revert the latest commit!

LindaGuiga:
Just a side note on my latest commit and comment: the shift issue on main was causing a couple of tests to fail in evm-tests (`shr_-1_256` / `shr_2^255_256`  / `shr_2^255_257`/`exp_4` for instance) and the tests pass with the fix.

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,338 @@
...
+        // otherwise, the output is 0. We use the logic in divmod.rs to achieve that.
+        divmod::generate_divmod(lv, nv, IS_SHR, INPUT_REGISTER_1, INPUT_REGISTER_2);
+    }
+}
+
+/// Evaluates the constraints for an SHL opcode.
+/// The logic is the same as the one for MUL. The only difference is that
+/// the inputs are in `INPUT_REGISTER_1`  and `INPUT_REGISTER_2` instead of
+/// `INPUT_REGISTER_0` and `INPUT_REGISTER_1`.
+pub fn eval_packed_shl<P: PackedField>(
```
unzvfu, line 84: Nit:
```suggestion
fn eval_packed_shl<P: PackedField>(
```

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,338 @@
...
+pub fn eval_packed_generic<P: PackedField>(
+    lv: &[P; NUM_ARITH_COLUMNS],
+    nv: &[P; NUM_ARITH_COLUMNS],
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    eval_packed_shl(lv, yield_constr);
+    eval_packed_shr(lv, nv, yield_constr);
+}
+
+pub fn eval_ext_circuit_shl<F: RichField + Extendable<D>, const D: usize>(
```
unzvfu, line 129: Nit:
```suggestion
fn eval_ext_circuit_shl<F: RichField + Extendable<D>, const D: usize>(
```

evm/src/arithmetic/shift.rs
```diff
@@ -0,0 +1,338 @@
...
+        builder,
+        lv,
+        is_shl,
+        input0_limbs,
+        shifted_limbs,
+        yield_constr,
+    );
+}
+
+pub fn eval_ext_circuit_shr<F: RichField + Extendable<D>, const D: usize>(
```
unzvfu, line 148: Nit:
```suggestion
fn eval_ext_circuit_shr<F: RichField + Extendable<D>, const D: usize>(
```

unzvfu (approved):
Looks great, thanks Linda! Just a few little nits about function visibility that will be easy to incorporate (not a big deal, but I think its a good habit to limit the visibility of helper functions).

And nice work catching the bug with big shifts!

PR #1271, opened 2023-10-03
Fix genesis state trie root when calling `prove_root`

wborgeaud:
`test_log_with_aggreg` was failing because it called `prove_root` with an empty genesis state trie not matching the non-empty `state_trie_before` used.

Also renamed `ExtraBlockData::genesis_state_root` to `ExtraBlockData::genesis_state_trie_root` to match the naming in `GenerationInputs`.


LindaGuiga (approved):
Looks great, thanks!

PR #1274, opened 2023-10-03
Add PROVER_INPUT stack constraints.

LindaGuiga:
This PR adds missing stack constraints for `PROVER_INPUT` by adding a `StackBehavior` with no pop, pushes and other channels disabled.


LindaGuiga:
Closing this as this is actually included in #1168.

PR #1276, opened 2023-10-04
Fix encoding for empty recipient

Nashtare:
The current custom encoding for empty `to` fields is erroneous, as it will attempt to end early the RLP processing, leading the final RLP string to miss the length of the final element (the `s` field of the signature) in the prefix computation.
This can be tested for instance by getting an arbitrary transaction, replace the `to` field with `AddressOption(None)` encode it, and try to decode it with an online tool like https://toolkit.abdk.consulting/ethereum#rlp.


wborgeaud (approved):
Looks good to me!

PR #1280, opened 2023-10-06
Fix shift constraint

Nashtare:
This fixes the shift constraints for displacement > $2^{32}$.

The issue is that when calling `append_shift` in `witness/operations.rs`, we don't perform a READ if the displacement does not fit in the lowest limb, but this isn't properly represented in constraints.
We also didn't compute the general column for the inverse of the high_limbs_sum, which is necessary for satisfying some of the constraints.

I've also added constraint comments on the recursive side for easy reading and consistency with how the other modules are documented.

@unzvfu Would you mind confirming?


unzvfu (approved):
Looks good!

PR #1281, opened 2023-10-07
Remove some dead_code in EVM crate

Nashtare:
We thought it'd be probably good to cleanup the codebase with what's most likely not going to be used anymore.
That's not all, but that's a start at least.


Nashtare:
Note that clippy is complaining about `shift_mut` in the Cpu general columns being unused but this is actually the source of an issue addressed in #1280 and would go away after #1280 is merged.

wborgeaud (approved):
Looks good to me!

PR #1282, opened 2023-10-09
Fix revert account destroyed with nonexistent target

LindaGuiga:
We noticed that the `create2collisionSelfdestructedRevert` tests from the evm-tests test suite were failing with a Kernel panic. This is because in `revert_account_destroyed`, we access the `target`'s balance to undo the balance change carried out in `SELF_DESTRUCT`. But in the tests, the `target` account had to be created when carrying out a `SELF_DESTRUCT`, and the revert therefore already deleted it. In that case, `mpt_read_state_trie` returns 0 (corresponding to a nonexistent account), and we should not try to change the `target`'s balance.

This PR therefore ensures that when a `target` account is nonexistent, we do not try to change its balance.

### DIFF
--- a/evm/src/cpu/kernel/asm/journal/account_destroyed.asm
+++ b/evm/src/cpu/kernel/asm/journal/account_destroyed.asm
@@ -17,12 +17,13 @@ revert_account_destroyed_contd:
     // Remove `prev_balance` from `target`'s balance.
     // stack: target, address, prev_balance, retdest
     %mpt_read_state_trie
-    %add_const(1)
+    // Only remove `prev_balance` if `target` exists.
     // stack: target_balance_ptr, address, prev_balance, retdest
-    DUP3
-    DUP2 %mload_trie_data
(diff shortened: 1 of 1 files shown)


wborgeaud (commented):
Good catch!
Your fix should work, but I think the underlying issue here is that in `sys_selfdestruct` we do `%journal_add_account_destroyed` before doing `%journal_add_account_created` (in `%add_eth`). 
It should be the other way around and looking at the `revm` implementation, that's [what they do too](https://github.com/bluealloy/revm/blob/41987aa5b17047b10c5c833157732bf6b7d34b0f/crates/revm/src/journaled_state.rs#L417).
I've pushed this change here https://github.com/0xPolygonZero/plonky2/tree/wb/testing_selfdestruct and it makes the `create2collisionSelfdestructedRevert` tests pass, although I haven't run the full suite with it yet. 
What do you think?

wborgeaud:
@LindaGuiga Great! Just opened #1287 with the fix.

PR #1283, opened 2023-10-09
Make sure success is 0 in contract failure

Nashtare:
`Syscalls` push a default 1 as txn `status` at the end of their execution, prior jumping to `terminate_common`. However the latter may end up in failure, while the `status` stored in the stack is never modified to be changed to 0.

Detected through all the `CREATE_ContractRETURNBigOffset` variants (now passing).
Also fixes `codesizeOOGInvalidSize` variants and most of previously failing `CreateOOGFromEOARefunds` variants.


wborgeaud (approved):
Good catch 👍

PR #1284, opened 2023-10-10
Fix failed receipt.

LindaGuiga:
We noticed a few evm-tests failing (such as Create2OOGFromCallRefunds_{16, 17}) when the logs are non-empty and the status of the transaction is 0. This is because we set the length of the logs to 0 but not their payload's length, which is later used for encoding the receipts. This PR fixes this by setting the logs' payload length to 0 when the status is 0.


Nashtare:
Also fixes the similar variants for `Create`, and the failing `CreateOOGFromEOARefunds` ones :)

wborgeaud (approved):
Nice 👍

PR #1285, opened 2023-10-10
Fix sys_blockhash

hratoanina:
Wrong stack at the end of `sys_blockhash`. Fixes (at least) one `stEIP150singleCodeGasPrices/gasCostBerlin` test.


wborgeaud (approved):
Good catch!

PR #1286, opened 2023-10-10
Add journal entry for logs

LindaGuiga:
We noticed a few tests failing, such as `logInOOG_Call` due to logs happening within a CALL, and not reverting. To fix this, this PR adds logs to the journal so they can be reverted when necessary.

### DIFF
--- a/evm/src/cpu/kernel/asm/core/log.asm
+++ b/evm/src/cpu/kernel/asm/core/log.asm
@@ -265,22 +265,3 @@ one_byte_data:
     %jump(rlp_data_len)
 %%after:
 %endmacro
-
-%macro journal_add_log
-    %journal_add_2(@JOURNAL_ENTRY_LOG)
-%endmacro
-
-global revert_log:
-    // stack: entry_type, ptr, retdest
-    POP
-    // First, reduce the number of logs.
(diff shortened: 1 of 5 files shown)


evm/src/cpu/kernel/asm/core/log.asm
```diff
@@ -263,3 +265,22 @@ one_byte_data:
...
+    %mload_global_metadata(@GLOBAL_METADATA_LOGS_LEN)
+    %decrement
+    %mstore_global_metadata(@GLOBAL_METADATA_LOGS_LEN)
+    // stack: ptr, retdest
+    // Second, restore payload length.
+    %journal_load_2
+    // stack: prev_logs_data_len, prev_payload_len, retdest
+    %mstore_global_metadata(@GLOBAL_METADATA_LOGS_DATA_LEN)
+    %mstore_global_metadata(@GLOBAL_METADATA_LOGS_PAYLOAD_LEN)
+    JUMP
```
wborgeaud, line 286: nit: For other journal entries, we have a file `evm/src/cpu/kernel/asm/journal/X.asm` containing `journal_add_X` and `revert_X`. So maybe we can add a `evm/src/cpu/kernel/asm/journal/log.asm` with this code.

wborgeaud (approved):
Looks good to me!

PR #1287, opened 2023-10-11
Fix journal order in `sys_selfdestruct`

wborgeaud:
We have to call `%journal_add_account_destroyed` after `%journal_add_account_created`.
See #1282 for discussions.


LindaGuiga (approved):
Looks great, thanks!

PR #1288, opened 2023-10-12
Updated `mir-protocol` --> `0xPolygonZero`

BGluth:
- Recently the underlying org name on Github changed to `0xPolygonZero`.
- This was causing some issues with dependencies downstream where a dependency pointing to the old org name was seen as a different dependency that was using the new org name.
- Using a `[patch] section in `Cargo.toml` was running into issues where we couldn't get it to get all dependency versions using a single org name.


dlubarov (approved):
LGTM

PR #1289, opened 2023-10-12
Reduce visibility for a bunch of structs and methods in EVM crate

Nashtare:
As a follow-up of the previous cleanup, this reduces visibility of a lot of previously public structs and methods.
I added a `#[allow(unused)]` in both the CTL module (for the new duplexed operations that aren't used yet, but soon!) and in the curve module for bls operations.

I was wondering though:
- should we also limit the visibility of the `Stark` trait? It currently forces `(Recursive)ConstraintConsumer` and `Lookup` types to be public (although the latter already hides all its fields)
- should we limit the visibility of all the internal `AllRecursiveCircuits` fields? For now only the STARK shrinking tables are private.


dvdplm (approved):
In general I think that anything we can do to reduce the number of public items, the better. Smaller API surfaces are strictly better in terms of maintainability.

PR #1290, opened 2023-10-12
Update `check_ctls` with extra looking values

Nashtare:
Update `check_ctls` utility tester and add it to the prover (behind a conditional attribute).


wborgeaud (approved):
Looks good to me!

PR #1291, opened 2023-10-12
Remove values of last memory channel

hratoanina:
This PR removes eight CPU columns (the eight value columns of the last memory channel).

The core idea is that all operations using five memory channels always have two channels sharing the same values. For example, for `MSTORE_GENERAL`, the value written in memory is the one popped from the stack. Another example is syscalls, which only push, but also use four other memory channels (channel 0 is reserved for the top of the stack, and the next three read the syscall address): if the stack isn't empty, you also need to write the current top of the stack in memory, which is already present in channel 0.

This required a lot of refactoring to make sure that every other instruction only used at most four memory channel, making  extensive use of https://github.com/0xPolygonZero/plonky2/pull/1229 to reduce the amount of channels used by some operations. The most modified one is `MSTORE_32BYTES`: to remove the length as argument, we added 32 individual instructions (similar to DUP or SWAP) and handled them separately in the kernel. This adds a few CPU cycles per call to `%mstore_unpacking`, but the overhead is negligible.

I'm not sure who would be best to tag?
@npwardberkeley @nbgl @wborgeaud


evm/src/cpu/decode.rs
```diff
@@ -39,13 +39,12 @@ const OPCODES: [(u8, usize, bool, usize); 17] = [
...
     (0x5b, 0, false, COL_MAP.op.jumpdest),
     (0x5f, 0, false, COL_MAP.op.push0),
-    (0x60, 5, false, COL_MAP.op.push), // 0x60-0x7f
-    (0x80, 4, false, COL_MAP.op.dup),  // 0x80-0x8f
-    (0x90, 4, false, COL_MAP.op.swap), // 0x90-0x9f
-    (0xee, 0, true, COL_MAP.op.mstore_32bytes),
+    (0x60, 5, false, COL_MAP.op.push),          // 0x60-0x7f
+    (0x80, 4, false, COL_MAP.op.dup),           // 0x80-0x8f
+    (0x90, 4, false, COL_MAP.op.swap),          // 0x90-0x9f
+    (0xc0, 5, true, COL_MAP.op.mstore_32bytes), // 0xc0-0xdf
```
Nashtare, line 45: nit: I'd move it above to have the opcode range aligned.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -415,23 +415,23 @@ impl<'a> Interpreter<'a> {
...
-            0xf7 => self.run_set_context(),                             // "SET_CONTEXT",
-            0xf8 => self.run_mload_32bytes(),                           // "MLOAD_32BYTES",
-            0xf9 => todo!(),                                            // "EXIT_KERNEL",
-            0xfa => todo!(),                                            // "STATICCALL",
-            0xfb => self.run_mload_general(),                           // "MLOAD_GENERAL",
-            0xfc => self.run_mstore_general(),                          // "MSTORE_GENERAL",
-            0xfd => todo!(),                                            // "REVERT",
-            0xfe => bail!("Executed INVALID"),                          // "INVALID",
-            0xff => todo!(),                                            // "SELFDESTRUCT",
+            x if (0xc0..0xe0).contains(&x) => self.run_mstore_32bytes(x - 0xc0 + 1), // "MSTORE_32BYTES",
```
Nashtare, line 418: nit: I'd also move this to have proper opcode ordering

evm/src/cpu/kernel/opcodes.rs
```diff
@@ -4,7 +4,7 @@ pub fn get_push_opcode(n: u8) -> u8 {
     0x5f + n
 }
 
-/// The opcode of a standard instruction (not a `PUSH`).
+/// The opcode of a standard instruction (not a `PUSH` or a `MSTORE_32BYTES`).
```
Nashtare, line 7: I'm not sure this comment is accurate. We do have the different `MSTORE_32BYTES` opcodes, unlike the `PUSH` operations, so I'd leave it unchanged.

Nashtare (commented):
Not a review, just dropped a minor comment.
Also I've attempted at reducing unnecessary public visibility for a bunch of structs and methods in #1289, could be nice if the new ones here were not made `pub` unless really necessary :)

Nashtare:
@pgebheim This induces quite some refactoring, so may be worth integrating in the audit version if time allows? (Gain not negligible, 8 columns on CPU removed)

pgebheim:
I'm 100% down with you including this -- as long as we can get it in before audit start and we document that the patch has been added.

evm/src/generation/mod.rs
```diff
@@ -317,6 +318,13 @@ fn simulate_cpu<F: RichField + Extendable<D>, const D: usize>(
                     break;
                 }
             }
+
+            // First CPU row's memory channel 0 must contain the kernel hash to make
+            // the Keccak CTL check pass.
+            let first_cpu_row = &mut state.traces.cpu[bootstrapping_len];
+            first_cpu_row.mem_channels[0].value = KERNEL.code_hash.map(F::from_canonical_u32);
+            first_cpu_row.mem_channels[0].value.reverse();
```
LindaGuiga, line 326: I believe this is no longer necessary, since it is dealt with by the update of the stack_top in `generate_bootstrap_kernel` + the copying of the previous top of the stack into `mem_channels[0]`. (The update of the top of the stack was not there was not there when you first opened the PR).

evm/src/witness/util.rs
```diff
@@ -200,6 +195,26 @@ pub(crate) fn mem_write_gp_log_and_fill<F: Field>(
...
+    address: MemoryAddress,
+    state: &GenerationState<F>,
+    row: &mut CpuColumnsView<F>,
+    val: U256,
+) -> MemoryOp {
+    let op = mem_write_log(MemoryChannel::PartialChannel, address, state, val);
+
+    let channel = &mut row.partial_channel;
+    assert!(channel.used.is_zero());
+    // channel.used = F::ONE;
```
LindaGuiga, line 208: Maybe you should uncomment this, since I believe you're setting it to one every time you actually call this function (and then you could remove ` row.partial_channel.used = F::ONE;` in the statement `if let Some(log) = write`, in `push_with_write`).

evm/src/witness/operation.rs
```diff
@@ -858,7 +861,14 @@ pub(crate) fn generate_mstore_general<F: Field>(
             .try_into()
             .map_err(|_| MemoryError(VirtTooLarge { virt }))?,
     };
-    let log_write = mem_write_gp_log_and_fill(4, address, state, &mut row, val);
+    let log_write = mem_write_partial_log_and_fill(address, state, &mut row, val);
+    // Write in partial channel.
+    let channel = &mut row.partial_channel;
+    channel.used = F::ONE;
+    channel.is_read = F::ZERO;
```
LindaGuiga, line 868: Isn't all this done in `mem_write_partial_log_and_fill`?

LindaGuiga (commented):
Looks good! But I'd like to make another pass at the changes, just to be sure I'm not missing anything!

evm/src/cpu/kernel/asm/memory/packing.asm
```diff
@@ -41,12 +41,243 @@ global mload_packing_u64_LE:
...
+    //stack: BYTES_PER_JUMP, len, context, segment, offset, value, retdest
+    MUL
+    //stack: jump_offset, context, segment, offset, value, retdest
+    PUSH mstore_unpacking_0
+    //stack: mstore_unpacking_0, jump_offset, context, segment, offset, value, retdest
+    ADD
+    //stack: address_unpacking, context, segment, offset, value, retdest
+    JUMP
+
+mstore_unpacking_0:
```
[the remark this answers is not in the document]
Nashtare: Are we sure this will properly panic? Because otherwise it will yield an unverifiable proofs, as the Bytepacking table cannot handle empty segments, which would fail the CTL verification.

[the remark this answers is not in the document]
Nashtare: additionally I'd add indentation to the %rep below for clarity, as is usually done.
```asm
    %rep 3
        PUSH 0
    %endrep
```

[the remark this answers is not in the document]
hratoanina: I believe you're never supposed to call `mstore_unpacking` with length `0`, but I might be wrong. @Nashtare?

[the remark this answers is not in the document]
Nashtare: We're never supposed, but I don't think there is such handling as I mentioned above. So we may as well add this logic here, by skipping completely the read/write if len==0?
For now it is only enforced through `memcpy`/`memset`.

evm/src/cpu/memio.rs
```diff
@@ -90,7 +96,7 @@ fn eval_ext_circuit_load<F: RichField + Extendable<D>, const D: usize>(
     }
 
     // Disable remaining memory channels, if any.
-    for &channel in &lv.mem_channels[4..NUM_GP_CHANNELS] {
+    for &channel in &lv.mem_channels[4..] {
```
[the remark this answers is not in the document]
hratoanina: Hmm, I'd rather keep them just to be safe, but I don't have a strong opinion against it. Apparently the ones for `contextops` were removed in another PR, and I'm not sure about bringing them back in this one.

LindaGuiga (approved):
Looks good, thanks!

PR #1292, opened 2023-10-13
Remerge context flags

hratoanina:
https://github.com/0xPolygonZero/plonky2/pull/1215 split context flags. This PR merges them again.


evm/src/cpu/contextops.rs
```diff
@@ -197,66 +218,38 @@ fn eval_ext_circuit_set<F: RichField + Extendable<D>, const D: usize>(
         yield_constr.constraint(builder, constr);
     }
 
-    // The next row's stack top is loaded from memory (if the stack isn't empty).
+    // The new top is loaded is in memory channel 3, if the stack isn't empty (see eval_packed).
```
LindaGuiga, line 221: Same as above

evm/src/cpu/modfp254.rs
```diff
@@ -24,7 +24,7 @@ pub fn eval_packed<P: PackedField>(
     // We want to use all the same logic as the usual mod operations, but without needing to read
     // the modulus from the stack. We simply constrain `mem_channels[1]` to be our prime (that's
     // where the modulus goes in the generalized operations).
-    let channel_val = lv.mem_channels[2].value;
+    let channel_val = lv.mem_channels[1].value;
```
LindaGuiga, line 27: Is this supposed to be changed?

evm/src/witness/operation.rs
```diff
@@ -318,7 +318,22 @@ pub(crate) fn generate_get_context<F: Field>(
     state: &mut GenerationState<F>,
     mut row: CpuColumnsView<F>,
 ) -> Result<(), ProgramError> {
-    push_with_write(state, &mut row, state.registers.context.into())?;
+    // Same logic as push_with_write, but we have to use channel 3 for stack constraint reasons.
```
LindaGuiga, line 321: Actually, since we are no longer storing the top of the stack in channel 4 (I think that was the reason), can't we use channel 4 both here and for the constraints now?

evm/src/cpu/contextops.rs
```diff
@@ -268,6 +261,33 @@ pub fn eval_packed<P: PackedField>(
...
+    yield_constr.constraint(new_filter * (channel.is_read - lv.opcode_bits[0]));
+    // In both cases, next row's context works.
+    yield_constr.constraint(new_filter * (channel.addr_context - nv.context));
+    // Same segment for both.
+    yield_constr.constraint(
+        new_filter * (channel.addr_segment - P::Scalar::from_canonical_u64(Segment::Stack as u64)),
+    );
+    // The address is one less than stack_len.
+    let addr_virtual = stack_len - P::ONES;
+    yield_constr.constraint(new_filter * (channel.addr_virtual - addr_virtual));
```
LindaGuiga, line 290: I might have missed it, but are you constraining the new stack length to be either `lv.stack_len-1` or `lv.stack_len`?

evm/src/witness/operation.rs
hratoanina: That's true, though it will have to be 3 after https://github.com/0xPolygonZero/plonky2/pull/1291 is merged. We might as well leave it as is.

LindaGuiga (approved):

PR #1295, opened 2023-10-16
Add some documentation in EVM crate

Nashtare:
@pgebheim That's the branch we started adding documentation on.
We didn't touch the kernel at all, nor the specs pdf (apart from adding an empty section for the new table).


pgebheim:
@Nashtare @LindaGuiga Would you be okay with merging this as is and then adding more docs?

Nashtare:
Sure, let me just rebase first.

Nashtare:
@pgebheim should be ready for merging if you want!

dvdplm (commented):
It'd be good to use the "`[SomeItemInScope]`"-syntax to leverage rustdoc auto-linking feature. So rather than 
```
/// Something something returns `CrossTableLookups`.
```

…you'd write 

```
/// Something something returns [`CrossTableLookups`].
```

Nashtare:
@dvdplm Agreed, though apart from the serialization module inside `plonky2` crate, this is never used, and the `plonky2_evm` crate hasn't been published on `crates.io`, so I'd rather go for a dedicated PR updating _all_ references with that suggestion (not just on `plonky2_evm`). Does that make sense to you?
cc @pgebheim

PR #1296, opened 2023-10-17
Remove new_stack_top_channel from StackBehavior

LindaGuiga:
Since #1229 introduced multi-row CTLs, it is no longer necessary to have the outputs of operations (such as arithmetic ones) on the same row in the CPU, since we can use the next row for CTLs and constraints. This PR thus removes the need for the `new_stack_top_channel` field in `Stackbehavior`.


Nashtare (approved):
LGTM, thanks!

PR #1297, opened 2023-10-18
Check gas in sys_stop

### LINKED ISSUE #1106: No out of gas check in `sys_stop`
https://github.com/mir-protocol/plonky2/blob/3de92d9ed1721cec133e4e1e1b3ec7facb756ccf/evm/src/cpu/kernel/asm/core/terminate.asm#L4

The `sys_stop` function does not check if the current gas usage exceeds the context gas limit. This can happen since simple opcodes which are directly evaluated in the CPU Stark update the gas usage, but do not check that it stays below the limit. This makes, for example, some of the `gasMeasure` tests of the EVM test suite fail.
To enforce the check, adding the line `%charge_gas_const(0)` does the trick, but the issue might be more widespread and I find the hotfix pretty ugly. Do you have any suggestions?

On a related note, this might enable a minor DOS attack: a very long code with only simple opcodes which would trigger an OOG error very early would be evaluated in its entirety by the EVM, making the CPU trace very long for no reason.

Nashtare:
Implements @hratoanina's suggestion in #1106 to handle gas overflowing the gaslimit caused by native instructions.

closes #1106

### DIFF
--- a/evm/src/cpu/kernel/asm/core/terminate.asm
+++ b/evm/src/cpu/kernel/asm/core/terminate.asm
@@ -6,6 +6,10 @@ global sys_stop:
     // Set the parent context's return data size to 0.
     %mstore_parent_context_metadata(@CTX_METADATA_RETURNDATA_SIZE, 0)
 
+    // This makes sure the gas used hasn't overflowed the gaslimit.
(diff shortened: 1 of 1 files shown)


wborgeaud (approved):
LGTM 👍

PR #1299, opened 2023-10-20
Pad according to RATE rather than WIDTH

LindaGuiga:
I noticed that when padding a hash, we pad to a multiple of `WIDTH` instead of `RATE`. I believe this is a mistake that this PR fixes. @dlubarov  Would you mind letting me know whether I'm missing something?


dlubarov (approved):
Good catch!

PR #1300, opened 2023-10-21
refactor: remove usage of unstable `generic_const_exprs` in starky

bingcicle:
This PR mostly lifts https://github.com/0xPolygonZero/plonky2/pull/1246 down into the lib level into the starky crate, getting rid of the unstable `generic_const_exprs` feature from the crate entirely. This should work towards the https://github.com/0xPolygonZero/plonky2/issues/417 as well.

This would probably warrant a minor version bump in the starky crate to 0.2.0, since this would be an API change for anyone relying on starky.


starky/src/lib.rs
```diff
@@ -1,7 +1,6 @@
 #![allow(incomplete_features)]
```
Nashtare, line 1:
```suggestion
```

I believe this can also be removed now :)

bingcicle: You're right! Thanks for the catch

starky/src/evaluation_frame.rs
```diff
@@ -0,0 +1,66 @@
...
+    const COLUMNS: usize;
+    const PUBLIC_INPUTS: usize;
+
+    /// Returns the local values (i.e. current row) for this evaluation frame.
+    fn get_local_values(&self) -> &[T];
+    /// Returns the next values (i.e. next row) for this evaluation frame.
+    fn get_next_values(&self) -> &[T];
+
+    fn get_public_inputs(&self) -> &[U];
+    /// Outputs a new evaluation frame from the provided local and next values.
```
Nashtare, line 19:
```suggestion
    fn get_public_inputs(&self) -> &[U];
    
    /// Outputs a new evaluation frame from the provided local and next values.
```

minor nit, but I'd add a separation here.

Nashtare (approved):
LGTM! Thanks for doing this :)

PR #1302, opened 2023-10-21
Remove kernel memory zeroizing in SHA256 precompile

Nashtare:
I originally wanted to amortize the `%zero_out_kernel_general` macro by leveraging the `MSTORE_32BYTES` instruction instead, but it seems that it is not needed at all anymore.

@npwardberkeley When looking at the git history for this part I noticed it was needed for something you were working on a while back. Can you confirm the removal is sound? Thanks 🙏


wborgeaud (approved):
Yes this was fixed in #1029, so this is ok to delete.

PR #1303, opened 2023-10-21
Amortize bloom filter reset at end of txn

Nashtare:
We currently zeroize one byte of the bloom filter at a time.
This improves the reset by proceeding by chunks of 32 bytes and calling `MSTORE_32BYTES` with value 0.

Saves around 2.2k CPU cycles / 8.8k Memory rows.


wborgeaud (approved):
Nice 👍

PR #1304, opened 2023-10-21
Add macro for copying sequences of bytes

Nashtare:
The current `%memcpy` is quite wasteful for sequence of bytes, which is being used twice for transaction encoding.
I suggest having a dedicated macro `memcpy_bytes` for those, which packs the bytes by sequences of length 32 (or smaller for the last chunk), and rely on `MLOAD_32BYTES` / `MSTORE_32BYTES` to amortize the CPU / Memory overhead.

On `add11_yml`, I am getting a mild 4.7% reduction in CPU cycle count, but for larger transactions, the impact is better.
For the more extreme example (`Shanghai/creationTxInitCodeSizeLimit_d0g0v0_Shanghai.json`), this is pruning almost ***30%*** of both CPU and Memory traces, with number of CPU cycles going from 8_758_697 to 6_192_428 and Memory rows going from 23_644_152 to 16_807_024.

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/memcpy.asm
+++ b/evm/src/cpu/kernel/asm/memory/memcpy.asm
@@ -54,7 +54,7 @@ memcpy_finish:
 %%after:
 %endmacro
 
-// Similar logic than memcpy, but optimized for copying sequences of bytes.
+// Similar logic to memcpy, but optimized for copying sequences of bytes.
 global memcpy_bytes:
     // stack: DST, SRC, count, retdest
     DUP7
--- a/evm/src/cpu/kernel/asm/memory/memcpy.asm
+++ b/evm/src/cpu/kernel/asm/memory/memcpy.asm
@@ -53,3 +53,74 @@ memcpy_finish:
     %jump(memcpy)
 %%after:
 %endmacro
+
+// Similar logic than memcpy, but optimized for copying sequences of bytes.
+global memcpy_bytes:
+    // stack: DST, SRC, count, retdest
+    DUP7
+    // stack: count, DST, SRC, count, retdest
+    %lt_const(0x20)
+    // stack: count < 32, DST, SRC, count, retdest
+    %jumpi(memcpy_bytes_finish)
+    
+    // We will pack 32 bytes into a U256 from the source, and then unpack it at the destination.
+    // Copy the next chunk of bytes.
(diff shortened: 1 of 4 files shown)


wborgeaud (approved):
Looks great!

We could also call `memcpy_bytes` [here](https://github.com/0xPolygonZero/plonky2/blob/05006deb44eac3f8dfb26cf7bbd158975c8e9b56/evm/src/cpu/kernel/asm/memory/syscalls.asm#L155) for `{CALLDATA,CODE,RETURNDATA}COPY`.

evm/src/cpu/kernel/asm/memory/memcpy.asm
```diff
@@ -53,3 +53,74 @@ memcpy_finish:
     %jump(memcpy)
 %%after:
 %endmacro
+
+// Similar logic than memcpy, but optimized for copying sequences of bytes.
```
dvdplm, line 57:
```suggestion
// Similar logic to memcpy, but optimized for copying sequences of bytes.
```

dvdplm (approved):

Nashtare:
Good point @wborgeaud! Will do before merging

Nashtare:
@wborgeaud made me realize we could also use this trick instead of calling `memset` when we fill with zeroes (or any 1-byte value in fact). Will do in a follow-up PR.

Nashtare:
Actually @wborgeaud we may have an issue here (I noticed a few tests failing with the updated runner). The issue is that we don't exactly respect the specs of the yellowpaper when it comes to those syscalls, which in turn leads to possible wrong behavior with `mstore_packing`, as the byte_packing  CTL expects byte values. More specifically, following the YP specs (p34):

- `CODECOPY` and `EXTCODECOPY` should stop copying if we go over the actual code length (i.e. specified size on the stack too large) -> we should add a check for that within `%wcopy` (especially if we implement the TODO to have `EXTCODECOPY` rely on `memcpy`). Tests like `codecopy_d0g0v0_Shanghai` fall into this (code of length 20, copying 64 elements).

- `CALLDATACOPY` and `RETURNDATACOPY` should pad with zeroes if we copy over the actual data length -> this may require different handling than the others above (`RETURNDATACOPY` already doesn't use `%wcopy` because of the additional overflow check). An additional call to memset after with the overflowing range?

PR #1305, opened 2023-10-23
Speed-up memset and fix it to write 0 values

Nashtare:
I realized in #1304 that a similar logic could be done with `memset`.
As it is always used to write zeroes, I just replaced the original label to be effective everywhere.

An example on `stPreCompiledContracts/modexp_d0g0v0_Shanghai` test:

before: `{ arithmetic_len: 141396, byte_packing_len: 3644, cpu_len: 899924, keccak_len: 12600, keccak_sponge_len: 525, logic_len: 3782, memory_len: 2447097 }`

after: `{ arithmetic_len: 125741, byte_packing_len: 11724, cpu_len: 740850, keccak_len: 12600, keccak_sponge_len: 525, logic_len: 3782, memory_len: 2028462 }`

i.e. 17.7% reduction in CPU / 17.1% reduction in Memory, pre padding.

### DIFF
--- a/evm/src/cpu/kernel/asm/account_code.asm
+++ b/evm/src/cpu/kernel/asm/account_code.asm
@@ -190,7 +190,7 @@ extcodecopy_large_offset:
     // offset is larger than the code size. So we just have to write zeros.
     // stack: code_size, size, offset, dest_offset, retdest
     GET_CONTEXT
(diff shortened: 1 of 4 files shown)


wborgeaud (approved):
Looks good to me!

PR #1306, opened 2023-10-23
Handle empty case for `memcpy_bytes`

Nashtare:
An oversight from #1304 (that isn't part of the last comment there), but copying empty sequences (either through a dumb call to one of the syscalls, or if the sequence length is a multiple of 32 bytes), would result in CTL failure as the byte_packing table isn't supposed to have empty sequences.

example: `codecopy_d3g0v0_Shanghai.json`


wborgeaud (approved):
Good catch 👍

PR #1308, opened 2023-10-24
Fix some edge cases with code_copy

Nashtare:
More detailed comment in [here](https://github.com/0xPolygonZero/plonky2/pull/1304#issuecomment-1776137373).

The YP specifies that we should stop copying for CODECOPY / EXTCODECOPY (while we currently pad with zeroes using memset when the offset is too large, and we don't check that offset + copy_size can go over the codesize, which can fail if we copy non-byte elements). The current failure on `main` is dues to the `test_extcodecopy` which generates a random code and random offsets / sizes to copy to. I think it would need a rewrite, to handle those specific edge cases manually and have deterministic failure.

@wborgeaud Was there a particular rationale for always padding with zeroes? This should be unrelated to the test failing so I could remove it, though it feels wasteful and it wasn't clear why we would need it. We do need padding with zeroes for CALLDATACOPY and RETURNDATACOPY though, which isn't implemented yet.

### DIFF
--- a/evm/src/cpu/kernel/asm/account_code.asm
+++ b/evm/src/cpu/kernel/asm/account_code.asm
@@ -195,11 +195,11 @@ extcodecopy_end:
     JUMP
 
 extcodecopy_large_offset:
-    // offset is larger than the code size. So we just have to write zeros.
+    // offset is larger than the code size.
+    // The YP specifies we should not do anything outside of the code's bounds. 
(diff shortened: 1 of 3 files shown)


Nashtare:
I've updated the `test_extcodecopy` test to make sure we don't touch memory parts we shouldn't. It's still non-deterministic (though running 10 times as opposed to once).

PR #1310, opened 2023-10-25
Add missing constraints for DUP/SWAP

hratoanina:
Some memory channels weren't disabled for the DUP and SWAP instructions.


LindaGuiga (approved):
Looks good to me, thanks!

PR #1311, opened 2023-10-25
Create Poseidon tables

LindaGuiga:
This PR adds PoseidonSpongeStark and PoseidonStark to enable the use of Poseidon on the STARK side. It is assumed here that if there is a swap in the inputs, it is done in the kernel, prior to calling `POSEIDON_GENERAL`. But if necessary, it should be easy to add a `SWAP` column to enable the swap in the table. Since we read the input elements from `MemoryStark`, we assume that the input elements are 32-bits long at most. @wborgeaud Please let me know what you think of this implementation / if there's anything wrong!


pgebheim:
Hey @LindaGuiga I added @wborgeaud as a reviewer.

evm/src/poseidon/poseidon_stark.rs
```diff
@@ -0,0 +1,597 @@
...
+        let num_rows = (inputs.len()).max(min_rows).next_power_of_two();
+
+        let mut rows = Vec::with_capacity(num_rows);
+        for input in inputs.iter() {
+            let inps_for_perm = input
+                .iter()
+                .map(|&elt| F::from_canonical_u64(elt))
+                .collect::<Vec<F>>()
+                .try_into()
+                .unwrap();
```
Nashtare, line 56: This is erroneous. As we are currently storing the first 64 bits of the U256 read at a given address, we have no guarantee that this in canonical form, and may lead to possible issues.
Though it depends on how data to be hashed may be represented, so pending further discussion.

evm/src/poseidon_sponge/poseidon_sponge_stark.rs
```diff
@@ -0,0 +1,657 @@
...
+            Column::linear_combination([
+                (cols.state_rate_limbs[2 * i], F::ONE),
+                (
+                    cols.state_rate_limbs[2 * i + 1],
+                    F::from_canonical_u64(1 << 32),
+                ),
+            ])
+        })
+        .collect();
+    inputs.extend(Column::singles(&cols.state_capacity));
```
Nashtare, line 57: Similar concern as in PoseidonStark trace generation: if we allow for full 64-bit elements per Poseidon state cell, we may have either:
- undefined behavior if we call `from_canonical_u64`
- collisions if we call `from_noncanonical_u64` (as hashing `[p+1]` and `[1]` would yield the same digest).

evm/src/witness/operation.rs
```diff
@@ -162,6 +165,61 @@ pub(crate) fn generate_keccak_general<F: Field>(
...
+    let mut padded_input = input.clone();
+    if padded_input.len() % POSEIDON_SPONGE_RATE == POSEIDON_SPONGE_RATE - 1 {
+        padded_input.push(1);
+    } else {
+        padded_input.push(1);
+        while (padded_input.len() + 1) % plonky2::hash::poseidon::SPONGE_RATE != 0 {
+            padded_input.push(0);
+        }
+        padded_input.push(1);
+    }
```
Nashtare, line 200: We probably want padding to be done either directly on the kernel side, otherwise we would need 2 distinct instructions: 1 for hashing arbitrarily long data (leaves), and 1 for compressing 2 digests into 1 (i.e. without padding), which cannot happen here without an extra permutation.

evm/src/witness/traces.rs
```diff
@@ -76,6 +84,12 @@ impl<T: Copy> Traces<T> {
                 .iter()
                 .map(|op| op.input.len() / keccak_sponge::columns::KECCAK_RATE_BYTES + 1)
                 .sum(),
+            poseidon_len: self.poseidon_inputs.len(),
+            poseidon_sponge_len: self
+                .poseidon_sponge_ops
+                .iter()
+                .map(|op| op.input.len() / poseidon::columns::POSEIDON_SPONGE_RATE + 1)
```
Nashtare, line 91: nit: we may need to adjust this, as we will *mostly* use it for compression within SMTs, with something like `if op.input.len() == SPONGE_RATE { 1 } else { op.input.len() / poseidon::columns::POSEIDON_SPONGE_RATE + 1 }`

evm/src/poseidon/poseidon_stark.rs
```diff
@@ -0,0 +1,597 @@
...
+            <F as Poseidon>::constant_layer_field(&mut state, round_ctr);
+
+            for i in 0..POSEIDON_SPONGE_WIDTH {
+                // We do not need to store the first full_sbox_0 inputs, since they are
+                // the permutation's inputs.
+                if r != 0 {
+                    row[full_sbox_0(r, i)] = state[i];
+                }
+                // Generate x^3 and x^6 for the SBox layer constraints.
+                row[reg_cubed_full(r, i)] = state[i] * state[i] * state[i];
```
Nashtare, line 94: nit:
```suggestion
                row[reg_cubed_full(r, i)] = state[i].cube();
```

evm/src/poseidon/columns.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const START_POWER_6: usize = POSEIDON_SPONGE_WIDTH + 2 * HALF_N_FULL_ROUNDS * POSEIDON_SPONGE_WIDTH;
+
+/// Holds x^6 for all elements in full rounds.
+pub fn reg_power_6_full(round: usize, i: usize) -> usize {
+    debug_assert!(i < POSEIDON_SPONGE_WIDTH);
+    debug_assert!(round < 2 * HALF_N_FULL_ROUNDS);
+    START_POWER_6 + POSEIDON_SPONGE_WIDTH * round + i
+}
```
Nashtare, line 34: Random thought here, but as `x^6` is only quadratic with respect to `x^3` which is stored previously, can't we amortize the number of columns here by not storing it directly?

Nashtare (commented):
Just skimmed through quickly, not a full review.

One main comment for @wborgeaud: were the SMTs envisioned solely with Poseidon? What about a Keccak at the leaf level / Poseidon on the upper layers? Keccak has the advantage of having a bigger throughput (1088 bits for rate vs 512 at most here, regardless of my security concerns), is much faster natively, and we would still need to keep its table anyway for contract execution. This would also allow the current `generate_poseidon_op` to be specific to node compression, as opposed to being generic, with padding rule and whatnot.

Nashtare:
> Just skimmed through quickly, not a full review.
> 
> One main comment for @wborgeaud: were the SMTs envisioned solely with Poseidon? What about a Keccak at the leaf level / Poseidon on the upper layers? Keccak has the advantage of having a bigger throughput (1088 bits for rate vs 512 at most here, regardless of my security concerns), is much faster natively, and we would still need to keep its table anyway for contract execution. This would also allow the current `generate_poseidon_op` to be specific to node compression, as opposed to being generic, with padding rule and whatnot.

One addition to my comment above, is that "if" we were going down that path, then as we would only use Poseidon for compression, we may want to use a dedicated compression function, with only a width of 8 elements, shortening greatly the PoseidonStark table (ballpark of 100 columns removal?) and speeding up native execution (we could use the fast FFT-friendly size-8 MDS matrix of RPO).
Though if as discussed we wanted to have recursive verification done within the execution, that would require some substantial changes in other places.

pgebheim:
We should take a look at what Hermez chose to do and start with maximal compatibility with their approach, and then make changes from that starting point in collaboration with them.

The mid-term goal past POC is to create an SMT standard that tooling providers can target,which is used by as many provers and chains as possible.

pgebheim:
Also CC: @bobbinth and @mandrigin for viz

wborgeaud:
> One main comment for @wborgeaud: were the SMTs envisioned solely with Poseidon? What about a Keccak at the leaf level / Poseidon on the upper layers? Keccak has the advantage of having a bigger throughput (1088 bits for rate vs 512 at most here, regardless of my security concerns), is much faster natively, and we would still need to keep its table anyway for contract execution. This would also allow the current generate_poseidon_op to be specific to node compression, as opposed to being generic, with padding rule and whatnot.

That's a good idea, but I think a big benefit of having a generic Poseidon implementation would be to use it for code hashes. Computing them with Keccak is pretty expensive and this is something we have to do quite often, like for every call and also for cheap opcodes like EXTCODESIZE. For the latter we could just add the code size to the account data, but not sure it's possible to avoid it for calls.

Nashtare:
> That's a good idea, but I think a big benefit of having a generic Poseidon implementation would be to use it for code hashes. Computing them with Keccak is pretty expensive and this is something we have to do quite often, like for every call and also for cheap opcodes like EXTCODESIZE. For the latter we could just add the code size to the account data, but not sure it's possible to avoid it for calls.

Ah I think our understanding differed here. I assumed we would only use Poseidon for trees with possibly some additional data on account leaves, and would keep the YP specs for the rest. Isn't that what a type-2 should look like?

wborgeaud:
> Ah I think our understanding differed here. I assumed we would only use Poseidon for trees with possibly some additional data on account leaves, and would keep the YP specs for the rest. Isn't that what a type-2 should look like?

Yes it would still be a type2, we would just add a `poseidon_codehash` field to accounts and use it instead to validate the code in calls and EXTCODE{SIZE,COPY,...}.

wborgeaud (commented):
I see that you're using a sponge with XOR, and not the approach we use in Plonky2 (simply set the first `rate` elements to the new input chunk). Is there a reason behind this? The Plonky2 approach wouldn't require using the logic table, and since we do one permutation per row, we could also consider merging the Poseidon and PoseidonSponge tables.

Looks good otherwise!

Nashtare:
I think there may be an issue with the Poseidon table, at least with the current approach.

We're taking 32-bit limbs as input, turned into field elements (canonical) and then obtain 4 `F` elements as output (i.e. 256 bits), and store the equivalent output as 8 (32-bit) limbs on the CPU side and enforce equality through the CTLs.

However because we don't operate on bits/bytes but on field elements, we don't necessarily have unicity of the output with this approach. A prover could tamper with the output limbs on the CPU side, storing the alternative representation of a field element that fits in a u64 (for instance $p+1 = 1$ with limbs $[2, 2^{32} - 1] = [1, 0]$, which lead to inconsistencies (double-hashing of the same initial value could lead to different outcomes).

One way to deal with this would be to enforce that _only_ canonical encodings are considered valid output for the CPU, with a range-check on smaller limbs (16bit?) done to enforce this. Do you have any take on this @wborgeaud?

wborgeaud:
> I think there may be an issue with the Poseidon table, at least with the current approach.
> We're taking 32-bit limbs as input, turned into field elements (canonical) and then obtain 4 `F` elements as output (i.e. 256 bits), and store the equivalent output as 8 (32-bit) limbs on the CPU side and enforce equality through the CTLs.
> However because we don't operate on bits/bytes but on field elements, we don't necessarily have unicity of the output with this approach. A prover could tamper with the output limbs on the CPU side, storing the alternative representation of a field element that fits in a u64 (for instance p+1=1 with limbs [2,232−1]=[1,0], which lead to inconsistencies (double-hashing of the same initial value could lead to different outcomes).

I agree that this is definitely an issue.

> One way to deal with this would be to enforce that _only_ canonical encodings are considered valid output for the CPU, with a range-check on smaller limbs (16bit?) done to enforce this. Do you have any take on this @wborgeaud?

Isn't there a more efficient way to do this range check? Like, we have 
$$2^{32} x + y \geq p \iff x = 2^{32}-1 \wedge y>0.$$ 
So given two 32-bit limbs $x,y$ we could check this with the constraint $(1-(2^{32}-x)z)y=0$.

LindaGuiga:
@wborgeaud  Thank you! That does sound like a better idea to deal with the issue indeed!

LindaGuiga:
Thank you for your comments!! 
Since I removed `PoseidonSpongeStark`, there are quite a few changes to the PR, so I'd really appreciate it if you could review this again @wborgeaud @Nashtare!

evm/src/poseidon/poseidon_stark.rs
```diff
@@ -0,0 +1,960 @@
...
+            tmp_row
+        }
+        .into();
+        while rows.len() < num_rows {
+            rows.push(padding_row);
+        }
+        rows
+    }
+
+    fn generate_rows_for_op(&self, op: PoseidonOp) -> Vec<[F; NUM_COLUMNS]> {
```
wborgeaud, line 144: nit: Could just return an iterator of `[F; NUM_COLUMNS]` instead of allocating a vector.

evm/src/poseidon/poseidon_stark.rs
```diff
@@ -0,0 +1,960 @@
...
+
+        Self::generate_commons(&mut row, input, op, already_absorbed_elements);
+        row
+    }
+
+    fn generate_perm(row: &mut PoseidonColumnsView<F>, input: [F; POSEIDON_SPONGE_WIDTH]) {
+        // Populate the round input for the first round.
+        row.input.copy_from_slice(&input);
+
+        let mut state: [F; POSEIDON_SPONGE_WIDTH] = input;
```
wborgeaud, line 224: nit: Is the type hint necessary?

evm/src/witness/operation.rs
```diff
@@ -155,6 +159,74 @@ pub(crate) fn generate_keccak_general<F: Field>(
...
+                state.traces.clock(),
+                address,
+                MemoryOpKind::Read,
+                val.0[0].into(),
+            ));
+
+            val.0[0] as u32
+        })
+        .collect_vec();
+    log::debug!("Hashing {:?}", input);
```
wborgeaud, line 190: nit: Maybe change to "Poseidon hashing {:?}" to differentiate with `generate_keccak_general`.

wborgeaud (approved):
Looks great, thanks!

evm/src/poseidon/poseidon_stark.rs
LindaGuiga: I tried changing this to return an iterator, but I got an error because we are taking the inputs by reference (and therefore the compiler wants me to collect the `map` into a vector anyway)... I might be doing it wrong so please let me know if you had way of doing this in mind. I've left this as is for now since it's done the same way in `KeccakSpongeStark`.

wborgeaud: Ah ok, no problem then! Shouldn't be an issue.

PR #1312, opened 2023-10-26
Implement EIP-5656 (MCOPY)

### LINKED ISSUE #1267: Support MCOPY EIP-5656
Make opcode public

Nashtare:
Relies on a similar logic than the `wcopy`-related syscalls (and hence may benefit partly from their incoming refactoring).

@wborgeaud apologies for the harassment the past few days 🙏🏼 This isn't targeting main, so no need to deal with it urgently!

closes #1267

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/syscalls.asm
+++ b/evm/src/cpu/kernel/asm/memory/syscalls.asm
@@ -221,20 +221,20 @@ global sys_mcopy:
     // stack: Gverylow, kexit_info, dest_offset, offset, size
     DUP5 %num_bytes_to_num_words %mul_const(@GAS_COPY) ADD %charge_gas
 
+    // stack: kexit_info, dest_offset, offset, size
+    DUP4
+    // stack: size, kexit_info, dest_offset, offset, size
+    ISZERO %jumpi(returndatacopy_empty) // If size is empty, just pop the stack and exit the kernel
+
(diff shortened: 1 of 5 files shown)


evm/src/cpu/kernel/asm/memory/syscalls.asm
```diff
@@ -213,3 +213,67 @@ global sys_returndatacopy:
...
+    %stack (kexit_info, dest_offset, offset, size) -> (dest_offset, size, kexit_info, dest_offset, offset, size)
+    %add_or_fault
+    // stack: expanded_num_bytes, kexit_info, dest_offset, offset, size, kexit_info
+    DUP1 %ensure_reasonable_offset
+    %update_mem_bytes
+
+    // stack:  kexit_info, dest_offset, offset, size
+    DUP4
+    // stack:  size, kexit_info, dest_offset, offset, size
+    ISZERO %jumpi(returndatacopy_empty) // If size is empty, just pop the stack and exit the kernel
```
hratoanina, line 233: Very nit: can't we do the check before verifying that the offset is reasonable?

hratoanina (approved):

evm/src/cpu/kernel/asm/memory/syscalls.asm
Nashtare: Yes! I had originally copied the `returndatacopy` code which has it also there, but it can be moved before

PR #1314, opened 2023-10-26
Refactor `%wcopy` and unify {ext}codecopy related code

Nashtare:
cf discussion in here: https://github.com/0xPolygonZero/plonky2/pull/1309/files#r1371598536

Moved `extcodecopy` with the other `wcopy` related syscalls, though I could leave it in `account_code.asm` but would need to have all internal labels defined globally.

Below a non fully exhaustive list of tests for the related syscalls

Screenshot from 2023-10-26 16-16-32


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1126,6 +1126,25 @@ impl<'a> Interpreter<'a> {
         }
     }
 
+    fn run_exit_kernel(&mut self) {
+        let _dummy = self.pop(); // not sure why we need this extra pop on the interpreter
```
wborgeaud, line 1130: Hm are you sure about that? This is what I get when running `test_extcodecopy` with some `dbg!`s
```
[evm/src/cpu/kernel/interpreter.rs:1130] self.stack() = [
    3735928559,
    18191040829150600853596117748455817273864626078056772018372608,
]
[evm/src/cpu/kernel/interpreter.rs:1133] _dummy = 18191040829150600853596117748455817273864626078056772018372608
[evm/src/cpu/kernel/interpreter.rs:1133] kexit_info = 3735928559
```
It looks like with this extra pop, `kexit_info=0xdeadbeef` and `_dummy` holds the actual kernel exit info.

wborgeaud (approved):
Very clean 👍

evm/src/cpu/kernel/interpreter.rs
Nashtare: Yeah I probably messed it up, but only popping the stack gives me a
```asm
At jumped_to_0
Error: Executed PANIC, stack=[3735928559], memory=[150, 235, 124, 56, 22, 20, 241, 50, 222, 39, 71, 97, 32, 238, 205, 48, 235, 96, 223, 187]
```
which I don't really understand.

Nashtare: Ah nevermind, it's because I set a dummy value for kexit info which messed up the pc in exit kernel.

PR #1315, opened 2023-10-27
Start of SMT implementation for Type 2 zkEVM

wborgeaud:
Replace the state MPT with an SMT using Keccak (for now).

A notable difference with the MPT implementation is that we pass the serialized SMT to `TrieInputs` (instead of a `HashedPartialTrie` for MPTs). That way we can directly copy the SMT data to memory without any additional logic.

TODO in future PRs:
- Implement deletion for the SMT
- Replace the transactions and receipts MPTs with SMTs.

Benchmarks: In `test_log_with_aggreg`, hashing the state SMT takes 1617 CPU cycles, whereas hashing the state MPT takes 12037 CPU cycles.
Will try to add more comparison data.


Nashtare (commented):
this isn't a full review, I'd need to also look more closely at the new dependency for SMTs, but so far so good!

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -52,19 +51,18 @@ fn prepare_interpreter(
...
-    ));
-    // Next, execute mpt_insert_state_trie.
-    interpreter.generation_state.registers.program_counter = mpt_insert_state_trie;
+    let k = keccak(address.to_fixed_bytes());
+    // Next, execute smt_insert_state.
+    interpreter.generation_state.registers.program_counter = smt_insert_state;
     let trie_data = interpreter.get_trie_data_mut();
     if trie_data.is_empty() {
-        // In the assembly we skip over 0, knowing trie_data[0] = 0 by default.
+        // In the assembly we skip over 0, knowing trie_data[0:4] = 0 by default.
```
LindaGuiga, line 59: I might be missing something, but isn't it only the first 2 elements that are set to 0 by default? (Looking at `load_state_smt`, it seems that we get the trie data size, add 2 for an empty hash node, and then store the trie root)

evm/src/cpu/kernel/tests/balance.rs
```diff
@@ -43,19 +42,18 @@ fn prepare_interpreter(
...
-    ));
-    // Next, execute mpt_insert_state_trie.
-    interpreter.generation_state.registers.program_counter = mpt_insert_state_trie;
+    let k = keccak(address.to_fixed_bytes());
+    // Next, execute smt_insert_state.
+    interpreter.generation_state.registers.program_counter = smt_insert_state;
     let trie_data = interpreter.get_trie_data_mut();
     if trie_data.is_empty() {
-        // In the assembly we skip over 0, knowing trie_data[0] = 0 by default.
+        // In the assembly we skip over 0, knowing trie_data[0:4] = 0 by default.
```
LindaGuiga, line 50: Same question as in another comment about the elements set to 0 by default.

evm/src/cpu/kernel/asm/smt/hash.asm
```diff
@@ -0,0 +1,171 @@
...
+    %mstore_unpacking POP
+
+    %stack (storage_root) -> (0, @SEGMENT_KERNEL_GENERAL, 40, storage_root, 32)
+    %mstore_unpacking POP
+
+    %stack (balance) -> (0, @SEGMENT_KERNEL_GENERAL, 8, balance, 32)
+    %mstore_unpacking POP
+
+    %stack (nonce) -> (0, @SEGMENT_KERNEL_GENERAL, 0, nonce)
+    %mstore_unpacking_u64_LE
```
Nashtare, line 130: nit: possible optimization, probably out of scope of this PR, but couldn't we reorder the way we do the storing here? As `mstore_unpacking` outputs the updated offset, there may be a way to leverage this as it would directly be at the top of the stack followed by the next item to write.

LindaGuiga (commented):
Looks great, thank you! I just have a few nits and some questions!

evm/src/cpu/kernel/asm/smt/read.asm
```diff
@@ -0,0 +1,81 @@
...
+    %jump(smt_read_state)
+%%after:
+%endmacro
+
+// Return the data at the given key in the SMT at `trie_data[node_ptr]`.
+// Pseudocode:
+// ```
+// read( HashNode { h }, key) = if h == 0 then 0 else PANIC
+// read( InternalNode { left, right }, key) = if key&1 { read( right, key>>1) } else { read( left, key>>1) }
+// read( Leaf { key', value_ptr }, key) = if key == key' then value_ptr' else 0
```
Nashtare, line 24: very nit: 
```suggestion
// read( HashNode { h }, key ) = if h == 0 then 0 else PANIC
// read( InternalNode { left, right }, key ) = if key&1 { read( right, key>>1 ) } else { read( left, key>>1 ) }
// read( Leaf { key', value_ptr }, key) = if key == key' then value_ptr' else 0
```

Nashtare (commented):
I noticed Linda's review, so I deleted my few redundant comments. It looks good to me, though I have the same concern about using 32B keys instead of 20B ones.

evm/src/cpu/kernel/asm/mpt/accounts.asm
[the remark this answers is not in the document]
wborgeaud: Good point. Done.

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -270,3 +277,6 @@ const CODE_SIZE_LIMIT: [(&str, u64); 3] = [
 
 const MAX_NONCE: (&str, u64) = ("MAX_NONCE", 0xffffffffffffffff);
 const CALL_STACK_LIMIT: (&str, u64) = ("CALL_STACK_LIMIT", 1024);
+
+// Holds a flag that is set to 1 when hashing storage SMTs. Used in `smt_hash`.
+const SMT_IS_STORAGE: (&str, u64) = ("SMT_IS_STORAGE", 13371337);
```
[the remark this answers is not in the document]
wborgeaud: It's mostly random, but chosen to be large enough to not interfere with other uses of `KernelGeneral` (it shouldn't matter in theory, but just to be safe).

evm/src/cpu/kernel/asm/smt/hash.asm
```diff
@@ -0,0 +1,171 @@
...
+    // stack: retdest
+    PUSH 0 %mstore_kernel_general(@SMT_IS_STORAGE) // is_storage flag.
+    %mload_global_metadata(@GLOBAL_METADATA_STATE_TRIE_ROOT)
+
+// Root hash of SMT stored at `trie_data[ptr]`.
+// Pseudocode:
+// ```
+// hash( HashNode { h } ) = h
+// hash( InternalNode { left, right } ) = keccak(1 || hash(left) || hash(right))
+// hash( Leaf { key, val_hash } ) = keccak(0 || key || val_hash)
```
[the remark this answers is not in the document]
wborgeaud: Only for domain separation. I wasn't aware of this trick, that's nice! I added a comment to use this when we switch to Poseidon.

evm/src/cpu/kernel/asm/smt/insert.asm
```diff
@@ -0,0 +1,106 @@
...
+    JUMP
+
+// Insert a key-value pair in the SMT at `trie_data[node_ptr]`.
+// `value_ptr`  should point to a an empty slot reserved for `rem_key`, followed by the actual value.
+// Pseudocode:
+// ```
+// insert( HashNode { h }, key, value_ptr ) = if h == 0 then Leaf { key, value_ptr } else PANIC
+// insert( InternalNode { left, right }, key, value_ptr ) = if key&1 { insert( right, key>>1, value_ptr ) } else { insert( left, key>>1, value_ptr ) }
+// insert( Leaf { key', value_ptr' }, key, value_ptr ) =
+// with internal = new InternalNode { insert(internal, key', value_ptr'), insert(internal, key, value_ptr), return internal }
```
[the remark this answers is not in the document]
wborgeaud: I reworded this comment to make it clearer.
Basically, when inserting a leaf L where there is already another leaf L', we replace L' with a new internal node, and then reinsert L' and insert L.

evm/src/cpu/kernel/tests/account_code.rs
wborgeaud: Ah this is because I wasn't calling `interpreter.generation_state.state_smt_prover_inputs = state_smt_prover_inputs_reversed(&trie_inputs);`. I've added it and now I don't need to manually extend the trie data.

evm/src/cpu/kernel/asm/smt/hash.asm
```diff
@@ -0,0 +1,171 @@
...
+    KECCAK_GENERAL
+    // stack: hash, key, retdest
+
+    // Leaf flag
+    PUSH 0 %mstore_kernel_general(0)
+
+    %stack (hash) -> (0, @SEGMENT_KERNEL_GENERAL, 33, hash, 32)
+    %mstore_unpacking POP
+
+    %stack (key) -> (0, @SEGMENT_KERNEL_GENERAL, 1, key, 32)
```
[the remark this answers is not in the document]
wborgeaud: I'm not sure there's much benefit in truncating the key once it's hashed. (I discuss why it's hashed in the comment below)

evm/src/cpu/kernel/asm/smt/hash.asm
wborgeaud: Good point. I added a TODO.

LindaGuiga (approved):

PR #1316, opened 2023-10-30
Move empty check inside final iteration

Nashtare:
There are currently 2 checks on each loop iteration of `memcpy_bytes`:
- check whether the sequence is empty (needed as `BytePackingStark` expects non-empty sequences)
- check whether the sequence is smaller than 32 bytes (to do a final packing/unpacking combo and exit)

The latter can contain the former case, so that we don't need to check twice on every iteration.


LindaGuiga (approved):
Looks good to me!

PR #1318, opened 2023-10-30
Constrain uninitialized memory to 0

hratoanina:
Missing constraints in the Memory Stark.
We do not check that memory is initialized at 0, that is to say that a read at a new address without a prior write must have a value of 0.
An immediate exploit is to bypass JUMPDEST analysis in case of a jump outside of the contract code, and subsequently arbitrary code execution in the uninitialized memory. In https://github.com/topos-protocol/plonky2/tree/code_injection, the code of the contract is a jump to address 0x80 which should raise an invalid jump exception, but the transaction is accepted and the proof is valid.


hratoanina:
@nbgl Assigning you, since I presumed you'd be interested in this exploit!

nbgl (approved):
Looks good!!

PR #1321, opened 2023-11-02
Add test for selfdestruct

wborgeaud:
Mainly to have a test that performs an account deletion.


pgebheim:
Added to Rc1 for tracking

PR #1322, opened 2023-11-03
Add withdrawals

wborgeaud:
Add withdrawals as a public input array of `address, amount`. See EIP-4895.

### DIFF
--- a/evm/tests/withdrawals.rs
+++ b/evm/tests/withdrawals.rs
@@ -11,10 +11,11 @@ use plonky2::plonk::config::PoseidonGoldilocksConfig;
 use plonky2::util::timing::TimingTree;
 use plonky2_evm::all_stark::AllStark;
 use plonky2_evm::config::StarkConfig;
-use plonky2_evm::fixed_recursive_verifier::AllRecursiveCircuits;
 use plonky2_evm::generation::mpt::AccountRlp;
 use plonky2_evm::generation::{GenerationInputs, TrieInputs};
 use plonky2_evm::proof::{BlockHashes, BlockMetadata, TrieRoots};
+use plonky2_evm::prover::prove;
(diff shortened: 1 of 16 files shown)


pgebheim:
Adding to RC1

evm/tests/withdrawals.rs
```diff
@@ -0,0 +1,104 @@
...
+        &all_stark,
+        &[16..17, 10..11, 15..16, 14..15, 9..10, 12..13, 18..19], // Minimal ranges to prove an empty list
+        &config,
+    );
+
+    let mut timing = TimingTree::new("prove", log::Level::Info);
+    let (root_proof, _public_values) =
+        all_circuits.prove_root(&all_stark, &config, inputs, &mut timing)?;
+    timing.filter(Duration::from_millis(100)).print();
+    all_circuits.verify_root(root_proof.clone())
```
Nashtare, line 99: Non-blocking, but is there any reason why we're using AllRecursiveCircuits here? And if so, wouldn't it deprecate `empty_txn_list`? As this one is run in the CI, we could combine the root_proof with a dummy one to still showcase `prove_aggreg` / `prove_block` in the CI.

Nashtare (approved):
Thanks @wborgeaud, looks good!

Just for my understanding, will we want at some point to include the `withdrawals_root` in the `BlockMetadata`? In which case we'd need to change the array layout to another MPT.

evm/tests/withdrawals.rs
wborgeaud: No reason really, I removed it.

wborgeaud:
> Just for my understanding, will we want at some point to include the withdrawals_root in the BlockMetadata? In which case we'd need to change the array layout to another MPT.

Yes for sure! This is just a quick and dirty fix to be able to prove L1 blocks.

PR #1323, opened 2023-11-03
SMT deletion

wborgeaud:


pgebheim:
It looks like this needs to be rebased on feat/type2

wborgeaud:
Cleaned in #1330

PR #1327, opened 2023-11-03
Implement EIP-6780 (new SELFDESTRUCT)

hratoanina:
Tracking issue: https://github.com/0xPolygonZero/plonky2/issues/1268

Be advised that some EVM tests don't pass. However, I believe that the tests aren't correct.
Out of select tests (everything [in here ](https://github.com/topos-protocol/evm-tests-suite-parsed/tree/selfdestruct/selfDestruct/GeneralStateTests) except in the Cancun folder, which deals with MEMCOPY tests), only three don't pass:
`doubleSelfdestructTest_d4g0v0_Cancun`, `doubleSelfdestructTest_d5g0v0_Cancun` and `doubleSelfdestructTest_d6g0v0_Cancun`. 
An issue was opened here: https://github.com/ethereum/tests/issues/1318

### DIFF
--- a/evm/src/cpu/kernel/asm/core/selfdestruct_list.asm
+++ b/evm/src/cpu/kernel/asm/core/selfdestruct_list.asm
@@ -12,8 +12,43 @@
     %mstore_global_metadata(@GLOBAL_METADATA_SELFDESTRUCT_LIST_LEN) // Store new length.
 %endmacro
 
+global maybe_insert_selfdestruct_list:
+    // stack: addr, retdest
+    %mload_global_metadata(@GLOBAL_METADATA_CREATED_CONTRACTS_LEN)
+    // stack: nb_created_contracts, addr, retdest
+    PUSH 0
+maybe_insert_selfdestruct_list_loop:
+    %stack (i, nb_created_contracts, addr, retdest) -> (i, nb_created_contracts, i, nb_created_contracts, addr, retdest)
+    EQ %jumpi(contract_not_created)
+    // stack: i, nb_created_contracts, addr, retdest
+    DUP1 %mload_kernel(@SEGMENT_CREATED_CONTRACTS)
+    // stack: addr_created_contract, i, nb_created_contracts, addr, retdest
+    DUP4
+    // stack: addr, addr_created_contract, i, nb_created_contracts, addr, retdest
+    EQ %jumpi(contract_just_created)
+    // stack: i, nb_created_contracts, addr, retdest
+    %increment
+    %jump(maybe_insert_selfdestruct_list_loop)
+contract_just_created:
+    // stack: i, nb_created_contracts, addr, retdest
+    %pop2
+    // stack: addr, retdest
+    %insert_selfdestruct_list
+    // stack: retdest
+    JUMP
+contract_not_created:
+    // stack: i, nb_created_contracts, addr, retdest
+    %pop3
+    JUMP
+
+%macro maybe_insert_selfdestruct_list
+    %stack (addr) -> (addr, %%after)
+    %jump(maybe_insert_selfdestruct_list)
+%%after:
+%endmacro 
+
 /// Remove one occurrence of the address from the list.
-/// Panics if the address is not in the list.
+/// No effect if the address is not in the list.
 global remove_selfdestruct_list:
     // stack: addr, retdest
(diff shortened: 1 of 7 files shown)


evm/src/cpu/kernel/asm/core/selfdestruct_list.asm
```diff
@@ -12,16 +12,51 @@
     %mstore_global_metadata(@GLOBAL_METADATA_SELFDESTRUCT_LIST_LEN) // Store new length.
 %endmacro
 
+global maybe_insert_selfdestruct_list:
```
wborgeaud, line 15: nit: A comment would be useful here.

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -81,9 +81,10 @@ global sys_selfdestruct:
     %charge_gas
     %stack (kexit_info, balance, address, recipient) -> (balance, address, recipient, kexit_info)
 
-    // Insert address into the selfdestruct set.
+    // EIP-6780: insert address into the selfdestruct set if contact has been created
```
wborgeaud, line 84:
```suggestion
    // EIP-6780: insert address into the selfdestruct set if contract has been created
```

evm/src/cpu/kernel/asm/core/selfdestruct_list.asm
```diff
@@ -40,6 +75,10 @@ remove_selfdestruct_list_found:
     SWAP1
     %mstore_kernel(@SEGMENT_SELFDESTRUCT_LIST) // Store the last address at the position of the removed address.
     JUMP
+remove_self_dstruct_not_found:
```
wborgeaud, line 78:
```suggestion
remove_selfdestruct_not_found:
```

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -95,14 +96,9 @@ global sys_selfdestruct:
...
     %mstore_trie_data
 
+    // Send the balance to the recipient. 
     %stack (balance, address, recipient, kexit_info) ->
-        (address, recipient, address, recipient, balance, kexit_info)
-
-    // If the recipient is the same as the address, then we're done.
-    // Otherwise, send the balance to the recipient.
-    // stack: address, recipient, address, recipient, balance, kexit_info
-    EQ %jumpi(sys_selfdestruct_journal_add)
```
wborgeaud, line 104: Don't we still need this in case we're in a selfdestruct of a contract that was created in the same txn?
Looking at this in the EIP
image

wborgeaud (approved):
Looks good to me!

evm/src/cpu/kernel/asm/core/terminate.asm
hratoanina: If we're in that case, the account will be in the selfdestruct list and will be deleted anyway, which erases all of the node data (oncluding its balance).
If we keep these lines, there will be a discrepancy in case of revertion.

wborgeaud: Right, but what if we call `BALANCE` on this address before the end of the tx?

hratoanina: Oops, you're right! I'll have to put this `add_eth` inside the "has_been_created" logic in `maybe_insert_selfdestruct_list`.

hratoanina:
@wborgeaud I updated the logic, could you take another look please?

evm/src/cpu/kernel/asm/core/terminate.asm
```diff
@@ -205,3 +223,41 @@ global terminate_common:
...
+// Returns 1 if the address is present in SEGMENT_CREATED_CONTRACTS, meaning that it has been
+// created during this transaction. Returns 0 otherwise.
+// Pre stack: addr
+// Post stack: is_just_created
+%macro contract_just_created
+    // stack: addr
+    %mload_global_metadata(@GLOBAL_METADATA_CREATED_CONTRACTS_LEN)
+    // stack: nb_created_contracts, addr
+    PUSH 0
+%%maybe_insert_selfdestruct_list_loop:
```
wborgeaud, line 239: nit: This name doesn't really make sense since this macro doesn't insert in the selfdestruct list.

wborgeaud (approved):
Looks good to me!

PR #1328, opened 2023-11-03
Implement out of gas exception

hratoanina:
Closes https://github.com/0xPolygonZero/plonky2/issues/1298

It's only been tested with William's (corrected) infinite loop test, but it passes fine.

One issue is the `gas_cost_for_opcode` table in the kernel. It should use the gas constants (`@GAS_VERYLOW` etc.) but apparently the parser doesn't like anything else than literals after the `BYTES` keyword. @wborgeaud Would you happen to know how to get around it?

### DIFF
--- a/evm/src/cpu/kernel/parser.rs
+++ b/evm/src/cpu/kernel/parser.rs
@@ -4,7 +4,7 @@ use ethereum_types::U256;
 use pest::iterators::Pair;
 use pest::Parser;
 
-use super::ast::StackPlaceholder;
+use super::ast::{BytesTarget, StackPlaceholder};
 use crate::cpu::kernel::ast::{File, Item, PushTarget, StackReplacement};
 
 /// Parses EVM assembly code.
@@ -38,7 +38,7 @@ fn parse_item(item: Pair<Rule>) -> Item {
         Rule::macro_label_decl => {
             Item::MacroLabelDeclaration(item.into_inner().next().unwrap().as_str().into())
         }
-        Rule::bytes_item => Item::Bytes(item.into_inner().map(parse_literal_u8).collect()),
+        Rule::bytes_item => Item::Bytes(item.into_inner().map(parse_bytes_target).collect()),
         Rule::jumptable_item => {
             Item::Jumptable(item.into_inner().map(|i| i.as_str().into()).collect())
         }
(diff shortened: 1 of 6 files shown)


Nashtare:
@pgebheim I've added it to RC1 as per Monday's discussion.

wborgeaud:
> One issue is the `gas_cost_for_opcode` table in the kernel. It should use the gas constants (`@GAS_VERYLOW` etc.) but apparently the parser doesn't like anything else than literals after the `BYTES` keyword. @wborgeaud Would you happen to know how to get around it?

You could do something like this in `evm_asm.pest` and add a `fn parse_bytes_target()`.
```
bytes_item = { ^"BYTES " ~ bytes_target ~ ("," ~ bytes_target)* }
bytes_target = { literal | constant }
```

hratoanina:
Sorry about the force-push, I messed up the previous commit.
The opcode gas table now uses gas constants. I'm really not familiar with the parser logic, so please tell me if there's anything wrong.

evm/src/cpu/kernel/parser.rs
```diff
@@ -167,6 +167,16 @@ fn parse_push_target(target: Pair<Rule>) -> PushTarget {
     }
 }
 
+fn parse_bytes_target(target: Pair<Rule>) -> BytesTarget {
+    assert_eq!(target.as_rule(), Rule::bytes_target);
+    let inner = target.into_inner().next().unwrap();
```
dvdplm, line 172:
```suggestion
    let inner = target.into_inner().next().expect("CANNOT FAIL BECAUSE …");
```

evm/src/cpu/kernel/parser.rs
```diff
@@ -167,6 +167,16 @@ fn parse_push_target(target: Pair<Rule>) -> PushTarget {
...
+    match inner.as_rule() {
+        Rule::literal => BytesTarget::Literal(parse_literal_u8(inner)),
+        Rule::constant => BytesTarget::Constant(inner.into_inner().next().unwrap().as_str().into()),
```
dvdplm, line 175:
```suggestion
        Rule::constant => BytesTarget::Constant(inner.into_inner().next().expect("CANNOT FAIL BECAUSE …").as_str().into()),
```

evm/src/cpu/kernel/parser.rs
```diff
@@ -167,6 +167,16 @@ fn parse_push_target(target: Pair<Rule>) -> PushTarget {
...
+        _ => panic!("Unexpected {:?}", inner.as_rule()),
```
dvdplm, line 176:
```suggestion
        _ => panic!("Unexpected rule {:?}", inner.as_rule()),
```

wborgeaud (approved):
Looks good to me!

PR #1329, opened 2023-11-04
Fix typos in comments

shuoer86:


evm/src/cpu/stack.rs
```diff
@@ -75,7 +75,7 @@ pub(crate) const JUMPDEST_OP: StackBehavior = StackBehavior {
 
 // AUDITORS: If the value below is `None`, then the operation must be manually checked to ensure
 // that every general-purpose memory channel is either disabled or has its read flag and address
-// propertly constrained. The same applies  when `disable_other_channels` is set to `false`,
+// property constrained. The same applies  when `disable_other_channels` is set to `false`,
```
Nashtare, line 78:
```suggestion
// properly constrained. The same applies  when `disable_other_channels` is set to `false`,
```

Nashtare (commented):
Thanks! There's one change that needs a fix

evm/src/cpu/stack.rs
shuoer86: changed. thank you for your review.

shuoer86:
@Nashtare changed. Thank you for your review.

Nashtare (approved):

PR #1333, opened 2023-11-06
Fix storage SMTs

wborgeaud:


wborgeaud:
Opening another PR superseding this one.

PR #1334, opened 2023-11-06
Remove `len` column in `KeccakSpongeStark`

LindaGuiga:
We noticed that in `KeccakSpongeStark`, we are storing the `len` in an extra column and check that the final number of absorbed bytes is equal to the length, and that the length corresponds to the value in `CpuStark`. 
This should actually be doable without storing the length in `KeccakSpongeStark`: we can directly reconstruct it using `already_absorbed_bytes` and `is_final_input_len`, and then compare that linear combination to the value in `CpuStark`. 
This results in one less column in `KeccakSpongeStark` and fewer constraints in this PR.

### DIFF
--- a/evm/src/keccak_sponge/keccak_sponge_stark.rs
+++ b/evm/src/keccak_sponge/keccak_sponge_stark.rs
@@ -1,5 +1,5 @@
 use std::borrow::Borrow;
-use std::iter::{once, repeat};
+use std::iter::{self, once, repeat};
 use std::marker::PhantomData;
 use std::mem::size_of;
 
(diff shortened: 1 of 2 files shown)


evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -41,15 +41,23 @@ pub(crate) fn ctl_looked_data<F: Field>() -> Vec<Column<F>> {
...
-    Column::singles([
-        cols.context,
-        cols.segment,
-        cols.virt,
-        cols.len,
-        cols.timestamp,
-    ])
-    .chain(outputs)
-    .collect()
+    // The length of the inputs is already_absorbed_byres + is_final_input_len.
```
Nashtare, line 44:
```suggestion
    // The length of the inputs is `already_absorbed_bytes + is_final_input_len`.
```

Nashtare (approved):
LGTM, thanks! Just a minor nit in a comment.

PR #1335, opened 2023-11-06
restore `no-std` support

muraca:
`no-std` support for `plonky2` and `starky` was added by #806, but nobody noticed some breaking changes later on.
This PR restores `no-std` support, and introduces `wasm32-unknown-unknown` as an additional target for tests, to avoid making the same mistake again.

https://github.com/0xPolygonZero/plonky2/pull/907#issuecomment-1462939316

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -29,6 +29,7 @@ jobs:
           profile: minimal
           toolchain: nightly
           target: ${{ matrix.target }}
+          default: true
           override: true
 
       - name: rust-cache
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -10,11 +10,14 @@ on:
     branches:
       - "**"
 
-jobs:  
+jobs:
   test:
     name: Test Suite
     runs-on: ubuntu-latest
     if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
+    strategy:
+      matrix:
+        target: [x86_64-unknown-linux-gnu, wasm32-unknown-unknown]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
@@ -25,6 +28,7 @@ jobs:
         with:
           profile: minimal
           toolchain: nightly
+          target: ${{ matrix.target }}
           override: true
 
       - name: rust-cache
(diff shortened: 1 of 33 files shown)


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -10,11 +10,14 @@ on:
...
 
-jobs:  
+jobs:
   test:
     name: Test Suite
     runs-on: ubuntu-latest
     if: "! contains(toJSON(github.event.commits.*.message), '[skip-ci]')"
+    strategy:
+      matrix:
+        target: [x86_64-unknown-linux-gnu, wasm32-unknown-unknown]
```
Nashtare, line 20: Is this doing what's expected here?
I'd assume both targets to be used for the runs of `cargo check` / `cargo test` below, but if so, `cargo check` for the `plonky2_evm` crate should fail because this is definitely not `no_std`.

muraca: I was aware of the `plonky2_evm` but it was late and I wanted to run the CI first, to see if the matrix was working as expected first. It is not.

Nashtare:
The CI isn't running with your changes: https://github.com/0xPolygonZero/plonky2/actions/runs/6782203615.
You may want to try it out locally against your own fork first.

Nashtare:
In addition, `cargo test` cannot be run against a `no-std` target as it also runs `plonky2_evm` tests which use `std`.

Nashtare:
This doesn't seem to be doing what's expected still.

image

If we were running against a `wasm32-unknown-unknown` target, the current would fail as we didn't deactivate the default features which include `std`.

I'd recommend you to update the CI `until` you have a failing job for wasm complaining about std, and only then remove the default features to have it go green.

Nashtare:
Thanks @muraca! Tiny final nit, the job is called "Test suite" but we do not run the tests against wasm, so a different name would be preferable.

muraca:
@Nashtare sorry for the mess. 
Looking back at it, I think it would be better to have separate jobs instead of the matrix, since we only have to perform 2 additional checks.
What do you think?

Nashtare:
@muraca Yeah, having a dedicated job for wasm32 seems better :)

Nashtare:
Hmm, the current command run in the CI job for wasm, `cargo check --manifest-path plonky2/Cargo.toml --target wasm32-unknown-unknown` doesn't trigger any error on main, although it should if not deactivating the `default-features`. I'll need to investigate it a bit more.

Nashtare:
@muraca Could you add `--no-default-features` to the `no-std` check? This is what's currently missing to enforce that we can properly compile in no-std environments (it fails on `main` without your fixes). Thanks!

Nashtare (approved):
Looks good! Thank you!

PR #1339, opened 2023-11-07
Implement EIP 4788 (Beacon block root)

### LINKED ISSUE #1272: Support EIP-4788: Beacon block root in the EVM
For Cancun upgrade.

Nashtare:
closes #1272

### DIFF
--- a/evm/src/proof.rs
+++ b/evm/src/proof.rs
@@ -109,6 +109,8 @@ pub struct BlockMetadata {
     pub block_gas_used: U256,
     /// The blob base fee. It must fit in a `u64`.
     pub block_blob_base_fee: U256,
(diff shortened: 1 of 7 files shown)


wborgeaud (commented):
My understanding from the EIP is that we also need to update the beacon roots contract with the new beacon block root at the beginning of every block.
So for us we could do it in `main.asm` under the condition that `TxnNumberBefore=0`.

wborgeaud:
> Ah my bad, I had assumed this would be one of those things were we'd just provide it as is but not any check. I'll update it!

I think that if we don't change the storage of the beacon roots contract, we won't be able to verify the new state trie after the block, so I don't think there's a way to avoid it.

PR #1340, opened 2023-11-07
minor: Move empty check of `memset` inside final iteration

Nashtare:
Same trick as in `memcpy`.
Additionally, performs `LT` against 33 instead of 32, as otherwise for a chunk of exactly 32 bytes, we'd go through the loop twice.


hratoanina (approved):
LGTM

PR #1341, opened 2023-11-08
Remove logic for multiple txns at once

Nashtare:
Following discussions on not supporting txn grouping in a single proof for now, this removes the corresponding logic. Now the prover only processes 0 (empty proofs) or a single txn.


evm/src/generation/state.rs
```diff
@@ -64,7 +62,8 @@ impl<F: Field> GenerationState<F> {
         log::debug!("Input storage_tries: {:?}", &inputs.tries.storage_tries);
         log::debug!("Input contract_code: {:?}", &inputs.contract_code);
         let mpt_prover_inputs = all_mpt_prover_inputs_reversed(&inputs.tries)?;
-        let rlp_prover_inputs = all_rlp_prover_inputs_reversed(&inputs.signed_txns);
+        let rlp_prover_inputs =
+            all_rlp_prover_inputs_reversed(&inputs.signed_txn.clone().unwrap_or_default());
```
wborgeaud, line 66: nit: Could avoid the clone with something like
```rust
all_rlp_prover_inputs_reversed(inputs.signed_txn.as_ref().unwrap_or(&vec![]));
```
or 
```rust
let rlp_prover_inputs = inputs
    .signed_txn
    .as_ref()
    .map(|tx| all_rlp_prover_inputs_reversed(tx))
    .unwrap_or_default();
```
```

wborgeaud (approved):
Looks good to me!

PR #1342, opened 2023-11-08
Range-check keccak sponge inputs to bytes

Nashtare:
Range-checks the `KeccakSpongeStark` `block_bytes` values to be fitting in a `u8`.

This, along with what's already there on `Arithmetic` / `BytePacking` tables, and assuming the TODO for `PUSH` constraints is implemented, should be enough to ensure that all memory values being written are indeed of the correct form (i.e. each decomposed limb holds at most 32 bits).

### DIFF
--- a/evm/src/keccak_sponge/columns.rs
+++ b/evm/src/keccak_sponge/columns.rs
@@ -1,5 +1,6 @@
 use std::borrow::{Borrow, BorrowMut};
 use std::mem::{size_of, transmute};
+use std::ops::Range;
 
 use crate::util::{indices_arr, transmute_no_compile_time_size_checks};
 
@@ -72,11 +73,34 @@ pub(crate) struct KeccakSpongeColumnsView<T: Copy> {
     /// The first part of the state of the sponge, seen as bytes, after the permutation is applied.
     /// This also represents the output digest of the Keccak sponge during the squeezing phase.
     pub updated_digest_state_bytes: [T; KECCAK_DIGEST_BYTES],
+
+    /// The counter column (used for the range check) starts from 0 and increments.
+    pub range_counter: T,
+    /// The frequencies column used in logUp.
+    pub rc_frequencies: T,
 }
 
 // `u8` is guaranteed to have a `size_of` of 1.
 /// Number of columns in `KeccakSpongeStark`.
 pub const NUM_KECCAK_SPONGE_COLUMNS: usize = size_of::<KeccakSpongeColumnsView<u8>>();
+
+// Indices for LogUp range-check.
+// They are on the last registers of this table.
+pub(crate) const RC_FREQUENCIES: usize = NUM_KECCAK_SPONGE_COLUMNS - 1;
+pub(crate) const RANGE_COUNTER: usize = RC_FREQUENCIES - 1;
+
+/// Indices for the range-checked values, i.e. the `block_bytes` section.
+// TODO: Find a better way to access those indices
+pub(crate) const fn get_block_bytes_range() -> Range<usize> {
+    let start = 6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S;
+    start..start + KECCAK_RATE_BYTES
+}
+
+/// Return the index for the targeted `block_bytes` element.
+pub(crate) const fn get_single_block_bytes_value(i: usize) -> usize {
+    debug_assert!(i < KECCAK_RATE_BYTES);
+    get_block_bytes_range().start + i
+}
 
 impl<T: Copy> From<[T; NUM_KECCAK_SPONGE_COLUMNS]> for KeccakSpongeColumnsView<T> {
     fn from(value: [T; NUM_KECCAK_SPONGE_COLUMNS]) -> Self {
--- a/evm/src/keccak_sponge/columns.rs
+++ b/evm/src/keccak_sponge/columns.rs
@@ -89,11 +89,12 @@ pub const NUM_KECCAK_SPONGE_COLUMNS: usize = size_of::<KeccakSpongeColumnsView<u
 pub(crate) const RC_FREQUENCIES: usize = NUM_KECCAK_SPONGE_COLUMNS - 1;
 pub(crate) const RANGE_COUNTER: usize = RC_FREQUENCIES - 1;
 
+pub(crate) const BLOCK_BYTES_START: usize =
+    6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S;
 /// Indices for the range-checked values, i.e. the `block_bytes` section.
 // TODO: Find a better way to access those indices
 pub(crate) const fn get_block_bytes_range() -> Range<usize> {
-    let start = 6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S;
-    start..start + KECCAK_RATE_BYTES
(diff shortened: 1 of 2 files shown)


evm/src/keccak_sponge/columns.rs
```diff
@@ -72,12 +73,35 @@ pub(crate) struct KeccakSpongeColumnsView<T: Copy> {
...
+// They are on the last registers of this table.
+pub(crate) const RC_FREQUENCIES: usize = NUM_KECCAK_SPONGE_COLUMNS - 1;
+pub(crate) const RANGE_COUNTER: usize = RC_FREQUENCIES - 1;
+
+/// Indices for the range-checked values, i.e. the `block_bytes` section.
+// TODO: Find a better way to access those indices
+pub(crate) const fn get_block_bytes_range() -> Range<usize> {
+    let start = 6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S;
+    start..start + KECCAK_RATE_BYTES
+}
```
Nashtare, line 97: I'm not super happy with this, as it can be broken easily if forgotten during a trace update.
One way to handle it properly may be to use an Enum for the different components inside the trace, and implement `Index` on the `KeccakSpongeColumnsView` with it?

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -477,6 +480,38 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
...
+
+        let n_rows = cols[0].len();
+        debug_assert!(cols.iter().all(|col| col.len() == n_rows));
+
+        for i in 0..BYTE_RANGE_MAX {
+            cols[RANGE_COUNTER][i] = F::from_canonical_usize(i);
+        }
+        for i in BYTE_RANGE_MAX..n_rows {
+            cols[RANGE_COUNTER][i] = F::from_canonical_usize(BYTE_RANGE_MAX - 1);
+        }
```
dvdplm, line 496:
```suggestion
        let filler = F::from_canonical_u8(u8::MAX);
        for i in BYTE_RANGE_MAX..n_rows {
                    cols[RANGE_COUNTER][i] = filler;
        }
```

…to avoid creating a field element for every iteration? (Although I bet the compiler does that optimization anyway)

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -477,6 +480,38 @@ impl<F: RichField + Extendable<D>, const D: usize> KeccakSpongeStark<F, D> {
...
+
+        // For each column c in cols, generate the range-check
+        // permutations and put them in the corresponding range-check
+        // columns rc_c and rc_c+1.
+        for col in 0..KECCAK_RATE_BYTES {
+            for i in 0..n_rows {
+                let c = get_single_block_bytes_value(col);
```
dvdplm, line 503: Why is this in the inner loop?

evm/src/keccak_sponge/keccak_sponge_stark.rs
Nashtare: Yeah, it's being used in a lot of places and I'd bet this is optimized away. One advantage to the current approach is that in the general case, it is clearly apparent which value we're writing to the trace / constraining, while saving in a temporary variable may not be clearly understandable.

evm/src/keccak_sponge/keccak_sponge_stark.rs
```diff
@@ -12,17 +12,21 @@ use plonky2::hash::hash_types::RichField;
...
 use crate::cross_table_lookup::Column;
 use crate::evaluation_frame::{StarkEvaluationFrame, StarkFrame};
 use crate::keccak_sponge::columns::*;
+use crate::lookup::Lookup;
 use crate::stark::Stark;
-use crate::util::trace_rows_to_poly_values;
 use crate::witness::memory::MemoryAddress;
 
+/// Strict upper bound for the individual bytes range-check.
+const BYTE_RANGE_MAX: usize = 256;
```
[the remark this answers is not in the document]
Nashtare: I'd rather keep it as is, maybe it's just a matter of personal preference but that way it matches the other range-check implementations. 
And the for loops below would look a bit odd? `for i in 0..=u8::MAX {}` and then `for i in u8::MAX+1..num_rows {}`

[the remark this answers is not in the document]
dvdplm: 🤷 yeah, maybe it's more readable with its own constant.

evm/src/keccak_sponge/columns.rs
dvdplm: I don't know enough to have much to say here, but if we're talking about a set of sections with varying sizes I think a custom `Index` impl would be a bit odd. If the sections were all of the same size, then perhaps yes.

How about this:

```rust
use std::ops::Range;
pub enum Sections {
    BlockBytes,
    Other,
    Moar,
}

impl Sections {
    const fn range(&self) -> Range<usize> {
        match self {
            // This is a bit awkwardly long, but it's `const`
            Self::BlockBytes => {
                (6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S)
                    ..(6 + KECCAK_RATE_BYTES
                        + KECCAK_RATE_U32S
                        + KECCAK_CAPACITY_U32S
                        + KECCAK_RATE_BYTES)
            }
            Self::Other => 1..11,
            Self::Moar => 129..890,
        }
    }
}

pub(crate) const fn single_block_bytes_val(i: usize) -> usize {
    debug_assert!(i < KECCAK_RATE_BYTES);
    // Using  it is nice
    Sections::BlockBytes.range().start + i
}
```

Another idea is to use a module and constants:

```rust
#[allow(non_snake_case, non_upper_case_globals)]
pub mod Sections {
     pub const BlockBytes: Range<usize> = 123..456,
     pub const Other:  Range<usize> = 789..1000,
     pub const Moar:  Range<usize> = 9..10
}
```

evm/src/keccak_sponge/columns.rs
```diff
@@ -72,12 +73,35 @@ pub(crate) struct KeccakSpongeColumnsView<T: Copy> {
...
 
+// Indices for LogUp range-check.
+// They are on the last registers of this table.
+pub(crate) const RC_FREQUENCIES: usize = NUM_KECCAK_SPONGE_COLUMNS - 1;
+pub(crate) const RANGE_COUNTER: usize = RC_FREQUENCIES - 1;
+
+/// Indices for the range-checked values, i.e. the `block_bytes` section.
+// TODO: Find a better way to access those indices
+pub(crate) const fn get_block_bytes_range() -> Range<usize> {
+    let start = 6 + KECCAK_RATE_BYTES + KECCAK_RATE_U32S + KECCAK_CAPACITY_U32S;
```
LindaGuiga, line 95: Maybe we can define this as a `const` outside of the function for clarity?

LindaGuiga (approved):
Looks great, thanks!

evm/src/keccak_sponge/columns.rs
Nashtare: I don't think this solves the problem in the end. Because altering the ordering / registers somehow will risk to mess this up. As we need the constraints for the milestone, I'll merge as is following @LindaGuiga's approval, but I'll come back to this topic, as a clearer way to handle it here (and in the other files) would be beneficial.

PR #1343, opened 2023-11-08
Constrain clock

hratoanina:
The `clock` register wasn't constrained. We might need to have a more complex logic to update it in future PRs, so I added the constraints in a dedicated file.

### DIFF
--- a/evm/src/cpu/clock.rs
+++ b/evm/src/cpu/clock.rs
@@ -1,0 +1,37 @@
+use plonky2::field::extension::Extendable;
+use plonky2::field::packed::PackedField;
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::CpuColumnsView;
+
+/// Check the correct updating of `clock`.
+pub fn eval_packed<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    nv: &CpuColumnsView<P>,
+    yield_constr: &mut ConstraintConsumer<P>,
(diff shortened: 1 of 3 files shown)


evm/src/cpu/clock.rs
```diff
@@ -0,0 +1,37 @@
...
+use plonky2::hash::hash_types::RichField;
+use plonky2::iop::ext_target::ExtensionTarget;
+
+use crate::constraint_consumer::{ConstraintConsumer, RecursiveConstraintConsumer};
+use crate::cpu::columns::CpuColumnsView;
+
+/// Check the correct updating of `clock`.
+pub fn eval_packed<P: PackedField>(
+    lv: &CpuColumnsView<P>,
+    nv: &CpuColumnsView<P>,
```
dvdplm, line 12:
```suggestion
    local_calues: &CpuColumnsView<P>,
    next_values: &CpuColumnsView<P>,
```

evm/src/cpu/clock.rs
```diff
@@ -0,0 +1,37 @@
...
+    // The clock is incremented by 1 at each row.
+    yield_constr.constraint_transition(nv.clock - lv.clock - P::ONES);
+}
+
+/// Circuit version of `eval_packed`.
+/// Check the correct updating of `clock`.
+pub fn eval_ext_circuit<F: RichField + Extendable<D>, const D: usize>(
+    builder: &mut plonky2::plonk::circuit_builder::CircuitBuilder<F, D>,
+    lv: &CpuColumnsView<ExtensionTarget<D>>,
+    nv: &CpuColumnsView<ExtensionTarget<D>>,
```
dvdplm, line 26: same here: better use explicit variable names

Nashtare (approved):

Issue #1338, opened 2023-11-07
Kernel benchmark


wborgeaud:
Here are the current table lengths for the ERC20 test:

| Table               | Length  | Width
|-------------------|---------|--|
| Memory        | 542 726  | 20 |
| Cpu           | 262 144  | 112 |
| Arithmetic    | 26 167   | 114 |
| Keccak       | 13 392   | 2431 |
| BytePacking  | 5 688    | 72 |
| Logic         | 4 180    | 523 |
| KeccakSponge | 558     | 437 |

The CPU is halted after 172,006 cycles. Here is a breakdown of the most expensive operations:
| Operation             | Cycles  |
|-----------------------|---------|
| Bootstrapping         | 12 922  |
| Load MPTs             | 1 155   |
| Hash state trie       | 10 218  |
| Hash txn trie         | 118     |
| Hash receipts trie    | 115     |
| Txn                   | 124 634 |
| - ECRECOVER           | 26 828  |
| - RLP decode txn      | 3 169   |
| - Process txn         | 4 898  |
| - Load code (388B)      | 5 443  |
| - Jumpdest check      | 7 567   |
| - SLOAD               | 400     |
| - CALL                | 64 572  |
| -- Load code (1824B)        | 25 691  |
| -- Jumpdest check          | 37 402  |
| - SLOAD               | 427     |
| - SSTORE              | 960     |
| - SLOAD               | 450     |
| - SSTORE              | 1 498   |
| - LOG3                | 1 376   |
| - RETURN              | 307     |
| Process receipt       | 2 862   |
| Hash state trie       | 13 436  |
| Hash txn trie         | 801     |
| Hash receipts trie    | 2 775   |

Some optimization ideas:
- **Batch ECDSA**: We could batch all the ECRECOVER calls in a block and perform one batch ECDSA verification after the txn proofs. This would amortize the cost of ECRECOVER across the block.
- **Preinitialize memory segments**: We could allow some memory segments to be preinitialized by the prover. That is, we would remove the constraint that memory is initialized at zero for these segments. This would remove the costly bootstrapping, MPT loading, and code loading, potentially saving ~45k cycles in this example. 
- **Optimize Jumpdest check**: We could cache the jumpdest analysis across the txn or block to avoid redundant work. For the type-2, we could add the jumpdest destination to the account data, or simply remove the check like in Polygon zkEVM.

Nashtare:
[...]

PR #1347, opened 2023-11-10
Refactor JUMPDEST analysis

hratoanina:
See #1338

We optimize JUMPDEST analysis by:
- using a table to read the number of bytes to skip
- merging the `%increment` with the table read

The second point makes a loop iteration encountering a JUMPDEST slightly more expensive, but all the others cheaper.

For the ERC20 test, we get the following improvements:
CPU trace: 171916 -> 159854
Memory trace: 545553 -> 517626


wborgeaud (approved):
Looks good to me!

PR #1350, opened 2023-11-13
chore: remove experimental fft python code

dvdplm:
Looks like this code isn't in use?


dvdplm:
Ping @Nashtare

Nashtare (commented):
Sorry I had missed your ping. I'm not sure keeping this `projects` folder would hurt though, in essence it's nice to highlight what's possible at a high-level so that external contributors could also give it a go if they want to.

PR #1352, opened 2023-11-13
Add push constraints

### LINKED ISSUE #1150: Add missing constraints

### LINKED ISSUE #1325: The top of the stack isn't always range-checked
Only writes into memory channels that are written in Memory are being range-checked. As top-of-the-stack is CPU only, and may remain at the top of the stack before being popped, it may not be range-checked by Memory.

Nashtare:
This constrains `PUSH` internal reads to be bytes, by delegating the check to `BytePackingStark`

closes #1150 
closes #1325 as this won't be an issue anymore after this.

### DIFF
--- a/evm/src/byte_packing/byte_packing_stark.rs
+++ b/evm/src/byte_packing/byte_packing_stark.rs
@@ -68,12 +68,11 @@ pub(crate) fn ctl_looked_data<F: Field>() -> Vec<Column<F>> {
     // obtain the corresponding limb.
     let outputs: Vec<Column<F>> = (0..8)
         .map(|i| {
-            let range = (value_bytes(i * 4)..value_bytes(i * 4) + 4).collect_vec();
+            let range = (value_bytes(i * 4)..value_bytes(i * 4) + 4);
             Column::linear_combination(
                 range
-                    .iter()
                     .enumerate()
-                    .map(|(j, &c)| (c, F::from_canonical_u64(1 << (8 * j)))),
+                    .map(|(j, c)| (c, F::from_canonical_u64(1 << (8 * j)))),
             )
         })
         .collect();
--- a/evm/src/cpu/cpu_stark.rs
+++ b/evm/src/cpu/cpu_stark.rs
@@ -157,6 +157,33 @@ pub(crate) fn ctl_filter_byte_unpacking<F: Field>() -> Column<F> {
     Column::single(COL_MAP.op.mstore_32bytes)
 }
 
+/// Creates the vector of `Columns` corresponding to the contents of the CPU registers when performing a `PUSH`.
+/// `PUSH` internal reads are done by calling `BytePackingStark`.
+pub(crate) fn ctl_data_byte_packing_push<F: Field>() -> Vec<Column<F>> {
+    let context = Column::single(COL_MAP.code_context);
+    let segment = Column::constant(F::from_canonical_usize(Segment::Code as usize));
+    // The initial offset if `pc + 1`.
+    let virt =
+        Column::linear_combination_with_constant([(COL_MAP.program_counter, F::ONE)], F::ONE);
+    let val = Column::singles_next_row(COL_MAP.mem_channels[0].value);
+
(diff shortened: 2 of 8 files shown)


evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -68,12 +68,11 @@ pub(crate) fn ctl_looked_data<F: Field>() -> Vec<Column<F>> {
...
     let outputs: Vec<Column<F>> = (0..8)
         .map(|i| {
-            let range = (value_bytes(i * 4)..value_bytes(i * 4) + 4).collect_vec();
+            let range = (value_bytes(i * 4)..value_bytes(i * 4) + 4);
             Column::linear_combination(
                 range
-                    .iter()
                     .enumerate()
-                    .map(|(j, &c)| (c, F::from_canonical_u64(1 << (8 * j)))),
+                    .map(|(j, c)| (c, F::from_canonical_u64(1 << (8 * j)))),
```
Nashtare, line 75: Technically unrelated, but as I was working on the `BytePackingStark` CTL, I noticed this collect that could be removed.

evm/src/cpu/cpu_stark.rs
```diff
@@ -157,6 +157,33 @@ pub(crate) fn ctl_filter_byte_unpacking<F: Field>() -> Column<F> {
...
+    // We fetch the length from the `PUSH` opcode lower bits, that indicate `len - 1`.
+    let len = Column::le_bits_with_constant(&COL_MAP.opcode_bits[0..5], F::ONE);
+
+    let num_channels = F::from_canonical_usize(NUM_CHANNELS);
+    let timestamp = Column::linear_combination([(COL_MAP.clock, num_channels)]);
+
+    let mut res = vec![context, segment, virt, len, timestamp];
+    res.extend(val);
+
+    res
```
dvdplm, line 179: There's a `Vec::concat` in `itertools` I think. Might use that to make this into a one-liner (`Vec::concat(vec![…, …], val)`).

Q: is this a hot function you think?

Nashtare: Oh I didn't know. This is not a hot spot though

hratoanina (commented):
Looks good, but I think we can take care of the PUSH stack constraints in this PR
https://github.com/0xPolygonZero/plonky2/blob/b022ea0abec719434e62f3a973b722009008a91f/evm/src/cpu/stack.rs#L105

Nashtare:
Good point, just added it!

Nashtare:
@hratoanina Good point, just added it!

hratoanina (approved):

PR #1356, opened 2023-11-14
Fix parsing of non-legacy receipts

Nashtare:
The parser currently only supports legacy receipts, for which the byte sequence is an actual RLP encoding. For non-legacy receipts, similarly to non-legacy transactions, the txn type should be prepended.


LindaGuiga (approved):
Looks good to me, thanks!

PR #1357, opened 2023-11-15
Constrain is_keccak_sponge

hratoanina:


Nashtare (approved):
Still not sure the last constraint 
```rust
yield_constr.constraint_last_row(local_values.is_keccak_sponge);
```

is needed, but better be safe than sorry. LGTM otherwise!

PR #1358, opened 2023-11-15
Merge public values inside prove_aggreg

Nashtare:
This PR changes the arguments of `prove_aggregation` to pass both `PublicValues` associated to the two inner proofs being combined, so that the logic of selecting a field from one `pv` or the other is done internally and is simpler from outside.

cc @BGluth


LindaGuiga (approved):
Looks good to me, thanks!

PR #1360, opened 2023-11-15
Move stack_len_bounds_aux to general columns

hratoanina:
The only instructions using all general columns are `EQ` and `ISZERO`, which can't trigger a stack overflow. By only checking for stack overflow after instructions which actually can, we can use a general column for `stack_len_bounds_aux` and save a column.


evm/src/cpu/stack.rs
```diff
@@ -270,10 +299,22 @@ pub(crate) fn eval_packed<P: PackedField>(
     nv: &CpuColumnsView<P>,
     yield_constr: &mut ConstraintConsumer<P>,
 ) {
-    for (op, stack_behavior) in izip!(lv.op.into_iter(), STACK_BEHAVIORS.into_iter()) {
+    for (op, stack_behavior, might_overflow) in izip!(
+        lv.op.into_iter(),
+        STACK_BEHAVIORS.into_iter(),
```
dvdplm, line 304: I must be missing something. The `izip` macro will terminate when one of the sub-iterators return `None`, so this `for`-loop will stop when we get to `eq_iszero`? Is that the intended behaviour here? Or am I misunderstanding this completely?

evm/src/witness/transition.rs
```diff
@@ -351,17 +367,22 @@ fn try_perform_instruction<F: Field>(state: &mut GenerationState<F>) -> Result<(
...
+            let disallowed_len = F::from_canonical_usize(MAX_USER_STACK_SIZE + 1);
+            let diff = row.stack_len - disallowed_len;
+            if let Some(inv) = diff.try_inverse() {
+                last_row.general.stack_mut().stack_len_bounds_aux = inv;
+            } else {
+                // This is a stack overflow that should have been caught earlier.
+                return Err(ProgramError::InterpreterError);
+            }
         }
+        state.registers.check_overflow = false;
```
dvdplm, line 385: Is this the exact same code as above? Is there a way to avoid duplication perhaps?

evm/src/witness/transition.rs
```diff
@@ -227,11 +227,25 @@ fn get_op_special_length(op: Operation) -> Option<usize> {
...
+    match op {
+        Operation::Pc | Operation::Push(0) => true,
+        Operation::Push(_) => true,
+        // SWAP doesn't, but share a flag with DUP.
+        Operation::Dup(_) | Operation::Swap(_) => true,
+        // Doesn't directly push, but the syscall it's returning from might.
+        Operation::ExitKernel => true,
+        _ => false,
+    }
+}
```
dvdplm, line 242: How do we ensure that this code here does not go out of sync with the `MIGHT_OVERFLOW` constant?

evm/src/cpu/stack.rs
dvdplm: Ok, ignore me. I get it.

I wonder though: can we refactor the `STACK_BEHAVIOR` to not be a list of `Option<StackBehavior>` and add the info in `MIGHT_OVERFLOW` to the `StackBehavior` type instead? It seems like most ops have some kind of "special flair", so perhaps it is wise to collect all that metadata into one type?

hratoanina: That's a good idea, and I think I even went for this approach the first time I implemented it. I think I changed my mind because there are other similar structs like `KEEPS_CONTEXT` in `contextops.rs` and `SIMPLE_OPCODES` in `gas.rs`.
Do you think it would make sense to have a single struct called `INSTRUCTION_BEHAVIOR` or something with all these combined?

dvdplm: Intuitively yes, I think so. Given this is a hot struct (right?), it should be as small as possible so we need to be careful that the layout is cache friendly, but I bet that traversing several smaller lists/structs is costlier. Hard numbers would be good ofc.

evm/src/witness/transition.rs
hratoanina: Ah, we don't, that's right. I'll turn it into a function like `get_op_special_length` or `fill_op_flag`.

evm/src/witness/transition.rs
hratoanina: Indeed, and there's more duplicated code between `try_perform_instruction` and `generate_exception`, I'll make a function for them.

evm/src/cpu/stack.rs
hratoanina: This is noted, but this requires enough changes to tackle it in another PR in my opinion.

evm/src/cpu/stack.rs
```diff
@@ -13,6 +13,35 @@ use crate::cpu::columns::CpuColumnsView;
...
+    jumpdest_keccak_general: false,
+    prover_input: false,
+    jumps: false,
+    pc_push0: true,
+    push: true,
+    dup_swap: true,
+    context_op: false,
+    mload_32bytes: false,
+    mstore_32bytes: false,
+    exit_kernel: true,
```
LindaGuiga, line 39: I think you should add a comment to explain why `exit_kernel` is true here

PR #1362, opened 2023-11-16
Update Memory in specs

hratoanina:


evm/spec/tables/memory.tex
```diff
@@ -11,40 +11,40 @@ \subsection{Memory}
   \item $v$, the value being read or written
   \item $\tau$, the timestamp of the operation
 \end{enumerate}
-The memory table should be ordered by $(a, \tau)$. Note that the correctness memory could be checked as follows:
+The memory table should be ordered by $(a, \tau)$. Note that the correctness of memory could be checked as follows:
```
Nashtare, line 14: nit: seems it's missing "the"?
```suggestion
The memory table should be ordered by $(a, \tau)$. Note that the correctness of the memory could be checked as follows:
```

Nashtare (approved):
LGTM, thanks!

PR #1369, opened 2023-11-20
Check is_kernel_mode when halting

hratoanina:
Extra check to prevent malicious halting while executing user code.


LindaGuiga (approved):
LGTM thanks!

PR #1370, opened 2023-11-21
Initialize blockhashes

LindaGuiga:
@BGluth This PR initializes the block hashes of the genesis block.

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -1035,7 +1035,8 @@ where
         } else {
             // Initialize genesis_state_trie, state_root_after and the block number for correct connection between blocks.
             // Initialize `state_root_after`.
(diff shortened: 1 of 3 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -1035,7 +1035,8 @@ where
         } else {
             // Initialize genesis_state_trie, state_root_after and the block number for correct connection between blocks.
```
Nashtare, line 1036: I think we're missing an update to the comment here

BGluth:
I might not be the best to review plonky2 logic, but I'm excited for when this PR lands! :laughing:

Nashtare:
@pgebheim added to the RC1 milestone as this is a bugfix for the block proof logic

hratoanina (approved):

pgebheim:
Perfect thanks.

PR #1372, opened 2023-11-21
Add exceptions to specs

hratoanina:


evm/spec/exceptions.tex
```diff
@@ -0,0 +1,44 @@
+\section{Exceptions}
+\label{exceptions}
+
+Sometimes, when executing user code (i.e. contract or transaction code), the EVM halts exceptionally (i.e. outside of a STOP, a RETURN or a REVERT).
+When this happens, the CPU table invokes a special instruction with a dedicated operation flag \texttt{exception}.
+Exceptions can only happen in user mode; triggering an exception in kernel mode would make the proof unverifiable.
+No matter the exception, the handling is the same:
+
+-- The opcode which will trigger the exception is not executed. The operation flag set is \texttt{exception} instead of the opcode's flag.
```
LindaGuiga, line 9:
```suggestion
-- The opcode which would trigger the exception is not executed. The operation flag set is \texttt{exception} instead of the opcode's flag.
```
or:
```suggestion
-- The opcode which triggers the exception is not executed. The operation flag set is \texttt{exception} instead of the opcode's flag.
```

evm/spec/exceptions.tex
```diff
@@ -0,0 +1,44 @@
...
+
+-- A value is pushed to the stack which contains the current program counter (to retrieve the faulty opcode) and the current value of \texttt{gas\_used}.
```
LindaGuiga, line 11: I find it a bit hard to read. Maybe something like:
"We push a value to the stack containing: the current program counter (to retrieve the faulty opcode), and the current value of \texttt{gas\_used}."

evm/src/cpu/kernel/asm/core/exception.asm
```diff
@@ -1,4 +1,5 @@
 // These exception codes are arbitrary and assigned by us.
+// Note that exceptions can only be triggered in user mode.
```
LindaGuiga, line 2: Maybe add a comment to say the user mode is checked by the constraints?

LindaGuiga (approved):
LGTM, thanks!

PR #1375, opened 2023-11-21
Add specs for the CPU table

hratoanina:


evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
 \subsection{CPU}
 \label{cpu}
 
-TODO
+The CPU is the central component of the zkEVM. Like any CPU, it reads instructions, executes them and modifies the state (its registers and the memory)
+accordingly. The constraining of some complex instructions (e.g. Keccak hashing) are delegated to the other tables.
```
LindaGuiga, line 5:
```suggestion
accordingly. The constraining of some complex instructions (e.g. Keccak hashing) is delegated to the other tables.
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+This section will only briefly present the CPU and its columns. Details about the CPU logic will be provided later.
+
+\subsubsection{CPU flow}
+
+An execution run is made of three distinct parts:
+\begin{itemize}
+    \item \textbf{Bootstrapping:} The CPU starts by writing all the kernel code in memory, and hash it. The hash is then compared to a public value shared with
```
LindaGuiga, line 12:
```suggestion
    \item \textbf{Bootstrapping:} The CPU starts by writing all the kernel code in memory, and hashing it. The hash is then compared to a public value shared with
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+the verifier to ensure that the kernel code is correct.
+    \item \textbf{CPU cycles:} The bulk of the execution. In each row, the CPU reads the current code at the PC address, and executes it. The current code can be the kernel code,
```
LindaGuiga, line 14: If we do not mention the PC before in the specs, maybe say "program counter (PC)"?

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+    \item \texttt{code\_context}: Indicates in which context the code to execute resides. It's equal to \texttt{context} in user mode, but is always 0 in kernel mode.
+    \item \texttt{program\_counter}: The address of the instruction to be read and executed.
+    \item \texttt{stack\_len}: The current length of the stack.
+    \item \texttt{stack\_len\_bounds\_aux}: Helper column used to check that the stack doesn't overflow in user mode.
+    \item \texttt{is\_kernel\_mode}: Boolean indicating whether we are in kernel (i.e. privileged) mode. This means we are executing kernel code, and we have access to
+privileged instructions.
+    \item \texttt{gas}: The current amount of gas used in the current context. It is eventually checked to be below the current gas limit. Must fit in 32 bits.
+    \item \texttt{is\_keccak\_sponge}: Boolean indicating whether we are executing a Keccak hash. This happens whenever a \texttt{KECCAK\_GENERAL} instruction is executed, or at the last
+cycle of bootstrapping to hash the kernel code.
+    \item \texttt{clock}: Required to generate timestamps for memory operations to order them. Starts at 0 and is incremented by 1 each row.
```
LindaGuiga, line 35:
```suggestion
    \item \texttt{clock}: Required to generate timestamps for memory operations to order them. Starts at 0 and is incremented by 1 at each row.
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+    \item \texttt{opcode\_bits}: 8 boolean columns, which are the bit decomposition of the opcode being read at PC.
+\end{itemize}
+
+\subsubsection{CPU columns}
```
LindaGuiga, line 39: You already have the subsubsection above
```suggestion
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+
+\paragraph*{Operation flags:} Boolean flags. During CPU cycles, each row executes a single instruction, which sets one and only one operation flag. No flag is set during
+bootstrapping and padding. The decoding constraints ensure that the flag set corresponds to the opcode being read.
+There isn't a 1-to-1 correspondance between instructions and flags. For efficiency, the same flag can represent different, unrelated instructions (e.g. \texttt{eq\_iszero}, which represents
```
LindaGuiga, line 43: I would use "is activated by" instead of represents, but as you prefer

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,55 @@
...
+(resp. \texttt{eq\_iszero * opcode\_bits[0]}).
+
+\paragraph*{Memory columns:} The CPU interacts with the EVM memory via its memory channels. At each row, a memory channel can execute a write, a read, or be disabled. A full memory channel is composed of:
+\begin{itemize}
+    \item  \texttt{used}: Boolean flag. If it's set to 1, a memory operation is executed in this channel at this row. If it's set to 0, no operation is done but its columns might be reused for other purposes.
+    \item  \texttt{is\_read}: Boolean flag indicating if a memory operation is a read or a write.
+    \item  3 \texttt{address} columns. A memory address is made of three parts: \texttt{context}, \texttt{segment} and \texttt{virtual}.
+    \item  8 \texttt{value} columns. EVM words are 256 bits long, and they are broken down in 8 32-bit limbs.
+\end{itemize}
+The last memory channel is a partial channel: it doesn't have its own \texttt{value} columns and shares them with the first full memory channel. This allows us to save eight columns.
```
LindaGuiga, line 55: I think the General columns are missing

LindaGuiga (commented):
Thanks!

LindaGuiga (approved):

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+performing some memory operations, and updating the PC.
+\item \textbf{Padding:} At the end of the execution, we need to pad the length to the next power of two. Halting happens when the PC reaches the special halting label
+in the kernel. Constraints ensure that every subsequent row is a padding row and that you can't resume execution.
+\end{itemize}
+
+\subsubsection{CPU columns}
+
+\paragraph*{Registers:} \begin{itemize}
+    \item \texttt{is\_bootstrap\_kernel}: Boolean indicating whether this is a bootstrapping row or not. It must be 1 at the first row, then switch to 0 until the end.
+    \item \texttt{context}: Indicates which context we are in. 0 for the kernel, and a positive integer for every user context. Incremented by 1 at every call.
```
dvdplm, line 25: Is the context stateful? If I execute a kernel call mid-program then `context` is set to `0` but when I come back to userland execution, is it restored to be a positive integer or does it restart from `1`?

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+    \item \texttt{code\_context}: Indicates in which context the code to execute resides. It's equal to \texttt{context} in user mode, but is always 0 in kernel mode.
+    \item \texttt{program\_counter}: The address of the instruction to be read and executed.
+    \item \texttt{stack\_len}: The current length of the stack.
+    \item \texttt{stack\_len\_bounds\_aux}: Helper column used to check that the stack doesn't overflow in user mode.
+    \item \texttt{is\_kernel\_mode}: Boolean indicating whether we are in kernel (i.e. privileged) mode. This means we are executing kernel code, and we have access to
```
dvdplm, line 30: I think it would be useful to discuss the relationship to `context` and why there are two flags to indicate kernel mode.

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+privileged instructions.
+    \item \texttt{gas}: The current amount of gas used in the current context. It is eventually checked to be below the current gas limit. Must fit in 32 bits.
+    \item \texttt{is\_keccak\_sponge}: Boolean indicating whether we are executing a Keccak hash. This happens whenever a \texttt{KECCAK\_GENERAL} instruction is executed, or at the last
+cycle of bootstrapping to hash the kernel code.
+    \item \texttt{clock}: Required to generate timestamps for memory operations to order them. Starts at 0 and is incremented by 1 at each row.
```
dvdplm, line 35: These are not actual timestamps though. Perhaps rephrase this to explain that "clock" here is simply a monotonic counter used to enforce correct ordering of memory accesses?

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+    \item \texttt{opcode\_bits}: 8 boolean columns, which are the bit decomposition of the opcode being read at PC.
```
dvdplm, line 36:
```suggestion
    \item \texttt{opcode\_bits}: 8 boolean columns, which are the bit decomposition of the opcode being read at the current PC.
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+\end{itemize}
+
+\paragraph*{Operation flags:} Boolean flags. During CPU cycles, each row executes a single instruction, which sets one and only one operation flag. No flag is set during
```
dvdplm, line 39:
```suggestion
\paragraph*{Operation flags:} Boolean flags. During CPU cycles phase, each row executes a single instruction, which sets one and only one operation flag. No flag is set during
```

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
+1 for a JUMP, and $\texttt{condition} \neq 0$ for a JUMPI. To check if the condition is actually non-zero for a JUMPI, \texttt{cond\_sum\_pinv} stores the modular inverse of
+\texttt{condition} (or 0 if it's zero).
+    \item  \texttt{Shift}: For shifts, the logic differs depending on whether the displacement is lower than $2^{32}$, i.e. if it fits in a single value limb.
+To check if this is not the case, we must check that at least one of the seven high limbs is not zero. The general column \texttt{high\_limb\_sum\_inv} holds the modular inverse
+of the sum of the seven high limbs, and is used to check it's non-zero like the previous cases.
+Contrary to the logic operations, we do not need to check limbs individually: each limb has been range-checked to 32 bits, meaning that it's not possible for the sum to
+overflow and be zero if some of the limbs are non-zero.
+    \item  \texttt{Stack}: The last three columns are used by popping-only (resp. pushing-only) instructions to check if the stack is empty after (resp. was empty
```
dvdplm, line 71: Not sure what the two `resp.` parentheses mean? Do you mean "as opposed to"?

evm/spec/tables/cpu.tex
```diff
@@ -1,4 +1,73 @@
...
-TODO
+The CPU is the central component of the zkEVM. Like any CPU, it reads instructions, executes them and modifies the state (its registers and the memory)
+accordingly. The constraining of some complex instructions (e.g. Keccak hashing) is delegated to the other tables.
+This section will only briefly present the CPU and its columns. Details about the CPU logic will be provided later.
+
+\subsubsection{CPU flow}
+
+An execution run is made of three distinct parts:
+\begin{itemize}
+    \item \textbf{Bootstrapping:} The CPU starts by writing all the kernel code in memory, and hashing it. The hash is then compared to a public value shared with
```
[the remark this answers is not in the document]
hratoanina: This indeed happens for each run of the EVM.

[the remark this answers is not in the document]
dvdplm: What is, exactly, "one run of the EVM"? One EVM program? One op-code? One transaction? What is the unit of work here?

evm/spec/tables/cpu.tex
hratoanina: It's for "respectively", as in "The last three columns are used by popping-only instructions to check if the stack is empty after the instruction, and are used by pushing-only instructions to check if the stack was empty before the instruction."
It's pretty common in math at least, I'm not sure about computer science (and English in general).

evm/spec/tables/cpu.tex
hratoanina: When you execute a syscall, `context` doesn't change but `is_kernel_mode` is set to `1` and `code_context` is set to `0`.
You're only in `context` `0` before executing the transaction (to check the validity of the tries before, to parse the transaction etc.) and after the transaction (to check the validity of the trees after, to update the receipt trie, to halt...). I'll add something about it.

dvdplm: Aha, ok, that clears it up. Yes, do please add this info! :)

hratoanina:
@dvdplm I applied your comments but I just noticed I missed one about what an EVM execution run is. It's the complete execution of the state transition caused by the transaction, i.e. the entire execution in a proof. Should I rephrase?

PR #1377, opened 2023-11-21
Starting the specs for the CPU logic

LindaGuiga:


evm/spec/cpulogic.tex
```diff
@@ -0,0 +1,170 @@
...
+
+Since we proceed one transaction at a time, a few intermediary values need to be provided by the prover. Indeed, to prove that the registers in the EVM state are currently updated, we need to have access to their initial values. When aggregating proofs, we can also constrain those values to match from one transaction to the next. Let us consider the example of the transaction number. Let $n$ be the number of transactions executed so far in the current block. If the current proof is not a dummy one (we are indeed executing a transaction), then the transaction number should be updated: $n := n+1$. Otherwise, the number reamins unchanged. We can easily constrain this update. When aggregating the previous transaction proof ($lhs$) with the current one ($rhs$), we also need to check that the output transaction number of $lhs$ is the same as the input transaction number of $rhs$.
```
hratoanina, line 16:
```suggestion
Since we process one transaction at a time, a few intermediary values need to be provided by the prover. Indeed, to prove that the registers in the EVM state are correctly updated, we need to have access to their initial values. When aggregating proofs, we can also constrain those values to match from one transaction to the next. Let us consider the example of the transaction number. Let $n$ be the number of transactions executed so far in the current block. If the current proof is not a dummy one (we are indeed executing a transaction), then the transaction number should be updated: $n := n+1$. Otherwise, the number remains unchanged. We can easily constrain this update. When aggregating the previous transaction proof ($lhs$) with the current one ($rhs$), we also need to check that the output transaction number of $lhs$ is the same as the input transaction number of $rhs$. 
```

evm/spec/cpulogic.tex
```diff
@@ -0,0 +1,170 @@
...
+
+Those prover inputs are stored in memory prior to entering the kernel, and are used in the kernel to assert correct updates. The list of prover inputs necessary to the kernel is the following:
+\begin{enumerate}
+    \item the previous transaction number: $t_n$,
+    \item the gas used before executing the current transaction: $g\_u_0$,
+    \item the gas used after executing the current transaction: $g\_u_1$,
+    \item the block bloom filter before executing the current transaction: $b\_f_0$
+    \item the block bloom filter after executing the current transaction: $b\_f_1$,
+    \item the state, transaction and receipts MPTs before executing the current transaction: $\texttt{tries}_0$,
+    \item the state, transaction and receipts MPTs after executing the current transaction: $\texttt{tries}_1$,
```
hratoanina, line 26:
```suggestion
```
We only have the hashes for the tries after.

evm/spec/cpulogic.tex
```diff
@@ -0,0 +1,170 @@
...
+    \item the hash of all MPTs before executing the current transaction: $\texttt{digests}_0$,
+    \item the hash of all MPTs after executing the current transaction: $\texttt{digests}_1$.
+\end{enumerate}
+
+\paragraph*{Initialization:} The first step consists in initializing:
+\begin{itemize}
+    \item The shift table: it maps the number of bit shifts $s$ with its shifted value $1 << s$. Note that $0 \leq s \leq 255$.
+    \item The block bloom filter: the current block bloom filter is initialized with $b\_f_0$.
+    \item The initial MPTs: the initial state, transaction and receipt tries $\texttt{tries}_0$ are loaded from memory and hashed. The hashes are then compared to $digests\_0$.
```
hratoanina, line 35:
```suggestion
    \item The initial MPTs: the initial state, transaction and receipt tries $\texttt{tries}_0$ are loaded from memory and hashed. The hashes are then compared to $\texttt{digests}_0$.
```

hratoanina (approved):
LGTM.
The MPT hashing will probably be moved to a dedicated MPT section later.

PR #1378, opened 2023-11-22
Explain difference between simple opcodes and syscalls

LindaGuiga:


evm/spec/cpulogic.tex
```diff
@@ -64,6 +64,13 @@ \subsection{Kernel}
 \end{itemize}
 Finally, we hash the output of the RLP encoding, stored in ``RlpRaw'' -- unless it is already a hash.
 
+\subsection{Simple opcodes VS Syscalls}
+For simplicity and efficiency, EVM opcodes are categorized into two groups: ``simple opcodes'' and ``syscalls''. Simple opcodes are generated directly in Rust, in \href{https://github.com/0xPolygonZero/plonky2/blob/main/evm/src/witness/operation.rs}{operation.rs}. Every call to a simple opcode adds exactly one row to the \href{https://github.com/0xPolygonZero/plonky2/blob/main/evm/spec/tables/cpu.tex}{cpu table}. Syscalls are more complex structures written with simple opcodes, in the kernel.
+
+Whenever we encounter a syscall, we switch to kernel mode and execute its associated code. At the end of each syscall, we run EXIT\_KERNEL, which resets the kernel mode to its state right before the syscall. It also sets the PC to point the opcode right after the syscall.
```
hratoanina, line 70:
```suggestion
Whenever we encounter a syscall, we switch to kernel mode and execute its associated code. At the end of each syscall, we run EXIT\_KERNEL, which resets the kernel mode to its state right before the syscall. It also sets the PC to point to the opcode right after the syscall.
```

hratoanina (approved):

PR #1380, opened 2023-11-22
Add range check constraints for the looked table

LindaGuiga:
@pgebheim We noticed that there were some missing constraints to ensure that the tables against which we perform a range check are correctly constructed. This PR addresses this issue.


hratoanina (approved):
Looks good!

PR #1382, opened 2023-11-24
Fix genesis block number in `prove_block`

Nashtare:
Fix genesis block number on the `prove_block` side + add sanity check.

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -1059,25 +1059,18 @@ where
             }
 
             // Initialize block hashes.
-            let block_hashes_keys = TrieRootsTarget::SIZE * 2 + BlockMetadataTarget::SIZE
-                ..TrieRootsTarget::SIZE * 2
-                    + BlockMetadataTarget::SIZE
-                    + BlockHashesTarget::BLOCK_HASHES_SIZE
-                    - 8;
-
-            for i in 0..public_values.block_hashes.prev_hashes.len() - 1 {
-                let targets = h256_limbs::<F>(public_values.block_hashes.prev_hashes[i]);
-                for j in 0..8 {
-                    nonzero_pis.insert(block_hashes_keys.start + 8 * (i + 1) + j, targets[j]);
-                }
-            }
+            // Only current_hash needs to be set, as the previous hashes should be 0.
+            let block_hashes_current_start = TrieRootsTarget::SIZE * 2
+                + BlockMetadataTarget::SIZE
+                + BlockHashesTarget::BLOCK_HASHES_SIZE
+                - 8;
             let cur_targets = h256_limbs::<F>(public_values.block_hashes.prev_hashes[255]);
             for i in 0..8 {
-                nonzero_pis.insert(block_hashes_keys.end + i, cur_targets[i]);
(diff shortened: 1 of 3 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -819,7 +820,7 @@ where
         let has_not_parent_block = builder.sub(one, has_parent_block.target);
 
         // Check that the genesis block number is 0.
-        let gen_block_constr = builder.mul(has_not_parent_block, rhs.block_metadata.block_number);
+        let gen_block_constr = builder.mul(has_not_parent_block, lhs.block_metadata.block_number);
```
dvdplm, line 823: This seems like something the test suite should have caught? I see the test changes below but I'm unsure if this change is covered now?

evm/src/fixed_recursive_verifier.rs
```diff
@@ -1034,6 +1035,14 @@ where
                 .set_proof_with_pis_target(&self.block.parent_block_proof, parent_block_proof);
         } else {
             // Initialize genesis_state_trie, state_root_after, block hashes and the block number for correct connection between blocks.
+
+            if public_values.trie_roots_before.state_root
+                != public_values.extra_block_data.genesis_state_trie_root
+            {
+                return Err(anyhow::Error::msg(
+                    "Inconsistent pre-state for first block with genesis state.",
```
dvdplm, line 1043: Include the hashes in the error message? "Inconsistent pre-state for first block with genesis state. Expected {}, found {}" or so?

evm/src/fixed_recursive_verifier.rs
```diff
@@ -1071,14 +1080,18 @@ where
...
             let cur_targets = h256_limbs::<F>(public_values.block_hashes.prev_hashes[255]);
             for i in 0..8 {
-                nonzero_pis.insert(block_hashes_keys.end + i, cur_targets[i]);
+                nonzero_pis.insert(block_hashes_current_start + i, cur_targets[i]);
             }
 
             // Initialize the block number.
-            let block_number_key = TrieRootsTarget::SIZE * 2 + 6;
-            nonzero_pis.insert(block_number_key, F::NEG_ONE);
+            let block_number_key = TrieRootsTarget::SIZE * 2 + 3;
```
dvdplm, line 1093: A comment explaining the `3` would be good.  And a test?

wborgeaud (approved):
Looks good to me!

evm/src/fixed_recursive_verifier.rs
Nashtare: The test suite didn't catch this because tests were all badly configured (with the first block having index 0 instead of 1), and wasn't caught when running the Ethereum test suite either because we don't use recursion there (too slow)

evm/src/fixed_recursive_verifier.rs
Nashtare: I was discussing with the team of a way to refactor all those indices which are quite error prone. The 3 refers to the position to this field in `BlockMetadata` but tests not calling `prove_block` won't need it hence why it wasn't caught earlier.

evm/src/fixed_recursive_verifier.rs
Nashtare: I used `{:?}` format instead otherwise hashes would be pruned and only display first and last few bytes.

PR #1383, opened 2023-11-24
Add specs for MPTs

4l0n50:


pgebheim:
Amazing @wborgeaud can we get this merged before Monday?

evm/spec/mpts.tex
```diff
@@ -1,9 +1,16 @@
-\section{Merkle Patricia tries}
+\section{Merkle Patricia Tries}
 \label{tries}
+The \emph{EVM World state} is a representation of the different accounts at a particular time, as well as the last processed transactions together with their receipts. The world state is represented using \emph{Merkle Patricia Tries} (MPTs) \cite[App.~D]{yellowpaper}, and there are three different tries: the state trie, the transaction trie and the receipt trie.
+
+For each transaction we need to show that the prover knows preimages of the hashed initial and final EVM states.  At the onset of its execution, the kernel stores these three tries within the {\tt Segment::TrieData} segment. The prover loads the initial tries from the inputs into the memory. Subsequently, the tries are modified during transaction execution, inserting new nodes or the deleting of existing nodes.
```
dvdplm, line 5:
```suggestion
For each transaction we need to show that the prover knows preimages of the hashed initial and final EVM states.  When the kernel starts execution, it stores these three tries within the {\tt Segment::TrieData} segment. The prover loads the initial tries from the inputs into memory. Subsequently, the tries are modified during transaction execution, inserting new nodes or deleting existing nodes. 
```

evm/spec/mpts.tex
```diff
@@ -1,9 +1,16 @@
...
+
+For each transaction we need to show that the prover knows preimages of the hashed initial and final EVM states.  At the onset of its execution, the kernel stores these three tries within the {\tt Segment::TrieData} segment. The prover loads the initial tries from the inputs into the memory. Subsequently, the tries are modified during transaction execution, inserting new nodes or the deleting of existing nodes. 
+
+A MPT is composed of five different nodes: branch, extension, leaf, empty and digest nodes. Branch and leaf nodes might contain a payload whose format depends on the particular trie. The nodes are encoded, primarily using RLP encoding and Hex-prefix encoding (see \cite{yellowpaper} App. B and C, respectively). The resulting encoding is then hashed, following a strategy similar to that of normal Merkle trees, to generate the trie hashes.
```
dvdplm, line 7:
```suggestion
An MPT is composed of five different nodes: branch, extension, leaf, empty and digest nodes. Branch and leaf nodes might contain a payload whose format depends on the particular trie. The nodes are encoded, primarily using RLP encoding and Hex-prefix encoding (see \cite{yellowpaper} App. B and C, respectively). The resulting encoding is then hashed, following a strategy similar to that of normal Merkle trees, to generate the trie hashes.
```

evm/spec/mpts.tex
```diff
@@ -1,9 +1,16 @@
...
+
+Insertion and deletion is performed in the same way as other MPTs implementations, however we don't modify the MPT in memory but create a new one with the modifications. In the rest of this section we describe how the MPTs are represented in memory, how they are given as input, and how MPTs are hashed.
```
dvdplm, line 9:
```suggestion
Insertion and deletion is performed in the same way as other MPT implementations, however we don't modify the MPT in memory but create a new one with the modifications. In the rest of this section we describe how the MPTs are represented in memory, how they are given as input, and how MPTs are hashed.
```

evm/spec/mpts.tex
```diff
@@ -1,9 +1,16 @@
...
 
 \subsection{Internal memory format}
 
-Withour our zkEVM's kernel memory,
+The tries are stored in the kernel memory, specifically in the {\tt Segment:TrieData} segment. Each node type is stored as
```
dvdplm, line 13:
```suggestion
The tries are stored in kernel memory, specifically in the {\tt Segment:TrieData} segment. Each node type is stored as
```

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
   \item A digest node is encoded as $(\texttt{MPT\_NODE\_HASH}, d)$, where $d$ is a Keccak256 digest.
 \end{enumerate}
 
+On the other hand the values or payloads are represented differently depending on the particular trie.
+
+\subsubsection{State trie}
+The state trie payload contains the account data. Each account is stored in 4 contiguous memory addresses containing
```
dvdplm, line 25: Are the 4 memory segments all the same size? I.e. 4 x U256?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+\begin{enumerate}
+	\item the nonce,
+	\item the balance,
+	\item a pointer to the account's storage trie,
+	\item a hash of the account's code.
```
dvdplm, line 30: Is this the hash of `MPT_NODE_EMPTY` when there is no code?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+\end{enumerate}
+The storage trie payload in turn is a single word.
```
dvdplm, line 32: This seems to say that the storage payload cannot be bigger than one 256-bit word. Is that the case?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+
+\subsubsection{Transaction Trie}
+The transaction trie nodes contain the length of the RLP encoded transaction, followed by the bytes of the RLP encoding of the transaction.
```
dvdplm, line 35: Isn't the length contained in the RLP encoding itself? Is the length repeated?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+
+\subsubsection{Receipt Trie}
+The payload of the recipts trie is a receipt. Each receipt is stored as
+\begin{enumerate}
+	\item the length in words of the payload,
+	\item the status,
```
dvdplm, line 41: What are the possible statuses? Worth mentioning?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+
 
 \subsection{Prover input format}
 
 The initial state of each trie is given by the prover as a nondeterministic input tape. This tape has a slightly different format:
 \begin{enumerate}
   \item An empty node is encoded as $(\texttt{MPT\_NODE\_EMPTY})$.
-  \item A branch node is encoded as $(\texttt{MPT\_NODE\_BRANCH}, v_?, c_1, \dots, c_{16})$. Here $v_?$ consists of a flag indicating whether a value is present,\todo{In the current implementation, we use a length prefix rather than a is-present prefix, but we plan to change that.} followed by the actual value payload if one is present. Each $c_i$ is the encoding of a child node.
+  \item A branch node is encoded as $(\texttt{MPT\_NODE\_BRANCH}, v_?, c_1, \dots, c_{16})$. Here $v_?$ consists of a flag indicating whether a value is present, followed by the actual value payload if one is present. Each $c_i$ is the encoding of a child node.
   \item An extension node is encoded as $(\texttt{MPT\_NODE\_EXTENSION}, k, c)$, $k$ represents the part of the key associated with this extension, and is encoded as a 2-tuple $(\texttt{packed\_nibbles}, \texttt{num\_nibbles})$. $c$ is a pointer to a child node.
```
dvdplm, line 57:
```suggestion
  \item An extension node is encoded as $(\texttt{MPT\_NODE\_EXTENSION}, k, c)$, where $k$ represents the part of the key associated with this extension, and is encoded as a 2-tuple $(\texttt{packed\_nibbles}, \texttt{num\_nibbles})$. $c$ is a pointer to a child node.
```

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
   \item A leaf node is encoded as $(\texttt{MPT\_NODE\_LEAF}, k, v)$, where $k$ is a 2-tuple as above, and $v$ is a value payload.
   \item A digest node is encoded as $(\texttt{MPT\_NODE\_HASH}, d)$, where $d$ is a Keccak256 digest.
 \end{enumerate}
 Nodes are thus given in depth-first order, enabling natural recursive methods for encoding and decoding this format.
+The payload of state and receipt tries is given in the natural sequential way. The transaction an receipt payloads contain variable size data, thus the input is slightly different. The prover input for for the transactions is the transaction RLP encoding preceeded by its lenght. For the receipts is in the natural sequential way, except that topics and data are preceeded by their lengths, respectively.
```
dvdplm, line 62:
```suggestion
The payload of state and receipt tries is given in the natural sequential way. The transaction and receipt payloads contain variable size data, thus the input is slightly different. The prover input for for the transactions is the transaction RLP encoding preceded by its length. For the receipts is in the natural sequential way, except that topics and data are preceded by their lengths, respectively.
```

"For the receipts is in the natural sequential way" – not sure what you mean to say here; can you rephrase?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+
+\subsection{Encoding and Hashing}
+
+Encoding is recursively performed starting from the trie root. Leaf, branch and extension nodes are encoded as the RLP encoding of list containing the hex prefix encoding of the node key as well as
```
dvdplm, line 66:
```suggestion
Encoding is done recursively starting from the trie root. Leaf, branch and extension nodes are encoded as the RLP encoding of list containing the hex prefix encoding of the node key as well as
```

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+
+\begin{description}
+	\item[Leaf Node:] the encoding of the the payload,
+	\item[Branch Node:] the hash or encoding of the 16 children and the encoding of the payload,
```
dvdplm, line 70: How is hash or encoding of children decided? Do the RLP encodings of 16 children really fit in a branch node?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+	\item[Extension Node:] the hash or encoding of the child and the encoding of the payload.
+\end{description}
+For the rest of the nodes we have:
+\begin{description}
+	\item[Empty Node:] the encoding of an empty node is {\tt 0x80},
+	\item[Digest Node:] the encoding of a digest node stored as $({\tt MPT\_HASH\_NODE}, d)$ is $d$.
```
dvdplm, line 76: I don't think the digest node was described previously?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+The payloads in turn are RLP encoded as follows
+\begin{description}
+	\item[State Trie:] Encoded as a list containing nonce, balance, storage trie hash and code hash.
+	\item[Storage Trie:] The RLP encoding of the value (thus the double RLP encoding)
+	\item[Transaction Trie:] The RLP encoded transaction.
+	\item[Receipt Trie:] Depending on the transaction type it's encoded as ${\sf RLP}({\sf RLP}({\tt receipt}))$ for Legacy transactions or ${\sf RLP}({\tt txn\_type}||{\sf RLP}({\tt receipt}))$ for transactions of type 1 or 2. Each receipt is encoded as a list containing:
+	\begin{enumerate}
+		\item the status,
+		\item the cumulative gas used,
+		\item the bloom filter, stored as a list of length 256.
```
dvdplm, line 88: Maybe it's out of scope for this text, but I don't think the purpose of the bloom is explained anywhere?

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+		\item the list of topics
+		\item the data string.
+	\end{enumerate}
+\end{description}
+
+Once a node is encoded is written on the {\tt Segment::RlpRaw} segment as a sequence of bytes. Then the RLP encoded data is hashed if the length of the data is more than 32 bytes. Otherwise we return the encoding. Further details can be found in the \href{https://github.com/0xPolygonZero/plonky2/tree/main/evm/src/cpu/mpt/hash}{mpt hash folder}.
```
dvdplm, line 94:
```suggestion
Once a node is encoded it is written to the {\tt Segment::RlpRaw} segment as a sequence of bytes. Then the RLP encoded data is hashed if the length of the data is more than 32 bytes. Otherwise we return the encoding. Further details can be found in the \href{https://github.com/0xPolygonZero/plonky2/tree/main/evm/src/cpu/mpt/hash}{mpt hash folder}.
```

evm/spec/mpts.tex
4l0n50: Yes, in 4.1, 5th item

evm/spec/mpts.tex
4l0n50: Idk. but as you mention, for the purpose of this section it's enough to say it is a string of 256 bytes

evm/spec/mpts.tex
```diff
@@ -12,15 +19,76 @@ \subsection{Internal memory format}
...
+	\item[Empty Node:] the encoding of an empty node is {\tt 0x80},
+	\item[Digest Node:] the encoding of a digest node stored as $({\tt MPT\_HASH\_NODE}, d)$ is $d$.
+\end{description}
+
+The payloads in turn are RLP encoded as follows
+\begin{description}
+	\item[State Trie:] Encoded as a list containing nonce, balance, storage trie hash and code hash.
+	\item[Storage Trie:] The RLP encoding of the value (thus the double RLP encoding)
+	\item[Transaction Trie:] The RLP encoded transaction.
+	\item[Receipt Trie:] Depending on the transaction type it's encoded as ${\sf RLP}({\sf RLP}({\tt receipt}))$ for Legacy transactions or ${\sf RLP}({\tt txn\_type}||{\sf RLP}({\tt receipt}))$ for transactions of type 1 or 2. Each receipt is encoded as a list containing:
```
wborgeaud, line 84: Shouldn't it be just `{\tt txn\_type}||{\sf RLP}({\tt receipt})` for type 1 or 2 txns?

wborgeaud (approved):
Looks good to me!

evm/spec/mpts.tex
4l0n50: Indeed, but adding the length explicitly makes things slightly easier at a negligible cost.

evm/spec/mpts.tex
4l0n50: For this part of document is enough to know that it fits in a 256 bits. Idk about the actual values

evm/spec/mpts.tex
```diff
@@ -1,9 +1,16 @@
...
+
+For each transaction we need to show that the prover knows preimages of the hashed initial and final EVM states.  When the kernel starts execution, it stores these three tries within the {\tt Segment::TrieData} segment. The prover loads the initial tries from the inputs into memory. Subsequently, the tries are modified during transaction execution, inserting new nodes or deleting existing nodes. 
+
+An MPT is composed of five different nodes: branch, extension, leaf, empty and digest nodes. Branch and leaf nodes might contain a payload whose format depends on the particular trie. The nodes are encoded, primarily using RLP encoding and Hex-prefix encoding (see \cite{yellowpaper} App. B and C, respectively). The resulting encoding is then hashed, following a strategy similar to that of normal Merkle trees, to generate the trie hashes.
+
+Insertion and deletion is performed in the same way as other MPT implementations, however we don't modify the MPT in memory but create a new one with the modifications. In the rest of this section we describe how the MPTs are represented in memory, how they are given as input, and how MPTs are hashed.
```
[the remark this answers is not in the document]
4l0n50: To be sure, is it just the insertion of extension nodes where the modified node is copied? (at least that's what I can see from the code in [#119](https://github.com/0xPolygonZero/plonky2/pull/1190.)

evm/spec/mpts.tex
4l0n50: Hmmm, at least the comment on the code explicitly says 
`// There is a double encoding! What we compute is:
    // either RLP(RLP(receipt)) for Legacy transactions or RLP(txn_type||RLP(receipt)) for transactions of type 1 or 2`
    @LindaGuiga or @hratoanina?

LindaGuiga: Indeed, we are doubly encoding (it didn't work with a simple encoding). I believe the value is the simple RLP encoding, and the actual node encoding is then the RLP encoding of the value (and therefore the double encoding). @hratoanina?

evm/spec/mpts.tex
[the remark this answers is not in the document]
wborgeaud: Yeah I think that's basically it

evm/spec/mpts.tex
wborgeaud: > Doesn't the `185` indicate the start of an RLP encoding for (0x2 || RLP(receipt)) here? (with `249` starting the RLP encoding of the receipt)?

Sure, but that's just because the RLP of a list is the concatenation of the RLP of the items of the list (prefixed by the length).

PR #1385, opened 2023-11-24
chore: run `cargo fmt` on all files

dvdplm:
After `rustc 1.76.0-nightly (a1a37735c 2023-11-23)` it looks like `rustfmt` has changed. This PR is the result of running `cargo fmt` in the root. No other changes.


dvdplm:
Turns out this was a bug where macos `rustfmt` behaved differently from linux. Bug https://github.com/rust-lang/rustfmt/issues/5964

PR #1388, opened 2023-11-28
Remove extra rows in BytePackingStark

LindaGuiga:
This PR changes the way `BytePackingStark` generates its rows: it generates only one row per operation (or no row for an empty byte sequence).
It reduces greatly the number of rows: for add11_yml, we go from 119,388 to 20,028 rows in `BytePackingStark`.
This also leads to fewer constraints and one less column.


Nashtare:
Are empty byte sequences possible now?

LindaGuiga:
> > or no row for an empty byte sequence
> 
> Are empty byte sequences possible now?

When running add11, I noticed some operations had empty byte sequences (I did not change the way operations are added, so I assume they were already there). But no row was added in that case either, because there was a `for` loop on the sequence's elements

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -1,28 +1,22 @@
 //! This crate enforces the correctness of reading and writing sequences
 //! of bytes in Big-Endian ordering from and to the memory.
 //!
-//! The trace layout consists in N consecutive rows for an `N` byte sequence,
-//! with the byte values being cumulatively written to the trace as they are
-//! being processed.
+//! The trace layout consists in one row for an `N` byte sequence (where `N` > 0).
```
Nashtare, line 4:
```suggestion
//! The trace layout consists in one row for an `N` byte sequence (where 32 ≥ `N` > 0).
```

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -1,28 +1,22 @@
...
 //!
-//! At row `i` of such a group (starting from 0), the `i`-th byte flag will be activated
-//! (to indicate which byte we are going to be processing), but all bytes with index
-//! 0 to `i` may have non-zero values, as they have already been processed.
+//! At each row the `i`-th byte flag will be activated to indicate a sequence of
+//! length i.
```
Nashtare, line 7: Wouldn't it be length i + 1? (that's what's said in the specs)

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -165,12 +171,17 @@ impl<F: RichField + Extendable<D>, const D: usize> BytePackingStark<F, D> {
         ops: Vec<BytePackingOp>,
         min_rows: usize,
     ) -> Vec<[F; NUM_COLUMNS]> {
-        let base_len: usize = ops.iter().map(|op| op.bytes.len()).sum();
+        let base_len: usize = ops
+            .iter()
+            .map(|op| if !op.bytes.is_empty() { 1 } else { 0 })
+            .sum();
```
Nashtare, line 177:
```suggestion
        let base_len: usize = ops.iter().map(|op| usize::from(!op.bytes.is_empty())).sum();
```

evm/src/byte_packing/byte_packing_stark.rs
```diff
@@ -194,34 +205,28 @@ impl<F: RichField + Extendable<D>, const D: usize> BytePackingStark<F, D> {
...
 
-            // Update those fields for the next row
-            row[index_bytes(i)] = F::ZERO;
-            row[ADDR_VIRTUAL] -= F::ONE;
+            Some(row)
+        } else {
+            None
         }
-
-        rows
```
Nashtare, line 224: Why doing all of this if the sequence is empty and we return `None` in the end?
We should probably just filter ahead empty sequences so that this always returns a row (for non-empty ones).

evm/src/witness/traces.rs
```diff
@@ -69,7 +69,11 @@ impl<T: Copy> Traces<T> {
                     Operation::RangeCheckOperation { .. } => 1,
                 })
                 .sum(),
-            byte_packing_len: self.byte_packing_ops.iter().map(|op| op.bytes.len()).sum(),
+            byte_packing_len: self
+                .byte_packing_ops
+                .iter()
+                .map(|op| if !op.bytes.is_empty() { 1 } else { 0 })
```
Nashtare, line 75:
```suggestion
                .map(|op| usize::from(!op.bytes.is_empty()))
```

Nashtare (changes requested):
Thanks! I think you forgot to update the ranges for empty_txn_list and log_opcodes, the tests are currently failing.

Nashtare (approved):

PR #1390, opened 2023-11-28
Remove bootstrapping

hratoanina:
By removing the constraint of zero-initializing for the code segment of the kernel context, we can non-deterministically initialize it with the correct values, and then check them with a call to KECCAK_GENERAL. Removes ~17k CPU rows, ~70k Memory rows and the `is_bootstrap_kernel` CPU column, at the cost of one extra Memory column.

https://github.com/0xPolygonZero/plonky2/issues/1338


evm/src/generation/mod.rs
```diff
@@ -149,6 +149,17 @@ fn apply_metadata_and_tries_memops<F: RichField + Extendable<D>, const D: usize>
...
         ),
+        (
+            GlobalMetadata::KernelHash,
+            KERNEL
+                .code_hash
+                .iter()
+                .enumerate()
+                .fold(0.into(), |acc, (i, &elt)| {
+                    acc + (U256::from(elt) << (224 - 32 * i))
+                }),
```
wborgeaud, line 160: Why not something like `U256::from_big_endian(&KERNEL.code_hash)`?

evm/src/generation/mod.rs
```diff
@@ -216,6 +227,19 @@ fn apply_metadata_and_tries_memops<F: RichField + Extendable<D>, const D: usize>
     state.traces.memory_ops.extend(ops);
 }
 
+fn initialize_kernel_code<F: RichField + Extendable<D>, const D: usize>(
+    state: &mut GenerationState<F>,
+) {
+    for (i, &byte) in enumerate(KERNEL.code.iter()) {
+        let address = MemoryAddress {
+            context: 0,
+            segment: 0,
```
wborgeaud, line 236: nit: Can use `Segment::Code as usize` for clarity.

evm/src/memory/memory_stark.rs
```diff
@@ -458,28 +463,33 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F
...
+            // the first operation of a new address is a read, then its value must be 0.
+            let first_read_value = builder.mul_extension(next_values_limbs[i], new_first_read);
+            let first_read_user_context_constraint =
+                builder.mul_extension(next_addr_context, first_read_value);
+            yield_constr.constraint_transition(builder, first_read_user_context_constraint);
+            // For any segment outside the code segment, memory is initialized with 0. Note that the previous constraint
+            // ensure that only the code segment at context 0 is not zero-initialized.
+            let kernel_context_initializing_constraint =
+                builder.mul_extension(next_addr_segment, first_read_value);
+            yield_constr.constraint_transition(builder, kernel_context_initializing_constraint);
```
wborgeaud, line 491: This is fine for now, but ideally I think we should allow non-zero initialized memory for all code segments. That way we wouldn't have to load contract code from prover inputs. Obviously we would need to take care of some security issues when doing that.

evm/spec/tables/memory.tex
```diff
@@ -64,6 +65,15 @@ \subsubsection{Timestamps}
...
   \item Some tables other than CPU can generate memory operations, like KeccakSponge. When this happens, these operations all have the timestamp of the CPU row of the instruction which invoked the table (for KeccakSponge, KECCAK\_GENERAL).
 \end{itemize}
+
+\subsubsection{Memory initialization}
+
+By default, all memory is zero-initialized. However, to save numerous writes, we allow some specific segments to be initialized with arbitrary values.
+
+\begin{itemize}
+  \item The read-only kernel code (in segment 0, context 0) is initialized with its correct values. It's checked by hashing the segment and verifying
+that the hash value matches a verifyer-provided one.
```
wborgeaud, line 78:
```suggestion
that the hash value matches a verifier-provided one.
```

wborgeaud (approved):
LGTM

Nashtare:
You'd need to update the preprocessed ranges accordingly for `empty_txn_list` following reduction on CPU length.
In addition, it seems `log_opcode` is currently failing on this branch (not on main). It would need to be fixed before merging.

Nashtare (changes requested):
You'd need to update the preprocessed ranges accordingly for `empty_txn_list` following reduction on CPU length.
In addition, it seems `log_opcode` is currently failing on this branch (not on main). It would need to be fixed before merging.

evm/src/memory/memory_stark.rs
hratoanina: Some must be zero-initialized for soundness (typically the main memory segment of any non-zero context), so I'd very much rather have all segments zero-initialized by default, with specific exceptions (like the code segment) made, as long as the constraints don't become too complicated and it doesn't add a prohibitive amount of extra columns.

evm/src/memory/memory_stark.rs
```diff
@@ -325,24 +330,24 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F
...
+                .constraint_transition(next_addr_segment * new_first_read * next_values_limbs[i]);
         }
-
-        // Check the range column: First value must be 0,
-        // and intermediate rows must increment by 1.
-        let rc1 = local_values[COUNTER];
-        let rc2 = next_values[COUNTER];
-        yield_constr.constraint_first_row(rc1);
-        let incr = rc2 - rc1;
-        yield_constr.constraint_transition(incr - P::Scalar::ONES);
```
LindaGuiga, line 345: I don't think these constraints should be removed.

evm/src/memory/memory_stark.rs
```diff
@@ -458,28 +463,33 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for MemoryStark<F
...
         }
-
-        // Check the range column: First value must be 0,
-        // and intermediate rows must increment by 1.
-        let rc1 = local_values[COUNTER];
-        let rc2 = next_values[COUNTER];
-        yield_constr.constraint_first_row(builder, rc1);
-        let incr = builder.sub_extension(rc2, rc1);
-        let t = builder.sub_extension(incr, one);
-        yield_constr.constraint_transition(builder, t);
```
LindaGuiga, line 482: I don't think these constraints should be removed.

evm/src/memory/memory_stark.rs
wborgeaud: Right, I was talking about the code segments, i.e., `segment=0`. If I'm correctly understanding your constraints, this allows non-zero initialized memory only for the kernel code segment `context=0, segment=0`.

hratoanina: Oh yeah, sorry. Indeed, we should unconstrain the code segment for all contexts in the future.

evm/src/generation/mod.rs
hratoanina: `KERNEL.code_hash` is an array of `u32`, do you have an easier way to do it?

Nashtare: IIRC, this was stored in limbs only because of KeccakSpongeStark. Now that we do the check inside the Kernel, we can change `code_hash` type to be the actual keccak output (i.e. `[u8; 32]`) and not have issues with the conversion above anymore.

evm/src/cpu/kernel/asm/main.asm
```diff
@@ -1,5 +1,17 @@
 global main:
-    // First, initialise the shift table
+    // First, hash the kernel code
+    %mload_global_metadata(@GLOBAL_METADATA_KERNEL_LEN)
+    PUSH 0
+    DUP1
+    DUP1
```
Nashtare, line 6: nit: I think it's cheaper on Memory side to call `PUSH 0` again instead of `DUP1`?

evm/src/memory/columns.rs
```diff
@@ -35,8 +35,11 @@ pub(crate) const CONTEXT_FIRST_CHANGE: usize = VALUE_START + VALUE_LIMBS;
 pub(crate) const SEGMENT_FIRST_CHANGE: usize = CONTEXT_FIRST_CHANGE + 1;
 pub(crate) const VIRTUAL_FIRST_CHANGE: usize = SEGMENT_FIRST_CHANGE + 1;
 
+// Used to zero-initialize (or not) the memory.
+pub(crate) const NEW_FIRST_READ: usize = VIRTUAL_FIRST_CHANGE + 1;
```
Nashtare, line 39: nit: may deserve a clearer comment / var naming? A bit hard to parse without looking at the rest of the code.

Nashtare (commented):
I think the failure in `log_opcode` is due to the fact that you forgot to update the memory bus with the new `GlobalMetadata` related to the Kernel on the recursive side. Not sure why it isn't caught in the `empty_txn_list` case though.

hratoanina:
@wborgeaud I changed the zero-initializing constraint to exclude all code segments. This might make it unsound for now, but I'll work on code loading soon.

Nashtare (approved):
Thanks. Can you fetch latest `main` before you merge this PR? as the ranges for log_opcodes would break `main` otherwise

PR #1392, opened 2023-11-29
`PolynomialBatch::from_values` takes refs

dvdplm:
Fixes a TODO in the code to make `PolynomialBatch::from_values` take `PolynomialValues` by ref instead of by value. 
Doing this doesn't really fix anything given that the subsequent FFT operations require ownership of the values. Still, 
I think it's better to have the clone happen in a single place than spread all over.

### DIFF
--- a/plonky2/src/fri/oracle.rs
+++ b/plonky2/src/fri/oracle.rs
@@ -55,7 +55,7 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
 {
     /// Creates a list polynomial commitment for the polynomials interpolating the values in `values`.
     pub fn from_values(
-        values: Vec<PolynomialValues<F>>,
+        values: &[PolynomialValues<F>],
         rate_bits: usize,
         blinding: bool,
         cap_height: usize,
@@ -65,7 +65,11 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
         let coeffs = timed!(
             timing,
             "IFFT",
-            values.into_par_iter().map(|v| v.ifft()).collect::<Vec<_>>()
+            values
+                .into_par_iter()
+                .cloned()
+                .map(|v| v.ifft())
+                .collect::<Vec<_>>()
         );
 
         Self::from_coeffs(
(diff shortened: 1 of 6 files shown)


plonky2/src/fri/oracle.rs
```diff
@@ -65,7 +65,11 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
         let coeffs = timed!(
             timing,
             "IFFT",
-            values.into_par_iter().map(|v| v.ifft()).collect::<Vec<_>>()
+            values
+                .into_par_iter()
+                .cloned()
```
dvdplm, line 70: I don't think this is easy to get rid of. :/

plonky2/src/fri/oracle.rs
```diff
@@ -55,7 +55,7 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize>
 {
     /// Creates a list polynomial commitment for the polynomials interpolating the values in `values`.
     pub fn from_values(
-        values: Vec<PolynomialValues<F>>,
+        values: &[PolynomialValues<F>],
```
wborgeaud, line 58: I think it's better to keep a `Vec<_>` here so that it's explicit that this function requires cloning the data.

dvdplm: You know what, I think you're right. I made an attempt to take this further but it's really built to take ownership of the data. If we want less cloning we should look elsewhere. Sigh.

dvdplm: @Nashtare Unless you think differently I'll revert this change and just remove the TODO.

Nashtare:
Sorry @dvdplm I had missed the review request.
Given that the initial changes here reverted to just removing the TODOs without addressing anything, I am wondering if we shouldn't at least leave them for now, in the hope that maybe some more consequent refactoring on this part of the prover may actually remove the heavy cloning?

Nashtare (commented):
Sorry @dvdplm I had missed the review request.
Given that the initial changes here reverted to just removing the TODOs without addressing anything, I am wondering if we shouldn't at least leave them for now, in the hope that maybe some more consequent refactoring on this part of the prover may actually remove the heavy cloning?

dvdplm:
I think that after this investigation it's clear that it'd require a pretty big re-architecting to fix this so if anything there could be a ticket to look into what the best way to go about that is. And measure if it's actually a problem. To me, a `TODO:` should be about small things and local to the place in the code where they appear (and preferably always be matched with a ticket nr), otherwise it just becomes misleading. 
It's like having a `TODO: Make plonky2 fast`, which is a true `todo` but also not something that belong to a single line of code or even single source file.

Nashtare (approved):

PR #1393, opened 2023-11-29
Add exceptions handling to the interpreter

LindaGuiga:
This PR adds exceptions handling to the interpreter. It also adds a test with a stack underflow exception in the user code (in add11.rs).


evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -369,26 +397,35 @@ fn sload() -> Result<()> {
...
+        .expect("The stack length should not be empty.");
     // Now, execute mpt_hash_state_trie. We check that the state trie has not changed.
     let mpt_hash_state_trie = KERNEL.global_labels["mpt_hash_state_trie"];
     interpreter.generation_state.registers.program_counter = mpt_hash_state_trie;
     interpreter.set_is_kernel(true);
     interpreter.set_context(0);
-    interpreter.push(0xDEADBEEFu32.into());
+    interpreter
+        .push(0xDEADBEEFu32.into())
+        .expect("Stack should not overflow.");
```
Nashtare, line 428: nit: Maybe rather 
```suggestion
        .expect("The stack should not overflow.");
```
to match `expect` messages above?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -103,6 +123,12 @@ pub(crate) fn run<'a>(
     Ok(interpreter)
 }
 
+enum InterpreterMemOpKind {
+    Push(usize),                      // We need to provide the context
+    Pop(U256, usize), // If you pop a certain value, you need to push it back to the correct context when reverting
+    Write(U256, usize, usize, usize), // If we write a value at a certain address, we need to write the old value back when reverting
```
Nashtare, line 129: very nit: because of the automatic space alignement, it may be better to have them as 
```suggestion
    /// We need to provide the context.
    Push(usize),
    /// If we pop a certain value, we need to push it back to the correct context when reverting.
    Pop(U256, usize),
    /// If we write a value at a certain address, we need to write the old value back when reverting.
    Write(U256, usize, usize, usize),
```
note that I've added an extra `/` and updated the second comment for consistency with the others

Nashtare: Also may be nice to have a descriptive documentation at the top of the enum

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -51,6 +60,17 @@ pub(crate) struct Interpreter<'a> {
     pub(crate) debug_offsets: Vec<usize>,
     running: bool,
     opcode_count: [usize; 0x100],
+    memops: Vec<InterpreterMemOpKind>,
+}
+
+struct InterpreterRegistersState {
```
Nashtare, line 66: nit: maybe add a top-level struct documentation

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -51,6 +60,17 @@ pub(crate) struct Interpreter<'a> {
...
+    kernel_mode: bool,
+    context: usize,
+    registers: RegistersState,
+}
+struct InterpreterCheckpoint {
```
Nashtare, line 71: nit: missing space + maybe add a top-level struct documentation

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -482,6 +627,24 @@ impl<'a> Interpreter<'a> {
...
+            let gas_limit =
+                TryInto::<u64>::try_into(self.generation_state.memory.get(gas_limit_address));
+            match gas_limit {
+                Ok(limit) => {
+                    if self.generation_state.registers.gas_used > limit {
+                        return Err(ProgramError::OutOfGas);
+                    }
+                }
+                Err(_) => return Err(ProgramError::IntegerTooLarge),
+            }
```
Nashtare, line 645:
```suggestion
            let gas_limit = u256_to_usize(self.generation_state.memory.get(gas_limit_address))? as u64;
            if self.generation_state.registers.gas_used > limit {
                return Err(ProgramError::OutOfGas);
            }
```

Nashtare (approved):
Thank you, looks good! Only nitpicky comments.

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -383,6 +369,9 @@ fn sstore() -> Result<()> {
     interpreter
         .push(0xDEADBEEFu32.into())
         .expect("The stack should not overflow");
+    interpreter
+        .push(1.into()) // Initia length of the trie data segment, unused.
```
Nashtare, line 373:
```suggestion
        .push(1.into()) // Initial length of the trie data segment, unused.
```

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -469,7 +458,10 @@ fn sload() -> Result<()> {
     interpreter.set_context(0);
     interpreter
         .push(0xDEADBEEFu32.into())
-        .expect("Stack should not overflow.");
+        .expect("The stack should not overflow.");
+    interpreter
+        .push(1.into()) // Initia length of the trie data segment, unused.
```
Nashtare, line 463:
```suggestion
        .push(1.into()) // Initial length of the trie data segment, unused.
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -103,6 +128,16 @@ pub(crate) fn run<'a>(
     Ok(interpreter)
 }
 
+/// Different types of Memory operations in the interpreter, and the data required to revert them.
+enum InterpreterMemOpKind {
+    /// We need to provide the context.
+    Push(usize),
+    /// If we pop a certain value, we need to push it back to the correct context when reverting.
+    Pop(U256, usize),
+    /// If we write a value at a certain address, we need to write the old value back when reverting
```
Nashtare, line 137: sorry, my OCD is kicking in again 😅 (for consistency)
```suggestion
    /// If we write a value at a certain address, we need to write the old value back when reverting.
```

PR #1395, opened 2023-11-30
Remove intermediary block bloom filters

Nashtare:
This PR removes the computation of the intermediary bloom filters pre & post transaction execution.
The block bloom is kept as PI and consistency check is maintained as part of `BlockMetadata::connect` within aggregation proofs. The transaction bloom logic is unchanged.

### DIFF
--- a/evm/src/memory/segments.rs
+++ b/evm/src/memory/segments.rs
@@ -50,27 +50,25 @@ pub(crate) enum Segment {
     SelfDestructList = 25,
     /// Contains the bloom filter of a transaction.
     TxnBloom = 26,
-    /// Contains the computed bloom filter of a block.
-    BlockBloom = 27,
-    /// Contains the final block bloom.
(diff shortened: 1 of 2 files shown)


evm/src/memory/segments.rs
```diff
@@ -52,9 +52,7 @@ pub(crate) enum Segment {
     TxnBloom = 26,
     /// Contains the computed bloom filter of a block.
     BlockBloom = 27,
```
hratoanina, line 54: I think this segment can be removed now?

hratoanina (approved):
Looks good!

evm/src/memory/segments.rs
Nashtare: Ah indeed, good catch

Nashtare:
PI is public inputs. We keep the bloom filter included in the block header (the final bloom), but we don't construct it internally anymore

PR #1397, opened 2023-11-30
Constrain first offset of a segment

hratoanina:
There seems to be an issue with the Memory constraints. The rangecheck ensures that, when the offset changes (but not the context and the segment), the new offset is somewhat close to the previous one. However, when changing context and/or segment, we don't check that the *initial* offset is small. Without this check, it's possible to jump from `(context_a, segment_a, offset_a)` to `(context_b, segment_b, -1)`, then to `(context_b, segment_b, 0)` without breaking any constraint.

This fix introduces a new column to hold `offset` when we change context and/or segment, and 0 otherwise. Range-checking this column should restore soundness.
It's a bit of a waste since this column is literally `addr_offset` with filter column `context_first_change + segment_first_change`, but our lookup implementation doesn't support filters. It could be added as a feature, but this would require some non-trivial changes.

Opening it as a draft PR to discuss it and see if I'm missing something.


Nashtare:
cc @pgebheim, we may need to inform the auditors on this.

evm/spec/tables/memory.tex
```diff
@@ -43,8 +43,11 @@ \subsection{Memory}
   \item $e_i (e_i - 1) = 0$,
   \item $(1 - e_i) (a_{i + 1} - a_i) = 0$,
   \item $c_i < 2^{32}$.
+  \item $e_i (v_{i + 1}) < 2^{32}$
```
LindaGuiga, line 46: `v` is the value being read or written. I think you should replace `v` by `a` (the address) or something like `a_{virtual}`.

evm/spec/tables/memory.tex
```diff
@@ -43,8 +43,11 @@ \subsection{Memory}
...
 \end{enumerate}
-The last constraint emulates a comparison between two addresses or timestamps by bounding their difference; this assumes that all addresses and timestamps fit in 32 bits and that the field is larger than that.
+The third constraint emulates a comparison between two addresses or timestamps by bounding their difference; this assumes that all addresses and timestamps fit in 32 bits and that the field is larger than that.
+
+The fourth constraint checks that the first value of a new address is indeed the smallest in history. Without this check, addresses could start at -1 (i.e. $p - 2$) and then increase properly.
```
LindaGuiga, line 50: It sounds a bit like you are talking about the actual value, and the address itself here.
```suggestion
The fourth constraint checks that the virtual address of a new address is indeed the smallest in history. Without this check, addresses could start at -1 (i.e. $p - 2$) and then increase properly.
```

evm/src/memory/memory_stark.rs
```diff
@@ -127,6 +127,11 @@ pub(crate) fn generate_first_change_flags_and_rc<F: RichField>(
         let address_changed =
             row[CONTEXT_FIRST_CHANGE] + row[SEGMENT_FIRST_CHANGE] + row[VIRTUAL_FIRST_CHANGE];
         row[INITIALIZE_AUX] = next_segment * address_changed * next_is_read;
+
+        let next_row = trace_rows[idx + 1].as_mut_slice();
+        if context_first_change || segment_first_change {
+            next_row[FIRST_OFFSET] = next_virt;
```
LindaGuiga, line 133: Is there a reason why we're storing it in the next row and not the current one? (it's fine, I was just wondering)

LindaGuiga (approved):
LGTM, thanks!

evm/spec/tables/memory.tex
```diff
@@ -56,6 +56,9 @@ \subsubsection{Virtual memory}
 \end{enumerate}
 The comparisons now involve several columns, which requires some minor adaptations to the technique described above; we will leave these as an exercise to the reader.
 
+Note that an additional constarint check is required: whenever we change the context or the segment, the virtual address must be range-checked to $2^{32}$.
```
LindaGuiga, line 59:
```suggestion
Note that an additional constraint check is required: whenever we change the context or the segment, the virtual address must be range-checked to $2^{32}$.
```

evm/src/memory/memory_stark.rs
```diff
@@ -76,7 +76,7 @@ impl MemoryOp {
     }
 }
 
-/// Generates the `_FIRST_CHANGE` columns and the `RANGE_CHECK` column in the trace.
+/// Generates the `_FIRST_CHANGE` columns, the `RANGE_CHECK` column and the `FIRST_OFFSET` column in the trace.
```
LindaGuiga, line 79:
```suggestion
/// Generates the `_FIRST_CHANGE` columns and the `RANGE_CHECK` column in the trace.
```

LindaGuiga (approved):

PR #1398, opened 2023-11-30
Use logUp for CTLs

hratoanina:
This PR changes the CTL argument for logUp. This doesn't bring any benefits on its own, but it allows further optimizations: degree 2 CTL filters and CTL bundling.


evm/spec/framework.tex
```diff
@@ -151,4 +110,46 @@ \subsubsection{Constraints}
...
+  \item constructs a running sum $Z_i^l$ for each table looking into $S$ (called looking sums here),
+  \item constructs a running sum $Z^S$ for $S$ (called looked sum here),
+  \item sends the final value for each running sum $Z_{i, 0}^l$ and $Z^S_0$ to the verifier,
+  \item sends a commitment to $Z_i^l$  and $Z^S$ to the verifier.
+\end{enumerate}
+Then, for each STARK $S$, the verifier:
+\begin{enumerate}
+  \item computes the sum $Z = \sum_i Z_{i, 0}^l$,
+  \item checks that $Z = Z^S_0$,
+  \item checks that each $Z_i^l$  and $Z^S$ was correctly constructed.
```
Nashtare, line 154: can't it be left to its original position?

evm/src/cross_table_lookup.rs
```diff
@@ -623,11 +623,11 @@ fn partial_products<F: Field>(
                 .iter()
                 .map(|c| c.eval_table(trace, i))
                 .collect::<Vec<_>>();
-            partial_prod *= challenge.combine(evals.iter());
+            partial_sum += challenge.combine(evals.iter()).inverse();
```
Nashtare, line 626: this looks heavy. Every `inverse` with Goldilocks is about 80 multiplications IIRC, better to do something with batch_inverse, I think that's what's done in `lookup.rs`?

Nashtare: also I think some methods may be partial duplicates now of the logUp implementation used in lookups for range-checks. May be nice to unify things if possible under common utility methods

evm/spec/framework.tex
hratoanina: It could, but since logUp is described in details in the Range-check section I wanted to move the CTLs afterwards.

Nashtare: Seems to me like CTLs are much more core than RCs, and should hence be introduced where they were originally (eventually with a footnote saying logUp is described in more depth later), but up to you

evm/src/cross_table_lookup.rs
hratoanina: I added batch inversion.

Mutualizing methods is actually non-trivial, since the logic between CTLs and range-checks is not identical. This should be done at some point though (this would make CTL batching much easier, as well as filters for range-checks), but it's complex enough to be a PR on its own.

hratoanina:
I reverted back to the previous order, with the CTLs before range checks. In the future it would be nicer to have a dedicated section for logUp anyway.

PR #1399, opened 2023-12-01
Refactor encode funcs

4l0n50:
This PR changes how branch and empty nodes are encoded:
- The encoding of an empty node is stored at position @ENCODED_EMPTY_NODE_POS of the rlp segment.
- For encoding a branch node we write the child encoded values directly to the rlp segment, instead of storing first the values on SEGMENT_TRIE_ENCODED_CHILD
It saves ~2500 cycles with add11_yml.


evm/src/cpu/kernel/asm/mpt/hash/hash.asm
```diff
@@ -71,7 +71,7 @@ global encode_or_hash_node:
 encode_or_hash_concrete_node:
     %stack (node_type, node_ptr, encode_value) -> (node_type, node_ptr, encode_value, maybe_hash_node)
     %jump(encode_node)
-maybe_hash_node:
+global maybe_hash_node:
```
Nashtare, line 74: no need for a global

evm/src/cpu/kernel/asm/mpt/hash/hash.asm
```diff
@@ -84,7 +84,7 @@ maybe_hash_node:
     KECCAK_GENERAL
     %stack (hash, retdest) -> (retdest, hash, 32)
     JUMP
-pack_small_rlp:
+global pack_small_rlp:
```
Nashtare, line 87: same as above

evm/src/cpu/kernel/asm/mpt/hash/hash.asm
```diff
@@ -99,7 +99,7 @@ after_packed_small_rlp:
 //
 // Pre stack: node_type, node_ptr, encode_value, retdest
 // Post stack: result_ptr, result_len
-encode_node:
+global encode_node:
```
Nashtare, line 102: same as above.

evm/src/cpu/kernel/asm/mpt/hash/hash.asm
```diff
@@ -117,52 +117,27 @@ encode_node:
...
 
-    // Get the next unused offset within the encoded child buffers.
-    // Then immediately increment the next unused offset by 16, so any
-    // recursive calls will use nonoverlapping offsets.
-    // TODO: Allocate a block of RLP memory instead?
-    %mload_global_metadata(@GLOBAL_METADATA_TRIE_ENCODED_CHILD_SIZE)
-    DUP1 %add_const(16)
-    %mstore_global_metadata(@GLOBAL_METADATA_TRIE_ENCODED_CHILD_SIZE)
-    // stack: base_offset, node_payload_ptr, encode_value, retdest
+    //Allocate a block of RLP memory
```
Nashtare, line 128: nit
```suggestion
    // Allocate a block of RLP memory
```

evm/src/cpu/kernel/asm/mpt/util.asm
```diff
@@ -10,6 +10,17 @@
...
 %endmacro
 
+%macro initialize_rlp_segment
+    PUSH 0x80
+    // We need to alloc a two rlp blocks because first positions will be used for storing the txn rlp.
+    // The encoding of the empty node will be stored at ENCODED_EMPTY_NODE_POS = 0x10009
+    %alloc_rlp_block
+    POP
+    %alloc_rlp_block
+global debug_position:
```
Nashtare, line 20: this is fine because it is called only once, but you shouldn't ever have `global` labels within macros, so I'd suggest removing it.

evm/src/cpu/kernel/asm/mpt/util.asm
```diff
@@ -10,6 +10,17 @@
     // stack: (empty)
 %endmacro
 
+%macro initialize_rlp_segment
+    PUSH 0x80
+    // We need to alloc a two rlp blocks because first positions will be used for storing the txn rlp.
```
Nashtare, line 15:
```suggestion
    // We need to alloc two rlp blocks as the first will be used for storing the txn rlp.
```

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -86,12 +86,17 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
...
 }
 
-const MISC_CONSTANTS: [(&str, [u8; 32]); 1] = [
+const MISC_CONSTANTS: [(&str, [u8; 32]); 2] = [
     // Base for limbs used in bignum arithmetic.
     (
         "BIGNUM_LIMB_BASE",
         hex!("0000000000000000000000000000000100000000000000000000000000000000"),
     ),
+    // Position in SEGMENT_RLP_RAW where the empty node encoding is stored
```
Nashtare, line 95:
```suggestion
    // Position in SEGMENT_RLP_RAW where the empty node encoding is stored.
```

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -86,12 +86,17 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
...
+    (
+        "ENCODED_EMPTY_NODE_POS",
+        hex!("0000000000000000000000000000000000000000000000000000000000010009"),
```
Nashtare, line 98: Maybe a note on where the `0x10009` comes from?

evm/src/cpu/kernel/constants/global_metadata.rs
```diff
@@ -30,68 +30,64 @@ pub(crate) enum GlobalMetadata {
...
-    BlockTimestamp = 15,
-    BlockNumber = 16,
-    BlockDifficulty = 17,
-    BlockRandom = 18,
-    BlockGasLimit = 19,
-    BlockChainId = 20,
-    BlockBaseFee = 21,
-    BlockGasUsed = 22,
+    BlockBeneficiary = 13,
+    BlockTimestamp = 14,
```
dvdplm, line 35: If you want to avoid the hassle of renaming, Rust wil auto increment the discriminant for you. See https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants

evm/src/cpu/kernel/constants/global_metadata.rs
```diff
@@ -30,68 +30,64 @@ pub(crate) enum GlobalMetadata {
...
 
-    KernelHash = 46,
-    KernelLen = 47,
+    KernelHash = 45,
+    KernelLen = 46,
 }
 
 impl GlobalMetadata {
-    pub(crate) const COUNT: usize = 48;
+    pub(crate) const COUNT: usize = 47;
```
dvdplm, line 90: There is a nightly-only `#![feature(variant_count)]` that enables `std::mem::variant_count` such that you could have `const COUNT: usize = variant_count::<Self>()` here instead. But not sure how we look at such unstable features.

Nashtare: The plan was to remove as many unstable features as possible, so I'd recommend against this suggestion for the sake of consistency.

4l0n50:
Sorry, I had to force push when I rebased to main

evm/src/generation/mod.rs
```diff
@@ -35,7 +35,7 @@ pub mod mpt;
 pub(crate) mod prover_input;
 pub(crate) mod rlp;
 pub(crate) mod state;
-mod trie_extractor;
+pub(crate) mod trie_extractor;
```
LindaGuiga, line 38: Do we need this to be `pub(crate)`?

LindaGuiga (approved):

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -86,12 +87,23 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
...
 
-const MISC_CONSTANTS: [(&str, [u8; 32]); 1] = [
+const MISC_CONSTANTS: [(&str, [u8; 32]); 3] = [
     // Base for limbs used in bignum arithmetic.
     (
         "BIGNUM_LIMB_BASE",
         hex!("0000000000000000000000000000000100000000000000000000000000000000"),
     ),
+    // Position in SEGMENT_RLP_RAW where the empty node encoding is stored. It is
+    // equal to usize::MAX so that all rlp pointers all much smalled than that
```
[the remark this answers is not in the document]
Nashtare: yeah we should write `u64::MAX` here, although we do have the assumption that `usize` is represented with 8 bytes across the lib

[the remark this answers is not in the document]
Nashtare: additionally: 
```suggestion
    // equal to usize::MAX so that all rlp pointers all much smaller than that.
```

evm/src/cpu/kernel/asm/main.asm
```diff
@@ -13,6 +13,11 @@ global main:
 
     // Initialise the shift table
     %shift_table_init
+    // Encode constant nodes
+    %initialize_rlp_segment
+
+    // Encode constant nodes
+    %initialize_rlp_segment
```
Nashtare, line 20: Is it a duplicate? If not we may want extra documentation, this is a bit confusing.

evm/src/cpu/kernel/constants/global_metadata.rs
Nashtare: Implemented in #1426 FYI.

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -86,12 +87,23 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
...
 
-const MISC_CONSTANTS: [(&str, [u8; 32]); 1] = [
+const MISC_CONSTANTS: [(&str, [u8; 32]); 3] = [
     // Base for limbs used in bignum arithmetic.
     (
         "BIGNUM_LIMB_BASE",
         hex!("0000000000000000000000000000000100000000000000000000000000000000"),
     ),
+    // Position in SEGMENT_RLP_RAW where the empty node encoding is stored. It is
+    // equal to usize::MAX so that all rlp pointers all much smaller than that.
```
Nashtare, line 97:
```suggestion
    // equal to usize::MAX so that all rlp pointers are much smaller than that.
```

Nashtare: it also still mentions `usize` while `u64` may be better (not platform specific) even if you're using `usize` when comparing with `0xFFFFFFFF` below? non-blocking though

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -201,21 +201,36 @@ fn test_extcodecopy() -> Result<()> {
...
-        [Segment::ContextMetadata as usize]
-        .set(GasLimit as usize, U256::from(1000000000000u64));
+    interpreter.generation_state.memory.set(
+        MemoryAddress {
+            context,
+            segment: Segment::ContextMetadata as usize,
+            virt: GasLimit as usize,
+        },
+        U256::from(1000000000000u64),
+    );
```
Nashtare, line 211: non-blocking, but not sure I understand why this was needed?

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -201,21 +201,36 @@ fn test_extcodecopy() -> Result<()> {
...
+            U256::from(rng.gen::<u8>()),
+        );
+        interpreter.generation_state.memory.set(
+            MemoryAddress {
+                context,
+                segment: Segment::KernelAccountCode as usize,
+                virt: i,
+            },
+            U256::from(rng.gen::<u8>()),
+        );
```
Nashtare, line 233: same comment as above

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -250,9 +265,11 @@ fn test_extcodecopy() -> Result<()> {
...
     // Check that the code was correctly copied to memory.
     for i in 0..size {
-        let memory = interpreter.generation_state.memory.contexts[context].segments
-            [Segment::MainMemory as usize]
-            .get(dest_offset + i);
+        let memory = interpreter.generation_state.memory.get(MemoryAddress {
+            context,
+            segment: Segment::MainMemory as usize,
+            virt: dest_offset + i,
+        });
```
Nashtare, line 272: same here, not sure it changes anything?

evm/src/cpu/kernel/tests/account_code.rs
```diff
@@ -277,13 +294,22 @@ fn prepare_interpreter_all_accounts(
...
+        U256::from_big_endian(&addr),
+    );
+    interpreter.generation_state.memory.set(
+        MemoryAddress {
+            context: 1,
+            segment: Segment::ContextMetadata as usize,
+            virt: ContextMetadata::GasLimit as usize,
+        },
+        100_000.into(),
+    );
```
Nashtare, line 312: same here

evm/src/cpu/kernel/tests/add11.rs
```diff
@@ -199,8 +200,14 @@ fn test_add11_yml() {
...
-    interpreter.generation_state.memory.contexts[0].segments[Segment::ContextMetadata as usize]
-        .set(ContextMetadata::GasLimit as usize, 1_000_000.into());
+    interpreter.generation_state.memory.set(
+        MemoryAddress {
+            context: 0,
+            segment: Segment::ContextMetadata as usize,
+            virt: ContextMetadata::GasLimit as usize,
+        },
+        1_000_000.into(),
+    );
```
Nashtare, line 210: same here

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -86,12 +87,23 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
...
 
-const MISC_CONSTANTS: [(&str, [u8; 32]); 1] = [
+const MISC_CONSTANTS: [(&str, [u8; 32]); 3] = [
     // Base for limbs used in bignum arithmetic.
     (
         "BIGNUM_LIMB_BASE",
         hex!("0000000000000000000000000000000100000000000000000000000000000000"),
     ),
+    // Position in SEGMENT_RLP_RAW where the empty node encoding is stored. It is
+    // equal to u32::MAX so that all rlp pointers are much smaller than that
```
Nashtare, line 97:
```suggestion
    // equal to u32::MAX so that all rlp pointers are much smaller than that.
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1171,6 +1173,18 @@ impl<'a> Interpreter<'a> {
         }
         self.generation_state.registers.context = context;
     }
+
+    /// Writes the encoding of 0 to position @ENCODED_EMPTY_NODE_POS
```
Nashtare, line 1177:
```suggestion
    /// Writes the encoding of 0 to position @ENCODED_EMPTY_NODE_POS.
```

Nashtare (approved):
Two minor comments for my OCD 😃

PR #1400, opened 2023-12-01
Fix kernel codehash discrepancy

Nashtare:
Attempt at fixing the Kernel hash discrepancy across runs.

### How to reproduce it

Print the `KERNEL` code hash in say, `prove_with_outputs()`, and then run:

```bash
for i in {0..10}; do cargo test --release --test simple_transfer; done
```

### Error

Upon inspection of the different code segments per parsed file, it appears that the inconsistency was coming from `hex_prefix.asm`, more particularly on the `%stack` macro which for some reasons seems to not be deterministic. I've replaced them by hardcoding what the optimizer had deduced.

On this branch, I've ran the above command 40 times and got the same hash every time.

### Other changes

I also:
- changed a variable name, that was identical to a `label` and could have messed things up
- removed the duplicate copy of the kernel code in `initialize_kernel_code()` as this is already being done when initializing `GenerationState`

### DIFF
--- a/evm/src/cpu/kernel/asm/mpt/hex_prefix.asm
+++ b/evm/src/cpu/kernel/asm/mpt/hex_prefix.asm
@@ -56,7 +56,7 @@ first_byte:
     SWAP1
     JUMP
     
-remaining_bytes:
+global remaining_bytes:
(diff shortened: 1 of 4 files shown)


LindaGuiga (approved):
LGTM, thanks!

PR #1401, opened 2023-12-01
Fix set_context constraints

hratoanina:
Some constraints for SET_CONTEXT were incorrect. Instead of checking that the new stack top was what was read in memory channel 3, we were only checking the first limb and forcing the other 7 to be 0.

Additionally, the zero-checking of the high limbs of value for one-limb quantities (like the stack length, or the context) wasn't done.

### DIFF
--- a/evm/src/cpu/contextops.rs
+++ b/evm/src/cpu/contextops.rs
@@ -163,14 +163,14 @@ fn eval_packed_set<P: PackedField>(
 
     // The next row's context is read from stack_top.
     yield_constr.constraint(filter * (stack_top[0] - nv.context));
-    for limb in &stack_top[1..] {
-        yield_constr.constraint(filter * *limb);
+    for &limb in &stack_top[1..] {
+        yield_constr.constraint(filter * limb);
     }
 
     // The old SP is decremented (since the new context was popped) and written to memory.
     yield_constr.constraint(filter * (write_old_sp_channel.value[0] - local_sp_dec));
-    for limb in &write_old_sp_channel.value[1..] {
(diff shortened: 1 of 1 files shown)


evm/src/cpu/contextops.rs
```diff
@@ -163,6 +163,9 @@ fn eval_packed_set<P: PackedField>(
 
     // The next row's context is read from stack_top.
     yield_constr.constraint(filter * (stack_top[0] - nv.context));
+    for limb in &stack_top[1..] {
+        yield_constr.constraint(filter * *limb);
+    }
```
Nashtare, line 168: nit: a tad clearer this way
```suggestion
    for &limb in &stack_top[1..] {
        yield_constr.constraint(filter * limb);
    }
```

evm/src/cpu/contextops.rs
```diff
@@ -177,6 +180,9 @@ fn eval_packed_set<P: PackedField>(
 
     // The new SP is loaded from memory.
     yield_constr.constraint(filter * (read_new_sp_channel.value[0] - nv.stack_len));
+    for limb in &read_new_sp_channel.value[1..] {
+        yield_constr.constraint(filter * *limb);
+    }
```
Nashtare, line 185: same as above:

```suggestion
    for &limb in &read_new_sp_channel.value[1..] {
        yield_constr.constraint(filter * limb);
    }
```

Nashtare: actually valid for all the changes below too

Nashtare (approved):
Thanks!

PR #1403, opened 2023-12-04
Pacify clippy

Nashtare:
Popping up from latest nightly.


dvdplm (approved):
This isn't really a new lint though right?

PR #1404, opened 2023-12-04
Implement degree 2 filters

hratoanina:
Support for arbitrary degree 2 filters for CTLs. Makes it possible to remove several columns.


evm/src/cross_table_lookup.rs
```diff
@@ -315,27 +315,111 @@ impl<F: Field> Column<F> {
...
+            + self
+                .constants
+                .iter()
+                .map(|col| col.eval_table(table, row))
+                .sum()
+    }
+}
+
 /// A `Table` with a linear combination of columns and a filter.
 /// `filter_column` is used to determine the rows to select in `Table`.
```
LindaGuiga, line 407:
```suggestion
/// `filter` is used to determine the rows to select in `Table`.
```

LindaGuiga (approved):
LGTM, thanks! I just think it would be nice to add a line in the specs to say that the filters are limited to degree 2.

PR #1405, opened 2023-12-04
Implement `PublicValues` retrieval from public inputs

Nashtare:
This allows to retrieve the `PublicValues` stored in the public inputs of a proof, regardless of its aggregation level.
I haven't touched the existing API for the prover, as we may still prefer to just keep the returned `PublicValues` but this at least allows a verifier to recover them if needed (as they are kind of "decoupled" from their proofs for now).

### DIFF
--- a/evm/src/proof.rs
+++ b/evm/src/proof.rs
@@ -63,7 +63,7 @@ pub struct PublicValues {
 }
 
 impl PublicValues {
-    /// Extracts public values from the given public inputs of a proofs.
+    /// Extracts public values from the given public inputs of a proof.
     /// Public values are always the first public inputs added to the circuit,
     /// so we can start extracting at index 0.
     pub fn from_public_inputs<F: RichField>(pis: &[F]) -> Self {
--- a/evm/src/proof.rs
+++ b/evm/src/proof.rs
@@ -66,7 +66,7 @@ impl PublicValues {
     /// Extracts public values from the given public inputs of a proofs.
     /// Public values are always the first public inputs added to the circuit,
     /// so we can start extracting at index 0.
-    pub fn from_public_inputs<F: RichField + Extendable<D>, const D: usize>(pis: &[F]) -> Self {
+    pub fn from_public_inputs<F: RichField>(pis: &[F]) -> Self {
         assert!(
             pis.len()
                 > TrieRootsTarget::SIZE * 2
@@ -116,7 +116,7 @@ pub struct TrieRoots {
 }
 
 impl TrieRoots {
-    pub fn from_public_inputs<F: RichField + Extendable<D>, const D: usize>(pis: &[F]) -> Self {
+    pub fn from_public_inputs<F: RichField>(pis: &[F]) -> Self {
         assert!(pis.len() == TrieRootsTarget::SIZE);
 
         let state_root = get_h256(&pis[0..8]);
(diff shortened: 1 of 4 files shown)


evm/src/proof.rs
```diff
@@ -61,8 +62,50 @@ pub struct PublicValues {
     pub extra_block_data: ExtraBlockData,
 }
 
+impl PublicValues {
+    /// Extracts public values from the given public inputs of a proofs.
```
dvdplm, line 66:
```suggestion
    /// Extracts public values from the given public inputs of a proof.
```

evm/src/proof.rs
```diff
@@ -123,9 +196,40 @@ pub struct BlockMetadata {
...
+        let block_timestamp = pis[5].to_canonical_u64().into();
+        let block_number = pis[6].to_canonical_u64().into();
+        let block_difficulty = pis[7].to_canonical_u64().into();
+        let block_random = get_h256(&pis[8..16]);
+        let block_gaslimit = pis[16].to_canonical_u64().into();
+        let block_chain_id = pis[17].to_canonical_u64().into();
+        let block_base_fee =
+            (pis[18].to_canonical_u64() + (pis[19].to_canonical_u64() << 32)).into();
+        let block_gas_used = pis[20].to_canonical_u64().into();
+        let block_bloom = core::array::from_fn(|i| h2u(get_h256(&pis[21 + 8 * i..29 + 8 * i])));
```
dvdplm, line 213: These ranges are bit awkward and it looks a bit brittle. If one of us make a change here, would we ever notice? Can we at least move the extraction logic to `PublicValues`? Have `pis.block_beneficiary()` etc and hope the compiler optimizes it?

evm/src/util.rs
```diff
@@ -210,3 +210,25 @@ pub(crate) fn biguint_to_mem_vec(x: BigUint) -> Vec<U256> {
...
+        &slice
+            .iter()
+            .rev()
+            .map(|x| x.to_canonical_u64() as u32)
+            .flat_map(|limb| limb.to_be_bytes())
+            .collect_vec(),
+    )
+}
+
+pub(crate) fn get_h256<F: RichField>(slice: &[F]) -> H256 {
```
dvdplm, line 225: nit: I'd prefer `extract` over `get` but perhaps that's just me. 

A macro here would be nice!

dvdplm (approved):
lgtm

I would like to see a better solution to the brittleness of accessing ranges in slices but it's a bit out of scope for this PR as it's pretty pervasive to the whole code base.

evm/src/proof.rs
Nashtare: Yeah I agree the range accesses are quite error-prone (we use the same in the target versions to construct the proofs, so this should normally be easily detected if there was a breaking change).
Ideally I'd like to review the way we construct those objects, maybe something similar to the different views (like `CpuColumnsView`?) to remove the need for numeric accesses.

evm/src/util.rs
Nashtare: > I'd prefer extract over get

Hmm, it seems `get` is a bit more accurate no? extracting makes me think we take a portion of the provided slice, while here we expect the whole slice to be used to recover the hash. But maybe that's my English being plain wrong :D 

> A macro here would be nice!

For inlining? I like the idea, though I think this may be out of scope of this PR, as several other methods in the `util` module could be turned to macros. WDYT?

muursh: >Hmm, it seems get is a bit more accurate no?

`get` makes far more sense

PR #1406, opened 2023-12-04
Implement MPT preinitialization

LindaGuiga:
This PR preinitializes the `TrieData` segment with the initial tries.
The tries are still checked by asserting the equality of hashes, but we also need to constrain the length of the segment.
This is done by computing the expected length in `mpt_hash` and asserting its equality with the `GlobalMetadata` value.

Please let me know what you think! @wborgeaud


evm/src/generation/mod.rs
```diff
@@ -191,6 +207,21 @@ fn apply_metadata_and_tries_memops<F: RichField + Extendable<D>, const D: usize>
...
+    state: &mut GenerationState<F>,
+    trie_inputs: &TrieInputs,
+) -> (TrieRootPtrs, usize) {
+    let (trie_roots_ptrs, trie_data) =
+        load_all_mpts(trie_inputs).expect("Invalid MPT data for preinitialization");
+
+    for (i, data) in trie_data.iter().enumerate() {
+        let trie_addr = MemoryAddress::new(0, Segment::TrieData, i);
+        state.memory.set(trie_addr, data.into());
+    }
```
Nashtare, line 220: nit: similarly to how the Kernel code is initialized, it may be a tad cleaner to do

```suggestion
        let trie_data_u256s = trie_data.iter().map(|&x| x.into()).collect();
        state.memory.contexts[0].segments[Segment::TrieData as usize].content = trie_data_u256s;
```

evm/src/cpu/kernel/asm/main.asm
```diff
@@ -14,14 +14,18 @@ global main:
...
 
-    // Second, load all MPT data from the prover.
-    PUSH hash_initial_tries
-    %jump(load_all_mpts)
-
 global hash_initial_tries:
-    %mpt_hash_state_trie   %mload_global_metadata(@GLOBAL_METADATA_STATE_TRIE_DIGEST_BEFORE)    %assert_eq
+    // We compute the length of the trie data segment in `mpt_hash` so that we
+    // can check the value provided by the prover.
+    PUSH 1 // Initial length
```
Nashtare, line 20: nit: maybe emphasize why the segment is initially of length 1 (i.e. containing 0) for clarity?

evm/src/cpu/kernel/asm/mpt/hash/hash_trie_specific.asm
```diff
@@ -1,272 +1,312 @@
...
     SWAP1 POP
-    // stack: rlp_pos, retdest
-    SWAP1
+    // stack: rlp_pos, cur_len'', retdest
+    %stack(rlp_pos, new_len, retdest) -> (retdest, rlp_pos, new_len)
     JUMP
 
 encode_nonzero_receipt_type:
-    // stack: txn_type, rlp_pos, value_ptr, retdest
+    // We have a nonlegacy receipt, so the type is also store in the trie data segment.
```
Nashtare, line 307: nit:
```suggestion
    // We have a nonlegacy receipt, so the type is also stored in the trie data segment.
```

evm/src/cpu/kernel/asm/mpt/hash/hash_trie_specific.asm
```diff
@@ -1,272 +1,312 @@
...
+    // stack: rlp_pos, payload_len_ptr, cur_len'', retdest
     SWAP1 POP
-    // stack: rlp_pos, retdest
-    SWAP1
+    // stack: rlp_pos, cur_len'', retdest
+    %stack(rlp_pos, new_len, retdest) -> (retdest, rlp_pos, new_len)
     JUMP
 
 encode_nonzero_receipt_type:
-    // stack: txn_type, rlp_pos, value_ptr, retdest
```
Nashtare, line 269: minor, but I'd keep a stack description when reaching this label for extra clarity (even though the next sequence of operations is fairly simple to follow)

evm/src/cpu/kernel/asm/mpt/hash/hash_trie_specific.asm
```diff
@@ -285,15 +325,19 @@ encode_nonzero_receipt_type:
     %jump(encode_receipt_after_type)
 
 global encode_storage_value:
-    // stack: rlp_pos, value_ptr, retdest
+    // stack: rlp_pos, value_ptr, cur_len, retdest
     SWAP1 %mload_trie_data SWAP1
-    // stack: rlp_pos, value, retdest
+
+    // A storage value is a scalar, so we only need to add 1 to the trie data length
```
Nashtare, line 331: nit: 
```suggestion
    // A storage value is a scalar, so we only need to add 1 to the trie data length.
```

evm/src/cpu/kernel/asm/mpt/hash/hash_trie_specific.asm
```diff
@@ -1,272 +1,312 @@
...
+    %stack(new_rlp_pos, new_len, retdest) -> (retdest, new_rlp_pos, new_len)
     JUMP
 
 // We assume a receipt in memory is stored as:
 // [payload_len, status, cum_gas_used, bloom, logs_payload_len, num_logs, [logs]].
 // A log is [payload_len, address, num_topics, [topics], data_len, [data]].
 global encode_receipt:
-    // stack: rlp_pos, value_ptr, retdest
+    // stack: rlp_pos, value_ptr, cur_len, retdest
+    // First, we add 261 to the trie data length for all values before the logs besides the type.
```
Nashtare, line 152: minor, but you may want to break down the 261 here for clarity

Nashtare (approved):
Thanks, it looks good!
Just make sure the specs are up to date if there are mentions on how trie data are handled.

FYI, the test-runner got ~1200 tests passing fine.

wborgeaud (approved):
Looks good to me!
Can you just explain why we need to constrain the trie data length and root pointers?

Nashtare:
I noticed these tests that failed, but it seems they are also broken on past main (rev: `5572da30d7ab818594cf8659839fa832dfcf1d3d`).

```call_outsize_then_create_successful_then_returndatasize_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:41.292820022Z
call_then_create_successful_then_returndatasize_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:41.953556675Z
create_callprecompile_returndatasize_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:41.733849328Z
returndatacopy_0_0_following_successful_create_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:41.622385899Z
returndatacopy_afterFailing_create_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:53.370883213Z
returndatacopy_following_revert_in_create_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:41.512688150Z
returndatasize_following_successful_create_d0g0v0_Shanghai,Ignored,2023-12-06T10:35:55.859143861Z```

**Comment by LindaGuiga** [comment]
> Looks good to me! Can you just explain why we need to constrain the trie data length and root pointers?

I can remove the length and pointers for the `extra_block_data` and replace that with a memory store in the Kernel, if that' s what you mean? (I still need to have a memory read since `GlobalMetadata` cannot yet be preinitialized given the constraints).

File: evm/src/cpu/kernel/asm/main.asm [hunk 7]
```diff
@@ -14,14 +14,20 @@ global main:
...
+    // the null pointer `0` when the tries are empty.
+    PUSH 1
+    %mpt_hash_state_trie  %mload_global_metadata(@GLOBAL_METADATA_STATE_TRIE_DIGEST_BEFORE)    %assert_eq
+    // stack: trie_data_len
     %mpt_hash_txn_trie     %mload_global_metadata(@GLOBAL_METADATA_TXN_TRIE_DIGEST_BEFORE)      %assert_eq
+    // stack: trie_data_len
     %mpt_hash_receipt_trie %mload_global_metadata(@GLOBAL_METADATA_RECEIPT_TRIE_DIGEST_BEFORE)  %assert_eq
+    // stack: trie_data_full_len
+    %mload_global_metadata(@GLOBAL_METADATA_TRIE_DATA_SIZE)
+    %assert_eq
```
wborgeaud, line 30: > I can remove the length and pointers for the `extra_block_data` and replace that with a memory store in the Kernel, if that' s what you mean? (I still need to have a memory read since `GlobalMetadata` cannot yet be preinitialized given the constraints).

Yeah removing them from `extra_block_data` would remove complexity. But also, I don't really understand why we need to check the trie data length here.

LindaGuiga: Can't there be an issue if, say, we set nondeterministically the length to 0 (but still have a nonempty trie)? `mpt_hash` doesn't use the trie data size, so it could lead to a nonempty hash, with the rest of the code assuming an empty trie data segment, right?

wborgeaud: Ah right, makes sense thanks!

PR #1409, opened 2023-12-06
Preinitialize all code segments

hratoanina:
#1338 

Exclude the code segment for all contexts from zero-initializing. Code is non-deterministically loaded into the segment, then hashed for verification. Applies to initial txn code loading, contract code loading after a call, EXTCODESIZE and EXTCODECOPY.

For the first two cases, we write an extra 33 zeros after code_size for soundness reasons: if user-executed code is malformed and ends with a PUSH, we must make sure that the extra bytes read (and the next location of PC) is a 0, accordingly to the specs.


hratoanina:
For the ERC20 test, against current main:
CPU rows: 136k -> 105k
Memory rows: 577k -> 479k

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -76,53 +76,62 @@ global sys_extcodesize:
...
+        // stack: 0, ctx, segment, i, ctx, segment, i, retdest, code_size
+        MSTORE_GENERAL
+        // stack: ctx, segment, i, retdest, code_size
+        DUP3 %increment
+        // stack: i+1, ctx, segment, i, retdest, code_size
+        SWAP3 POP
+        // stack: ctx, segment, i+1, retdest, code_size
+    %endrep
+    // stack: ctx, segment, code_size+32, retdest, code_size
+    %pop3
```
Nashtare, line 136: This saves around 700 CPU cycles.
```suggestion
    // stack: ctx, segment, i, retdest, code_size
    %stack (ctx, segment, i, retdest) -> (ctx, segment, i, 0, segment, ctx, retdest)
    MSTORE_32BYTES_32
    // stack: i', segment, ctx, retdest, code_size
    SWAP2
    PUSH 0
    // stack: 0, ctx, segment, i', retdest, code_size
    MSTORE_GENERAL
    // stack: retdest, code_size
```

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -76,53 +76,62 @@ global sys_extcodesize:
...
 load_code_non_existent_account:
-    %stack (codehash, ctx, segment, retdest) -> (retdest, 0)
+    // Write 0 at address 0 for soundness.
+    // stack: codehash, ctx, retdest
+    %stack (codehash, ctx, retdest) -> (0, ctx, @SEGMENT_CODE, 0, retdest, 0)
+    MSTORE_GENERAL
+    // stack: retdest, 0
+    JUMP
+
+// Identical to load_code, but adds 33 zeros after code_size for soundness reasons.
```
Nashtare, line 113: Maybe we want to quickly describe the soundness issue as you did in the PR description? Or you would refer to it in the specs?

Nashtare (approved):
LGTM, thank you! Just a tiny optimization for the code padding, and the `memory initialization` section of the specs would need an update :)

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -76,53 +76,54 @@ global sys_extcodesize:
...
+// Post stack: code_size
+global load_code_padded:
+    %stack (address, ctx, retdest) -> (address, ctx, load_code_padded_ctd, ctx, retdest)
+    %jump(load_code)
+
+load_code_padded_ctd:
+    %stack (code_size, ctx, retdest) -> (ctx, @SEGMENT_CODE, code_size, 0, ctx, retdest, code_size)
+    MSTORE_32BYTES_32
+    // stack: last_offset, ctx, retdest, code_size
+    %stack (last_offset, ctx) -> (0, ctx, @SEGMENT_CODE, last_offset)
```
Nashtare, line 126: in general we prefer not to have several calls to %stack, as grouping them can save some operations. I haven't compared with my suggestion, but this *may* be costing more CPU cycles? But that's mostly nitpicking as a general guideline, as the diff would be just a few cycles at most.

hratoanina: I think there was a 2-cycle difference when I tried your method.

PR #1410, opened 2023-12-06
Remove is_keccak_sponge

hratoanina:


evm/src/cpu/cpu_stark.rs
```diff
@@ -49,7 +49,13 @@ pub(crate) fn ctl_data_keccak_sponge<F: Field>() -> Vec<Column<F>> {
 
 /// CTL filter for a call to the Keccak sponge.
 pub(crate) fn ctl_filter_keccak_sponge<F: Field>() -> Filter<F> {
-    Filter::new_simple(Column::single(COL_MAP.is_keccak_sponge))
+    Filter::new(
+        vec![(
+            Column::single(COL_MAP.op.jumpdest_keccak_general),
+            Column::linear_combination_with_constant([(COL_MAP.opcode_bits[1], -F::ONE)], F::ONE),
```
Nashtare, line 55: Nit: the approach is used several times in constraints already but I think it wouldn't hurt just emphasizing here again the opcode bit is used to get rid of jumpdest?

Nashtare (approved):

PR #1412, opened 2023-12-07
Replace increment macro with an opcode.

LindaGuiga:
This PR replaces all `%increment` (and `%add_const(1)`) calls by an INCREMENT custom opcode. This saves 1435 cycles on the CPU side (and 7356 cycles on memory side) for erc20, but this comes at the expense of one additional `ArithmeticStark` column.


Nashtare:
I hadn't realized the gains would be that light (makes sense, it's halving the overhead of all increments), hence we may want to consider all possibilities around custom instructions before inserting new ones, especially if it incurs additional columns elsewhere?

LindaGuiga:
> I hadn't realized the gains would be that light (makes sense, it's halving the overhead of all increments), hence we may want to consider all possibilities around custom instructions before inserting new ones, especially if it incurs additional columns elsewhere?

It was made even lighter than expected because of the new pre-loading optimizations I think (originally, the gains were supposed to be higher than that). I agree it now makes sense to look into all possible instructions.

Nashtare:
@LindaGuiga Should we close this?

LindaGuiga:
I think it makes sense to close it yes. I don't think this is worth it as is.

PR #1413, opened 2023-12-07
Use mstore_32bytes to optimize decode_int_given_len

LindaGuiga:
This removes 1677 CPU cycles for erc20.


Nashtare (approved):
looks good! Thanks!

PR #1414, opened 2023-12-07
Merge MSTORE_32BYTES and MLOAD_32BYTES columns

hratoanina:


evm/src/cpu/byte_unpacking.rs
```diff
@@ -32,7 +34,8 @@ pub(crate) fn eval_ext_circuit<F: RichField + Extendable<D>, const D: usize>(
     nv: &CpuColumnsView<ExtensionTarget<D>>,
     yield_constr: &mut RecursiveConstraintConsumer<F, D>,
 ) {
-    let filter = lv.op.mstore_32bytes;
+    let filter =
```
Nashtare, line 37: nit: we tend to duplicate comments across native and circuit version for ease of reading
```suggestion
    // The MSTORE_32BYTES opcodes are differentiated from MLOAD_32BYTES
    // by the 5th bit set to 0.
    let filter =
```

evm/src/cpu/decode.rs
```diff
@@ -133,17 +132,22 @@ pub(crate) fn eval_packed_generic<P: PackedField>(
...
+    };
+
     // Manually check lv.op.m_op_constr
-    let opcode: P = lv
-        .opcode_bits
-        .into_iter()
-        .enumerate()
-        .map(|(i, bit)| bit * P::Scalar::from_canonical_u64(1 << i))
-        .sum();
+    let opcode_full = opcode_high_bits(8);
```
Nashtare, line 146: `opcode` is self-sufficient no? i.e. we do get the actual opcode here.

Nashtare (approved):
Thanks, LGTM!

PR #1417, opened 2023-12-08
Merge `push` and  `prover_input` flags

LindaGuiga:


evm/src/cpu/decode.rs
```diff
@@ -73,6 +73,23 @@ const fn bits_from_opcode(opcode: u8) -> [bool; 8] {
...
+    builder: &mut CircuitBuilder<F, D>,
+    lv: &CpuColumnsView<ExtensionTarget<D>>,
+    block_length: usize,
+) -> ExtensionTarget<D> {
+    let zero = builder.zero_extension();
+    lv.opcode_bits
+        .into_iter()
+        .enumerate()
+        .rev()
+        .take(block_length)
```
hratoanina, line 86: Nit: `block_length` is misleading. The argument is the number of bits you take, which is `8 - block_length`.

evm/src/cpu/decode.rs
```diff
@@ -133,17 +150,22 @@ pub(crate) fn eval_packed_generic<P: PackedField>(
...
+    };
+
     // Manually check lv.op.m_op_constr
-    let opcode: P = lv
-        .opcode_bits
-        .into_iter()
-        .enumerate()
-        .map(|(i, bit)| bit * P::Scalar::from_canonical_u64(1 << i))
-        .sum();
+    let opcode_full: P = opcode_high_bits(8);
```
hratoanina, line 164: Nit: Same comment than Robin: we can just call it `opcode` if it's full.

evm/src/cpu/gas.rs
```diff
@@ -27,10 +27,9 @@ const SIMPLE_OPCODES: OpsColumnsView<Option<u32>> = OpsColumnsView {
     not_pop: None, // This is handled manually below
     shift: G_VERYLOW,
     jumpdest_keccak_general: None, // This is handled manually below.
-    prover_input: KERNEL_ONLY_INSTR,
-    jumps: None, // Combined flag handled separately.
+    push_prover_input: None,       // This handled manually below.
```
hratoanina, line 30:
```suggestion
    push_prover_input: None,       // This is handled manually below.
```

evm/src/cpu/gas.rs
```diff
@@ -113,6 +112,14 @@ fn eval_packed_accumulate<P: PackedField>(
     yield_constr.constraint_transition(
         lv.op.jumpdest_keccak_general * (gas_diff - jumpdest_keccak_general_gas_cost),
     );
+
+    // For PROVER_INPUT and PUSH operations.
+    // PUSH operations are differentiated from PROVER_INPUT by their 6th bit set to 1.
```
hratoanina, line 117: Nit: I say 5th in that context. I don't know which convention has been used throughout the document, but we should harmonize.

hratoanina (approved):
Looks good!

evm/src/cpu/gas.rs
LindaGuiga: In the rest of the file, they use 6th in this context.

PR #1418, opened 2023-12-08
Add `Checkpoint` heights

Nashtare:
This changes the notion of `genesis` block for a `checkpoint` block, allowing to pass a `None` argument not only for the first block post-genesis but at any chosen checkpoint height.

- removed constraint on the `checkpoint` block number to be 0. It is still checked to be 1 less than the current block proof.
- binds `block_proof` output `PublicValues` to what's encoded, similarly to how it's done at the aggregation level, instead of letting them be arbitrary.
- added an example of changing the checkpoint for a new block in `log_opcode`

@LindaGuiga Would you mind reviewing?

cc @BGluth, this would also allow us to not have to restart from block 1 whenever we want to test block N :)

### DIFF
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -37,7 +37,7 @@ use crate::generation::GenerationInputs;
 use crate::get_challenges::observe_public_values_target;
 use crate::proof::{
     BlockHashesTarget, BlockMetadataTarget, ExtraBlockData, ExtraBlockDataTarget, PublicValues,
-    PublicValuesTarget, StarkProofWithMetadata, TrieRootsTarget,
+    PublicValuesTarget, StarkProofWithMetadata, TrieRoots, TrieRootsTarget,
 };
 use crate::prover::prove;
 use crate::recursive_verifier::{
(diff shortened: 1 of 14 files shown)


evm/src/fixed_recursive_verifier.rs
```diff
@@ -1130,6 +1180,14 @@ where
                 nonzero_pis.insert(block_hashes_current_start + i, cur_targets[i]);
             }
 
+            // Initialize the checkpoint block number.
+            // Subtraction would result in an invalid proof for genesis, but we shouldn't try proving this block anyway.
```
LindaGuiga, line 1184: I'm not sure I understand this comment: if the genesis block is on the rhs, then the parent block has block number -1, but that also means that the lhs is a dummy block proof, right?

evm/src/fixed_recursive_verifier.rs
```diff
@@ -1145,13 +1203,26 @@ where
         block_inputs
             .set_verifier_data_target(&self.block.cyclic_vk, &self.block.circuit.verifier_only);
 
-        set_public_value_targets(&mut block_inputs, &self.block.public_values, &public_values)
-            .map_err(|_| {
-                anyhow::Error::msg("Invalid conversion when setting public values targets.")
-            })?;
+        // This is basically identical to this block public values, apart from the `trie_roots_before`
+        // that may come from the previous proof, if any.
```
LindaGuiga, line 1207: Shouldn't only the state trie root come from the previous proof? Aren't the other tries initialized at the beginning of each block?

LindaGuiga (approved):
LGTM thanks! I just had a couple questions!

evm/src/fixed_recursive_verifier.rs
Nashtare: Genesis being on the rhs means we're trying to prove it, but we can't.

PR #1424, opened 2023-12-13
Regenerate tries upon Kernel failure during `hash_final_tries`

Nashtare:
Adds some debugging output to proof logs in `debug` mode in case of `KernelPanic` error during the final phase, i.e. when computing the final trie roots and checking them against the provided ones.

Code mostly stolen from @4l0n50's previous work and refactored a little.

cc @BGluth


BGluth:
Having the entire trie instead of just the root will be pretty helpful for debugging root mismatches. Glad you added this!

LindaGuiga (approved):
LGTM, thanks!

PR #1426, opened 2023-12-14
Address bundling

### LINKED ISSUE #1324: Bundle memory addresses in a single `U256`
Addresses in the kernel are currently represented as 3 elements: `context`, `segment`, and `virtual` (an offset).
These values are small (32 bits at most for virtual), few values for context and segment.

This incurs quite some overhead on both CPU and Memory when manipulating addresses. For instance, copying memory addresses requires 3 `DUP`, deleting one 3 `POP`, etc..

It would be beneficial to have instead memory addresses represented as $2^{64} \cdot ctx + 2^{32}\cdot seg + off$.
By scaling the hardcoded values in the `Segment` enum, and increasing the context by $2^{64}$ instead of 1 when entering a new one, this turns a memory creation from 3 `POP` to 3 `POP` + 2 `ADD`.
Given that **many** operations take place in the kernel, with $ctx == 0$, 1 `ADD` can be removed, same reasoning goes for $seg == CODE == 0$. Hence the overhead on address creation should be low enough for actual gains to occur, especially for memory heavy transitions involving `memcpy` / `memset` / `rlp_` / ...

Nashtare:
This PR refactors how addresses are internally handled by the Kernel.
Instead of dealing with three distinct words, each storing `context`, `segment`, and `virtual`, addresses are now represented using a single `U256` word as $virt + 2^{32} seg + 2^{64} ctx$.

In addition, some memory offsets like `GlobalMetadata`, `ContextMetadata`, `TxnFields`, which set size is known in advance, are already scaled by their respective segments, to allow for faster address construction (as we now `2 ADD` in addition of the existing 3 pushes when inserting an arbitrary address on the stack).

`context` and `segment` values are automatically scaled by their respective shift to reduce the overhead on the kernel side.

As a follow-up, this refactoring enables almost immediate drop of an entire memory channel, effectively removing 13 CPU columns.

For ERC20 against latest main, the trace lengths are:
`TraceCheckpoint { arithmetic_len: 18374, byte_packing_len: 26159, cpu_len: 102230, keccak_len: 12768, keccak_sponge_len: 532, logic_len: 3705, memory_len: 463214 }`
while on this branch, they are:
`TraceCheckpoint { arithmetic_len: 21766, byte_packing_len: 24570, cpu_len: 94271, keccak_len: 13920, keccak_sponge_len: 580, logic_len: 3945, memory_len: 447213 }`

Note:
- there are some inefficient operations in the kernel, following the major refactoring around stack operations, etc, which could be dealt with in the future, but opening the PR from now as it is quite massive already.
- I've run 8000 tests in `witness-only` mode with the test-runner, without any failure. Now running in `proving` mode just to make sure everything works as expected (got ~1.5k tests ok so far).
- I haven't updated the specs yet, will do after getting some feedback on the approach.

closes #1324


evm/tests/log_opcode.rs
```diff
@@ -444,7 +444,7 @@ fn test_log_with_aggreg() -> anyhow::Result<()> {
     // Preprocess all circuits.
     let all_circuits = AllRecursiveCircuits::<F, C, D>::new(
         &all_stark,
-        &[16..17, 13..16, 15..18, 14..15, 9..10, 12..13, 17..20],
+        &[16..17, 13..16, 15..18, 14..15, 10..11, 12..13, 17..20],
```
dvdplm, line 447: What was this change about?

Nashtare: KeccakSponge table uses 3 or 4 more rows here, but we were already on the edge of 512 (2^9), hence the bump

evm/src/cpu/kernel/constants/mod.rs
```diff
@@ -58,16 +58,19 @@ pub(crate) fn evm_constants() -> HashMap<String, U256> {
     c.insert(CALL_STACK_LIMIT.0.into(), U256::from(CALL_STACK_LIMIT.1));
 
     for segment in Segment::all() {
-        c.insert(segment.var_name().into(), (segment as u32).into());
+        c.insert(segment.var_name().into(), (segment as u64).into());
```
dvdplm, line 61: Why cast to `u64` here and `usize` below? I think the discriminant representation is `isize` (unless `#[repr(SOMETHING ELSE)]` is used) so perhaps use that? Not sure though!

evm/src/cpu/kernel/constants/txn_fields.rs
```diff
@@ -1,28 +1,36 @@
...
+    R = Segment::TxnFields as usize + 12,
+    S = Segment::TxnFields as usize + 13,
+    Origin = Segment::TxnFields as usize + 14,
 
     /// The actual computed gas price for this transaction in the block.
     /// This is not technically a transaction field, as it depends on the block's base fee.
-    ComputedFeePerGas = 15,
-    ComputedPriorityFeePerGas = 16,
+    ComputedFeePerGas = Segment::TxnFields as usize + 15,
+    ComputedPriorityFeePerGas = Segment::TxnFields as usize + 16,
```
dvdplm, line 33: This is fine, but if you want to you can also just set the first one and let the rest of the discriminant be "auto increment". See https://doc.rust-lang.org/reference/items/enumerations.html#implicit-discriminants

evm/src/cpu/kernel/constants/global_metadata.rs
```diff
@@ -1,93 +1,100 @@
...
+    LogsLen = Segment::GlobalMetadata as usize + 41,
+    LogsDataLen = Segment::GlobalMetadata as usize + 42,
+    LogsPayloadLen = Segment::GlobalMetadata as usize + 43,
+    TxnNumberBefore = Segment::GlobalMetadata as usize + 44,
+    TxnNumberAfter = Segment::GlobalMetadata as usize + 45,
 
-    KernelHash = 46,
-    KernelLen = 47,
+    KernelHash = Segment::GlobalMetadata as usize + 46,
+    KernelLen = Segment::GlobalMetadata as usize + 47,
```
dvdplm, line 97: Same comment as for `txn_fields`

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -177,52 +177,95 @@ impl<'a> Interpreter<'a> {
     }
 
     pub(crate) fn get_txn_field(&self, field: NormalizedTxnField) -> U256 {
-        self.generation_state.memory.contexts[0].segments[Segment::TxnFields as usize]
-            .get(field as usize)
+        // Those fields are already scaled by their respective segment.
```
dvdplm, line 180:
```suggestion
        // These fields are already scaled by their respective segment.
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -842,9 +895,10 @@ impl<'a> Interpreter<'a> {
     }
 
     fn run_mload_general(&mut self) {
-        let context = self.pop().as_usize();
-        let segment = Segment::all()[self.pop().as_usize()];
-        let offset = self.pop().as_usize();
+        let addr = self.pop();
+        let offset = addr.low_u32() as usize;
+        let segment = Segment::all()[(addr >> SEGMENT_SCALING_FACTOR).low_u32() as usize];
+        let context = (addr >> CONTEXT_SCALING_FACTOR).low_u32() as usize;
```
dvdplm, line 901: This seems sufficiently common that you could consider a helper or a macro. Maybe `let (addr, offset, segment, context) = unpackify!(self.pop())`?

evm/src/cpu/kernel/constants/mod.rs
Nashtare: Yeah right I could use `usize` for consistency with what's below. Note however that I've had weird behaviors with the discriminant representation, like the one for `Segment` is now `usize`, but I still need to cast the segment variants to `usize`, unless I am missing something with how discriminants work.

evm/src/cpu/contextops.rs
```diff
@@ -83,9 +84,13 @@ fn eval_packed_get<P: PackedField>(
     // If the opcode is GET_CONTEXT, then lv.opcode_bits[0] = 0.
     let filter = lv.op.context_op * (P::ONES - lv.opcode_bits[0]);
     let new_stack_top = nv.mem_channels[0].value;
-    yield_constr.constraint(filter * (new_stack_top[0] - lv.context));
-    for &limb in &new_stack_top[1..] {
-        yield_constr.constraint(filter * limb);
+    // Context is scaled by 2^64, hence stored in the 3rd limb.
+    yield_constr.constraint(filter * (new_stack_top[2] - lv.context));
+
+    for (i, &limb) in new_stack_top.iter().enumerate() {
```
dvdplm, line 90: There are always three limbs here yes? If so, why bother iterating? Just `yield_constr.constraint()` twice up front with explicit index `0` and `1`?

evm/src/cpu/kernel/constants/txn_fields.rs
Nashtare: Oh that'd be much nicer, let me give it a try!

evm/src/cpu/contextops.rs
Nashtare: There are 8 limbs per memory channel (i.e. holding a `U256` split into 8 `u32` limbs).

evm/src/cpu/cpu_stark.rs
```diff
@@ -145,27 +143,29 @@ pub(crate) fn ctl_filter_byte_packing<F: Field>() -> Filter<F> {
...
-    let segment = Column::single(COL_MAP.mem_channels[1].value[0]);
-    let virt = Column::single(COL_MAP.mem_channels[2].value[0]);
-    let val = Column::singles(COL_MAP.mem_channels[3].value);
+    let (context, segment, virt) = get_addr(&COL_MAP);
+    let mut res = vec![
+        Column::single(context),
+        Column::single(segment),
+        Column::single(virt),
+    ];
+    let val = Column::singles(COL_MAP.mem_channels[1].value);
```
dvdplm, line 155: Perhaps move this to line 168, closer to `res.extend()`?

evm/src/generation/mod.rs
```diff
@@ -151,7 +151,8 @@ fn apply_metadata_and_tries_memops<F: RichField + Extendable<D>, const D: usize>
         .map(|(field, val)| {
             mem_write_log(
                 channel,
-                MemoryAddress::new(0, Segment::GlobalMetadata, field as usize),
+                // Those fields are already scaled by their segment, and are in context 0 (kernel).
```
dvdplm, line 154:
```suggestion
                // These fields are already scaled by their segment, and are in context 0 (kernel).
```

evm/src/witness/memory.rs
```diff
@@ -69,6 +71,14 @@ impl MemoryAddress {
         })
     }
 
+    pub(crate) fn new_bundle(addr: U256) -> Result<Self, ProgramError> {
```
dvdplm, line 74: Docs perhaps?

dvdplm: Why do we need to convert the limbs into `U256` and then run the checks in `new_u256s()` and then convert to `usize`? Isn't there a way to split `addr` into the three `usize` we need and check `bits()`/`Segment::COUNT` on them directly? Save a round trip?

evm/src/cpu/kernel/constants/mod.rs
dvdplm: Discriminants are a bit under-specified in Rust imo and not super nice to work with. Can we perhaps use an explicit `repr(u64)` so that we can rest in peace wrt to weird platform specifics or any changes to how they are represented internally by the compiler in the future?

dvdplm: I guess I was also wondering "why wasn't u32 enough?"

evm/src/cpu/contextops.rs
dvdplm: Ok, fair enough then (I guess you could have used `filter()` to exclude index `2` but this is fine too.

Nashtare: filter would be cleaner though, will update it!

evm/src/cpu/kernel/constants/mod.rs
Nashtare: > Can we perhaps use an explicit repr(u64) so that we can rest in peace wrt to weird platform specifics or any changes to how they are represented internally by the compiler in the future?
I'm not opposed to that (though we've been so extensively using `usize` as an equivalent of `u64` in the rest of the codebase that I don't think this alone would bring peace of heart..

As for:
> I guess I was also wondering "why wasn't u32 enough?"

That's because now, each `Segment` variant is scaled by `1 << 32`, which doesn't fit in a `u32` (obviously), hence the change in type.

evm/src/cpu/cpu_stark.rs
```diff
@@ -225,17 +225,17 @@ pub(crate) const MEM_CODE_CHANNEL_IDX: usize = 0;
...
 /// - `segment`, shifted by 2^32 (i.e. at index 1)
 /// - `virtual`, not shifted (i.e. at index 0)
-pub(crate) const fn get_addr<T: Copy>(lv: &CpuColumnsView<T>) -> (T, T, T) {
-    let addr_context = lv.mem_channels[0].value[2];
-    let addr_segment = lv.mem_channels[0].value[1];
-    let addr_virtual = lv.mem_channels[0].value[0];
+pub(crate) const fn get_addr<T: Copy>(lv: &CpuColumnsView<T>, mem_channel: usize) -> (T, T, T) {
+    let addr_context = lv.mem_channels[mem_channel].value[2];
+    let addr_segment = lv.mem_channels[mem_channel].value[1];
+    let addr_virtual = lv.mem_channels[mem_channel].value[0];
```
dvdplm, line 238: What I had in mind here was actually:

```rust
pub(crate) const fn get_addr<T: Copy, const IDX: usize>(lv: &CpuColumnsView<T>) -> (T, T, T) {
    let addr_context = lv.mem_channels[IDX].value[2];
    let addr_segment = lv.mem_channels[IDX].value[1];
    let addr_virtual = lv.mem_channels[IDX].value[0];
    (addr_context, addr_segment, addr_virtual)
}
```
…but I think what you did is more readable; `get_addr::<_, 0>(lv)` is just awkward to type.

4l0n50 (approved):
I'm just a little concerned about how error prone are all those >>, <<. Maybe also adding already scaled down constants? Other than that, looks good to me

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1281,6 +1271,17 @@ fn get_mnemonic(opcode: u8) -> &'static str {
     }
 }
 
+#[macro_use]
+macro_rules! unpack_address {
```
dvdplm, line 1275: ❤️

evm/src/witness/memory.rs
```diff
@@ -71,6 +71,9 @@ impl MemoryAddress {
         })
     }
 
+    /// Creates a new `MemoryAddress` from a bundled address fitting a `U256`.
+    /// It will recover the virtual offset asthe lowest 32-bit limb, the segment
```
dvdplm, line 75: s/asthe/as the/

evm/src/cpu/kernel/asm/account_code.asm
```diff
@@ -94,18 +96,19 @@ load_code_ctd:
     DUP1 ISZERO %jumpi(load_code_non_existent_account)
     // Load the code non-deterministically in memory and return the length.
     PROVER_INPUT(account_code)
-    %stack (code_size, codehash, ctx, retdest) -> (ctx, @SEGMENT_CODE, 0, code_size, codehash, retdest, code_size)
+    %stack (code_size, codehash, ctx, retdest) -> (ctx, code_size, codehash, retdest, code_size)
     // Check that the hash of the loaded code equals `codehash`.
+    // ctx == DST, as SEGMENT_CODE == offset == 0.
```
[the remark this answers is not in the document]
Nashtare: Yeah I can't see why it would ever change (and we would have issues with memory stark preinitialization) but I guess I could add a comment saying this segment should always be zero.

hratoanina (commented):
Review of the Rust code (not the assembly files).

evm/src/cpu/kernel/constants/global_metadata.rs
```diff
@@ -98,7 +105,7 @@ impl GlobalMetadata {
             Self::LargestContext,
             Self::MemorySize,
             Self::TrieDataSize,
-            Self::RlpDataSize,
+            Self::RlpDataPos,
```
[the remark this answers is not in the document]
Nashtare: Probably a leftover from the initial work of changing all offset operations to address operations. Will revert it.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1447,14 +1469,28 @@ fn get_mnemonic(opcode: u8) -> &'static str {
...
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
 
+    use ethereum_types::U256;
+
     use crate::cpu::kernel::constants::context_metadata::ContextMetadata;
     use crate::cpu::kernel::interpreter::{run, Interpreter};
-    use crate::memory::segments::Segment;
+    use crate::memory::segments::{Segment, SEGMENT_SCALING_FACTOR};
```
hratoanina, line 1491: Same as above.

hratoanina (approved):
Mostly nits, thanks a lot for this one!

evm/spec/cpulogic.tex
```diff
@@ -93,29 +92,49 @@ \subsection{Privileged instructions}
...
+
+
 \subsection{Stack handling}
 \label{stackhandling}
 
 \subsubsection{Top of the stack}
 
 The majority of memory operations involve the stack. The stack is a segment in memory, and stack operations (popping or pushing) use the memory channels.
-Every CPU instruction performs between 0 and 4 pops, and may push at most once. However, for efficiency purposes, we hold the top of the stack in
+Every CPU instruction performs between 0 and 2 pops, and may push at most once. However, for efficiency purposes, we hold the top of the stack in
```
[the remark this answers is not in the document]
Nashtare: Woops good catch!

PR #1427, opened 2023-12-14
fix: make `from_noncanonical_biguint` work for zero

shuklaayush:
Fixes `index out of bounds: the len is 0 but the index is 0` error for `GoldilocksField::from_noncanonical_biguint(BigUint::zero())`


wborgeaud (approved):
Good catch, thanks!

PR #1428, opened 2023-12-15
Change context to current context for BN precompiles

wborgeaud:
Storing data at context=0 for every call to the BN multiplication precompile causes bugs when the precompile is called more than once.
For example, [this tx](https://etherscan.io/tx/0x78a7318c6cb4df2d26421e6027e665ae75af3db2f6caebee101ace36f698727f)  (a zkSync proof verification) fails.
Fixed by storing data in the current context instead.


Nashtare (approved):
Thank you, looks good!

Nashtare:
Side note @wborgeaud, shouldn't we need to do similar changes for `SEGMENT_KERNEL_ECDSA_TABLE` and `SEGMENT_KERNEL_BN_PAIRING`?

wborgeaud:
> Side note @wborgeaud, shouldn't we need to do similar changes for `SEGMENT_KERNEL_ECDSA_TABLE` and `SEGMENT_KERNEL_BN_PAIRING`?

Hm I think we should be safe with the ECDSA code, not sure about the pairing but it's also called multiple times in some txns. But I guess it doesn't hurt to change the context to be on the safe side. Will address this in a separate PR.

PR #1430, opened 2023-12-15
Add IS_READ column to BytePacking CTL

hratoanina:
In the CPU <-> BytePacking CTL, we were not checking whether the operation was a byte packing (with memory writes) or a byte unpacking (with memory reads). This could lead to unwanted writes in memory.


Nashtare:
Unless that's a bug on my end, it seems you've included this commit by mistake on the ERC-721 PR?

PR #1433, opened 2023-12-18
Filter range checks

LindaGuiga:
This PR enables the use of filters and linear combinations of columns for range-checks. This is useful to remove a memory column from PR #1397.


hratoanina (approved):
Looks good, thanks!

PR #1437, opened 2023-12-21
add Debug trait to PartitionWitness to enable trace information output

wizicer:
With this minor change, user can visualize their trace in following tool without code modification in plonky2 repo.

https://plonk.pro/play/index.html#/analyzer/plonky2


Nashtare (approved):
Thank you!

PR #1438, opened 2023-12-21
Add alternative method to prove txs without pre-loaded table circuits

### LINKED ISSUE #1394: Prover initialization and `AllRecursiveCircuits` API
At the moment, a prover will call `AllRecursiveCircuits` with a set of ranges for the different tables it aims at supporting, do the preprocessing, and be ready to prove incoming transaction payloads.
However, this comes at the disadvantage that:
- we need to be able to support any scenario (even though most txs may be small, say `ERC20` / `ERC721`), and this implies specifying fairly large ranges for the different tables (especially the main ones like `MemoryStark`)
- this incurs a non-negligible overhead on the RAM consumption of the prover, as loading all these preprocessed circuits can take about 20GB with large ranges.

Given that we would most likely want "smaller" workers for most of the time, and some "beefy" ones available in case we encounter an OOM issue upon particularly heavy contract deployment / contract call, I think we may want to review the way prover initialization is done.

The idea is to:
- pre-load and save to disk ***all*** the preprocessed circuits we *may* need to use, along with the upper circuits (`root`, `aggreg`, `block`). These upper circuits would be stored in a dedicated file, as would be every table - degree_size individual pair.
Workers, upon bootstrapping, would only load the upper circuits (common to all), and engage in generating the `AllStark` proof.
They would then only load from disk the 7 different circuits they need to recursively shrink the obtained STARK proofs, and carry on with the txn process.

This has the following advantages:
- faster bootstrapping time: workers would only fetch ~2GB of data instead of ~20GB. Additionally workers entering an aggreg/block proof job don't even need to fetch additional circuits. (This could even be used for an aggreg-tier like worker with even smaller hardware requirements)
- lower memory requirements for workers: if we were to grant each 32GB of RAM, they now have ~30GB of memory available instead of ~12, allowing to prove an even wider set of txns with a "smaller" hardware config.

This would require changing the `AllRecursiveCircuits` API slightly, and making some internal methods public so that the degree-read + circuit fetching from disk can be handled outside of `plonky2`.

Nashtare:
This PR:
- Allows to write/load only upper circuits when serializing/deserializing the prover state (i.e. by skipping all the STARK table circuits).
- Adds an alternative method to prove txns (`prove_root_light()`) by passing an additional path to a folder containing all different table circuits. After generating the initial STARK, the prover will read only the necessary circuit files, deserializes them and process to recursion. The files are expected to be stored at `initial_path/TableName_size.ser`.
- Also removes the `expand` methods, which were only added previously for easier testing but ended up never being really useful.

cc @cpubot: For zero-bin / eth-tx-proof, we would:

- still be generating the initial prover state as usual.
- instead of writing the entire state in a single file, we would write the upper circuits to a single file, which would be the default prover state (and sufficient for aggreg/block proofs).
- for each individual table, write them to a dedicated file within the same folder, with the format above.
- pass the serializers and the folder containing serialized tables when calling `generate_txn_proof`, with new signature below:

_Important note_: when writing the individual tables to file, we *MUST* prepend the byte payload by a byte indicating the position of this table in the initial range specified when constructing the prover state. Say we specified ranges 16..28 for the `Arithmetic` table, we should write first 3 in the file `Arithmetic_19.ser` before writing the actual serialized circuit. This is necessary to be able to prove transactions recursively.

```rust
pub fn generate_txn_proof(
    p_state: &ProverState,
    preprocessed_tables_path: &Path,
    start_info: TxnProofGenIR,
    gate_serializer: &dyn GateSerializer<GoldilocksField, 2>,
    generator_serializer: &dyn WitnessGeneratorSerializer<GoldilocksField, 2>,
    abort_signal: Option<Arc<AtomicBool>>,
) -> ProofGenResult<GeneratedTxnProof> {
```

Please let me know if there's anything that you'd like changed. I can also help with the integration in the decoder and zero-bin.

closes #1394

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -566,7 +566,7 @@ impl GrandProductChallenge<Target> {
 
 /// Like `PermutationChallenge`, but with `num_challenges` copies to boost soundness.
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub(crate) struct GrandProductChallengeSet<T: Copy + Eq + PartialEq + Debug> {
+pub struct GrandProductChallengeSet<T: Copy + Eq + PartialEq + Debug> {
     pub(crate) challenges: Vec<GrandProductChallenge<T>>,
 }
 
--- a/evm/src/fixed_recursive_verifier.rs
+++ b/evm/src/fixed_recursive_verifier.rs
@@ -1,9 +1,11 @@
 use core::mem::{self, MaybeUninit};
 use std::collections::BTreeMap;
 use std::ops::Range;
+use std::path::Path;
 use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 
+use anyhow::anyhow;
 use eth_trie_utils::partial_trie::{HashedPartialTrie, Node, PartialTrie};
 use hashbrown::HashMap;
 use itertools::{zip_eq, Itertools};
@@ -23,6 +25,7 @@ use plonky2::plonk::config::{AlgebraicHasher, GenericConfig};
 use plonky2::plonk::proof::{ProofWithPublicInputs, ProofWithPublicInputsTarget};
 use plonky2::recursion::cyclic_recursion::check_cyclic_proof_verifier_data;
 use plonky2::recursion::dummy_circuit::cyclic_base_proof;
+use plonky2::util::serialization::gate_serialization::default;
 use plonky2::util::serialization::{
     Buffer, GateSerializer, IoResult, Read, WitnessGeneratorSerializer, Write,
 };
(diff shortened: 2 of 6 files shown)


LindaGuiga (approved):
LGTM, thanks!

Nashtare:
@cpubot I wasn't super fond of this either, but not sure on how to handle it best. With your suggestion, we would then have the logic related to disk IO and serialization in the decoder?

Nashtare:
I've updated the logic, let me know if that looks better now.

As for 
> which would be zero-bin in this case.

I think my concern was that we now have to to txn proofs in three steps, generate the initial STARK proof, load the necessary circuits for this one, and use the new method to finish (`prove_root_after_initial_stark` in my latest commit). This was initially wrapped and handled by plonky-block-proof-gen (proof-protocol-decoder now) hence why I'm not sure I see how this would cleanly be done within zero-bin, but I guess that's not a big issue.

PR #1439, opened 2023-12-21
Implement CTL bundling

LindaGuiga:
This PR bundles, within a CTL, the looking tables associated to the same STARK together. This is done using helper columns. 

Even though this adds helper columns when there are only 1 or 2 looking tables to bundle together, we can almost halve the number of commited columns for large groups of looking tables.
For example, `MemoryStark` has 136 looking tables coming from `KeccakSponge` reads, but we only have to commit to 69 polynomials (68 helper columns and 1 Z polynomial) with the bundling.


evm/src/cross_table_lookup.rs
```diff
@@ -466,16 +469,81 @@ impl<F: Field> CrossTableLookup<F> {
...
+        table: Table,
+        num_challenges: usize,
+        constraint_degree: usize,
+    ) -> usize {
+        let mut num_ctls = 0;
+        for ctl in ctls {
+            let all_tables = std::iter::once(&ctl.looked_table).chain(&ctl.looking_tables);
+            let num_appearances = all_tables.filter(|twc| twc.table == table).count();
+            if num_appearances > 0 {
+                num_ctls += ceil_div_usize(num_appearances, constraint_degree - 1) + 1;
```
Nashtare, line 486: not sure how impactful it is, but `constraint_degree - 1` being fixed here, we may want to consider dividing only once (same below)?

evm/src/cross_table_lookup.rs
```diff
@@ -509,12 +582,34 @@ impl<F: Field> CtlData<F> {
...
 
-    /// Returns all the cross-table lookup polynomials.
-    pub(crate) fn z_polys(&self) -> Vec<PolynomialValues<F>> {
-        self.zs_columns
-            .iter()
-            .map(|zs_columns| zs_columns.z.clone())
-            .collect()
+    /// Returns all the cross-table lookup helper polynomials.
+    pub(crate) fn ctl_helper_polys(&self) -> Vec<PolynomialValues<F>> {
+        let mut res = vec![];
```
Nashtare, line 587: pre-allocating all those vec![] may help a bit.

evm/src/cross_table_lookup.rs
```diff
@@ -658,180 +785,447 @@ pub(crate) fn cross_table_lookup_data<F: RichField, const D: usize>(
...
-                    .push(CtlZData {
-                        z,
-                        challenge,
-                        columns: table.columns.clone(),
-                        filter: table.filter.clone(),
-                    });
+
+            // Group looking tables by `Table`, since we group together all looking tables associated to the same STARK
+            // thanks to the helper columns.
+            let mut grouped_tables = HashMap::new();
```
Nashtare, line 805: pre-allocating may help

evm/src/cross_table_lookup.rs
```diff
@@ -658,180 +785,447 @@ pub(crate) fn cross_table_lookup_data<F: RichField, const D: usize>(
...
+pub(crate) fn get_helper_cols<F: Field>(
+    trace: &[PolynomialValues<F>],
+    degree: usize,
+    columns: &[&[Column<F>]],
+    filter_cols: &[&Option<Filter<F>>],
+    challenge: GrandProductChallenge<F>,
+    constraint_degree: usize,
+) -> Vec<PolynomialValues<F>> {
+    let num_helper_columns = ceil_div_usize(columns.len(), constraint_degree - 1) + 1;
+    let mut helper_columns: Vec<PolynomialValues<F>> = Vec::with_capacity(num_helper_columns);
```
Nashtare, line 859: nit: is the type hint needed?

evm/src/cross_table_lookup.rs
```diff
@@ -658,180 +785,447 @@ pub(crate) fn cross_table_lookup_data<F: RichField, const D: usize>(
...
+            CrossTableLookup {
+                looking_tables,
+                looked_table,
+            },
+            num_ctls,
+        ) in cross_table_lookups.iter().zip(num_helper_ctl_columns)
         {
             for &challenges in &ctl_challenges.challenges {
+                // Group looking tables by `Table`, since we bundle the looking tables taken from the same `Table` together thanks to helper columns.
+                let mut grouped_tables = HashMap::new();
```
Nashtare, line 1102: pre-allocating may help

evm/src/cross_table_lookup.rs
```diff
@@ -658,180 +785,447 @@ pub(crate) fn cross_table_lookup_data<F: RichField, const D: usize>(
...
+                for looking_table in looking_tables {
+                    let cur = grouped_tables
+                        .entry(looking_table.table as usize)
+                        .or_insert((vec![], vec![]));
```
Nashtare, line 1106: nit
```suggestion
                        .or_default();
```

evm/src/lookup.rs
```diff
@@ -64,6 +67,21 @@ pub(crate) fn lookup_helper_columns<F: Field>(
...
 
+    let looking_cols = lookup
+        .columns
+        .iter()
+        .map(|col| vec![col.clone()])
+        .collect::<Vec<Vec<Column<F>>>>();
+    let lookup_cols = looking_cols
+        .iter()
+        .map(|col| &col[..])
+        .collect::<Vec<&[Column<F>]>>();
```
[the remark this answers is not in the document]
LindaGuiga: The final type we need for `get_helper_cols` is `&[&[Column<F>]]`, and we only have a `Vec<Column<F>>`. The only way I found to reach the final type was through this intermediary step of creating a `Vec<Vec<_>>`. I could,  change the type in `get_helper_cols` but I'm afraid it would to cloning on the CTL side...

evm/src/cross_table_lookup.rs
LindaGuiga: If I understand what you mean correctly, I'm not sure we can do that here: since it's a `ceil_div_usize` for each CTL, we might have an issue if, say, we have 1 appearance in two CTLs: then if we divide at the end, we would get 1, but we should have gotten 2.

LindaGuiga:
Update: With the new optimization PRs, there is no additional column: when there are only 1 or 2 columns for a STARK, we do not create helper columns.

matthiasgoergens:
Why so complicated?  Why introduce a completely new helper, instead of just bundling eg every two lookups into a single partial sum?

PR #1440, opened 2023-12-27
Add initial constraint z polynomial

LindaGuiga:
I noticed that in the LogUp checks for range-checks, we do not check that the `Z` polynomial is initialized to 0.


Nashtare (approved):
Good catch, thanks!

muursh (approved):

PR #1441, opened 2024-01-02
adding link to `plonky2-gpu`

liangping:


Nashtare:
Unlike the other links to plonky2-related repos that have been developed by Polygon Zero, this repository is from a third-party, hence I wouldn't be so inclined to include it in the README here, but that is my sole opinion.

muursh:
> Unlike the other links to plonky2-related repos that have been developed by Polygon Zero, this repository is from a third-party, hence I wouldn't be so inclined to include it in the README here, but that is my sole opinion.

Yeah agreed here tbh. I think we can probably close this for now unless someone has a good argument for its inclusion

Issue #1376, opened 2023-11-21
Improve Rust Documentation for EVM


pgebheim:
## Goal

Create high-level rust documentation that encompasses the API for the zk-evm implementation.

## Tasks

Priority:
- [ ] Add module level documentation to evm/src/lib.rs - High level overview of how to use the evm proving system, link to sub-documentation for interfaces, data structures, and main interface. This content should be able to be used in the main evm README. [Check out this example from Paladin](https://github.com/0xPolygonZero/paladin/blob/main/paladin-core/src/lib.rs)
- [ ] Document prover.rs - Code examples of how major API entrypoints. [Example from Paladin](https://github.com/0xPolygonZero/paladin/blob/b2f04f0ef5112035f4e6a044aaad7c494bfe99e7/paladin-core/src/runtime/mod.rs#L218).
- [ ] Document the major types used within prover.rs during proof generation

Goal:
- [ ] Add documentation with example across the board for all public APIs (modules in the evm/src). Examples should represent test-cases such that examples in the documentation is actionable.

Nashtare:
@pgebheim does the linked PR satisfy what you had in mind? I'll let you close this or clarify if you want more specific doc

PR #1444, opened 2024-01-04
Add crate-level documentation

Nashtare:
Add crate-level documentation and API explanations for the evm crate.

Should address #1376


muursh (approved):
I'll hold back judgement as to whether it meets to desires of 1367 but what is here looks fine to me.

Nashtare:
I guess @pgebheim is the best to answer this question!

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! transaction proofs and aggregation proofs. We only need to specify for each child if it is an aggregation proof
+//! or a regular one.
+//!
+//! ```ignore
+//! let (proof_1, pv_1) =
+//!     prover_state.prove_root(all_stark, config, inputs_1, &mut timing, None);
+//! let (proof_2, pv_2) =
+//!     prover_state.prove_root(all_stark, config, inputs_2, &mut timing, None);
+//! let (proof_3, pv_3) =
+//!     prover_state.prove_root(all_stark, config, inputs_2, &mut timing, None);
```
dvdplm, line 81:
```suggestion
//!     prover_state.prove_root(all_stark, config, inputs_3, &mut timing, None);
```

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! let (agg_proof_1_2, pv_1_2) =
+//!     prover_state.prove_aggreg(false, proof_1, pv_1, false, proof_2, pv_2);
+//!
+//! // Now aggregate the newly generated aggregation proof with the last regular txn proof.
+//! let (agg_proof_1_3, pv_1_3) =
+//!     prover_state.prove_aggreg(true, agg_proof_1_2, pv_1_2, false, proof_3, pv_3);
+//! ```
+//!
+//! **Note**: The proofs provided to the [prove_aggreg](AllRecursiveCircuits::prove_aggreg) method *MUST* have contiguous states.
+//! Trying to combine `proof_1` and `proof_3` from the example above would fail.
```
dvdplm, line 93: This was a question I had: are we guaranteed to fail if/when states are not contiguous (but otherwise valid)?

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! ```ignore
+//! let mut timing = TimingTree::new("prove", log::Level::Debug);
+//! let kill_signal = None; // Useful only with distributed proving to kill hanging jobs.
+//! let (proof, public_values) =
+//!     prover_state.prove_root(all_stark, config, inputs, &mut timing, kill_signal);
+//! ```
+//!
+//! This outputs a transaction proof and its associated public values. These are necessary during the
+//! aggregation levels (see below). If one were to miss the public values, they are also retrievable directly
+//! from the proof's encoded public inputs, as such:
```
dvdplm, line 60: If the PVs are included in the proof, why are they so often passed around separately? Is it expensive to recover them from the proof's encoded public values?

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! ```ignore
+//! let previous_block_proof = { ... };
+//! let (block_proof, block_public_values) =
+//!     prover_state.prove_block(Some(&previous_block_proof), &agg_proof, agg_pv)?;
+//! ```
+//!
+//! ### Checkpoint heights
+//!
+//! Because the process of always providing a previous block proof when generating a proof for the current block may yield some
+//! undesirable issues, the plonky2 zk-EVM supports checkpoint heights. At given block heights, the prover does not have to pass
```
dvdplm, line 112: > undesirable issues

Such as?

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! ```ignore
+//! let prover_state = AllRecursiveCircuits::<F, C, D>::new(...);
+//!
+//! // Default serializers
+//! let gate_serializer = DefaultGateSerializer;
+//! let generator_serializer = DefaultGeneratorSerializer::<C, D> {
+//!     _phantom: PhantomData::<C>,
+//! };
+//!
+//! // Serializes the prover state to a sequence of bytes
```
dvdplm, line 133: Can you provide an order of magnitude for how big the proofs are when serialized?

plonky2/src/fri/mod.rs
```diff
@@ -15,6 +15,7 @@ mod validate_shape;
 pub mod verifier;
 pub mod witness_util;
 
+/// A configuration for the FRI protocol.
```
dvdplm, line 18: Should we provide a sane `Default` impl for this you think?

evm/src/config.rs
```diff
@@ -1,13 +1,16 @@
 use plonky2::fri::reduction_strategies::FriReductionStrategy;
 use plonky2::fri::{FriConfig, FriParams};
 
+/// A configuration containing the different parameters to be used by the STARK prover.
 pub struct StarkConfig {
```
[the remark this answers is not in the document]
Nashtare: That's a good suggestion, this is already done for `CircuitConfig` with the default `standard_recursion_config()`.

evm/src/fixed_recursive_verifier.rs
```diff
@@ -373,6 +398,16 @@ where
...
 
     /// Preprocess all recursive circuits used by the system.
+    ///
+    /// # Arguments
+    ///
+    /// - `all_stark`: a structure defining the logic of all STARK modules and their associated
+    /// cross-table lookups.
+    /// - `degree_bits_ranges`: the ranges to be supported for each STARK module. The values in these
+    /// ranges correspond to the binary log of the trace lengths of the STARK tables. Specifying a wide
+    /// enough range allows a prover to cover all possible scenarios from arbitrary transactions.
```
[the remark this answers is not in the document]
Nashtare: I was puzzled as well when writing it, would this work better?

```rust
/// - `degree_bits_ranges`: the logarithmic ranges to be supported for the recursive tables.
/// Transactions may yield arbitrary trace lengths for each STARK module (within some bounds),
/// unknown prior generating the witness to create a proof. Thus, for each STARK module, we
/// construct a map from `2^{degree_bits} = length` to a chain of shrinking recursion circuits,
/// starting from that length, for each `degree_bits` in the range specified for this STARK module.
/// Specifying a wide enough range allows a prover to cover all possible scenarios.
```

evm/src/lib.rs
Nashtare: Yes, because we try connecting the states (with some equality checks in the circuit), for instance that the gas used after txn 1 is equal to the gas used prior txn 2. So the prover would not be able to generate a proof.

evm/src/lib.rs
Nashtare: It's more as a matter of convenience than anything else. It shouldn't be too heavy to retrieve them from the encoded public inputs, but as they are there anyway, it cost nothing to attach them to the output proof. I guess to be a tad cleaner, we could have a custom proof type, like `EvmProof` that consists in a `ProofWithPublicInputs` and `PublicValues`, to hide the tuple behind.

evm/src/lib.rs
Nashtare: the proofs or the prover state?

evm/src/lib.rs
Nashtare: If proving is decentralized for instance, and a prover falls behind. Or if a prover was to crash and lose its DB, with previously generated proofs, it would need to start again from genesis.

plonky2/src/fri/mod.rs
Nashtare: This one may be less relevant, because typically `StarkConfig` and `CircuitConfig` may deserve their own `Default` impl, but they have drastically different `FriConfig`, one tailored for proving time, the other one for proof size and recursion efficiency. So I'd rather not here.

evm/src/lib.rs
Nashtare: The prover state will range between 20GB to 25GB when serialized.
The proofs are in the ballpark of ~150kB, regardless of the aggregation level, based on the current config (although the initial STARK proofs for a transaction, depend on the actual transaction execution, and may be several MBs large).

plonky2/src/fri/mod.rs
dvdplm: Fair, your call. The counter argument would be: if incompetent people try to use some code, then we should help them out by providing a default that is at least safe-ish so they don't accidentally run with unsafe parameters (I don't know if this can happen here).

evm/src/lib.rs
dvdplm: prover state! I'm picturing some poor devops person that is trying to figure out how to set up a proving machine. Could be useful to have a hunch of how much disk space is actually needed.

evm/src/lib.rs
dvdplm: decentralized? You mean distributed yeah?

So by "undesirable issues" you mean "avoiding restarting from scratch after a crash or missing previous data"? Might as well write that then! :)

Do you think we should mention that using the checkpoint feature is probably not a good idea for production systems (but awesome for development)?

plonky2/src/fri/mod.rs
Nashtare: We can't really have unsafe parameters here, because if for some reasons those were unsound (typically not matching the targeted security level), one could not create a `CircuitBuilder` instance, needed to construct circuits (it checks it).

The only risk is that someone may use a configuration not suited for their application, but well in general I wouldn't recommend people not familiar at all with this kind of things to even think using such library for anything else than fun/learning purposes 🙃

evm/src/lib.rs
dvdplm: It's aesthetically a bit annoying to pass around duplicate data, but if it's convenient as-is I'll shut up.

evm/src/lib.rs
Nashtare: The alternative is to split the prover state (it's been rendered possible with #1438), by loading only upper circuits needed after shrinking the STARK proofs and split the transaction proof process into three chunks:

- compute the initial STARK proofs
- load the related recursive table circuits
- finish the txn proof (basically what's in `prove_root` after the first line)

But this is out of scope of the plonky2_evm lib.

evm/src/fixed_recursive_verifier.rs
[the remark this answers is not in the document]
dvdplm: Much better! 🙏

dvdplm (approved):
(Would still be good to get a review from @4l0n50 before merging)

evm/src/lib.rs
Nashtare: I did mean decentralized. We could picture a rewarding scenario where first prover to submit a valid proof from checkpoint A to checkpoint B is rewarded.

Even in the case of centralized proving, the checkpoint feature is useful for the other issues I mentioned.

Nashtare: Checkpointing can really be seen as skipping unnecessary work. Once you've submitted your proof on your settlement layer, and it's been verified and approved, your state up to block N is verified. Hence no need to keep track to what previously happen, and locally do a proof from block N to block N+c for some constant c. Proper connection between the two proofs should be done on the settlement layer, to ensure that the starting state of the incoming proof was the final state of the previously submitted proof (much like what's being done when aggregating transactions, except that this is done internally in the circuits).

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//!
+//! ```ignore
+//! let previous_block_proof = { ... };
+//! let (block_proof, block_public_values) =
+//!     prover_state.prove_block(Some(&previous_block_proof), &agg_proof, agg_pv)?;
+//! ```
+//!
+//! ### Checkpoint heights
+//!
+//! Because the process of always providing a previous block proof when generating a proof for the current block may yield some
```
4l0n50, line 111:
```suggestion
//! The process of always providing a previous block proof when generating a proof for the current block may yield some
```

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! undesirable issues, the plonky2 zk-EVM supports checkpoint heights. At given block heights, the prover does not have to pass
```
4l0n50, line 112:
```suggestion
//! undesirable issues. For this reason the plonky2 zk-EVM supports checkpoint heights. At given block heights, the prover does not have to pass
```

4l0n50: It's just because the "Because of..." is also used later

4l0n50 (commented):
Looks good to me. Just minor comments

evm/src/lib.rs
```diff
@@ -1,3 +1,153 @@
...
+//! // Now aggregate proofs for txn 1 and 2.
+//! let (agg_proof_1_2, pv_1_2) =
+//!     prover_state.prove_aggreg(false, proof_1, pv_1, false, proof_2, pv_2);
+//!
+//! // Now aggregate the newly generated aggregation proof with the last regular txn proof.
+//! let (agg_proof_1_3, pv_1_3) =
+//!     prover_state.prove_aggreg(true, agg_proof_1_2, pv_1_2, false, proof_3, pv_3);
+//! ```
+//!
+//! **Note**: The proofs provided to the [prove_aggreg](AllRecursiveCircuits::prove_aggreg) method *MUST* have contiguous states.
```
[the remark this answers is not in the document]
Nashtare: Typo on my side, it's actually `prove_aggregation`! Fixed it, thanks!

PR #1449, opened 2024-01-08
Fix `after_mpt_delete_extension_branch`

wborgeaud:
The current implementation of `after_mpt_delete_extension_branch` is buggy.
This PR fixes it and adds a test that fails on the current main.

### DIFF
--- a/evm/src/cpu/kernel/tests/mpt/delete.rs
+++ b/evm/src/cpu/kernel/tests/mpt/delete.rs
@@ -1,7 +1,8 @@
 use anyhow::{anyhow, Result};
 use eth_trie_utils::nibbles::Nibbles;
 use eth_trie_utils::partial_trie::{HashedPartialTrie, PartialTrie};
-use ethereum_types::{BigEndianHash, H256};
+use ethereum_types::{BigEndianHash, H256, U512};
+use rand::random;
 
 use crate::cpu::kernel::aggregator::KERNEL;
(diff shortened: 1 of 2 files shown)


evm/src/cpu/kernel/tests/mpt/delete.rs
```diff
@@ -48,6 +49,32 @@ fn mpt_delete_branch_into_hash() -> Result<()> {
...
+    };
+    let nibbles = Nibbles::from_bytes_be(&random::<[u8; 5]>()).unwrap();
+    let state_trie = Node::Extension {
+        nibbles,
+        child: branch.into(),
+    }
+    .into();
+    let key = nibbles.merge_nibbles(&Nibbles {
+        packed: U512::zero(),
+        count: 64 - 2 * 5,
```
LindaGuiga, line 73: I'm struggling to understand the `- 2 * 5` here. Is there a specific reason for it?

LindaGuiga (approved):
Looks good to me, thank you!

muursh (approved):

evm/src/cpu/kernel/tests/mpt/delete.rs
wborgeaud: `2*5` is just `nibbles.count`, so we basically add `64-nibbles.count` zeros to `nibbles` to get the key.
I replaced `2*5` with `nibbles.count` to make this clearer.

Nashtare (approved):

PR #1450, opened 2024-01-09
Remove full memory channel

hratoanina:
Following https://github.com/0xPolygonZero/plonky2/pull/1426, we can now remove a memory channel.

This allows us to remove 13 CPU columns (down to 85), at the cost of some CTL columns:
- For SET_CONTEXT, we use two CPU CTL columns to write the old stack length and read the new stack length.
- For syscalls and exceptions, we bundle the three contiguous jumptable reads with one CPU CTL column.


evm/src/cpu/contextops.rs
```diff
@@ -326,18 +234,26 @@ fn eval_ext_circuit_set<F: RichField + Extendable<D>, const D: usize>(
...
     }
 
+    // Unused channels.
+    for i in 1..NUM_GP_CHANNELS {
+        if i != 2 {
+            let channel = lv.mem_channels[i];
+            let constr = builder.mul_extension(filter, channel.used);
+            yield_constr.constraint(builder, constr);
+        }
+    }
```
dvdplm, line 256: This is something we do a fair bit (repeated code). Perhaps a candidate for moving to its own function? And it looks like it could be expensive too, should it use a parallel iterator (can it?)?

evm/src/all_stark.rs
```diff
@@ -254,6 +270,8 @@ fn ctl_memory<F: Field>() -> CrossTableLookup<F> {
     let all_lookers = iter::once(cpu_memory_code_read)
         .chain(cpu_memory_gp_ops)
         .chain(iter::once(cpu_push_write_ops))
+        .chain(iter::once(cpu_set_context_write))
+        .chain(iter::once(cpu_set_context_read))
         .chain(keccak_sponge_reads)
         .chain(byte_packing_ops)
         .collect();
```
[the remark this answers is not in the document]
Nashtare: No the order doesn't matter, and agreed it would be more readable that way.

evm/src/cpu/contextops.rs
```diff
@@ -96,7 +96,7 @@ fn eval_packed_get<P: PackedField>(
 
     // Unused channels.
     for i in 1..NUM_GP_CHANNELS {
-        if i != 3 {
+        if i != 2 {
```
Nashtare, line 99: here I'd just remove the `if` and do the `for` loop with `for i in 3..NUM_GP_CHANNELS`, eventually adding a comment for clarity if needed

evm/src/cpu/contextops.rs
```diff
@@ -138,7 +138,7 @@ fn eval_ext_circuit_get<F: RichField + Extendable<D>, const D: usize>(
...
```
Nashtare, line 141: same as above

evm/src/cpu/contextops.rs
```diff
@@ -200,17 +176,24 @@ fn eval_packed_set<P: PackedField>(
...
+        .zip(lv.mem_channels[2].value.iter())
     {
         yield_constr.constraint(
             lv.op.context_op * lv.general.stack().stack_inv_aux_2 * (limb_new_top - limb_read_top),
         );
     }
 
+    // Unused channels.
+    for i in 1..NUM_GP_CHANNELS {
+        if i != 2 {
```
Nashtare, line 192: same as above

evm/src/cpu/contextops.rs
```diff
@@ -326,18 +234,26 @@ fn eval_ext_circuit_set<F: RichField + Extendable<D>, const D: usize>(
...
     {
         let diff = builder.sub_extension(limb_new_top, limb_read_top);
         let prod = builder.mul_extension(lv.general.stack().stack_inv_aux_2, diff);
         let constr = builder.mul_extension(lv.op.context_op, prod);
         yield_constr.constraint(builder, constr);
     }
 
+    // Unused channels.
+    for i in 1..NUM_GP_CHANNELS {
+        if i != 2 {
```
Nashtare, line 251: same as above.

evm/src/cpu/contextops.rs
Nashtare: I agree with code rep, it could be wrapped in a utility function. As for parallel iterator, I think the rayon overhead would be greater than the gains here.

dvdplm: I had a stab at this with:

```rust
fn unused_channels<P: PackedField>(
    mem_channels: &[MemoryChannelView<P>],
    filter: P,
) {
    mem_channels
        .iter()
        .skip(2)
        .for_each(|channel| yield_constr.constraint(filter * channel.used))
}
```
…but there's something wrong so leaving it to the experts. :/

Nashtare: It should be `skip(3)`, we're starting at 1 and skipping when `i == 2`, effectively ignoring the first 3 elements. Otherwise looks good to me (actually I think it's even more correct, as the code above does not constrain the last `mem_channel` of `lv`, while yours does.

evm/src/cpu/contextops.rs
Nashtare: alternatively I'm not even sure to understand why we constrain only channels 1 and 3 to be unused (range is end-exclusive). There are other channels not used no?

hratoanina: 0 is the top of the stack and may or may not be in use. 2 is not used by GET_CONTEXT, but it's used by SET_CONTEXT and must not be constrained.
All other channels are unused, which is only channel 1 now (since channel 3 doesn't exist anymore).

hratoanina:
@Nashtare @dvdplm 
I introduced the disabling functions for this file only, but it can be extended to the whole codebase (and put in a new `utils` file). I'll do it in another PR to not crowd this one though.

Nashtare:
Why don't you put it already in `cpu_stark.rs`, similarly to where the utility `get_addr` is? Not sure having a PR just to move this around is useful..

evm/src/cpu/cpu_stark.rs
```diff
@@ -189,6 +190,40 @@ pub(crate) fn ctl_filter_byte_unpacking<F: Field>() -> Filter<F> {
     )
 }
 
+/// /// Creates the vector of `Columns` corresponding to three consecutive (byte) reads in memory.
```
Nashtare, line 193:
```suggestion
/// Creates the vector of `Columns` corresponding to three consecutive (byte) reads in memory.
```

evm/src/cpu/cpu_stark.rs
```diff
@@ -319,6 +401,42 @@ pub(crate) fn ctl_filter_partial_memory<F: Field>() -> Filter<F> {
...
+    Filter::new(
+        vec![(
+            Column::single(COL_MAP.op.context_op),
+            Column::single(COL_MAP.opcode_bits[0]),
+        )],
+        vec![],
+    )
+}
+
+pub(crate) fn disable_unused_channels<P: PackedField>(
```
Nashtare, line 416: nit: missing doc

evm/src/cpu/cpu_stark.rs
```diff
@@ -319,6 +401,42 @@ pub(crate) fn ctl_filter_partial_memory<F: Field>() -> Filter<F> {
...
+    filter: P,
+    channels: Vec<usize>,
+    yield_constr: &mut ConstraintConsumer<P>,
+) {
+    for i in channels {
+        yield_constr.constraint(filter * lv.mem_channels[i].used);
+    }
+}
+
+pub(crate) fn disable_unused_channels_circuit<F: RichField + Extendable<D>, const D: usize>(
```
Nashtare, line 427: nit: missing doc

Nashtare (approved):
Thank you, LGTM minus the minor nits!

PR #1452, opened 2024-01-09
Improve some calls to `%mstore_rlp`

Nashtare:
Following #1426, `%mstore_rlp`'s name is not really meaningful, as it only performs `SWAP MSTORE_GENERAL`.
Additionally, most calls can be changed to a single call to `MSTORE_GENERAL` instead by reordering previous stack operations. This addresses it and renames the last remaining calls to `%mstore_rlp` as `%swap_mstore` for clarity.

### DIFF
--- a/evm/src/cpu/kernel/asm/memory/core.asm
+++ b/evm/src/cpu/kernel/asm/memory/core.asm
@@ -410,7 +410,7 @@
 %endmacro
 
 // Store a single byte to @SEGMENT_RLP_RAW.
-%macro mstore_rlp
+%macro swap_mstore
     // stack: addr, value
     SWAP1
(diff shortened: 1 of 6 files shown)


evm/src/cpu/kernel/asm/memory/core.asm
```diff
@@ -410,7 +410,7 @@
 %endmacro
 
 // Store a single byte to @SEGMENT_RLP_RAW.
```
hratoanina, line 412: Should we keep the same comment?

Nashtare: Ah yeah good point, just removed it.

PR #1453, opened 2024-01-09
Add Boolean constraints for `ArithmeticStark`

Nashtare:
The arithmetic flags were not constrained to be boolean.
Also add a comment on the Memory side for clarity on the `IS_READ` boolean implicit constraint.


LindaGuiga (approved):
Looks good, thanks!

PR #1455, opened 2024-01-10
Interpreter GenerationInputs

LindaGuiga:
This PR adds an initilization method to the interpreter based on `GenerationInputs`. Given a new state, it updates it following `apply_metadata_and_tries_memops` on the STARK side. It also updates the `rlp` and `withdrawal` prover inputs, as well as the MPT roots, as is done when initializing a `GenerationState` on the STARK side.

This should make running a variety of EVM tests easier on the interpreter side.

### DIFF
--- a/evm/src/cpu/kernel/interpreter.rs
+++ b/evm/src/cpu/kernel/interpreter.rs
@@ -153,7 +153,7 @@ impl<'a> Interpreter<'a> {
         result
     }
 
-    /// Returns an `Interpreter` given `GenerationInputs`, and assuming we are
+    /// Returns an instance of `Interpreter` given `GenerationInputs`, and assuming we are
     /// initializing with the `KERNEL` code.
     pub(crate) fn new_with_generation_inputs_and_kernel(
         initial_offset: usize,
@@ -196,13 +196,12 @@ impl<'a> Interpreter<'a> {
         result
     }
 
-    /// Initialize interpreter state given `GenerationInputs`. This assumes that the
-    /// loaded code is the KERNEL code.
+    /// Initializes the interpreter state given `GenerationInputs`, using the KERNEL code.
     pub(crate) fn initialize_interpreter_state_with_kernel(&mut self, inputs: GenerationInputs) {
         self.initialize_interpreter_state(inputs, KERNEL.code_hash, KERNEL.code.len());
     }
 
-    /// Initialize interpreter state given `GenerationInputs`.
+    /// Initializes the interpreter state given `GenerationInputs`.
     pub(crate) fn initialize_interpreter_state(
         &mut self,
         inputs: GenerationInputs,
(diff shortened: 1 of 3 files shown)


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -181,6 +196,137 @@ impl<'a> Interpreter<'a> {
         result
     }
 
+    /// Initialize interpreter state given `GenerationInputs`. This assumes that the
```
Nashtare, line 199:
```suggestion
    /// Initializes interpreter state given `GenerationInputs`. This assumes that the
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -181,6 +196,137 @@ impl<'a> Interpreter<'a> {
...
+    /// loaded code is the KERNEL code.
```
Nashtare, line 200: Not sure to understand the second sentence. It's not really assuming per se as it's explicitely hardcoding the use of KERNEL code.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -181,6 +196,137 @@ impl<'a> Interpreter<'a> {
...
+    pub(crate) fn initialize_interpreter_state_with_kernel(&mut self, inputs: GenerationInputs) {
+        self.initialize_interpreter_state(inputs, KERNEL.code_hash, KERNEL.code.len());
+    }
+
+    /// Initialize interpreter state given `GenerationInputs`.
```
Nashtare, line 205:
```suggestion
    /// Initializes the interpreter state given `GenerationInputs`.
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -150,6 +153,18 @@ impl<'a> Interpreter<'a> {
         result
     }
 
+    /// Returns an `Interpreter` given `GenerationInputs`, and assuming we are
```
Nashtare, line 156:
```suggestion
    /// Returns an instance of `Interpreter` given `GenerationInputs`, and assuming we are
```

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -181,6 +196,137 @@ impl<'a> Interpreter<'a> {
...
+        self.generation_state.rlp_prover_inputs = rlp_prover_inputs;
+        self.generation_state.withdrawal_prover_inputs = withdrawal_prover_inputs;
+
+        // Set `GlobalMetadata` values.
+        let metadata = &inputs.block_metadata;
+        let global_metadata_to_set = [
+            (
+                GlobalMetadata::BlockBeneficiary,
+                U256::from_big_endian(&metadata.block_beneficiary.0),
+            ),
```
Nashtare, line 242: probably out of scope for this PR, but we're starting to have quite some duplication between the interpreter and the actual prover, so we may want to start unifying everything. Would be easier to maintain as well.

Nashtare (approved):
Thank you!

PR #1456, opened 2024-01-10
Use current context in ecrecover

wborgeaud:
Similar to #1428 for ecrecover. The current version (storing everything in context=0) should be safe, but it doesn't hurt to change to the current context and that way it's consistent across all precompiles.
Also rename `SEGMENT_KERNEL_BN_PAIRING` to `SEGMENT_BN_PAIRING` since it's already using the current context.


muursh (approved):

Nashtare (approved):
Thanks William! LGTM

PR #1457, opened 2024-01-11
Free up some CPU cycles

Nashtare:
Some light tweaks on the kernel side

- %decrement / %div / %mod -> try preventing the SWAP call
- %div_const(pow_two) -> %shr_const (no SWAP)
- sibling of `encode_rlp_scalar` with swapped inputs (as the `%stack` macro inside becomes actually cheaper)

### DIFF
--- a/evm/src/cpu/kernel/asm/hash/blake2/compression.asm
+++ b/evm/src/cpu/kernel/asm/hash/blake2/compression.asm
@@ -21,10 +21,10 @@ compression_loop:
     // stack: addr, cur_block, retdest
     POP
     // stack: cur_block, retdest
-    PUSH 0
     PUSH 1
+    PUSH 0
     %mload_current_general
-    // stack: num_blocks, cur_block, retdest
+    // stack: num_blocks, 1, cur_block, retdest
     SUB
     // stack: num_blocks - 1, cur_block, retdest
     DUP2
--- a/evm/src/cpu/kernel/asm/hash/blake2/compression.asm
+++ b/evm/src/cpu/kernel/asm/hash/blake2/compression.asm
@@ -22,9 +22,10 @@ compression_loop:
     POP
     // stack: cur_block, retdest
     PUSH 0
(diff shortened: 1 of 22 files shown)


evm/src/cpu/kernel/asm/hash/blake2/compression.asm
```diff
@@ -22,9 +22,10 @@ compression_loop:
     POP
     // stack: cur_block, retdest
     PUSH 0
+    PUSH 1
```
wborgeaud, line 25: Shouldn't this `PUSH 1` be before the `PUSH 0`?

evm/src/cpu/kernel/asm/rlp/num_bytes.asm
```diff
@@ -5,8 +5,12 @@ global num_bytes:
     DUP1 ISZERO %jumpi(return_1)
     // Non-deterministically guess the number of bits
     PROVER_INPUT(num_bits)
-    %stack (num_bits, x) -> (num_bits, x, num_bits)
-    %decrement
+    // stack: num_bits, x
+    SWAP1
+    PUSH 1
+    DUP3
+    // stack: num_bits, 1, x, num_bits
```
wborgeaud, line 12: nit: This is fine, but why not use a `%stack` macro here? I feel like it should be preferred for larger sequences of stack manipulation.

wborgeaud (approved):
LGTM, thanks!

evm/src/cpu/kernel/asm/rlp/num_bytes.asm
Nashtare: Can do indeed :)

evm/src/cpu/kernel/asm/hash/blake2/compression.asm
Nashtare: Oops good catch, also updated the comment above `SUB`

PR #1458, opened 2024-01-11
Constrain `syscall` / `exceptions` filter to be boolean

Nashtare:
One auditor noticed these two flags weren't properly constrained to be boolean.
All other flags are individually constrained in `decode` CPU submodule.


LindaGuiga (approved):
LGTM, thanks!

muursh (approved):

PR #1459, opened 2024-01-11
Add math rendering with Katex

Nashtare:
Add math rendering with `katex`.
Should be automatic for crates.io. For local rendering, one can use:

`RUSTDOCFLAGS="--html-in-header .cargo/katex-header.html" cargo doc --no-deps --open`

I just updated the documentation on the `gates` module and `CosetInterpolationGate` for some rendering idea.


muursh (approved):
lgtm

PR #1460, opened 2024-01-11
Packed rlp prover inputs

4l0n50:
In this small PR the rlp prover inputs are packed into U256 words. It saves 1600 cycles on erc20.


evm/src/cpu/kernel/asm/rlp/read_to_memory.asm
```diff
@@ -13,29 +13,26 @@ global read_rlp_to_memory:
...
-
 read_rlp_to_memory_loop:
     // stack: addr, final_addr, retdest
     DUP2
     DUP2
-    EQ
-    // stack: addr == final_addr, addr, final_addr, retdest
+    LT
+    ISZERO
+    // stack: addr >= final_addr, addr, final_addr, retdest
```
Nashtare, line 22: can't we just use `GE`?

evm/src/cpu/kernel/asm/rlp/read_to_memory.asm
```diff
@@ -13,29 +13,26 @@ global read_rlp_to_memory:
...
+    MSTORE_32BYTES_32
     // stack: addr', final_addr, retdest
     %jump(read_rlp_to_memory_loop)
 
 read_rlp_to_memory_finish:
     // stack: addr, final_addr, retdest
     // we recover the offset here
     PUSH @SEGMENT_RLP_RAW // ctx == virt == 0
-    DUP2 SUB
+    DUP3 SUB
```
Nashtare, line 35: this is changed but no stack doc is changed, so `pos` now is referring to the final offset instead of the current one?

Nashtare (approved):
LGTM, just minor nit

evm/src/cpu/kernel/asm/rlp/read_to_memory.asm
```diff
@@ -13,29 +13,26 @@ global read_rlp_to_memory:
...
-    // stack: addr, len, retdest
-    DUP1
-    PROVER_INPUT(rlp)
-    // stack: byte, addr, addr, final_addr, retdest
-    MSTORE_GENERAL
     // stack: addr, final_addr, retdest
-    %increment
+    PROVER_INPUT(rlp)
+    SWAP1
+    MSTORE_32BYTES_32
```
Nashtare, line 27: We need to make sure we don't have an empty blob (length 0), otherwise it would fail the CTL because of MSTORE_UNPACKING.

evm/src/cpu/kernel/asm/rlp/read_to_memory.asm
Nashtare: Nevermind, I saw 2 macros using GE so I thought we could use it, but it actually doesn't exist 🤦🏼 
So looks good! (the macros are never used hence why it didn't trigger compile errors)

evm/src/cpu/kernel/asm/rlp/read_to_memory.asm
4l0n50: In the original version the condition for exiting the loop was `addr == final_addr`. Since now is possible that `addr 
 > final_add`, which happens when the last chunk received from the prover inputs  is padded with 0s, the final positon to be returned is `final_addr`

PR #1467, opened 2024-01-15
Fix simulation for jumpdest analysis

4l0n50:
Fix an error in the user's code simulation used by jumpdest analysis.


Nashtare (approved):
LGTM, thanks for fixing it!

PR #1469, opened 2024-01-15
Remove some CPU cycles

Nashtare:
Removes some CPU cycles:

- all calls to `%mload_packing` can be directly replaced to `MLOAD_32BYTES`, no need for hook + jump
- Similarly for `%mstore_unpacking`, we can sometimes just call directly the variant if the size is known which skips going through the jumptable
- tweaked some very large stack operation in `hash.asm`, to reduce the needless overhead

Removes 1050 CPU cycles on ERC20 (much more with large MPTs)

### DIFF
--- a/evm/src/cpu/kernel/asm/core/precompiles/bn_add.asm
+++ b/evm/src/cpu/kernel/asm/core/precompiles/bn_add.asm
@@ -56,7 +56,6 @@ bn_add_return:
     %stack (parent_ctx, x, y) -> (parent_ctx, @SEGMENT_RETURNDATA, x, parent_ctx, y)
     %build_address_no_offset
     MSTORE_32BYTES_32
-bn_add_contd6:
     POP
     %stack (parent_ctx, y) -> (parent_ctx, @SEGMENT_RETURNDATA, 32, y)
     %build_address
--- a/evm/src/cpu/kernel/asm/core/precompiles/bn_add.asm
+++ b/evm/src/cpu/kernel/asm/core/precompiles/bn_add.asm
@@ -14,32 +14,32 @@ global precompile_bn_add:
 
     %charge_gas_const(@BN_ADD_GAS)
 
-    // Load x0, y0, x1, y1 from the call data using `mload_packing`.
+    // Load x0, y0, x1, y1 from the call data using `MLOAD_32BYTES`.
     PUSH bn_add_return
     // stack: bn_add_return, kexit_info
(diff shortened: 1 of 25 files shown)


evm/src/cpu/kernel/asm/core/precompiles/bn_add.asm
```diff
@@ -53,11 +53,12 @@ bn_add_return:
     // Store the result (x, y) to the parent's return data using `mstore_unpacking`.
     %mstore_parent_context_metadata(@CTX_METADATA_RETURNDATA_SIZE, 64)
     %mload_context_metadata(@CTX_METADATA_PARENT_CONTEXT)
-    %stack (parent_ctx, x, y) -> (parent_ctx, @SEGMENT_RETURNDATA, x, 32, bn_add_contd6, parent_ctx, y)
+    %stack (parent_ctx, x, y) -> (parent_ctx, @SEGMENT_RETURNDATA, x, parent_ctx, y)
     %build_address_no_offset
-    %jump(mstore_unpacking)
+    MSTORE_32BYTES_32
 bn_add_contd6:
```
LindaGuiga, line 59: nit: same as above: I think it can be removed

evm/src/cpu/kernel/asm/util/keccak.asm
```diff
@@ -55,12 +55,10 @@ sys_keccak256_empty:
...
     // @SEGMENT_KERNEL_GENERAL[32..64].
-    %stack (a) -> (@SEGMENT_KERNEL_GENERAL, a, 32, %%after_mstore_a)
-    %jump(mstore_unpacking)
-%%after_mstore_a:
-    %stack (addr, b) -> (addr, b, 32, %%after_mstore_b)
-    %jump(mstore_unpacking)
-%%after_mstore_b:
+    %stack (a) -> (@SEGMENT_KERNEL_GENERAL, a)
+    MSTORE_32BYTES_32
+    %stack (addr, b) -> (addr, b)
```
LindaGuiga, line 60:
```suggestion
    // stack: addr, b
```

LindaGuiga (approved):
LGTM, just a couple of nits, thank you!

evm/src/cpu/kernel/asm/util/keccak.asm
Nashtare: That indeed makes more sense 🙃

PR #1471, opened 2024-01-16
Fix interpreter jumps

LindaGuiga:
In the interpreter, `run_jump` and `run_jumpi` were not throwing the correct exceptions. This PR changes the two functions so that they correspond to the logic in `operation.rs`. 

Note that this also removes an unnecessary `assert_ne!` which suggests to calm `KECCAK256` if the segment is `MainMemory` in `KECCAK_GENERAL`; since the `sys_keccak256` calls `KECCAK_GENERAL` and the interpreter now relies directly on the kernel syscalls, the assert is no longer useful (and causes unnecessary failures).

### DIFF
--- a/evm/src/cpu/kernel/interpreter.rs
+++ b/evm/src/cpu/kernel/interpreter.rs
@@ -57,7 +57,6 @@ impl MemoryState {
 }
 
 pub(crate) struct Interpreter<'a> {
-    jumpdests: Vec<usize>,
     pub(crate) generation_state: GenerationState<F>,
     prover_inputs_map: &'a HashMap<usize, ProverInputFn>,
     pub(crate) halt_offsets: Vec<usize>,
@@ -173,7 +172,6 @@ impl<'a> Interpreter<'a> {
         prover_inputs: &'a HashMap<usize, ProverInputFn>,
     ) -> Self {
         let mut result = Self {
-            jumpdests: find_jumpdests(code),
             generation_state: GenerationState::new(GenerationInputs::default(), code)
                 .expect("Default inputs are known-good"),
             prover_inputs_map: prover_inputs,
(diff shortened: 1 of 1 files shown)


evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1129,14 +1165,20 @@ impl<'a> Interpreter<'a> {
...
-    fn jump_to(&mut self, offset: usize) -> anyhow::Result<(), ProgramError> {
-        // The JUMPDEST rule is not enforced in kernel mode.
-        if !self.is_kernel() && self.jumpdests.binary_search(&offset).is_err() {
-            return Err(ProgramError::InvalidJumpDestination);
-        }
-
+    fn jump_to(&mut self, offset: usize, is_jumpi: bool) -> anyhow::Result<(), ProgramError> {
         self.generation_state.registers.program_counter = offset;
 
+        if offset == KERNEL.global_labels["observe_new_address"] {
```
LindaGuiga, line 1171: I believe this is only for debugging purposes, but I thought it might also be useful to store them on the interpreter side.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1129,14 +1165,20 @@ impl<'a> Interpreter<'a> {
         Ok(())
     }
 
-    fn jump_to(&mut self, offset: usize) -> anyhow::Result<(), ProgramError> {
-        // The JUMPDEST rule is not enforced in kernel mode.
-        if !self.is_kernel() && self.jumpdests.binary_search(&offset).is_err() {
```
4l0n50, line 1134: It seems that `self.jumpdest` isn't needed anymore?

Nashtare: Same for `find_jumpdests()`

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1107,9 +1126,26 @@ impl<'a> Interpreter<'a> {
...
+            > x.low_u32() as usize
+        {
+            self.generation_state.memory.get(MemoryAddress {
+                context: self.context(),
+                segment: Segment::JumpdestBits.unscale(),
+                virt: x.low_u32() as usize,
+            })
+        } else {
+            0.into()
+        };
```
Nashtare, line 1144: As it's used both for jump and jumpi, it may be cleaner to have it into its own `set_jumpdest_bit(x)` function.

Nashtare (approved):
Thank! LGTM minus some nits

PR #1474, opened 2024-01-17
Fix bugs in jumpdest analysis

4l0n50:
This PR incorporates the following changes:
- Fix `test_jumpdest_analysis` which was previously not meaningful, as it only verified that `jumpdest_analysis` never aborted though that can't happen. Now, it verifies the correctness of `SEGMENT_JUMPDEST_BITS` and of  `jumpdest_table`.
- Introduce `test_packed_verification`.

The PR addresses the bugs uncovered during these tests, primarily:
- Utilizing the current context instead of the one in the stack in `PROVER_INPUT(jumpdest_table::)`.
- Correcting the prefix computation in the packed verification of `write_table_if_jumpdest`.


evm/src/cpu/kernel/tests/core/jumpdest_analysis.rs
```diff
@@ -37,10 +39,82 @@ fn test_jumpdest_analysis() -> Result<()> {
...
+                .iter()
+                .enumerate()
+                .filter(|&(_, &x)| x)
+                .map(|(i, _)| i),
+        ),
+    )]));
+
+    assert_eq!(
+        interpreter.generation_state.jumpdest_table,
+        Some(HashMap::from([(3, vec![0, 1, 0, 5, 0, 7])]))
```
dvdplm, line 55: A comment to justify the `vec![0, 1, 0, 5, 0, 7]`? Also: can it be a constant perhaps?

evm/src/cpu/kernel/asm/core/jumpdest_analysis.asm
```diff
@@ -282,8 +284,8 @@ global write_table_if_jumpdest:
     // check the remaining path
     %jump(verify_path_and_write_jumpdest_table)
 return:
-    // stack: proof_prefix_addr, jumpdest, ctx, retdest
-    %pop3
+    // stack: packed_opcodes, proof_prefix_addr, ctx, jumpdest, retdest
```
Nashtare, line 287: nit: the stack order does not match what's in the conditional jump sending us to `return`, i.e.
`jumpdest, ctx, proof_prefix_addr, retdest`.
It doesn't matter as we pop them anyway, but may be better to swap the elements for consistency and clarity.

evm/src/cpu/kernel/asm/core/jumpdest_analysis.asm
```diff
@@ -164,106 +164,108 @@ global write_table_if_jumpdest:
...
     AND
-    SWAP1
+    // stack: (in_range'|X⁷)³², (~has_prefix'|X⁷)³², packed_opcodes, proof_prefix_addr, ctx, jumpdest, retdest
 
-    // Compute in_range' = in_range AND
-    //   - (0xFF|X⁷)³² for bytes in odd positions
-    //   - (has_prefix => is_0_at_8 |X⁷)³² on the rest
+    // Compute in_range' as
+    //   - in_range' = in_range,                                    for odd positions
+    //   - in_rage' = in_range AND (has_prefix => is_0_at_8 |X⁷)³², for the rest
```
Nashtare, line 262: nit: 
```suggestion
    //   - in_range' = in_range AND (has_prefix => is_0_at_8 |X⁷)³², for the rest
```

Nashtare (approved):
Thank you, this looks good minor a few nit comments.

evm/src/cpu/kernel/tests/core/jumpdest_analysis.rs
Nashtare: I also think some description may be beneficial (though it also applies to other tests in the suite that are a bit cryptic).

evm/src/generation/prover_input.rs
```diff
@@ -251,38 +251,39 @@ impl<F: Field> GenerationState<F> {
 
     /// Returns the next used jump address.
     fn run_next_jumpdest_table_address(&mut self) -> Result<U256, ProgramError> {
-        let context = self.registers.context;
-        let code_len = u256_to_usize(self.get_code_len()?.into());
+        let context = u256_to_usize(stack_peek(self, 0)? >> CONTEXT_SCALING_FACTOR)?;
```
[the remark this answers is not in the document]
4l0n50: The test `test_jumpdest_analysis` also calls `jumpdest_analysis` with context 3 in the stack. Without this change it fails because the current context is 0 instead.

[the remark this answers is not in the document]
Nashtare: Ah ok it's because the test has a special scenario, but you cannot reach `jumpdest_analysis` (outside of this test) without calling `enter_new_ctx`. But yeah, harmless change anyway

evm/src/cpu/kernel/asm/core/jumpdest_analysis.asm
```diff
@@ -277,15 +277,19 @@ global write_table_if_jumpdest:
     // pos 0102030405060708091011121314151617181920212223242526272829303132
     PUSH 0x8080808080808080808080808080808080808080808080808080808080808080
     AND
-    %jump_neq_const(0x8080808080808080808080808080808080808080808080808080808080808080, return)
+    %jump_neq_const(0x8080808080808080808080808080808080808080808080808080808080808080, return_pop_opcode)
     POP
```
Nashtare, line 281: Shouldn't we remove the POP below then?

PR #1476, opened 2024-01-18
Speed-up `bn254` pairing operation

Nashtare:
Speeds-up pairing computation, mostly around refactoring how address handling for reads and writes is done, following #1426.

For a regular pairing operation, it saves 11.35% of the CPU cycles count (about 165k), with a reduction of around 550k Memory rows.

Ran the Ethereum test suite on related tests:  131/131 tests ✅


Nashtare:
**Note to reviewers**: I'd recommend hiding whitespaces when looking at the changes.

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
```diff
@@ -145,253 +164,259 @@
...
+    %create_bn254_pairing_address
+    DUP1
     %add_const(5)
-    // stack:                     ind5
-    %mload_bn254_pairing
-    // stack:                       x5
-    PUSH $ptr  
+    // stack:                 addr5, addr
+    MLOAD_GENERAL
+    // stack:                    x5, addr
```
LindaGuiga, line 234: Why not follow the same approach as for the macro above (`load_fp254_6`): load x1 to x4 using only DUPs (leaving addr at the back of the stack), then load x5, then SWAP and load x0?

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
```diff
@@ -145,253 +164,259 @@
...
-    // stack:                           x5
-    PUSH $ptr  
-    %add_const(5)
-    // stack:                     ind5, x5
-    %mstore_bn254_pairing
+    %create_bn254_pairing_address
+    // stack: addr, x0, x1, x2, x3, x4, x5
+    DUP1 DUP3
+    MSTORE_GENERAL
+    // stack: addr, x0, x1, x2, x3, x4, x5
```
LindaGuiga, line 361: Nit: maybe write addr1, addr2 etc in what follows, since it seems like it's been done for the other macros

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
```diff
@@ -145,253 +164,259 @@
...
+    %increment
+    DUP1 DUP6
+    MSTORE_GENERAL
+    // stack: addr, x0, x1, x2, x3, x4, x5
+    %increment
+    DUP1 DUP7
+    MSTORE_GENERAL
+    // stack: addr, x0, x1, x2, x3, x4, x5
+    %increment
+    %stack (addr, x0, x1, x2, x3, x4, x5) -> (x5, addr)
```
LindaGuiga, line 379: Nit: Using the same technique as below in `store_fp254_6_sh` (DUP, SWAP, MSTORE) would help remove these POPs/SWAPs, right?

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
```diff
@@ -863,265 +888,269 @@
 %endmacro
 
 %macro load_fp254_12
-    // stack:                                                          ptr
-    DUP1  
+    // stack:                                                          addr
```
LindaGuiga, line 891: Nit
```suggestion
    // stack:                                                          ptr
```

LindaGuiga (approved):
LGTM, just a couple of nits, thank you!

evm/src/cpu/kernel/asm/curve/bn254/curve_arithmetic/miller_loop.asm
```diff
@@ -284,10 +292,34 @@ after_add:
...
+    // stack: out
+    %create_bn254_pairing_address
+    // stack: addr
+    PUSH 0x10000000000000000000000
+    SWAP1
+    // stack: addr, 0x10000000000000000000000
+    MSTORE_32BYTES_12
+    POP
+    // stack:
+%endmacro
```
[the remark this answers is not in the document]
Nashtare: What do you mean?

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
Nashtare: This is post MSTORE_GENERAL, so we effectively always have `addr` on top of the stack (there is no intermediary stack description when the `addr_i` is present).

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
Nashtare: The other macro being cheaper, I just replaced it with 
```asm
%macro load_fp254_6(ptr)
    // stack:
    PUSH $ptr
    %load_fp254_6
    // stack: x0, x1, x2, x3, x4, x5
%endmacro
```
instead

Nashtare: The starting ptr isn't at the back of the stack but at the start, hence the difference with `load_fp254_6`

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
Nashtare: I've actually replaced it with a call to the upper macro, as it's cheaper:
```asm
%macro store_fp254_6(ptr)
    // stack:       x0, x1, x2, x3, x4, x5
    PUSH $ptr
    %store_fp254_6
    // stack:
%endmacro
```

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
LindaGuiga: I'm sorry, I'm not sure I understand: I meant that at this point, the address is at the back so for the rest of the macro, instead of having a SWAP and DUP every time, why not use the trick above?

evm/src/cpu/kernel/asm/curve/bn254/curve_arithmetic/miller_loop.asm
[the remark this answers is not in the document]
Nashtare: Ah oki, yes will add it

evm/src/cpu/kernel/asm/curve/bn254/field_arithmetic/util.asm
Nashtare: Hmm I'm confused, the address is on top of the stack here

LindaGuiga: The stack is: `x5, addr`, right?

PR #1477, opened 2024-01-18
Improve `blake2f` call

Nashtare:
Speed-up by around 15% the code portion of `blake2_f`.


evm/src/cpu/kernel/asm/hash/blake2/blake2_f.asm
```diff
@@ -55,14 +55,22 @@ global blake2_f:
     %rep 8
         SWAP1
         DUP2
-        %mstore_current_general
+        %swap_mstore
```
hratoanina, line 58: We can use a normal MSTORE:
```
DUP1
SWAP2
MSTORE_GENERAL
```

evm/src/cpu/kernel/asm/hash/blake2/blake2_f.asm
```diff
@@ -55,14 +55,22 @@ global blake2_f:
...
         %increment
     %endrep
     // stack: start + 8, rounds, t0, t1, flag, retdest
 
+    // Retrieve the offset in `start + 8`
```
hratoanina, line 63: Any reason why you retrieve the offset instead of keeping it in address form and replacing the `%mstore_current_general` l.82 with a `%swap_mstore`?

hratoanina (approved):
No issue spotted, but I see a lot of calls to `%swap_mstore` shortly after stack manipulation. Maybe modifying the stack so you don't need the extra swap is cheaper, but I wouldn't know without testing.
Thanks!

evm/src/cpu/kernel/asm/hash/blake2/blake2_f.asm
Nashtare: Not sure why actually, I changed it as you suggested!

Nashtare:
Updating the stack macro calls to get rid of the `%swap_mstore`, + the other suggestions bumped the reduction from 15.9% to 16.2%.

PR #1480, opened 2024-01-19
Improve SHA2 precompile

Nashtare:
Reduces the number of CPU cycles for SHA2 by 7.57% on average.

### DIFF
--- a/evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
+++ b/evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
@@ -8,10 +8,10 @@
     %build_address
 %endmacro
 
-// Precodition: stack contains address of one message block, followed by output address
+// Precondition: stack contains address of one message block, followed by output address
 // Postcondition: 256 bytes starting at given output address contain the 64 32-bit chunks
 //                of message schedule (in four-byte increments)
-global gen_message_schedule_from_block:
+gen_message_schedule_from_block:
     // stack: block_addr, output_addr, retdest
     DUP1
     // stack: block_addr, block_addr, output_addr, retdest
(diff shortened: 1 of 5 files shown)


evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
```diff
@@ -3,32 +3,35 @@
...
+    PUSH @SEGMENT_KERNEL_GENERAL
+    GET_CONTEXT
+    %build_address
 %endmacro
 
 // Precodition: stack contains address of one message block, followed by output address
 // Postcondition: 256 bytes starting at given output address contain the 64 32-bit chunks
 //                of message schedule (in four-byte increments)
-gen_message_schedule_from_block:
+global gen_message_schedule_from_block:
```
LindaGuiga, line 14: Are these added `global` labels necessary? (This one and the ones below, in the same file)

evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
```diff
@@ -3,32 +3,35 @@
...
     // stack: block_addr + 32, block[0], output_addr, retdest
-    %mload_current_general_u256
+    %mload_u256
     // stack: block[1], block[0], output_addr, retdest
     SWAP2
     // stack: output_addr, block[0], block[1], retdest
     %add_const(28)
     PUSH 8
     // stack: counter=8, output_addr + 28, block[0], block[1], retdest
     %jump(gen_message_schedule_from_block_0_loop)
```
LindaGuiga, line 33: Can't this also be removed? (since the label we jump to is in the next row)

evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
```diff
@@ -3,32 +3,35 @@
     // stack: num_blocks
     %mul_const(64)
     %add_const(2)
+    PUSH @SEGMENT_KERNEL_GENERAL
+    GET_CONTEXT
+    %build_address
 %endmacro
 
 // Precodition: stack contains address of one message block, followed by output address
```
LindaGuiga, line 11: Nit, not exactly part of the PR but since you corrected the typo in main.asm:
```suggestion
// Precondition: stack contains address of one message block, followed by output address
```

LindaGuiga (approved):
LGTM, thanks!

evm/src/cpu/kernel/asm/hash/sha2/message_schedule.asm
Nashtare: They were necessary for debugging when I had messed something up 🫣 will revert!

Nashtare: And I removed one too many global label 🤦🏼 Last commit adds it back

PR #1481, opened 2024-01-21
Use web-time crate

vivekvpandya:
Use web-time crate (https://github.com/daxpedda/web-time) instead of std version. It works for wasm32 targets too. If target support time in std then it's just an alias to that.

### DIFF
--- a/plonky2/Cargo.toml
+++ b/plonky2/Cargo.toml
@@ -15,7 +15,7 @@ default = ["gate_testing", "parallel", "rand_chacha", "std", "timing"]
 gate_testing = []
 parallel = ["hashbrown/rayon", "plonky2_maybe_rayon/parallel"]
 std = ["anyhow/std", "rand/std", "itertools/use_std"]
-timing = ["std"]
+timing = ["std", "dep:web-time"]
 
 [dependencies]
 ahash = { version = "0.8.3", default-features = false, features = ["compile-time-rng"] } # NOTE: Be sure to keep this version the same as the dependency in `hashbrown`.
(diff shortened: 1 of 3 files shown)


dlubarov:
Did you try with the `timing` feature off, or is that not a good option as you want the timing breakdown?

plonky2/Cargo.toml
```diff
@@ -34,6 +34,7 @@ serde = { version = "1.0", default-features = false, features = ["derive", "rc"]
 serde_json = "1.0"
 static_assertions = { version = "1.1.0", default-features = false }
 unroll = { version = "0.1.5", default-features = false }
+instant = "0.1"
```
dlubarov, line 37: I wonder if this breaks `no_std` builds? If so, could make it an optional dependency, with `timing` depending on `dep:instant`

vivekvpandya:
> Did you try with the `timing` feature off, or is that not a good option as you want the timing breakdown?

I tried with `default-features=false, features=["std"]` but it did not worked.

but in any case I think this PR is harmless.

Nashtare (commented):
> I tried with default-features=false, features=["std"] but it did not worked.

Yeah switching `std` feature on would break with most WASM targets, as you can't make OS assumptions.

OTOH, I'm not super in favor of bringing into the dependencies a crate that's clearly abandoned (last version was 3y ago, issues and PR opens about problems encountered and not addressed).

Nashtare:
If we're to add full timing support for wasm targets, then I'd suggest using [web-time](https://crates.io/crates/web-time), as this issue properly suggests: https://github.com/sebcrozet/instant/issues/52.

vivekvpandya:
@Nashtare anything more needed to be done for this PR? if not could you please merge it.

Nashtare (approved):

PR #1489, opened 2024-01-29
Simulate jumpdest data with the interpreter

4l0n50:
The jumpdest analysis data is now generated only using the intepreter


evm/src/cpu/kernel/tests/blake2_f.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::Result;
+use plonky2::field::goldilocks_field::GoldilocksField as F;
```
Nashtare, line 2: I'd rather see this everywhere (as is commonly done to shorten the associated type name size), than the `Interpreter<GoldilocksField>` there are now everywhere among test files.

evm/src/cpu/kernel/tests/core/jumpdest_analysis.rs
```diff
@@ -50,24 +47,30 @@ fn test_jumpdest_analysis() -> Result<()> {
         ),
     )]));
 
+    // The set_jumpdest_analysis_inputs method is never used
```
Nashtare, line 50:
```suggestion
    // The `set_jumpdest_analysis_inputs` method is never used.
```

evm/src/generation/prover_input.rs
```diff
@@ -319,6 +317,7 @@ impl<F: Field> GenerationState<F> {
         self.memory = memory;
 
         // Find proofs for all contexts
+        log::debug!("jumpdest_table = {:?}", jumpdest_table);
```
Nashtare, line 320: nit: I'd move it to the `trace` level.

evm/src/memory/segments.rs
```diff
@@ -1,5 +1,5 @@
 use ethereum_types::U256;
-use num::traits::AsPrimitive;
+// use num::traits::AsPrimitive;
```
Nashtare, line 2:
```suggestion
```

evm/src/witness/transition.rs
```diff
@@ -145,7 +145,7 @@ pub(crate) fn decode(registers: RegistersState, opcode: u8) -> Result<Operation,
         (0xf6, true) => Ok(Operation::GetContext),
         (0xf7, true) => Ok(Operation::SetContext),
         (0xf8, true) => Ok(Operation::Mload32Bytes),
-        (0xf9, true) => Ok(Operation::ExitKernel),
+        (0xf9, _) => Ok(Operation::ExitKernel),
```
Nashtare, line 148: Hmm, this is insufficient, without additional changes, to make sure the opcode isn't called outside of context 0. Why has it been updated?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -682,6 +750,18 @@ impl<'a> Interpreter<'a> {
     }
 
     fn run_opcode(&mut self) -> Result<(), ProgramError> {
+        // Jumpdest analysis is performed natively by the interpreter and not
+        // using the non-deterministic assemly code
```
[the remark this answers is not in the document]
4l0n50: Is not skipping but doing it in rust (natively?)

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -120,17 +122,38 @@ pub(crate) fn run_interpreter_with_memory(
...
+    } else {
+        let halt_pc = KERNEL.global_labels[final_label];
+        let initial_context = state.registers.context;
+        let mut interpreter =
+            Interpreter::new_with_state_and_halt_condition(state, halt_pc, initial_context);
+
+        log::debug!("Simulating CPU for jumpdest analysis.");
+
+        interpreter.run();
+        log::debug!("Simulated CPU for jumpdest analysis halted");
```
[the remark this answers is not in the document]
4l0n50: Right now the interpreter doesn't know at which cycle it is.  But I agree, the number of cycles may hint some problems in the simulation without having to print all the simulation logs.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -893,16 +903,25 @@ impl<'a, F: Field> Interpreter<'a, F> {
...
+                let mut stack = self.stack();
+                stack.reverse();
+                stack
+            });
         }
 
-        let op = decode(self.generation_state.registers, opcode)
-            // We default to prover inputs, as those are kernel-only instructions that charge nothing.
-            .unwrap_or(Operation::ProverInput);
-        self.generation_state.registers.gas_used += gas_to_charge(op);
```
Nashtare, line 823: Why has this been moved prior to the opcode execution? (this is different from the actual CPU)

4l0n50: Because if you have an invalid exit kernel you will still execute `run_exit_kernel` and set the kernel flag to `false`. Then when you call decode it doesn't return an error.

4l0n50: Maybe this is not the best solution, but the problem was that the interpreter was not erroring in invalid EXIT_KERNEL while it should

4l0n50: It was with the `undefinedOpcodeFirstByte_d0g0v0_Shanghai` test I think. And the problem is that `run_exit_kernel` doesn't check if it is or not in kernel mode.
 I think the solution should be that the interpreter's `run_opcode` fn does the match with an already decoded opcode like `perfom_op`. I wanted to minimize the changes in the interpreter though

Nashtare: Yeah this may be out of scope for this PR, but useful to keep track of such discrepancies nonetheless. Thanks for the info!

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1083,7 +1165,7 @@ impl<'a> Interpreter<'a> {
         self.push(syscall_info)
     }
 
-    fn set_jumpdest_bit(&mut self, x: U256) -> U256 {
+    fn get_jumpdest_bit(&self, x: U256) -> U256 {
```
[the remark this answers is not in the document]
dvdplm: That's indeed a bit weird. Hmm.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -56,10 +56,12 @@ impl MemoryState {
     }
 }
 
-pub(crate) struct Interpreter<'a> {
+pub(crate) struct Interpreter<'a, F: Field> {
```
dvdplm, line 59: This is what causes all the noise in this PR yes? Do you think having `new_with_state_and_halt_condition()` be a freestanding function instead could work so you don't have to add the generic here? So you'd have `fn new_interpreter_with_state<F: Field>(… … )` outside the `impl` for `Interpreter`?

evm/src/generation/prover_input.rs
```diff
@@ -411,30 +407,30 @@ fn get_proofs_and_jumpdests(
...
         (vec![], 0),
-        |(mut proofs, acc), (pos, opcode)| {
-            let has_prefix = if let Some(prefix_start) = pos.checked_sub(32) {
-                code[prefix_start..pos]
-                    .iter()
-                    .enumerate()
-                    .fold(true, |acc, (prefix_pos, &byte)| {
+        |(mut proofs, last_proof), (addr, opcode)| {
+            let has_prefix = if let Some(prefix_start) = addr.checked_sub(32) {
+                code[prefix_start..addr].iter().enumerate().fold(
```
dvdplm, line 412: In the case that the predicate returns `false`, then `acc` will be `false` in the next iteration, but once it's set to `false`, it can't ever flip back to `true` I think – is that what you want here? If that is correct, then why not use [std::iterator::any](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) here to short-circuit the iteration at the first `false`?

evm/src/cpu/kernel/tests/core/jumpdest_analysis.rs
```diff
@@ -28,16 +29,11 @@ fn test_jumpdest_analysis() -> Result<()> {
         add,
         jumpdest,
     ];
+    let code_len = code.len();
```
dvdplm, line 32: Is this needed? Just used once yeah?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1023,6 +1123,7 @@ impl<'a> Interpreter<'a> {
                     .byte(0)
             })
             .collect::<Vec<_>>();
+        #[cfg(test)]
         println!("Hashing {:?}", &bytes);
```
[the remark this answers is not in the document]
Nashtare: The thing is we do not want to print anything during the CPU simulation, so even setting it to the lowest logging level wouldn't be sufficient.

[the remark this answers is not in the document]
dvdplm: Why is that? For runtime-set levels (the default), it's a single `if` check ([here](https://github.com/rust-lang/log/blob/master/src/macros.rs#L35)). If that's too expensive then we can perhaps use the static level that removes all unwanted logging calls (see [this](https://docs.rs/log/0.4.17/log/#compile-time-filters), and I think `tracing` has a similar thing).
If the problem is the actual printing to `STDOUT`, then why not just reconfigure the logging to write to a file?

evm/src/generation/prover_input.rs
4l0n50: > but once it's set to false, it can't ever flip back to true

Yes, because you're computing a big and over 32 boolean vars. 

>  is that what you want here? If that is correct, then why not use [std::iterator::any](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) 

Sure, that looks much better. Though you would need `iterator::all` 

Thanks!

evm/src/cpu/kernel/interpreter.rs
4l0n50: But you would need to create a new `Intepreter<F>` within `new_interpreteter_with_state`. The problem is that right now `Interpreter` has a field representing the state of type `GenerationState<GoldilocksField>`, but we would like to run it in a context where the state is only `GenerationState<F>` for `F: Field`.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -396,9 +448,29 @@ impl<'a> Interpreter<'a> {
...
             let pc = self.generation_state.registers.program_counter;
-            if self.is_kernel() && self.halt_offsets.contains(&pc) {
+
+            if let Some(halt_context) = self.halt_context {
+                if self.is_kernel()
+                    && self.halt_offsets.contains(&pc)
+                    && halt_context == self.generation_state.registers.context
+                {
+                    log::debug!(
+                        "se acabo el show cuando el contepsto era {:?}",
```
dvdplm, line 458: I think you have a spelling mistake here: `contesto` yeah?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -396,9 +448,29 @@ impl<'a> Interpreter<'a> {
...
+                    return Ok(());
+                }
+            } else if self.halt_offsets.contains(&pc) {
                 return Ok(());
-            };
+            }
+
+            if !self.running {
+                log::debug!(
+                    "se pago con contepsto {:?}",
```
dvdplm, line 470: Same?

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -120,17 +123,46 @@ pub(crate) fn run_interpreter_with_memory(
...
+        None
+    } else {
+        let halt_pc = KERNEL.global_labels[final_label];
+        let initial_context = state.registers.context;
+        let mut interpreter =
+            Interpreter::new_with_state_and_halt_condition(state, halt_pc, initial_context);
+
+        log::debug!("Simulating CPU for jumpdest analysis.");
+
+        interpreter.run().unwrap();
```
dvdplm, line 153: I think this method should change to return `Result<Option<HashMap<usize, Vec<usize>>>, SomeError>`. Not ok to panic here I think.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -120,17 +123,46 @@ pub(crate) fn run_interpreter_with_memory(
...
     Ok(interpreter)
 }
 
+/// Simulates the CPU execution from `state` until the program counter reaches `final_label`  
+/// in the current context.
+pub(crate) fn simulate_cpu_and_get_user_jumps<F: Field>(
+    final_label: &str,
+    state: &GenerationState<F>,
+) -> Option<HashMap<usize, Vec<usize>>> {
+    if state.jumpdest_table.is_some() {
```
dvdplm, line 143: Using `match` is more idiomatic Rust.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -682,12 +749,42 @@ impl<'a> Interpreter<'a> {
...
+                "User instruction {:?}, stack = {:?}, ctx = {}",
+                op,
+                {
+                    let mut stack = self.stack();
+                    stack.reverse();
+                    stack
+                },
+                self.generation_state.registers.context
+            );
+        }
```
dvdplm, line 786: Not a fan of this, as discussed elsewhere with @Nashtare I think this should be `log::trace!`

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1078,56 +1172,95 @@ impl<'a> Interpreter<'a> {
             + (U256::from(self.generation_state.registers.gas_used) << 192);
         self.generation_state.registers.program_counter = new_program_counter;
 
+        if opcode == 244 {
+            println!(
```
dvdplm, line 1176: Same. Use the logger.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1078,56 +1172,95 @@ impl<'a> Interpreter<'a> {
...
-    fn run_jump(&mut self) -> anyhow::Result<(), ProgramError> {
-        let x = self.pop()?;
 
-        let jumpdest_bit = self.set_jumpdest_bit(x);
+    pub(crate) fn get_jumpdest_bits(&self, context: usize) -> Vec<bool> {
+        self.generation_state.memory.contexts[context].segments[Segment::JumpdestBits.unscale()]
+            .content
+            .iter()
+            .map(|x| x.bit(0))
+            .collect()
```
dvdplm, line 1212: Is this a hot function? I _think_ there are tricks to make this very fast and I do not think `bit()` from `uint` is optimal.

evm/src/cpu/kernel/interpreter.rs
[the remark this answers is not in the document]
Nashtare: Writing to file can be expensive, especially if we redirect all printed outputs during simulations.

evm/src/cpu/kernel/interpreter.rs
Nashtare: Yeah let's keep it to the universal language here 😆

evm/src/cpu/kernel/interpreter.rs
Nashtare: Yeah agree, let's gently error during simulations, although I think there may be other unwraps on the interpreter side

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -120,17 +123,46 @@ pub(crate) fn run_interpreter_with_memory(
...
+        let halt_pc = KERNEL.global_labels[final_label];
+        let initial_context = state.registers.context;
+        let mut interpreter =
+            Interpreter::new_with_state_and_halt_condition(state, halt_pc, initial_context);
+
+        log::debug!("Simulating CPU for jumpdest analysis.");
+
+        interpreter.run().unwrap();
+
+        log::debug!("jdt = {:?}", interpreter.jumpdest_table);
```
Nashtare, line 155: I guess this can either be removed or put one level lower, to the `trace` level?

Nashtare: additionally let's not use abbreviations
```suggestion
        log::debug!("Jumpdest table = {:?}", interpreter.jumpdest_table);
```

evm/src/cpu/kernel/interpreter.rs
4l0n50: ops!

evm/src/cpu/kernel/interpreter.rs
4l0n50: Yes, I was only doing this for debugging purposes

4l0n50: Actually, here I need to always return `interpreter.jumpdest_table`, even when the simulation fails.  I guess the best option is to ignore any error in `run()`

evm/src/cpu/kernel/interpreter.rs
4l0n50: Ok, but should I change it in this PR. I'm not sure if the interpreter is printing debug info in other places

evm/src/cpu/kernel/interpreter.rs
4l0n50: This are not necessary. I just forgot to remove them.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -423,13 +485,16 @@ impl<'a> Interpreter<'a> {
                 }
             }?;
         }
-        println!("Opcode count:");
-        for i in 0..0x100 {
-            if self.opcode_count[i] > 0 {
-                println!("{}: {}", get_mnemonic(i as u8), self.opcode_count[i])
+        #[cfg(test)]
```
Nashtare, line 488: What we can do, similarly to #1503, is to use `#[cfg(debug_assertions)]` instead. This will not change anything when we run tests in `debug` mode, and can still display everything in release mode iif we enable debug assertions, so that if we want to print CPU simulation for some reason (during debugging), just using an environment variable suffices instead of changing manually the code.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -682,12 +739,42 @@ impl<'a> Interpreter<'a> {
...
             .code()
             .get(self.generation_state.registers.program_counter)
             .byte(0);
         self.opcode_count[opcode as usize] += 1;
         self.incr(1);
+
+        let op = decode(self.generation_state.registers, opcode)?;
+        self.generation_state.registers.gas_used += gas_to_charge(op);
+
+        #[cfg(test)]
```
Nashtare, line 764: same as above, maybe better to use `#[cfg(debug_assertions)]` in the end

evm/src/cpu/kernel/interpreter.rs
Nashtare: Yeah as this is now part of the prover logs (through CPU simulations), replacing the `println` by `log::trace` sounds better.

evm/src/cpu/kernel/interpreter.rs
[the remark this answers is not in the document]
Nashtare: I'd suggest going for `#[cfg(debug_assertions)]`

evm/src/cpu/kernel/interpreter.rs
Nashtare: contract bytecode can be quite long, so we may want to have this optimized, to some extent. But I don't think that will be _the_  bottleneck anyway

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1083,51 +1167,77 @@ impl<'a> Interpreter<'a> {
...
+    }
+
+    fn run_jump(&mut self) -> anyhow::Result<(), ProgramError> {
+        let offset = self.pop()?;
 
         // Check that the destination is valid.
-        let x: u32 = x
+        let offset: usize = offset
             .try_into()
             .map_err(|_| ProgramError::InvalidJumpDestination)?;
```
Nashtare, line 1215: There's also a `u256_to_usize` method that returns a `ProgramError`, but `try_into` is fine too.

evm/src/cpu/kernel/interpreter.rs
```diff
@@ -1163,9 +1273,11 @@ impl<'a> Interpreter<'a> {
             self.generation_state.observe_contract(tip_h256)?;
         }
 
-        if self.halt_offsets.contains(&offset) {
-            self.running = false;
+        if !self.is_kernel() {
+            self.add_jumpdest_offset(offset);
+            log::debug!("Added {offset} to jdt")
```
Nashtare, line 1278: nit: We're already printing the final jumpdest table I think? So this may be redundant info that we can skip?
If we keep it, then maybe move it to `trace` and expand `jdt` for clarity

Nashtare (approved):
LGTM, modulo the few nitpicky remaining comments.

evm/src/cpu/kernel/interpreter.rs
4l0n50: Yes. I forgot to delete this.

evm/src/cpu/kernel/interpreter.rs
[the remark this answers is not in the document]
4l0n50: I will use `#[cfg(debug_assertions)]`.

evm/src/cpu/kernel/interpreter.rs
4l0n50: I assume `#[cfg(debug_assertions)]` is also ok.

evm/src/cpu/kernel/interpreter.rs
4l0n50: In the end this function is only used for testing that `jumpdest_analysis` correctly write the jumpdest bits.

PR #1493, opened 2024-01-31
Make CTLs more generic

Nashtare:
Make CTLs generic in the number of tables they are working on, instead of being pegged to the `Table` enum with `NUM_TABLES` variants, in preparation for the code migration.

### DIFF
--- a/evm/src/cross_table_lookup.rs
+++ b/evm/src/cross_table_lookup.rs
@@ -719,7 +719,7 @@ pub(crate) fn num_ctl_helper_columns_by_table<F: Field, const N: usize>(
             let sum = group.count();
             if sum > 2 {
                 // We only need helper columns if there are more than 2 columns.
-                num_by_table[table as usize] = ceil_div_usize(sum, constraint_degree - 1);
+                num_by_table[table] = ceil_div_usize(sum, constraint_degree - 1);
             }
         }
 
@@ -756,7 +756,7 @@ pub(crate) fn cross_table_lookup_data<'a, F: RichField, const D: usize, const N:
             );
 
             let mut z_looked = partial_sums(
-                &trace_poly_values[looked_table.table as usize],
+                &trace_poly_values[looked_table.table],
                 &[(&looked_table.columns, &looked_table.filter)],
                 challenge,
                 constraint_degree,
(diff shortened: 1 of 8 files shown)


evm/src/cross_table_lookup.rs
```diff
@@ -423,19 +423,22 @@ impl<F: Field> Filter<F> {
...
+/// An alias for `usize`, to represent the index of a STARK table in a multi-STARK setting.
+pub(crate) type TableIdx = usize;
+
+/// A `table` index with a linear combination of columns and a filter.
+/// `filter` is used to determine the rows to select in `table`.
+/// `columns` represents linear combinations of the columns of `table`.
 #[derive(Clone, Debug)]
 pub(crate) struct TableWithColumns<F: Field> {
-    table: Table,
+    table: TableIdx,
```
4l0n50, line 434: What about `:T where T: Into` , or another ad-hoc trait. That should work for any enum for representing the tables. Though you need to add a type parameter

Nashtare: This would require you to add everywhere calls to `table.into()` which I initially didn't want, or you'd need additional traits (from the top of my head, you'd need `Into<usize> + PartialEq + Eq + Debug` but even with this you'd still need to call conversions to `usize` here and there, so overall I'm not sure this brings much (and the user still needs to implement `From<Foo> for usize`, similarly to the current requirement to implement `Deref`.

4l0n50: Nevermind, I was assuming enums implement Into<usize>. But still, you may want to use whatever you want for your tables (an enum for example) and here you just want to be able to derive an index for each table

Nashtare: This would require you to add everywhere calls to `table.into()` which I initially didn't want, or you'd need additional traits (from the top of my head, you'd need `Into<usize> + PartialEq + Eq + Debug` but even with this you'd still need to call conversions to `usize` here and there, so overall I'm not sure this brings much (and the user still needs to implement `From<Foo> for usize`, similarly to the current requirement to implement `Deref`.

Nashtare: > But still, you may want to use whatever you want for your tables (an enum for example) and here you just want to be able to derive an index for each table

This is still possible with the current approach though (I kept the `Table` enum), you just need to implement `Deref`.

4l0n50: Ok, I see. then it's ok I think. I was just confused with the "hacky" implementation of deref .

evm/src/recursive_verifier.rs
```diff
@@ -232,7 +232,7 @@ where
     let (total_num_helpers, num_ctl_zs, num_helpers_by_ctl) =
         CrossTableLookup::num_ctl_helpers_zs_all(
             cross_table_lookups,
-            table,
+            table as usize,
```
4l0n50, line 235: `*table`?

4l0n50 (approved):
Looks good to me!

PR #1494, opened 2024-02-01
Fix BaseSumGenerator and BaseSplitGenerator Ids

utsavjnn:
`impl_generator_serializer!` matches generator Id to get correct types. You would need different Ids for different bases if you wanna use `impl_generator_serializer` for these generators with different bases.
For example if your circuit uses `BaseSplitGenerator<2>` and `BaseSplitGenerator<4>` something like the snippet below:
```
impl_generator_serializer! {
  CustomGeneratorSerializer
  BaseSplitGenerator<2>,
  BaseSplitGenerator<4>
}
```
would lead to incorrect generation of `ProverOnlyCircuitData` when reconstructed using `ProverOnlyCircuitData::from_bytes(..)`

CC: @ultrainstinct30

### DIFF
--- a/plonky2/src/gadgets/split_base.rs
+++ b/plonky2/src/gadgets/split_base.rs
@@ -1,7 +1,6 @@
 use alloc::string::String;
-use alloc::format;
-use alloc::vec;
 use alloc::vec::Vec;
+use alloc::{format, vec};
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
Good catch, thank you for this!

PR #1495, opened 2024-02-01
Reorganize lookup / ctl modules

Nashtare:
Just some code reorganization, to have the CTL module rely on the lookup one instead of the opposite, in preparation for the code migration.


evm/src/lookup.rs
```diff
@@ -9,16 +13,381 @@ use plonky2::hash::hash_types::RichField;
...
+    pub(crate) fn eval_table(&self, table: &[PolynomialValues<F>], row: usize) -> F {
+        let mut res = self
+            .linear_combination
+            .iter()
+            .map(|&(c, f)| table[c].values[row] * f)
+            .sum::<F>()
+            + self.constant;
+
+        // If we access the next row at the last row, for sanity, we consider the next row's values to be 0.
+        // If CTLs are correctly written, the filter should be 0 in that case anyway.
```
hratoanina, line 313: Nit:
```suggestion
        // If the lookups are correctly written, the filter should be 0 in that case anyway.
```

hratoanina (approved):

PR #1496, opened 2024-02-02
Switch permutation argument for logUp in `starky`

Nashtare:
Replaces the permutation argument in `starky` with logup, in continuation with the code migration of the zkEVM prover.


starky/src/stark.rs
```diff
@@ -147,79 +157,71 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+        let num_lookup_columns = self.num_lookup_helper_columns(config);
+        let num_auxiliary_polys = num_lookup_columns;
+        let auxiliary_oracle = FriOracleInfo {
+            num_polys: num_auxiliary_polys,
+            blinding: false,
+        };
+        let auxiliary_polys_info =
+            FriPolynomialInfo::from_range(AUXILIARY_ORACLE_INDEX, 0..num_auxiliary_polys);
+
+        let ctl_zs_info = FriPolynomialInfo::from_range(
```
LindaGuiga, line 175: nit: Same as above

starky/src/stark.rs
```diff
@@ -147,79 +157,71 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+            AUXILIARY_ORACLE_INDEX,
+            num_lookup_columns..num_lookup_columns,
```
LindaGuiga, line 177: If you keep this to prepare for the CTLs, then this should be:
```suggestion
            num_lookup_columns..num_auxiliary_polys,
```

starky/src/verifier.rs
```diff
@@ -181,24 +196,24 @@ where
...
-        ensure!(permutation_zs.len() == num_zs);
-        ensure!(permutation_zs_next.len() == num_zs);
+        ensure!(auxiliary_polys_cap.height() == cap_height);
+        ensure!(auxiliary_polys.len() == num_auxiliary);
+        ensure!(auxiliary_polys_next.len() == num_auxiliary);
     } else {
-        ensure!(permutation_zs_cap.is_none());
-        ensure!(permutation_zs.is_none());
-        ensure!(permutation_zs_next.is_none());
+        ensure!(auxiliary_polys_cap.is_none());
```
LindaGuiga, line 214: Nit: Aren't these checks (the `else` part) already ensured in `check_lookup_options`?

LindaGuiga (approved):
LGTM, thanks! Just a couple of nits/questions!

starky/src/stark.rs
```diff
@@ -94,49 +98,55 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
-                blinding: false,
-            });
-            polys
-        } else {
-            vec![]
         };
+        let trace_info = FriPolynomialInfo::from_range(TRACE_ORACLE_INDEX, 0..Self::COLUMNS);
+
+        let num_lookup_columns = self.num_lookup_helper_columns(config);
+        let num_auxiliary_polys = num_lookup_columns;
```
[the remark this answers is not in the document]
Nashtare: Yes exactly, I wanted to keep it as "consistent" with what's coming

starky/src/stark.rs
```diff
@@ -94,49 +98,55 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
+        let auxiliary_oracle = FriOracleInfo {
+            num_polys: num_auxiliary_polys,
+            blinding: false,
+        };
+        let auxiliary_polys_info =
+            FriPolynomialInfo::from_range(AUXILIARY_ORACLE_INDEX, 0..num_auxiliary_polys);
+
+        let ctl_zs_info = FriPolynomialInfo::from_range(
```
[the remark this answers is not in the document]
Nashtare: oops, leftover thanks!

starky/src/stark.rs
Nashtare: will remove as it's CTL related, as the other items above. Thanks for noticing!

evm/src/prover.rs
```diff
@@ -616,9 +616,7 @@ where
             );
             // Get the local and next row evaluations for the permutation argument, as well as the associated challenges.
             let lookup_vars = lookup_challenges.map(|challenges| LookupCheckVars {
-                local_values: auxiliary_polys_commitment.get_lde_values_packed(i_start, step)
-                    [..num_lookup_columns]
-                    .to_vec(),
+                local_values: auxiliary_polys_commitment.get_lde_values_packed(i_start, step),
```
dvdplm, line 619: Is this just to save on the allocation?

starky/src/get_challenges.rs
```diff
@@ -33,21 +30,16 @@ fn get_challenges<F, C, S, const D: usize>(
...
-        let tmp = get_n_permutation_challenge_sets(
-            &mut challenger,
-            num_challenges,
-            stark.permutation_batch_size(),
-        );
-        challenger.observe_cap(permutation_zs_cap);
+    let lookup_challenge_set = auxiliary_polys_cap.map(|auxiliary_polys_cap| {
+        let tmp = get_grand_product_challenge_set(&mut challenger, num_challenges);
+        challenger.observe_cap(auxiliary_polys_cap);
         tmp
```
dvdplm, line 43: Can you use something like `inspect` here to get rid of the `tmp` var? Perhaps not, given it's not an iterator. Still. Hmm.

starky/src/get_challenges.rs
```diff
@@ -79,27 +71,21 @@ where
 {
     // TODO: Should be used later in compression?
     #![allow(dead_code)]
```
dvdplm, line 73: Is this still dead code?

starky/src/get_challenges.rs
```diff
@@ -79,27 +71,21 @@ where
 {
     // TODO: Should be used later in compression?
```
dvdplm, line 72: Still relevant?

starky/src/get_challenges.rs
```diff
@@ -153,13 +136,8 @@ where
...
-        let tmp = get_n_permutation_challenge_sets_target(
-            builder,
-            &mut challenger,
-            num_challenges,
-            stark.permutation_batch_size(),
-        );
+    let lookup_challenge_set = auxiliary_polys_cap.map(|permutation_zs_cap| {
+        let tmp = get_grand_product_challenge_set_target(builder, &mut challenger, num_challenges);
         challenger.observe_cap(permutation_zs_cap);
         tmp
```
dvdplm, line 142: This is repeating code. Can we do better?

starky/src/get_challenges.rs
```diff
@@ -186,22 +164,17 @@ where
 }
 
 impl<const D: usize> StarkProofWithPublicInputsTarget<D> {
-    pub(crate) fn get_challenges<
-        F: RichField + Extendable<D>,
-        C: GenericConfig<D, F = F>,
-        S: Stark<F, D>,
-    >(
+    pub(crate) fn get_challenges<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>>(
```
dvdplm, line 167: Lots of generics here, can we use a `where` clause?

starky/src/prover.rs
```diff
@@ -70,63 +69,113 @@ where
     challenger.observe_cap(&trace_cap);
 
     // Permutation arguments.
```
dvdplm, line 71: Comment needs updating now?

starky/src/prover.rs
```diff
@@ -70,63 +69,113 @@ where
...
+    let num_lookup_columns = lookups
+        .iter()
+        .map(|l| l.num_helper_columns(constraint_degree))
+        .sum();
 
-        let permutation_zs_commitment = timed!(
+    let auxiliary_polys_commitment = stark.uses_lookups().then(|| {
+        let lookup_helper_columns = timed!(timing, "compute lookup helper columns", {
+            let challenges = lookup_challenges.as_ref().expect("We do have challenges.");
+            let mut columns = Vec::with_capacity(num_lookup_columns);
```
dvdplm, line 90: So the code on line 82 is only there to give the allocator a size hint? I wonder if that is worth the price of running though the `lookups` one extra time? Is it a small number perhaps? You could just "guesstimate" the size, and do `Vec::with_capacity(5000)` knowing it'll be mostly fine. If `columns` is going to survive this function call – seems like it will – then perhaps you can resize the vec before returning it?

starky/src/prover.rs
```diff
@@ -70,63 +69,113 @@ where
...
-        &stark,
-        &trace_commitment,
-        &permutation_zs_commitment_challenges,
-        public_inputs,
-        alphas,
-        degree_bits,
-        config,
+
+    #[cfg(test)]
+    {
```
dvdplm, line 131: Not a fan of abusing `test` mode for this. If we can, at some point, we should investigate the actual impact of doing `log::debug!` for these kinds of cases. I get why it's here, and it's needed for sure, I just don't think `test` is the way to go here.

evm/src/prover.rs
Nashtare: Yes correct

starky/src/get_challenges.rs
Nashtare: Still unused for now (see `TODO: Deal with the compressed stuff.`)

starky/src/get_challenges.rs
Nashtare: Still unused for now (see `TODO: Deal with the compressed stuff.`)

starky/src/get_challenges.rs
Nashtare: What do you mean this is repeating code?

starky/src/prover.rs
Nashtare: I kind of disagree here. Verbosity level should not over impact runtime (apart from the overhead incurred by extra I/O). This `check_constraints` is used for testing and debugging, but has a serious impact on prover performances, and should never be run outside of testing purposes I think. (adds 45% proving time for erc20 on my machine).

starky/src/get_challenges.rs
dvdplm: Perhaps file a ticket and add the ref here so we don't forget? Up to you.

starky/src/get_challenges.rs
dvdplm: My bad, it looked a lot like the code starting on line #40, but it's actually not the same.

starky/src/prover.rs
dvdplm: What I mean is this: we should check on what the overhead of e.g. `log::debug!("bla bla {}", expensive_calculation_here());` actually is when running with debug logging **off**. I am 99% sure that `expensive_calculation_here()` will not run. If am wrong, and the expensive computation does run, we should use something like the static debug logging filters that both `log` and `tracing` crates provide.

If that turns out to be the wrong path then it'd be better to have a proper feature, e.g. `expensive_debugging`, rather than using `test`.

Nashtare: Other places are affected with similar logic, hence I think it's kind of out of scope of this PR (which just moves stuff around really). But a conditional feature flag may be better indeed.

starky/src/lookup.rs
```diff
@@ -0,0 +1,1006 @@
...
+        constraint_degree,
+    );
+
+    // Add `1/(table+challenge)` to the helper columns.
+    // This is 1/phi_0(x) = 1/(x + t(x)) from the paper.
+    // Here, we don't include m(x) in the numerator, instead multiplying it with this column later.
+    let mut table = lookup.table_column.eval_all_rows(trace_poly_values);
+    for x in table.iter_mut() {
+        *x = challenge + *x;
+    }
```
dvdplm, line 616: Could just do this in a `map()` I think?

starky/src/lookup.rs
```diff
@@ -0,0 +1,1006 @@
...
+                        challenge.combine(evals.iter())
+                    } else {
+                        assert_eq!(f, F::ZERO, "Non-binary filter?");
+                        // Dummy value. Cannot be zero since it will be batch-inverted.
+                        F::ONE
+                    }
+                })
+                .collect::<Vec<F>>();
+
+            combined = F::batch_multiplicative_inverse(&combined);
```
dvdplm, line 829: Just a thought: this inverting function seems to be called often. I wonder if it could be a target for SIMD optimization? It'd be a bit of a challenge, but perhaps worth it at some point?

starky/src/prover.rs
```diff
@@ -307,3 +374,102 @@ where
...
+            consumer.accumulators()
+        })
+        .collect::<Vec<_>>();
+
+    // Assert that all constraints evaluate to 0 over our subgroup.
+    for v in constraint_values {
+        assert!(
+            v.iter().all(|x| x.is_zero()),
+            "Constraint failed in {}",
+            std::any::type_name::<S>()
```
dvdplm, line 472: Use `core::any` here

starky/src/lookup.rs
```diff
@@ -0,0 +1,1006 @@
+use alloc::vec;
```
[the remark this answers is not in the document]
Nashtare: The document there is about CTLs, not lookup themselves. Though we'll probably work on some header descriptions before actual version bump.

starky/src/verifier.rs
```diff
@@ -81,16 +82,30 @@ pub(crate) fn verify_stark_proof_with_challenges<
         l_0,
         l_last,
     );
-    let permutation_data = stark.uses_permutation_args().then(|| PermutationCheckVars {
-        local_zs: permutation_zs.as_ref().unwrap().clone(),
-        next_zs: permutation_zs_next.as_ref().unwrap().clone(),
-        permutation_challenge_sets: challenges.permutation_challenge_sets.unwrap(),
+
+    let num_lookup_columns = stark.num_lookup_helper_columns(config);
```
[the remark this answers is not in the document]
Nashtare: It will be used once CTLs are available on starky side, so I'd rather keep it as is even if "for now" it is only used as a non-empty check.

starky/src/prover.rs
dvdplm: Before we add a feature I'd like to see hard proof that `log::debug!("bla bla {}", expensive_calculation_here());` actually is slow when LOG_LEVEL excludes debug logs. :)

Nashtare: My concern was not about it being not run at a different log level. My concern was that the lowest level is TRACE (AFAIK) which we use for debugging failures. And I'd rather not clog debugging by adding something that crazily increases proving time, while trace logging already incurs a large overhead because of all the I/O operations (we're talking GBs file)

PR #1499, opened 2024-02-05
Add missing constraints mentioned by auditors

Nashtare:


evm/src/logic.rs
```diff
@@ -276,11 +284,21 @@ impl<F: RichField + Extendable<D>, const D: usize> Stark<F, D> for LogicStark<F,
...
-        // IS_AND, IS_OR, and IS_XOR come from the CPU table, so we assume they're valid.
         let is_and = lv[columns::IS_AND];
         let is_or = lv[columns::IS_OR];
         let is_xor = lv[columns::IS_XOR];
 
+        // Flags must be boolean.
+        for &flag in &[is_and, is_or, is_xor] {
+            let constraint = builder.mul_sub_extension(flag, flag, flag);
+            yield_constr.constraint(builder, constraint);
+        }
```
Nashtare, line 295: I believe that since logic flags are combined on the CPU side, this initial assumption is now invalid and we also need to constrain each individual flag to be boolean.

Nashtare, line 295: I believe that since logic flags are combined on the CPU side, this initial assumption is now invalid and we also need to constrain each individual flag to be boolean.

hratoanina (approved):
Looks good, thanks!

PR #1501, opened 2024-02-06
Fix no-std tests and add corresponding jobs in the CI

Nashtare:
We currently do check that `plonky` and `starky` crates compile well in `no-std` for WASM targets, but it'd be nice to also make sure we can actually run their test suites in `no-std` to make sure the internal logic isn't broken.\

This PR fixes a bunch of missing conditional imports to make tests pass in `no-std`, and adds two more CI jobs for testing the two mentioned crates.

### DIFF
--- a/plonky2/src/lookup_test.rs
+++ b/plonky2/src/lookup_test.rs
@@ -1,28 +1,34 @@
-static LOGGER_INITIALIZED: Once = Once::new();
-
-use alloc::sync::Arc;
-use std::sync::Once;
+#[cfg(not(feature = "std"))]
+use alloc::{sync::Arc, vec, vec::Vec};
+#[cfg(feature = "std")]
+use std::sync::{Arc, Once};
 
 use itertools::Itertools;
-use log::{Level, LevelFilter};
+use log::Level;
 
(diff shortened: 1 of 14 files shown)


plonky2/src/lookup_test.rs
```diff
@@ -1,28 +1,34 @@
-static LOGGER_INITIALIZED: Once = Once::new();
```
Nashtare, line 1: This file has a lot of changes, basically just noise related to removing duplicate exports from within individual tests to the module level. Only the feature gating of `LOGGER_INITIALIZED` matters here.

Nashtare:
@pgebheim Could you update the required jobs for PRs to be mergeable?

- I updated the `wasm` job for consistent naming with others
- I added a new `no-std` job that probably should be required

muursh:
> @pgebheim Could you update the required jobs for PRs to be mergeable?
> 
> * I updated the `wasm` job for consistent naming with others
> * I added a new `no-std` job that probably should be required

I'll do it once this is merged

muursh (approved):
Fine by me

PR #1502, opened 2024-02-08
Revert "Remove StarkProofWithMetadata (#1497)"

Nashtare:
Oversight from my end, we need to keep track of the initial challenger state for recursion with many tables (consistency check of the challenger across tables), so reverting past PR.


LindaGuiga (approved):
LGTM!

muursh (approved):

PR #1503, opened 2024-02-09
Update `starky` and leverage it as dependency for `plonky2_evm`

### LINKED ISSUE #1098: Duplicate code in evm and starky crate?
It seems there are lots of duplicated code between evm and starky crate, like constraint_consumer.rs, vanishing_polynomials.rs, get_challanges.rs and so on. 

Also it seems cross_table_lookup is also generic so it can also be put under starky crate.

Any specific reason why above is not done yet?

Nashtare:
Final PR for code migration from `plonky2_evm` prover-related code back to `starky`.

There's not clear API defined for multi-stark logic within starky yet, mostly because I couldn't get a satisfactory `MultiStark` handling arrays of `Stark` because of their associated generic types. We could either leave it for now, with expanded documentation on how one can build a multi-stark system given the available API, or work our way through a satisfiable generic approach.

Tagging it as draft for now as there still some cleanup / refactoring to do, but open to any suggestion! I'll also move back all the documentation that was done on the evm crate modules that have been moved to starky.

closes #1098


starky/src/config.rs
```diff
@@ -28,7 +37,7 @@ impl StarkConfig {
         }
     }
 
-    pub(crate) fn fri_params(&self, degree_bits: usize) -> FriParams {
+    pub fn fri_params(&self, degree_bits: usize) -> FriParams {
```
dvdplm, line 40: Missing docs.

dvdplm: Just a thought: The `fri_config.fri_params` is also `pub`. We want to minimize the public API we expose as much as possible. Could it be that we don't need both to be public?

starky/src/cross_table_lookup.rs
```diff
@@ -493,13 +493,25 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
...
         cross_table_lookups: &'a [CrossTableLookup<F>],
         ctl_challenges: &'a GrandProductChallengeSet<F>,
         num_lookup_columns: &[usize; N],
         num_helper_ctl_columns: &Vec<[usize; N]>,
     ) -> [Vec<Self>; N] {
+        let mut ctl_vars_per_table = [0; N].map(|_| vec![]);
+        // If there are no auxiliary polys in the proofs `openings`,
+        // return early. The verifier will reject the proofs when
+        // calling `validate_proof_shape`.
+        for p in proofs.iter() {
```
dvdplm, line 507: `if proofs.iter().any(|p| p.proof.openings.auxiliary_polys.is_none() || proof.openings.auxiliary_polys_next.is_none())) { return ctl_vars_per_table }` perhaps?

starky/src/cross_table_lookup.rs
```diff
@@ -729,15 +740,23 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
         // Get all cross-table lookup polynomial openings for each STARK proof.
         let ctl_zs = {
             let openings = &proof.openings;
-            let ctl_zs = openings.auxiliary_polys.iter().skip(num_lookup_columns);
+            let ctl_zs = openings
+                .auxiliary_polys
+                .as_ref()
+                .unwrap()
```
dvdplm, line 746: Same.

starky/src/cross_table_lookup.rs
```diff
@@ -729,15 +740,23 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
...
+                .iter()
+                .skip(num_lookup_columns);
             let ctl_zs_next = openings
                 .auxiliary_polys_next
+                .as_ref()
+                .unwrap()
```
dvdplm, line 752: same

starky/src/cross_table_lookup.rs
```diff
@@ -729,15 +740,23 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
...
                 .iter()
                 .skip(num_lookup_columns);
             ctl_zs.zip(ctl_zs_next).collect::<Vec<_>>()
         };
 
-        // Put each cross-table lookup polynomial into the correct table data: if a CTL polynomial is extracted from looking/looked table t, then we add it to the `CtlCheckVars` of table t.
+        // Put each cross-table lookup polynomial into the correct table data.
```
dvdplm, line 758: "the correct table data" sounds odd. Is it "correct table's data" or "correct data table" or "data section of the correct table"?

starky/src/cross_table_lookup.rs
```diff
@@ -729,15 +740,23 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
...
+        // If a CTL polynomial is extracted from looking/looked table t, then we add it to the `CtlCheckVars` of table t.
```
dvdplm, line 759:
```suggestion
        // If a CTL polynomial is extracted from looking/looked table `t`, then we add it to the `CtlCheckVars` of table `t`.
```

starky/src/cross_table_lookup.rs
```diff
@@ -1019,34 +1037,35 @@ pub(crate) fn verify_cross_table_lookups_circuit<
     debug_assert!(ctl_zs_openings.iter_mut().all(|iter| iter.next().is_none()));
 }
 
-#[cfg(test)]
-pub(crate) mod testutils {
-    use std::collections::HashMap;
+#[cfg(debug_assertions)]
```
dvdplm, line 1040: :+1

starky/src/cross_table_lookup.rs
```diff
@@ -1019,34 +1037,35 @@ pub(crate) fn verify_cross_table_lookups_circuit<
...
+pub mod testutils {
+    #[cfg(not(feature = "std"))]
```
dvdplm, line 1042: Can't you make `std` a required feature of `debug_assertions`? That is, is it really plausible that one would run `debug_assertions` in a no-std env?

starky/src/cross_table_lookup.rs
```diff
@@ -1019,34 +1037,35 @@ pub(crate) fn verify_cross_table_lookups_circuit<
...
+            check_ctl(trace_poly_values, ctl, i, extra_looking_values.get(&i));
         }
     }
 
     fn check_ctl<F: Field>(
         trace_poly_values: &[Vec<PolynomialValues<F>>],
         ctl: &CrossTableLookup<F>,
         ctl_index: usize,
-        extra_memory_looking_values: &[Vec<F>],
+        extra_looking_values: Option<&Vec<Vec<F>>>,
```
dvdplm, line 1068: `&Vec<…>` is a bit annoying here, especially given that we clone the values on line 1091 anyway. Can we do better? Explicit clone in the caller?

starky/src/proof.rs
```diff
@@ -55,6 +57,40 @@ pub struct StarkProofTarget<const D: usize> {
 }
 
 impl<const D: usize> StarkProofTarget<D> {
+    /// Serializes a STARK proof.
+    pub fn to_buffer(&self, buffer: &mut Vec<u8>) -> IoResult<()> {
```
dvdplm, line 61: `buffer` could be slice I think

starky/src/proof.rs
```diff
@@ -176,22 +279,94 @@ impl<F: RichField + Extendable<D>, const D: usize> StarkOpeningSet<F, D> {
...
     pub auxiliary_polys: Option<Vec<ExtensionTarget<D>>>,
     pub auxiliary_polys_next: Option<Vec<ExtensionTarget<D>>>,
+    pub ctl_zs_first: Option<Vec<Target>>,
     pub quotient_polys: Vec<ExtensionTarget<D>>,
 }
 
 impl<const D: usize> StarkOpeningSetTarget<D> {
-    pub(crate) fn to_fri_openings(&self) -> FriOpeningsTarget<D> {
+    /// Serializes a STARK's opening set.
+    pub(crate) fn to_buffer(&self, buffer: &mut Vec<u8>) -> IoResult<()> {
```
dvdplm, line 319: Same.

In idiomatic Rust, `to_` methods should do some sort of (possibly expensive) _conversion_, which is not really the case here. This method is instead performing the IO into a buffer. Assuming we have only one way of serializing a STARK, why not call this (and similar methods) "serialize" instead?

dvdplm: Alternatively make this into an actual `to_` conversion that takes `Bytes` and returns a `Result<Buffer, _>`?

starky/src/cross_table_lookup.rs
```diff
@@ -514,16 +526,15 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
             .map(|(p, &num_lookup)| {
                 let openings = &p.proof.openings;
 
-                let ctl_zs = &openings.auxiliary_polys[num_lookup..];
-                let ctl_zs_next = &openings.auxiliary_polys_next[num_lookup..];
+                let ctl_zs = &openings.auxiliary_polys.as_ref().unwrap()[num_lookup..];
+                let ctl_zs_next = &openings.auxiliary_polys_next.as_ref().unwrap()[num_lookup..];
```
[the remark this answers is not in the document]
Nashtare: We can't have CTLs without auxiliary polys. All the extra `Option` wrappers are here to account for single-stark + no-lookup scenarios that effectively wouldn't have such auxiliary polys, but if you need CTLs (i.e. multi-stark system), regardless of if you use lookups, you will have these.
I could replace them all by `expect` but I found it so verbose that I just stuck to `unwrap`.

starky/src/cross_table_lookup.rs
```diff
@@ -1019,34 +1037,35 @@ pub(crate) fn verify_cross_table_lookups_circuit<
     debug_assert!(ctl_zs_openings.iter_mut().all(|iter| iter.next().is_none()));
 }
 
-#[cfg(test)]
-pub(crate) mod testutils {
-    use std::collections::HashMap;
+#[cfg(debug_assertions)]
+pub mod testutils {
```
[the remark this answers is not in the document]
Nashtare: Yeah I'll add a one liner explanation of the module too

starky/src/cross_table_lookup.rs
Nashtare: Can this be enforced at the Cargo.toml level? Or should I just change the `#[cfg(debug_assertions)]` by `#[cfg(all(debug_assertions, feature = "std"))]`?

starky/src/cross_table_lookup.rs
Nashtare: Yeah I agree, I am not too bothered by performances because this is never to be run in an actual release mode anyway (without extra debug assertions). But I'll look at changing it

starky/src/proof.rs
Nashtare: Actually it can't, the `Write` trait for serialization is only implemented over `Vec<u8>`

starky/src/proof.rs
Nashtare: It initially came from the distinction from all the `to_bytes` / `from_bytes` methods, where we needed access to the input/output buffer, hence the function name change to `to_buffer` / `from_buffer`.

starky/src/config.rs
Nashtare: Given `hiding` (i.e. zero-knowledgeness) is hardcoded to `false` on the stark side, leaving users the possibility to wrongly call this method with `true` worries me a bit (as it could possibly happen), so I'd rather maintain the wrapper function at the stark level that obfuscates this parameter. Does that make sense?

starky/src/cross_table_lookup.rs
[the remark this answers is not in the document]
dvdplm: If I could, I'd go back in time and unimplement `unwrap()` (or make sure `--release` doesn't work when there is one present). When we are positive that it's ok to panic, let's use `expect("motivation here")` and if it's long and ugly, so be it.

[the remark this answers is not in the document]
Nashtare: Alright, will change to `expect`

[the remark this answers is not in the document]
Nashtare: Alright, will change to `expect`

starky/src/proof.rs
Nashtare: If we want to update all the serialization method names, I'm not against it *but* this is out of scope of this PR.

starky/src/proof.rs
dvdplm: Ugh. That's a bit surprising and annoying. Oh well, not this PR for sure... :)

starky/src/lib.rs
```diff
@@ -1,14 +1,333 @@
+//! A FRI-based STARK implementation over the Goldilocks field, with support
+//! for recursive proof verification through the plonky2 SNARK backend.
+//!
+//! This library is intended to provide all the necessary tools to prove,
+//! verify, and recursively verify STARK statements. While the library
+//! is tailored for a system with a single STARK, it also is flexible
+//! enough to support a multi-STARK system, i.e. a system of independent
+//! STARK statements possibly sharing common values. See section below for
+//! more information on how to define such a system. // TODO: ROBIN
```
LindaGuiga, line 9:
```suggestion
//! more information on how to define such a system.
```

starky/src/lib.rs
```diff
@@ -1,14 +1,333 @@
...
+//! }
+//!
+//! // Define witness generation.
+//! impl<F: RichField + Extendable<D>, const D: usize> FibonacciStark<F, D> {
+//!     // The first public input is `x0`.
+//!     const PI_INDEX_X0: usize = 0;
+//!     // The second public input is `x1`.
+//!     const PI_INDEX_X1: usize = 1;
+//!     // The third public input is the second element of the last row,
+//!     //which should be equal to the `num_rows`-th Fibonacci number.
```
LindaGuiga, line 56:
```suggestion
//!     // which should be equal to the `num_rows`-th Fibonacci number.
```

LindaGuiga, line 56:
```suggestion
//!     // which should be equal to the `num_rows`-th Fibonacci number.
```

starky/src/lib.rs
```diff
@@ -1,14 +1,333 @@
...
+//!
+//!     fn constraint_degree(&self) -> usize {
+//!         2
+//!     }
+//! }
+//! ```
+//!
+//! One can then instantiate a new `FibonacciStark` instance, generate an associated
+//! STARK trace, and generate a proof for it.
+//!
```
LindaGuiga, line 157: I think the witness generation and constraints are written twice in this file (and should therefore be removed from below). What follows should start at `fn fibonacci<F: Field>(n: usize, x0: F, x1: F) -> F {`

starky/src/lookup.rs
```diff
@@ -402,12 +398,24 @@ impl<F: Field> Column<F> {
 
 pub(crate) type ColumnFilter<'a, F> = (&'a [Column<F>], &'a Option<Filter<F>>);
 
+/// A `Lookup` defines a set of `columns`` whose values should appear in a
```
LindaGuiga, line 401:
```suggestion
/// A `Lookup` defines a set of `columns` whose values should appear in a
```

Also, should these be [Lookup] etc, as in the other comments?

starky/src/lookup.rs
```diff
@@ -419,6 +427,7 @@ pub struct Lookup<F: Field> {
 }
 
 impl<F: Field> Lookup<F> {
+    /// Outputs the number of helper columns needed by this `Lookup`.
```
LindaGuiga, line 430: Same question as above for `Lookup`. (There are a couple of others below as well, such as `GrandProductChallengeSet`)

starky/src/proof.rs
```diff
@@ -46,15 +52,57 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize> S
...
+        buffer.write_target_merkle_cap(&self.quotient_polys_cap)?;
+        buffer.write_target_fri_proof(&self.opening_proof)?;
+        self.openings.to_buffer(buffer)?;
+        Ok(())
+    }
+
+    /// Deserializes a STARK proof.
+    pub fn from_buffer(buffer: &mut Buffer) -> IoResult<Self> {
+        let trace_cap = buffer.read_target_merkle_cap()?;
+        let auxiliary_polys_cap = if buffer.read_bool()? {
```
LindaGuiga, line 88: I might be totally wrong, but doesn't this read the next value anyway? (so that if there's no bool, we'd read the next value but "discard" it when it should be part of the `quotient_polys_cap`?)

starky/src/proof.rs
```diff
@@ -46,15 +52,57 @@ impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize> S
...
 }
 
 impl<const D: usize> StarkProofTarget<D> {
+    /// Serializes a STARK proof.
+    pub fn to_buffer(&self, buffer: &mut Vec<u8>) -> IoResult<()> {
+        buffer.write_target_merkle_cap(&self.trace_cap)?;
+        if let Some(poly) = &self.auxiliary_polys_cap {
+            buffer.write_bool(true)?;
+            buffer.write_target_merkle_cap(poly)?;
+        }
```
LindaGuiga, line 78: Should there be an `else { buffer.write_bool(false)?; }` if my next comment below is true?

starky/src/prover.rs
```diff
@@ -68,54 +73,120 @@ where
...
+        &mut challenger,
+        public_inputs,
+        timing,
+    )
+}
+
+/// Generates a proof for a single STARK table, including:
+///
+/// - the initial state of the challenger,
+/// - all the requires Merkle caps,
```
LindaGuiga, line 92: nit 
```suggestion
/// - all the required Merkle caps,
```

LindaGuiga (approved):
LGTM besides a few nitpicks/questions/comments, thanks!

4l0n50 (commented):
I need to pass again, but looks good to me in general

starky/src/lib.rs
Nashtare: It wouldn't compile if this wasn't included. Rustdoc starting with ` # ` does not print the line but uses it for compiling / running doctest. You can see it doesn't appear twice by calling `cargo doc -p starky --no-deps --open`.

starky/src/lib.rs
```diff
@@ -1,14 +1,333 @@
+//! A FRI-based STARK implementation over the Goldilocks field, with support
+//! for recursive proof verification through the plonky2 SNARK backend.
+//!
+//! This library is intended to provide all the necessary tools to prove,
+//! verify, and recursively verify STARK statements. While the library
+//! is tailored for a system with a single STARK, it also is flexible
```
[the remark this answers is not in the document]
Nashtare: Starky on its own is tailored for single-STARK system. There is no clear API for building multi-STARK systems yet, like the one on the EVM side, although the pieces are there so that people can build them.

starky/src/proof.rs
Nashtare: Ah yeah good catch. Moved the boolean write outside of the `if` as we need it in both cases

starky/src/proof.rs
Nashtare: I'm not sure to understand this comment. If it refers to the problematic write of boolean you mentioned above then yes. Considering the above fixed, then there is no issue here anymore (we always will have a bool to read, and read the rest only if we do have auxiliary poly cap).

starky/src/proof.rs
```diff
@@ -95,69 +152,158 @@ pub struct CompressedStarkProof<
...
+    /// Proofs for all the different STARK modules.
+    pub stark_proofs: [StarkProofWithMetadata<F, C, D>; N],
+    /// Cross-table lookup challenges.
+    pub ctl_challenges: GrandProductChallengeSet<F>,
+}
 
+impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize, const N: usize>
+    MultiProof<F, C, D, N>
+{
+    /// Returns the degree (i.e. the trace length) of each STARK proofs,
```
4l0n50, line 200:
```suggestion
    /// Returns the degree (i.e. the trace length) of each STARK proof,
```

starky/src/proof.rs
```diff
@@ -95,69 +152,158 @@ pub struct CompressedStarkProof<
...
+    /// from their common [`StarkConfig`].
+    pub fn recover_degree_bits(&self, config: &StarkConfig) -> [usize; N] {
+        core::array::from_fn(|i| self.stark_proofs[i].proof.recover_degree_bits(config))
+    }
+}
+
+/// Randomness used for a STARK proof.
+#[derive(Debug)]
+pub struct StarkProofChallenges<F: RichField + Extendable<D>, const D: usize> {
+    /// Optional randomness used in any permutation arguments.
```
4l0n50, line 210:
```suggestion
    /// Optional randomness used in any permutation argument.
```

starky/src/proof.rs
```diff
@@ -95,69 +152,158 @@ pub struct CompressedStarkProof<
...
-
+    /// Randomness used in FRI.
     pub fri_challenges: FriChallenges<F, D>,
 }
 
-pub(crate) struct StarkProofChallengesTarget<const D: usize> {
+/// Circuit version of [`StarkProofChallenges`].
+#[derive(Debug)]
+pub struct StarkProofChallengesTarget<const D: usize> {
+    /// Optional `Target`'s randomness used in any permutation arguments.
```
4l0n50, line 223:
```suggestion
    /// Optional `Target`'s randomness used in any permutation argument.
```

starky/src/recursive_verifier.rs
```diff
@@ -277,13 +333,16 @@ pub fn set_stark_proof_with_pis_target<F, C: GenericConfig<D, F = F>, W, const D
         witness.set_target(pi_t, pi);
     }
 
-    set_stark_proof_target(witness, pt, proof);
+    set_stark_proof_target(witness, pt, proof, zero);
 }
 
+/// Set the targets in a `StarkProofTarget` totheir corresponding values in a
```
4l0n50, line 339:
```suggestion
/// Set the targets in a `StarkProofTarget` to their corresponding values in a
```

4l0n50 (approved):
LGTM, only nits.

evm/src/arithmetic/divmod.rs
```diff
@@ -321,52 +321,57 @@ mod tests {
...
-                // the output is defined to be zero when modulus is zero.
-                let mut lv = [F::default(); NUM_ARITH_COLUMNS]
-                    .map(|_| F::from_canonical_u16(rng.gen::<u16>()));
-                let mut nv = [F::default(); NUM_ARITH_COLUMNS]
-                    .map(|_| F::from_canonical_u16(rng.gen::<u16>()));
-
-                // Reset operation columns, then select one
-                for op in MODULAR_OPS {
-                    lv[op] = F::ZERO;
+                for corrupt_constraints in [false, true] {
```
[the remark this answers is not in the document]
Nashtare: They should probably be refactored with the other arithmetic tests anyway at some point.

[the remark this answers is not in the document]
LindaGuiga: Thanks!

PR #1512, opened 2024-02-10
Implement Hermez SMT

wborgeaud:
Implement the Hermez SMT and replace the state MPT with a state SMT.

Still TODO:
- Implement `PoseidonStark`, reusing code from #1311.
- Implement Hermez's Poseidon sponge and check codehash in `load_code`.
- Fix tests in `evm/src/cpu/kernel/tests/mpt/*`.
- Figure out what to do with SEFLDESTRUCT.

For some context on the last one, I think it's impossible to correctly implement SELFDESTRUCT with this SMT design, because we can't clear the storage of an account. Polygon zkEVM replaced SELFDESTRUCT with SENDALL, so we could do the same. Otherwise, we could fast track EIP-6780 (included in Cancun) which modifies SELFDESTRUCT in a way that we can implement with this SMT.


Nashtare:
Closing in favor of https://github.com/0xPolygonZero/zk_evm/pull/34

PR #1519, opened 2024-02-10
Remove risk of panics in interpreter

Nashtare:
Now that the interpreter is aimed at being used for `jumpdest_analysis`, and soon for continuations, we need to alleviate the remaining risks of `panic` that can occur.


muursh (approved):
:)

PR #1529, opened 2024-02-16
Add hash public input methods

hratoanina:
Methods needed for continuation.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -339,7 +339,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
 
     /// Adds a new `HashOutTarget`.
     pub fn add_virtual_hash(&mut self) -> HashOutTarget {
-        HashOutTarget::from_vec(self.add_virtual_target_arr<4>())
+        HashOutTarget::from_vec(self.add_virtual_target_arr::<4>())
     }
 
     /// Registers a new `HashOutTarget` as a public input, adding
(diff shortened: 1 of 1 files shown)


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -343,6 +343,13 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         HashOutTarget::from_vec(self.add_virtual_targets(4))
     }
 
+    /// Registers a new `HashOutTarget` as a public input. `NUM_HASH_OUT_ELTS`
+    /// being hardcoded to 4, it internally adds 4 virtual targets in a
+    /// vector fashion.
```
Nashtare, line 348: Nit (not likely, but in the eventuality it was to change in the future), perhaps would be easier to just have:

```suggestion
    /// Registers a new `HashOutTarget` as a public input, adding
    /// internally `NUM_HASH_OUT_ELTS` virtual targets.
```

Nashtare: (also, they are _not_ added in vector fashion, see comment below)

plonky2/src/plonk/circuit_builder.rs
```diff
@@ -343,6 +343,13 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
...
+    pub fn add_virtual_hash_public_input(&mut self) -> HashOutTarget {
+        HashOutTarget::from_vec(self.add_virtual_public_input_arr::<4>().into())
```
Nashtare, line 350: `HashOutTarget` stores elements internally as an array, and implements `From<[Target; NUM_HASH_OUT_ELTS]>`, so could just do:
```suggestion
        HashOutTarget::from(self.add_virtual_public_input_arr::<4>())
```
so that we don't need the allocation.

Nashtare (approved):

PR #1532, opened 2024-02-17
Do some additional cleanup pre-release

Nashtare:
- adds `Debug` implementation for all types, adding `#![deny(missing_debug_implementations)]` to `plonky2_field`, `plonky2` and `plonky2_util` crates (`starky` already has it, and `plonky2_maybe_rayon` only reexports stuff).
- hides internal methods of custom gates that should not be exposed, only maintaining the meaningful ones for new instance declaration or recursion purposes
- makes some more functions `const`


LindaGuiga (approved):
LGTM thanks!

PR #1540, opened 2024-02-23
chore: remove conditional compilation for debug_utils

dvdplm:
See [this PR](https://github.com/0xPolygonZero/zk_evm/pull/60) for the motivation.

### DIFF
--- a/starky/src/cross_table_lookup.rs
+++ b/starky/src/cross_table_lookup.rs
@@ -1041,12 +1041,10 @@ pub fn verify_cross_table_lookups_circuit<
     debug_assert!(ctl_zs_openings.iter_mut().all(|iter| iter.next().is_none()));
 }
 
-/// Debugging module, to assert correctness of the different CTLs of a multi-STARK system,
+/// Debugging module used to assert correctness of the different CTLs of a multi-STARK system,
 /// that can be used during the proof generation process.
 ///
-/// **Note**: this is an expensive check, hence is only available when the `debug_assertions`
-/// flag is activated, to not hinder performances with regular `release` build.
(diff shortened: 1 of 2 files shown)


starky/src/cross_table_lookup.rs
```diff
@@ -1041,12 +1041,10 @@ pub fn verify_cross_table_lookups_circuit<
...
 }
 
-/// Debugging module, to assert correctness of the different CTLs of a multi-STARK system,
+/// Debugging module used to assert correctness of the different CTLs of a multi-STARK system,
 /// that can be used during the proof generation process.
 ///
-/// **Note**: this is an expensive check, hence is only available when the `debug_assertions`
-/// flag is activated, to not hinder performances with regular `release` build.
-#[cfg(debug_assertions)]
+/// **Note**: This is an expensive check.
```
Nashtare, line 1047: Non-blocking, but shouldn't we write rustdoc above the flag, for consistency with the rest of the codebase? (derive implementations and whatnot)

Nashtare (commented):
Thanks David!
I'm not against the change, although I don't think this is the source of the issue on zk_evm side. If we're importing within the brackets that are conditionally compiled with `debug_assertions`, we should be fine IMO.

Also missing an entry in the CHANGELOG here

starky/src/cross_table_lookup.rs
dvdplm: The flag is removed though! Or am I misunderstanding?

dvdplm:
> It's missing an entry in the CHANGELOG however.

Updated, my bad.

PR #1547, opened 2024-02-28
Pacify clippy with latest `nightly` toolchain

Nashtare:
Clippy complaining about multiple bound locations and such


plonky2/src/hash/poseidon.rs
```diff
@@ -923,7 +923,7 @@ impl<F: RichField> AlgebraicHasher<F> for PoseidonHash {
 pub(crate) mod test_helpers {
     use super::*;
 
-    pub(crate) fn check_test_vectors<F: Field>(
```
Nashtare, line 926: `Poseidon` is a super trait of `Field`

muursh (approved):

PR #1552, opened 2024-03-08
Prove Starks without constraints

4l0n50:
This PR enable to implement starks without constraints.


starky/src/stark.rs
```diff
@@ -44,11 +44,13 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
-        vars: &Self::EvaluationFrame<FE, P, D2>,
-        yield_constr: &mut ConstraintConsumer<P>,
+        _vars: &Self::EvaluationFrame<FE, P, D2>,
+        _yield_constr: &mut ConstraintConsumer<P>,
     ) where
         FE: FieldExtension<D2, BaseField = F>,
-        P: PackedField<Scalar = FE>;
+        P: PackedField<Scalar = FE>,
+    {
+    }
```
Nashtare, line 53: To prevent any risk of misuse for downstream users of `starky`, I think we shouldn't have the blanket impl do this, and keep it as it was. If one (like in our usecase) was to need such unconstrained table, they would need to explicitly mention it in their implementation.

starky/src/stark.rs
```diff
@@ -74,17 +76,21 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
         &self,
-        builder: &mut CircuitBuilder<F, D>,
-        vars: &Self::EvaluationFrameTarget,
-        yield_constr: &mut RecursiveConstraintConsumer<F, D>,
-    );
+        _builder: &mut CircuitBuilder<F, D>,
+        _vars: &Self::EvaluationFrameTarget,
+        _yield_constr: &mut RecursiveConstraintConsumer<F, D>,
+    ) {
+    }
```
Nashtare, line 83: Same comment as above

starky/src/verifier.rs
```diff
@@ -164,19 +164,22 @@ where
...
     // `quotient_degree_factor`-sized chunk of the original evaluations.
-    for (i, chunk) in quotient_polys
-        .chunks(stark.quotient_degree_factor())
-        .enumerate()
-    {
-        ensure!(
-            vanishing_polys_zeta[i] == z_h_zeta * reduce_with_powers(chunk, zeta_pow_deg),
-            "Mismatch between evaluation and opening of quotient polynomial"
-        );
+    // TODO: this should be done only with the iterator
```
Nashtare, line 167: I'm not sure what you mean by this

plonky2/src/recursion/recursive_verifier.rs
```diff
@@ -1,3 +1,5 @@
+use std::vec;
```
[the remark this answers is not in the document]
4l0n50: Thanks! I didn't know why why the no-std test were failing

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,217 @@
+//! An example of generating and verifying STARK proofs for the Fibonacci sequence.
+//! The toy STARK system also includes two columns that are a permutation of the other,
+//! to highlight the use of the permutation argument with logUp.
```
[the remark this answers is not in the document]
4l0n50: It's just by-product of the copy-paste. I need to update the comment.

starky/src/verifier.rs
4l0n50: That it should be a way of avoiding the if let converting the option quotient_polys to iter. But I don't remember why it didn't worked when I tried.

starky/src/lookup.rs
```diff
@@ -666,9 +669,12 @@ pub(crate) fn eval_helper_columns<F, FE, P, const D: usize, const D2: usize>(
...
     if !helper_columns.is_empty() {
-        for (j, chunk) in columns.chunks(constraint_degree - 1).enumerate() {
-            let fs =
-                &filter[(constraint_degree - 1) * j..(constraint_degree - 1) * j + chunk.len()];
+        for (j, chunk) in columns
+            .chunks(constraint_degree.checked_sub(1).unwrap_or(1))
+            .enumerate()
+        {
+            let fs = &filter[(constraint_degree.checked_sub(1).unwrap_or(1)) * j
+                ..(constraint_degree.checked_sub(1).unwrap_or(1)) * j + chunk.len()];
```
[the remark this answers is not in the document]
4l0n50: Hmmm. And now that I think, Idk why it's `.unwrap_or(1)` and not `.unwrap_or(0)`. Both work :confused:

starky/src/permutation_stark.rs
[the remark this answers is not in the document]
4l0n50: Btw, I left `fibonacci_stark.rs` unchanged but I think we can remove the permutation fibonacci example now that it can "live" in its own stark.

starky/src/lookup.rs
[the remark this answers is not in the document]
Nashtare: `.saturating_sub(1)` would yield 0, while here we need to have a lower bound of 1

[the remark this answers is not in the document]
4l0n50: A yes you're right. It should be the size of the chunk which can't be 0.

starky/src/lookup.rs
```diff
@@ -719,9 +722,12 @@ pub(crate) fn eval_helper_columns_circuit<F: RichField + Extendable<D>, const D:
...
     if !helper_columns.is_empty() {
-        for (j, chunk) in columns.chunks(constraint_degree - 1).enumerate() {
-            let fs =
-                &filter[(constraint_degree - 1) * j..(constraint_degree - 1) * j + chunk.len()];
+        for (j, chunk) in columns
+            .chunks(constraint_degree.checked_sub(1).unwrap_or(1))
+            .enumerate()
+        {
+            let fs = &filter[(constraint_degree.saturating_sub(1)) * j
+                ..(constraint_degree.saturating_sub(1)) * j + chunk.len()];
```
LindaGuiga, line 730: Shouldn't this be as above? 
```suggestion
let chunk_size = constraint_degree.checked_sub(1).unwrap_or(1);
        for (chunk, (fs, &h)) in columns
            .chunks(chunk_size)
            .zip(filter.chunks(chunk_size).zip(helper_columns))
        {
```
Especially since, as mentioned in Robin's comments, `saturating_sub` would yield 0 but I think we need 1?

starky/src/prover.rs
```diff
@@ -353,6 +358,10 @@ where
     let rate_bits = config.fri_config.rate_bits;
     let total_num_helper_cols: usize = num_ctl_columns.iter().sum();
 
+    if stark.quotient_degree_factor() == 0 {
+        return None;
+    }
+
```
LindaGuiga, line 364: Nit: Can't this be at the very start at the function? (we don't need the degree, rate bits and total number of helper columns if we're not computing the polynomial)

starky/src/stark.rs
```diff
@@ -192,11 +201,17 @@ pub trait Stark<F: RichField + Extendable<D>, const D: usize>: Sync {
...
-            blinding: false,
-        });
+        let quotient_info = if num_quotient_polys > 0 {
+            let quotient_polys =
+                FriPolynomialInfo::from_range(oracles.len(), 0..num_quotient_polys);
+            oracles.push(FriOracleInfo {
+                num_polys: num_quotient_polys,
+                blinding: false,
+            });
+            quotient_polys
```
LindaGuiga, line 211: Very nit:
```suggestion
            FriPolynomialInfo::from_range(oracles.len(), 0..num_quotient_polys);
            oracles.push(FriOracleInfo {
                num_polys: num_quotient_polys,
                blinding: false,
            })
```

LindaGuiga (approved):
LGTM, thanks! Just a couple of nits!

pgebheim:
Does this need to be raised against 0xPolygonZero/zk_evm instead of this repo?

starky/src/lookup.rs
```diff
@@ -577,7 +580,7 @@ pub(crate) fn lookup_helper_columns<F: Field>(
     constraint_degree: usize,
 ) -> Vec<PolynomialValues<F>> {
     assert!(
-        constraint_degree == 2 || constraint_degree == 3,
+        constraint_degree <= 3,
         "TODO: Allow other constraint degrees."
```
Nashtare, line 584: Actually this was backported from the evm side but I'm not sure I understand the restriction. As long as `constraint_degree <= blowup_factor + 1`, this should be fine, no?

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl<F: RichField + Extendable<D>, const D: usize> PermutationStark<F, D> {
+    const fn new(num_rows: usize) -> Self {
+        Self {
+            num_rows,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Generate the trace using `x0, x1, 0, 1, 1` as initial state values.
```
Nashtare, line 38:
```suggestion
    /// Generate the trace using `x0, x0+1, 1` as initial state values.
```

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+        let mut trace_rows = (0..self.num_rows)
+            .scan([x0, x0 + F::ONE, F::ONE], |acc, _| {
+                let tmp = *acc;
+                acc[0] = tmp[0] + F::ONE;
+                acc[1] = tmp[1] + F::ONE;
+                // acc[2] (i.e. frequency column) remains unchanged, as we're permuting a strictly monotonous sequence.
+                Some(tmp)
+            })
+            .collect::<Vec<_>>();
+        trace_rows[self.num_rows - 1][1] = F::ZERO; // So that column 0 and 1 are permutation of one another.
```
Nashtare, line 49: This passes fine because the test sets `x0` to `0`, but it should be 
```suggestion
        trace_rows[self.num_rows - 1][1] = x0; // So that column 0 and 1 are permutation of one another.
```

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn lookups(&self) -> Vec<Lookup<F>> {
+        vec![Lookup {
+            columns: vec![Column::single(0)],
+            table_column: Column::single(1),
+            frequencies_column: Column::single(2),
+            filter_columns: vec![None; 1],
+        }]
+    }
+
+    // No constraints
```
Nashtare, line 79: nit, for clarity:
```suggestion
    // We don't constrain any register, for the sake of highlighting the permutation argument only.
```

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+        &self,
+        _vars: &Self::EvaluationFrame<FE, P, D2>,
+        _yield_constr: &mut ConstraintConsumer<P>,
+    ) where
+        FE: FieldExtension<D2, BaseField = F>,
+        P: PackedField<Scalar = FE>,
+    {
+    }
+
+    // No constraints
```
Nashtare, line 90: same as above

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+        type C = PoseidonGoldilocksConfig;
+        type F = <C as GenericConfig<D>>::F;
+        type S = PermutationStark<F, D>;
+
+        let config = StarkConfig::standard_fast_config();
+        let num_rows = 1 << 5;
+
+        let public_input = F::ZERO;
+
+        // Test first STARK
```
Nashtare, line 136: remove
```suggestion
```

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
...
+        const D: usize = 2;
+        type C = PoseidonGoldilocksConfig;
+        type F = <C as GenericConfig<D>>::F;
+        type S = PermutationStark<F, D>;
+
+        let config = StarkConfig::standard_fast_config();
+        let num_rows = 1 << 5;
+        let public_input = F::ZERO;
+
+        // Test first STARK
```
Nashtare, line 186:
```suggestion
```

starky/src/permutation_stark.rs
```diff
@@ -0,0 +1,238 @@
+//! An example of generating and verifying a STARK to highlight the use of the
+//! permutation argument with logUp.
```
Nashtare, line 2: Given that we have this example now, as you suggested, I think it'd make sense to remove the `permutation` from the Fibonacci example, as well as the needless columns (used only for the sake of permutation)

Nashtare (approved):

Nashtare:
@4l0n50 Could you add an entry to the CHANGELOG as well? Thanks!

4l0n50:
> Does this need to be raised against 0xPolygonZero/zk_evm instead of this repo?

It changes a little the the starky code and can be eventually used outside 0xPolygonZero/zk_evm, so I would say it goes here.

starky/src/lookup.rs
4l0n50: Maybe because <= 3 was a requirement when this code was in the zk_evm?

4l0n50: Agree, given that there are no zk_evms here.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- Make Starks without constraints provable [#1552](https://github.com/0xPolygonZero/plonky2/pull/1552)
```
Nashtare, line 11: Just to match previous format.
```suggestion
- Make Starks without constraints provable ([#1552](https://github.com/0xPolygonZero/plonky2/pull/1552))
```

4l0n50: You can still prove CTLs/lookups (which are some kind of fixed constraints though)

starky/src/lookup.rs
4l0n50: The only problem is that here we don't know the blowup_factor. But why is this assert here in the first place?

CHANGELOG.md
4l0n50: I think the proof of knowledge property appears once you add the OOD evaluation. But I'm not sure this will be the case if you don't have CTLs or lookups.

Nashtare: > You can still prove CTLs/lookups

Yeah though you don't necessarily have them. But I get your point.

starky/src/prover.rs
```diff
@@ -119,6 +119,12 @@ where
         "FRI total reduction arity is too large.",
     );
 
+    let constraint_degree = stark.constraint_degree();
+    assert!(
+        constraint_degree <= 2 << rate_bits + 1,
```
Nashtare, line 124:
```suggestion
        constraint_degree <= (1 << rate_bits) + 1,
```

starky/src/lookup.rs
```diff
@@ -885,10 +889,7 @@ pub(crate) fn eval_packed_lookups_generic<F, FE, P, S, const D: usize, const D2:
     let local_values = vars.get_local_values();
     let next_values = vars.get_next_values();
     let degree = stark.constraint_degree();
-    assert!(
-        degree == 2 || degree == 3,
-        "TODO: Allow other constraint degrees."
-    );
+    assert!(degree <= 3, "TODO: Allow other constraint degrees.");
```
Nashtare, line 892: needs to be updated still

starky/src/lookup.rs
```diff
@@ -958,10 +959,7 @@ pub(crate) fn eval_ext_lookups_circuit<
 
     let local_values = vars.get_local_values();
     let next_values = vars.get_next_values();
-    assert!(
-        degree == 2 || degree == 3,
-        "TODO: Allow other constraint degrees."
-    );
+    assert!(degree <= 3, "TODO: Allow other constraint degrees.");
```
Nashtare, line 962: same as above

starky/src/prover.rs
```diff
@@ -119,6 +119,12 @@ where
...
+        "The degree of the Stark constraints must be <= blowup_factor + 1".
```
Nashtare, line 125: this breaks compilation
```suggestion
        "The degree of the Stark constraints must be <= blowup_factor + 1",
```

Nashtare (changes requested):

CHANGELOG.md
4l0n50: It works without CTLs/lookups out-of-the-box. I added an example of a completely unconstrained stark

starky/src/lookup.rs
```diff
@@ -579,11 +579,6 @@ pub(crate) fn lookup_helper_columns<F: Field>(
     challenge: F,
     constraint_degree: usize,
 ) -> Vec<PolynomialValues<F>> {
-    assert!(
-        constraint_degree <= 3,
-        "TODO: Allow other constraint degrees."
-    );
```
Nashtare: I think we'd rather keep the assert, but with the updated condition instead.

4l0n50: I moved the assert to `prove_with_commitment` since you don't know the `rate_bits` there (and is also weird to check the constraint degree in the lookups)

starky/src/lookup.rs
Nashtare: I think it's just to not go through an expensive proof computation if the parameters were badly set. We can get rid of it, or adapt it to take into consideration the blowup factor.

Nashtare: But one of the two decisions needs to be taken before this is merged.

4l0n50: > I think it's just to not go through an expensive proof computation

Indeed. I mean, why is the assertion there if the file is about lookups. That's why I moved the assertion to `prove_with_commitment`, where you also happen to have access to the `rate_bits`.

Nashtare (approved):
LGTM!

PR #1555, opened 2024-03-14
Fix case of exactly 2 looking tables for CTLs

### LINKED ISSUE #1554: Fix CTLs with exactly two looking tables
Something wrong happens when declaring a CTL with exactly two looking tables (one looking table works, and three looking tables works). I encountered it while working on the zk_evm repository, but I'm pretty sure this comes from here. I suspect this comes from how the helper columns are constructed.

The error we get is: `Mismatch between evaluation and opening of quotient polynomial`.

Nashtare:
The way helper columns were computed was truncating the resulting vector for `n = 2`, resulting in quotient polynomial check failure. 

closes #1554

### DIFF
--- a/starky/src/cross_table_lookup.rs
+++ b/starky/src/cross_table_lookup.rs
@@ -123,7 +123,7 @@ impl<F: Field> CrossTableLookup<F> {
         for (i, ctl) in ctls.iter().enumerate() {
             let all_tables = once(&ctl.looked_table).chain(&ctl.looking_tables);
             let num_appearances = all_tables.filter(|twc| twc.table == table).count();
-            let is_helpers = num_appearances > 2;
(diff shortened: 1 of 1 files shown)


muursh (approved):

LindaGuiga (approved):
The fix looks good to me, thanks! But I think we shouldn't need helper columns for 2 columns (and that's why it was ignored). The constraints with two looking columns would still be of degree 3 without helper columns: `combin0 * combin1 * z - filter0 * combin1 - filter1 * combin0`. SO I think we could try to take that into account in another PR to save an extra column in that case!

Nashtare:
Yeah I noticed it while fixing, but would probably be requiring more refactor and I wanted a quick fix :')

hratoanina (approved):
Thanks!

PR #1565, opened 2024-04-05
Derive `Default` and `Eq` where possible

matthiasgoergens:
We can't use Rust's default deriving mechanism, until https://github.com/rust-lang/rust/issues/26925 and https://github.com/rust-lang/rust/issues/7671 are fixed.

But fortunately, there's a crate for that.


Nashtare (commented):
I am not really in favor of these changes. I don't see the problem of having a few custom implementations of `Default` or `Eq`, especially if it prevents including an additional dependency that hasn't been maintained for the past 3 years.

Is the current implementation problematic in any way for your usecase?

PR #1566, opened 2024-04-11
Add stdarch_x86_avx512 feature

jimpo:
Without this, compiling on AVX512 enabled machine gives an error:

```
error[E0658]: use of unstable library feature 'stdarch_x86_avx512'
```


Nashtare (approved):

Nashtare:
@jimpo Thanks for the fix! Would you mind merging latest `main`? clippy's been fixed there. We'll then be good to go when CI goes green

muursh (approved):

PR #1567, opened 2024-04-15
Simplify types: remove option from CTL filters

matthiasgoergens:
CTL filters can already express perfectly well the behaviour that we gave for `None`.  No need to complicated anything.


matthiasgoergens:
We might bump up the version number, because this is probably an incompatible change to the API.

Nashtare:
@matthiasgoergens starky v0.4.0 has been published, including these changes as well as compilation fix when AVX-512 instructions are enabled.

PR #1572, opened 2024-04-17
refactor: switch to `tiny-keccak` library

Daniel-Aaron-Bloom:
Improves performance by switching to the underlying library to do incremental hashing without allocation (i.e. no more temporary `Vec`s)


muursh (commented):
I'm hesitant over this. It's been a long time since `tiny-keccak` work done on it and I feel `keccak-hash` is a safer bet maintenance wise.

Daniel-Aaron-Bloom:
I chose `tiny-keccak` because `keccak-hash` utilizes it for the actual hashing implementation.

What makes you feel like the maintenance on `keccak-hash` will be better than `tiny-keccak`? It doesn't look like it's had any meaningful code updates [since 2020](https://github.com/paritytech/parity-common/commits/master/keccak-hash/src), and honestly as far as it's used here, it seems like just ~100 line wrapper around `tiny-keccak`.

I'd be happy to make PR to use something like `sha3` instead if maintenance is a concern.

muursh:
My main argument would be that I expect Parity to be more likely to maintain their crate, which makes my life easier long term. I'm open to a change if you have some convincing benchmarks or the like though

Nashtare (commented):
I would also tend to favor the library developed by Parity as opposed to `tiny-keccak` which hasn't been maintained in years.

muursh:
I'm going to close this for now given the consensus that `tiny-keccak` has a higher chance of being problematic at some point.

PR #1573, opened 2024-04-17
fix: remove trivial allocations and `try_into`

Daniel-Aaron-Bloom:
Use the following array tools to avoid allocations and `try_into` calls:

- `core::array::from_fn`/`core::array::try_from_fn`
- `[T]::first_chunk`
- `[T]::array_chunks`
- `[T; N]::map`
- `iter_fixed::IntoIteratorFixed`

Some of the `try_into` calls are more of a nit just because they look like allocations but are really just slice conversions, but it was easier to just get rid of all of them.

### DIFF
--- a/plonky2/src/hash/hash_types.rs
+++ b/plonky2/src/hash/hash_types.rs
@@ -90,14 +90,10 @@ impl<F: RichField> GenericHashOut<F> for HashOut<F> {
     }
 
     fn from_bytes(bytes: &[u8]) -> Self {
+        let bytes = bytes.as_chunks::<8>().0.first_chunk().unwrap();
+
         HashOut {
-            elements: bytes
-                .chunks(8)
-                .take(NUM_HASH_OUT_ELTS)
-                .map(|x| F::from_canonical_u64(u64::from_le_bytes(x.try_into().unwrap())))
-                .collect::<Vec<_>>()
-                .try_into()
-                .unwrap(),
+            elements: bytes.map(u64::from_le_bytes).map(F::from_canonical_u64),
         }
     }
 
--- a/plonky2/src/hash/hash_types.rs
+++ b/plonky2/src/hash/hash_types.rs
@@ -90,7 +90,8 @@ impl<F: RichField> GenericHashOut<F> for HashOut<F> {
     }
 
     fn from_bytes(bytes: &[u8]) -> Self {
-        let bytes = bytes.as_chunks::<8>().0.first_chunk().unwrap();
+        // TODO: replace with as_chunks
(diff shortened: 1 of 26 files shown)


Nashtare (commented):
As an overall comment, we've been working on removing unstable features from these libraries (with the current status of having `specialization` remaining for `plonky2_field`).

This PR isn't really aligned in that it brings two additional unstable features, and as such, I am not enclined in merging this. As a general rule, such addition should be justified by significant performance improvement to balance out the drawback.

Daniel-Aaron-Bloom:
Fair.

I would note though that `specialization` is probably the least likely feature to be stabilized in the near future, so depending on your plans to remove that from `plonky2_field`, the introduction of these features may or may not lengthen the timeline to stability (assuming that's your goal).

Other than micro-benchmarks, but there's no real perf impact of this change. I can trim it down to just the stable if there's any interest in that.

muursh:
> I would note though that `specialization` is probably the least likely feature to be stabilized in the near future, so depending on your plans to remove that from `plonky2_field`, the introduction of these features may or may not lengthen the timeline to stability (assuming that's your goal).

I'd rather avoid adding more unstable features no matter what tbh.

> Other than micro-benchmarks, but there's no real perf impact of this change. I can trim it down to just the stable if there's any interest in that.

I'm open to that.

plonky2/src/hash/hash_types.rs
```diff
@@ -90,14 +90,11 @@ impl<F: RichField> GenericHashOut<F> for HashOut<F> {
     }
 
     fn from_bytes(bytes: &[u8]) -> Self {
+        // TODO: replace with as_chunks
```
Nashtare, line 93: Same remark as the other TODO

plonky2/src/iop/ext_target.rs
```diff
@@ -150,9 +151,15 @@ pub fn flatten_target<const D: usize>(l: &[ExtensionTarget<D>]) -> Vec<Target> {
 }
 
 /// Batch every D-sized chunks into extension targets.
-pub fn unflatten_target<const D: usize>(l: &[Target]) -> Vec<ExtensionTarget<D>> {
+pub fn unflatten_target<const D: usize>(mut l: &[Target]) -> Vec<ExtensionTarget<D>> {
     debug_assert_eq!(l.len() % D, 0);
-    l.chunks_exact(D)
-        .map(|c| c.to_vec().try_into().unwrap())
-        .collect()
+    // TODO: replace with array_chunks
```
Nashtare, line 156: Same remark

Nashtare (commented):
I also have concerns about the introduction of another dependency, `iter_fixed`, which hasn't been maintained for the past 3 years.

Nashtare:
Closing for inactivity.

PR #1575, opened 2024-04-22
Relax restriction to a single looked table for CTL

matthiasgoergens:
EDIT: it's perhaps better to get https://github.com/0xPolygonZero/plonky2/pull/1577 in first, and then I rebase this PR.

In our use of plonky2, we have a few instances where we need both several looked and several looking tables.  Think of these lookups as many-to-many multiplexers, or a kind of bus.  Looking tables can be thought of as pushing values with an obligation to process them onto the bus, looked tables can be seen as pulling values to discharge that obligation.

My first attempt changed from `looked_table` to `looked_tables`, but then I noticed that thanks to logup, we don't need this complication: looked_tables are just looking_tables with negative multiplicities.

Thus we can get a simpler system by just removing `looked_table` completely, and giving people tools to negate the multiplicities of their lookup tables.

We also preserve the old `CrossTableLookup::new`, which applies the negation to the passed `looked_table` automatically to keep the API as compatible as possible.

This PR only touches cross table lookups.  We could do the same for the intra-table lookups in `starky/src/lookup.rs`.  But that's for a separate PR.

### DIFF
--- a/field/src/types.rs
+++ b/field/src/types.rs
@@ -556,6 +556,16 @@ pub trait PrimeField64: PrimeField + Field64 {
 
     fn to_noncanonical_u64(&self) -> u64;
 
+    /// The conversion to a canonical i64 is mostly useful for debugging,
+    /// as it tries to recover 'negative' values for better readability.
+    fn to_canonical_i64(&self) -> i64 {
+        i64::try_from(self.to_canonical_u64()).unwrap_or_else(|_| {
+            i64::try_from(self.neg().to_canonical_u64())
+                .expect("This conversion should never fail.")
+                .neg()
+        })
+    }
+
     #[inline(always)]
     fn to_canonical(&self) -> Self {
         Self::from_canonical_u64(self.to_canonical_u64())
--- a/field/src/types.rs
+++ b/field/src/types.rs
@@ -559,11 +559,12 @@ pub trait PrimeField64: PrimeField + Field64 {
     /// The conversion to a canonical i64 is mostly useful for debugging,
     /// as it tries to recover 'negative' values for better readability.
     fn to_canonical_i64(&self) -> i64 {
-        i64::try_from(self.to_canonical_u64()).unwrap_or_else(|_| {
-            i64::try_from(self.neg().to_canonical_u64())
-                .expect("This conversion should never fail.")
-                .neg()
-        })
+        let x = self.to_canonical_u64();
+        if x < Self::ORDER / 2 {
+            x as i64
+        } else {
+            x.wrapping_sub(Self::ORDER) as i64
+        }
     }
 
     #[inline(always)]
--- a/starky/src/cross_table_lookup.rs
+++ b/starky/src/cross_table_lookup.rs
@@ -31,7 +31,7 @@
 use alloc::{vec, vec::Vec};
 use core::cmp::min;
 use core::fmt::Debug;
-use core::iter::once;
+use core::ops::Neg;
 
 use anyhow::{ensure, Result};
 use itertools::Itertools;
@@ -80,30 +80,34 @@ impl<F: Field> TableWithColumns<F> {
     }
 }
 
+impl<F: Field> Neg for TableWithColumns<F> {
+    type Output = Self;
+
+    fn neg(self) -> Self {
+        Self::new(self.table, self.columns, -self.filter)
+    }
+}
+
 /// Cross-table lookup data consisting in the lookup table (`looked_table`) and all the tables that look into `looked_table` (`looking_tables`).
 /// Each `looking_table` corresponds to a STARK's table whose rows have been filtered out and whose columns have been through a linear combination (see `eval_table`). The concatenation of those smaller tables should result in the `looked_table`.
 #[derive(Clone, Debug)]
 pub struct CrossTableLookup<F: Field> {
-    /// Column linear combinations for all tables that are looking into the current table.
+    /// Column linear combinations for all tables that are participating in this lookup.
     pub(crate) looking_tables: Vec<TableWithColumns<F>>,
-    /// Column linear combination for the current table.
-    pub(crate) looked_table: TableWithColumns<F>,
 }
 
 impl<F: Field> CrossTableLookup<F> {
     /// Creates a new `CrossTableLookup` given some looking tables and a looked table.
     /// All tables should have the same width.
     pub fn new(
-        looking_tables: Vec<TableWithColumns<F>>,
+        mut looking_tables: Vec<TableWithColumns<F>>,
         looked_table: TableWithColumns<F>,
     ) -> Self {
         assert!(looking_tables
             .iter()
             .all(|twc| twc.columns.len() == looked_table.columns.len()));
-        Self {
-            looking_tables,
-            looked_table,
-        }
+        looking_tables.push(-looked_table);
+        Self { looking_tables }
     }
 
     /// Given a table, returns:
@@ -120,8 +124,11 @@ impl<F: Field> CrossTableLookup<F> {
         let mut num_ctls = 0;
         let mut num_helpers_by_ctl = vec![0; ctls.len()];
         for (i, ctl) in ctls.iter().enumerate() {
-            let all_tables = once(&ctl.looked_table).chain(&ctl.looking_tables);
-            let num_appearances = all_tables.filter(|twc| twc.table == table).count();
+            let num_appearances = ctl
+                .looking_tables
+                .iter()
+                .filter(|twc| twc.table == table)
+                .count();
             let is_helpers = num_appearances > 1;
             if is_helpers {
                 num_helpers_by_ctl[i] = num_appearances.div_ceil(constraint_degree - 1);
@@ -279,10 +286,7 @@ pub(crate) fn num_ctl_helper_columns_by_table<F: Field, const N: usize>(
 ) -> Vec<[usize; N]> {
     let mut res = vec![[0; N]; ctls.len()];
     for (i, ctl) in ctls.iter().enumerate() {
-        let CrossTableLookup {
-            looking_tables,
-            looked_table: _,
-        } = ctl;
+        let CrossTableLookup { looking_tables } = ctl;
         let mut num_by_table = [0; N];
 
         let grouped_lookups = looking_tables.iter().group_by(|&a| a.table);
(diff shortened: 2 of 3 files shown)


starky/src/cross_table_lookup.rs
```diff
@@ -923,17 +866,10 @@ pub fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize,
...
-        index,
-        CrossTableLookup {
-            looking_tables,
-            looked_table,
-        },
-    ) in cross_table_lookups.iter().enumerate()
-    {
+    for (index, CrossTableLookup { looking_tables }) in cross_table_lookups.iter().enumerate() {
         // Get elements looking into `looked_table` that are not associated to any STARK.
         let extra_sum_vec: &[F] = ctl_extra_looking_sums
```
matthiasgoergens, line 871: `ctl_extra_looking_sums` are now indexed the same as `cross_table_lookups`, not like looked tables.

starky/src/cross_table_lookup.rs
```diff
@@ -1041,26 +971,21 @@ pub mod debug_utils {
         }
     }
 
-    fn check_ctl<F: Field>(
+    fn check_ctl<F: PrimeField64>(
```
matthiasgoergens, line 974: I'm not sure these functions are actually in use anywhere?  They don't seem to have been updated to reflect the greater capabilities of logup vs the old product argument.

Nashtare:
> Thank you @matthiasgoergens , it's a good idea!
But I tried running this PR against the integration tests in zk_evm (for example add11_yml), and CTLs and Lookups failed. So I was wondering whether this was also the case for you or if maybe I had messed something up when testing?

This branch is based on an older version of #1577 which was broken (and that @matthiasgoergens now fixed). After rebasing this should go away I believe.

LindaGuiga:
> > Thank you @matthiasgoergens , it's a good idea!
> > But I tried running this PR against the integration tests in zk_evm (for example add11_yml), and CTLs and Lookups failed. So I was wondering whether this was also the case for you or if maybe I had messed something up when testing?
> 
> This branch is based on an older version of #1577 which was broken (and that @matthiasgoergens now fixed). After rebasing this should go away I believe.

Indeed, it works with those changes, thanks!

LindaGuiga (commented):
The changes look great (once it includes the fix from #1577 ), thank you @matthiasgoergens !

matthiasgoergens:
Could you please tell me how to run the zk_evm tests? Is it just 'cargo test' in there (after fixing the dependencies to point at my plonky2 branch)? Or something more involved? Thanks!

Btw, do you have a way to run some benchmarks for zk_evm? I mean, I found some micro-benchmarks in the repo, but I'm looking for something more substantial that covers more of the system (but hopefully doesn't run for longer than a minute or perhaps five at most for one iteration).

That's because I have some ideas for speeding up plonky2 that work for our system, but I want to make sure they don't make life worse for you.

(And chances are, if it helps two unrelated system, it's probably a good idea in general.)

Nashtare:
@matthiasgoergens For testing, yes you can just run `cargo --release --test --package evm_arithmetization` within `zk_evm` repo: https://github.com/0xpolygonzero/zk_evm/.

For benchmarking purposes, we usually would do it against real-life examples, with `zero-bin`: https://github.com/0xpolygonzero/zero-bin/. There you can use the `stdio` mode to pass a block trace as input, and generate an entire proof for the block. If you need an access endpoint / block traces to play with, feel free to reach me in DM.

field/src/types.rs
```diff
@@ -556,6 +556,16 @@ pub trait PrimeField64: PrimeField + Field64 {
 
     fn to_noncanonical_u64(&self) -> u64;
 
+    /// The conversion to a canonical i64 is mostly useful for debugging,
+    /// as it tries to recover 'negative' values for better readability.
+    fn to_canonical_i64(&self) -> i64 {
+        i64::try_from(self.to_canonical_u64()).unwrap_or_else(|_| {
```
Nashtare, line 562: I'm a bit confused about this method. Do we want to interpret integers in the range `[0, M/2[` where `M` is the modulus as being canonical positive integers, and `[M/2, M[` being negative integers?
If so, this is slightly incorrect, as for instance with Goldilocks, the negative range spans from `2^63 - 2^16` to `2^64 - 2^32`, but the first `2^16` elements of the range would appear as positive elements.

starky/src/cross_table_lookup.rs
```diff
@@ -324,12 +328,14 @@ pub(crate) fn cross_table_lookup_data<'a, F: RichField, const D: usize, const N:
...
-    {
-        log::debug!("Processing CTL for {:?}", looked_table.table);
+    for CrossTableLookup { looking_tables } in cross_table_lookups {
+        log::debug!(
+            "Processing CTL for {:?}",
+            looking_tables
+                .iter()
+                .map(|table| table.table)
+                .collect::<Vec<_>>()
+        );
```
Nashtare, line 338: It seems like printing all the involved tables will look pretty heavy compared to the previous display. Especially as we can have multiple looking columns from the same table.

Nashtare: Alternatively, as we always have the `looked` table being pushed last, we could either get back to the previous display as 
```suggestion
        log::debug!(
            "Processing CTL for table {:?}",
            looking_tables
                .last()
                .expect("We should always have tables")
                .table
        );
```

or have a more complete one also listing the `looking` ones (although from my point above, I find it meaningless to see that table `N` appears 50 times, if we can't print more specific information as to what we are looking for).

starky/src/cross_table_lookup.rs
```diff
@@ -997,18 +928,17 @@ pub fn verify_cross_table_lookups_circuit<
...
 
-            looking_zs_sum = builder.add(looking_zs_sum, extra_sum_vec[c]);
+            let looking_zs_sum = builder.add(looking_zs_sum, extra_sum_vec[c]);
+            let zero = builder.zero();
 
-            // Get the looked table CTL polynomial opening.
-            let looked_z = *ctl_zs_openings[looked_table.table].next().unwrap();
             // Verify that the combination of looking table openings is equal to the looked table opening.
-            builder.connect(looked_z, looking_zs_sum);
+            builder.connect(zero, looking_zs_sum);
```
Nashtare, line 941: nit: `CircuitBuilder` implements an `assert_zero` method, possibly clearer.

field/src/types.rs
matthiasgoergens: I mostly just want 'small' negative numbers like -1, or -100 to survive a round-trip into field elements and back.

And I want the result to be independent of whether the field element has had `to_canonical` applied to it or not.

So the implied semantics are basically: put the canonical representation of the number into a non-negative `i64`,  unless that doesn't fit; in that case, give me the negative `i64`.

However you are right that splitting the range somewhere around `M/2` instead of `i64::MAX` might be less confusing and easier to explain and would also give a reasonable answers (that survives round-tripping in the sense mentioned above) for fields with `M < i64::MAX`.

starky/src/cross_table_lookup.rs
matthiasgoergens: So in our use of CTL, we often don't have a single table that's somehow special that it would make to single out and sense to print here.

If you are afraid of having one table appear 50 times, we can consolidate repeated use of the same table and replace that with a count (as a number).

field/src/types.rs
Nashtare: Sorry I had missed it. Yes I'd tend to prefer the `M/2` route for clarity.

starky/src/cross_table_lookup.rs
Nashtare: Yeah, at least counting the occurences instead of printing all tables would be better (in our case we do have some CTLs with a fairly large number of looking tables).

> So in our use of CTL, we often don't have a single table that's somehow special

Fair, for logging it isn't super useful compared to the debugging function.

Nashtare:
@matthiasgoergens Could you address the few comments / rebase with latest `main` so we can get this into the next release?

matthiasgoergens:
> @matthiasgoergens Could you address the few comments / rebase with latest `main` so we can get this into the next release?

Please see the new version.

starky/src/cross_table_lookup.rs
matthiasgoergens: I'm keeping counts now.

starky/src/cross_table_lookup.rs
```diff
@@ -327,12 +331,17 @@ pub(crate) fn cross_table_lookup_data<'a, F: RichField, const D: usize, const N:
...
-        log::debug!("Processing CTL for {:?}", looked_table.table);
+    for CrossTableLookup { looking_tables } in cross_table_lookups {
+        log::debug!(
+            "Processing CTL for {:?}",
+            looking_tables
+                .iter()
+                .map(|table| table.table)
+                .counts()
+                .into_iter()
+                .sorted()
```
matthiasgoergens, line 342: We sort to keep the debug output deterministic.

starky/src/cross_table_lookup.rs
```diff
@@ -327,12 +331,17 @@ pub(crate) fn cross_table_lookup_data<'a, F: RichField, const D: usize, const N:
...
+                .collect::<Vec<_>>()
+        );
```
Nashtare, line 344: At this point I think I may prefer just specifying for which table the CTL is, i.e. 

```suggestion
        log::debug!(
            "Processing CTL for STARK table {:?}",
            looking_tables
                .last().
                .expect("We always have a looked table.")
                .table
        );
```

or otherwise removing it altogether. The current display doesn't carry much meaning (especially with the sorting which kind of hides which table is being looked at. For example this is our zkEVM output with this implem:

```bash
[DEBUG starky::cross_table_lookup] Processing CTL for [(0, 1), (2, 1)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(1, 1), (2, 4)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(2, 1), (4, 1)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(3, 1), (4, 1)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(3, 1), (4, 1)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(2, 1), (4, 5), (5, 1)]
[DEBUG starky::cross_table_lookup] Processing CTL for [(1, 32), (2, 7), (4, 136), (6, 1)]
```

gio256:
@matthiasgoergens thanks for this PR. 

Can I be of any help getting this across the line? Happy to take it from here or just get it caught up to `main`.

Nashtare:
Closing for inactivity.
@gio256 feel free to pick it up if you want to / have some free time.

Issue #1514, opened 2024-02-10
Increase testing in starky crate


Nashtare:
The starky crate doesn't have much unit tests, and would benefit from a stronger test suite covering the different setups that can be used.

shreyas-londhe:
Hi @Nashtare, I'm interested in working on this, can you please guide me on what needs to be done?

Nashtare:
Hey @shreyas-londhe! Thanks for your interest in this!
It's mostly about testing the cross-functionalities between different scenarios at the moment. We currently support testing simpler STARKs with:
- only regular constraints
- no regular constraints but a permutation argument

We do lack of CTLs test strentghening, which involve defining some simple system across 2 or 3 tables operating on common values. We'd probably need some consistency checks around challenge computations and quotient polynomial evaluation for badly used parameters cases (like not specifying a `requires_ctls` for a table in a multi-STARK for instance).

We'd probably also need some higher-degree test cases for STARKs, valid & invalid cases ideally. But not everything has to be done at once, this can be a long-running task :) Let me know if you have any additional question!

Nashtare:
@shreyas-londhe Do you need help on this?

PR #1577, opened 2024-04-22
Remove restriction to binary-only multiplicities

matthiasgoergens:
This is a simpler PR to prepare the ground for https://github.com/0xPolygonZero/plonky2/pull/1575

### DIFF
--- a/starky/src/lookup.rs
+++ b/starky/src/lookup.rs
@@ -9,7 +9,7 @@ use core::iter::repeat;
 
 use itertools::Itertools;
 use num_bigint::BigUint;
-use plonky2::field::batch_util::batch_add_inplace;
+use plonky2::field::batch_util::{batch_add_inplace, batch_multiply_inplace};
 use plonky2::field::extension::{Extendable, FieldExtension};
 use plonky2::field::packed::PackedField;
 use plonky2::field::polynomial::PolynomialValues;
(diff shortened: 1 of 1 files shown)


matthiasgoergens:
Btw, why does starky lack tests?

Nashtare:
> Btw, why does starky lack tests?

We have an open issue for it (#1514). All tests were specific to the zkEVM and migrated to the new repo. I agree this definitely needs some more testing. I have pinged the person who assigned themselves this task but no answer yet.

Nashtare:
Related to testing, have you tested this? Haven't reviewed yet but it's failing on our integration tests there https://github.com/0xPolygonZero/zk_evm/tree/develop.

matthiasgoergens:
I ran all the tests in the plonky2 repository.  I'm still running tests elsewhere.  Thanks for pointing to your other repository, I'll try to run those tests, too.

matthiasgoergens:
Here's your fix: https://github.com/0xPolygonZero/zk_evm/pull/183

Nashtare:
I already have a branch working with main. But moving deps to your current branch gives me an quotient polynomial evaluation error.

matthiasgoergens:
> I already have a branch working with main. But moving deps to your current branch gives me an quotient polynomial evaluation error.

Thanks.  I found the bug.

I cleaned up the code a bit while fixing the bug.

starky/src/lookup.rs
```diff
@@ -761,79 +761,35 @@ pub(crate) fn get_helper_cols<F: Field>(
...
-    let mut helper_columns = Vec::with_capacity(num_helper_columns);
-
-    for mut cols_filts in &columns_filters
-        .iter()
-        .chunks(constraint_degree.checked_sub(1).unwrap_or(1))
-    {
-        let (first_col, first_filter) = cols_filts.next().unwrap();
-
-        let mut filter_col = Vec::with_capacity(degree);
-        let first_combined = (0..degree)
```
matthiasgoergens, line 773: I removed the duplication of logic here, and replaced it with a single call to `reduce` further down.

starky/src/lookup.rs
```diff
@@ -761,79 +761,35 @@ pub(crate) fn get_helper_cols<F: Field>(
         .len()
         .div_ceil(constraint_degree.checked_sub(1).unwrap_or(1));
 
-    let mut helper_columns = Vec::with_capacity(num_helper_columns);
```
matthiasgoergens, line 764: Iterators in Rust have a [size hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint), so `collect` automatically does what we were trying to do manually via `with_capacity` here.

Nashtare:
> > I already have a branch working with main. But moving deps to your current branch gives me an quotient polynomial evaluation error.
> 
> Thanks. I found the bug.
> 
> I cleaned up the code a bit while fixing the bug.
> 
> ```
>  starky/src/lookup.rs | 84 ++++++++++++++++++++----------------------------------------------------------------
>  1 file changed, 20 insertions(+), 64 deletions(-)
> ```

For things touching `starky`, I usually run them against our `zk_evm` monorepo as sanity check, until we get some good coverage here. Priority wise it's a bit tough for us to dedicate time for it at the moment,  but it will be done sooner or later. In the meantime, as long as myself and other maintainers are testing against `zk_evm`, we should be fine.

starky/src/lookup.rs
```diff
@@ -761,79 +761,35 @@ pub(crate) fn get_helper_cols<F: Field>(
...
-                            .iter()
-                            .map(|c| c.eval_table(trace, d))
-                            .collect::<Vec<F>>();
-                        challenge.combine(evals.iter())
-                    } else {
-                        assert_eq!(f, F::ZERO, "Non-binary filter?");
-                        // Dummy value. Cannot be zero since it will be batch-inverted.
-                        F::ONE
-                    }
+                    let mut combined = F::batch_multiplicative_inverse(&combined);
```
LindaGuiga, line 780: If I remember correctly, the reason why the previous code did not use `batch_multiplicative_inverse` (and was therefore more complicated). was because the filter column is often (not always!) sparse, so we wanted to avoid inverting a lot of values that would be later zeroed out. However, I am not sure how much of an impact it would actually have, and this code does look much nicer!

matthiasgoergens: Yes, though the code in main already uses batch multiplicative inverse.

muursh (approved):
I'm good with this

PR #1579, opened 2024-04-22
update 2-adic generator to `0x64fdd1a46201e246`

benediamond:
- this is a first step towards addressing https://github.com/0xPolygonZero/plonky2/issues/850. note that `0x64fdd1a46201e246^(2^30) == 0x1000000000000` holds modulo p, as desired. we don't actually change the FFT in this PR; but this paves the way for that as a future change.
- both [Hermez](https://github.com/0xPolygonHermez/goldilocks/blob/f89eb016830f8c4301482d83691aed22e5a92742/src/goldilocks_base_field.cpp#L44) and [Miden](https://github.com/facebook/winterfell/blob/09525751727a283dfbf5e569a09909b82380e059/math/src/field/f64/mod.rs#L282) already currently use this new generator.

upon changing the main 2^32nd root of unity of Goldilocks, a few auxiliary values also need to change. these include:
- the multiplicative generator [here](https://github.com/benediamond/plonky2/blob/5055681898f327365e59e9b91393dd67192abc2b/field/src/goldilocks_field.rs#L80). this is a generator of Goldilocks' _entire_ multiplicative group of units, which moreover "lifts" the main 2-adic generator above in the sense that `MULTIPLICATIVE_GROUP_GENERATOR^((p - 1) / 2^32) == POWER_OF_TWO_GENERATOR` holds. this isn't too hard to construct; essentially, it amounts to finding a multiplicative generator of $\mathbb{F}_p^*$'s order-(p - 1) / 2^32 subgroup, which is not hard.
- in the extension fields, we need to do something analogous, but for 2-adic generators. for example, in the quadratic extension of Goldilocks given by `X^2 - 7`, we now get 2^33-order 2-adicity; the value [`EXT_POWER_OF_TWO_GENERATOR`](https://github.com/benediamond/plonky2/blob/5055681898f327365e59e9b91393dd67192abc2b/field/src/goldilocks_extensions.rs#L27) needs to be a primitive 2^33nd root of unity which _moreover_ lifts `POWER_OF_TWO_GENERATOR`, in the sense that `EXT_POWER_OF_TWO_GENERATOR^2 = POWER_OF_TWO_GENERATOR`. this is a bit tricky to do; we were able to construct this efficiently using essentially a Pohlig–Hellman-type idea.
- Finally, this 2-adic generator itself also needs to be lifted to a full multiplicative generator `EXT_MULTIPLICATIVE_GROUP_GENERATOR` of $\mathbb{F}_{p^2}^*$, i.e., so that `EXT_MULTIPLICATIVE_GROUP_GENERATOR^((p^2 - 1) / 2^33) == EXT_POWER_OF_TWO_GENERATOR` holds.

we also need to do something analogous for the quartic and quintic extensions.

### DIFF
--- a/field/src/goldilocks_extensions.rs
+++ b/field/src/goldilocks_extensions.rs
@@ -21,8 +21,7 @@ impl Extendable<2> for GoldilocksField {
     // DTH_ROOT = W^((ORDER - 1)/2)
     const DTH_ROOT: Self = Self(18446744069414584320);
 
-    const EXT_MULTIPLICATIVE_GROUP_GENERATOR: [Self; 2] =
-        [Self(0), Self(11713931119993638672)];
+    const EXT_MULTIPLICATIVE_GROUP_GENERATOR: [Self; 2] = [Self(0), Self(11713931119993638672)];
 
     const EXT_POWER_OF_TWO_GENERATOR: [Self; 2] = [Self(0), Self(7226896044987257365)];
 }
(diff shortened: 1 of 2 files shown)


Nashtare (approved):
LGTM, but I'd hold off merging this into `main` until we have reached a decision regarding version bump. As @dlubarov pointed out we may need to bump the `major` version for once, as this would be a hard proof backwards incompatibility.

muursh:
Yeah I'd agree that we'd want a major version bump for this

muursh (approved):

PR #1584, opened 2024-04-25
Fix `verify_cross_table_lookups` with no `ctl_extra_looking_sums`

matthiasgoergens:
This used to fail with an out of bounds error when `None` was passed for `ctl_extra_looking_sums`.

### DIFF
--- a/starky/src/cross_table_lookup.rs
+++ b/starky/src/cross_table_lookup.rs
@@ -931,10 +931,6 @@ pub fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize, 
         },
     ) in cross_table_lookups.iter().enumerate()
     {
-        // Get elements looking into `looked_table` that are not associated to any STARK.
-        let extra_sum_vec: &[F] = ctl_extra_looking_sums
-            .map(|v| v[looked_table.table].as_ref())
-            .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


starky/src/cross_table_lookup.rs
```diff
@@ -949,7 +945,9 @@ pub fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize,
                 .iter()
                 .map(|&table| *ctl_zs_openings[table].next().unwrap())
                 .sum::<F>()
-                + extra_sum_vec[c];
+                // Get elements looking into `looked_table` that are not associated to any STARK.
+                + ctl_extra_looking_sums
+                .map(|v| v[index][c]).unwrap_or_default();
```
LindaGuiga, line 950: Should this be:
```suggestion
                .map(|v| v[looked_table.table][c]).unwrap_or_default();
```
as in the circuit version? Or am I misunderstanding something?

LindaGuiga (commented):
Thank you for fixing this, @matthiasgoergens !

starky/src/cross_table_lookup.rs
matthiasgoergens: Yes, I think you are right.

Nashtare (approved):

PR #1585, opened 2024-04-25
Fix CTL generation of last row

matthiasgoergens:
Not all `correct` cross table lookups use padding for the last row.

### DIFF
--- a/starky/src/lookup.rs
+++ b/starky/src/lookup.rs
@@ -321,24 +321,16 @@ impl<F: Field> Column<F> {
 
     /// Evaluate on a row of a table given in column-major form.
     pub(crate) fn eval_table(&self, table: &[PolynomialValues<F>], row: usize) -> F {
-        let mut res = self
-            .linear_combination
+        self.linear_combination
             .iter()
             .map(|&(c, f)| table[c].values[row] * f)
             .sum::<F>()
-            + self.constant;
-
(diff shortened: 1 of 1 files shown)


starky/src/lookup.rs
```diff
@@ -321,24 +321,16 @@ impl<F: Field> Column<F> {
...
-            res += self
+            + self
                 .next_row_linear_combination
                 .iter()
-                .map(|&(c, f)| table[c].values[row + 1] * f)
-                .sum::<F>();
-        }
-
-        res
+                .map(|&(c, f)| table[c].values[(row + 1) % table[c].values.len()] * f)
```
matthiasgoergens, line 331: If you are worried about performance, you could also do `&  (table[c].values.len() - 1)`, taking advantage of the fact that we only work with powers of two.  But this is confusing, and modern processors are fast.

---

On a wider scale, I thought of doing linear combinations on whole columns at once via `batch_add_inplace` etc, but it's not really worth it, because CTL data generation is such a small part of overall runtime.

Nashtare: > If you are worried about performance, you could also do &  (table[c].values.len() - 1), taking advantage of the fact that we only work with powers of two.

I don't mind having it (with an eventual comment for clarity), but this isn't a bottleneck anyway so probably negligible as you mentioned.

Nashtare (approved):

muursh (approved):

PR #1586, opened 2024-04-26
Optimize reduce_polys_base

Daniel-Aaron-Bloom:
Updated version of #1034, because apparently removing the collect (as #1571 tried to) significantly worsens performance.


Nashtare (commented):
Unfortunately I can't seem to find any performance improvement, it actually worsens perfs further when bringing back the collect. Below is a comparison between `main`, #1571 and #1586.

Screenshot 2024-04-26 at 3 42 37 PM

Daniel-Aaron-Bloom:
Bummer.

On my end, `cargo run --release --example bench_recursion -- -vv --size=18` seems to show no real difference between any of them here so I don't disagree.

Feel free to close all related PRs.

PR #1587, opened 2024-05-03
Changes to prepare for dummy segment removal in zk_evm's continuations

LindaGuiga:
This PR makes some methods public so they can be used on the `zk_evm` side, and creates a couple of new methods.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -532,7 +532,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
     }
 
     /// If `condition`, enforces that two routable `Target` values are equal, using Plonk's permutation argument.
-    pub fn assert_equal_if(&mut self, condition: Target, x: Target, y: Target) {
+    pub fn conditional_assert_eq(&mut self, condition: Target, x: Target, y: Target) {
         let zero = self.zero();
         let diff = self.sub(x, y);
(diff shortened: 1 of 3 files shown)


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -531,6 +531,14 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
         }
     }
 
+    /// If `condition`, enforces that two routable `Target` values are equal, using Plonk's permutation argument.
+    pub fn assert_equal_if(&mut self, condition: Target, x: Target, y: Target) {
```
Nashtare, line 535: nit: maybe `conditional_assert_eq` would be preferable? No strong opinion though

plonky2/src/recursion/dummy_circuit.rs
```diff
@@ -121,7 +113,7 @@ pub(crate) fn dummy_circuit<
 }
 
 impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
-    pub(crate) fn dummy_proof_and_vk<C: GenericConfig<D, F = F> + 'static>(
+    pub fn dummy_proof_and_vk<C: GenericConfig<D, F = F> + 'static>(
```
Nashtare, line 116: seems not needed anymore, so no need to expose

Nashtare (approved):
LGTM!

PR #1591, opened 2024-05-20
Allow multiple `extra_looking_sums` for the same looked table

qope:
Currently, `ctl_extra_looking_sums` corresponds to the index of the looked table. The problem with this is that it is not possible to set another `extra_looking_sums` for the same looked table. For example, in zk_evm's `keccak_stark`, there are two CTLs  for the same looked table, and therefore, it is not possible to use those two CTLs as `extra_looking_sums`. I addressed this issue by changing the type of `ctl_extra_looking_sums` to a `HashMap` similar to `check_ctl`, and making it correspond one-to-one with the index of the CTL.

### DIFF
--- a/starky/src/cross_table_lookup.rs
+++ b/starky/src/cross_table_lookup.rs
@@ -34,6 +34,7 @@ use core::fmt::Debug;
 use core::iter::once;
 
 use anyhow::{ensure, Result};
+use hashbrown::HashMap;
 use itertools::Itertools;
 use plonky2::field::extension::{Extendable, FieldExtension};
 use plonky2::field::packed::PackedField;
@@ -916,10 +917,11 @@ pub(crate) fn eval_cross_table_lookup_checks_circuit<
 }
 
 /// Verifies all cross-table lookups.
+/// The key of `ctl_extra_looking_sums` is the corresponding CTL's position within `cross_table_lookups`.
 pub fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize, const N: usize>(
     cross_table_lookups: &[CrossTableLookup<F>],
     ctl_zs_first: [Vec<F>; N],
-    ctl_extra_looking_sums: Option<&[Vec<F>]>,
+    ctl_extra_looking_sums: &HashMap<usize, Vec<F>>,
     config: &StarkConfig,
 ) -> Result<()> {
     let mut ctl_zs_openings = ctl_zs_first.iter().map(|v| v.iter()).collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


starky/src/cross_table_lookup.rs
```diff
@@ -1029,10 +1034,11 @@ pub mod debug_utils {
     type MultiSet<F> = HashMap<Vec<F>, Vec<(TableIdx, usize)>>;
 
     /// Check that the provided traces and cross-table lookups are consistent.
+    /// The key of `extra_looking_values` is the corresponding CTL's position within `cross_table_lookups`.
     pub fn check_ctls<F: Field>(
         trace_poly_values: &[Vec<PolynomialValues<F>>],
         cross_table_lookups: &[CrossTableLookup<F>],
-        extra_looking_values: &HashMap<TableIdx, Vec<Vec<F>>>,
+        extra_looking_values: &HashMap<usize, Vec<Vec<F>>>,
```
Nashtare, line 1041: There's no need to change this.

starky/src/cross_table_lookup.rs
```diff
@@ -916,10 +917,11 @@ pub(crate) fn eval_cross_table_lookup_checks_circuit<
 }
 
 /// Verifies all cross-table lookups.
+/// The key of `ctl_extra_looking_sums` is the corresponding CTL's position within `cross_table_lookups`.
 pub fn verify_cross_table_lookups<F: RichField + Extendable<D>, const D: usize, const N: usize>(
     cross_table_lookups: &[CrossTableLookup<F>],
     ctl_zs_first: [Vec<F>; N],
-    ctl_extra_looking_sums: Option<&[Vec<F>]>,
+    ctl_extra_looking_sums: &HashMap<usize, Vec<F>>,
```
Nashtare, line 924: As we now refer to table indices, please use the associated type `TableIdx` for consistency. It's an alias for `usize`.

starky/src/cross_table_lookup.rs
```diff
@@ -973,15 +976,19 @@ pub fn verify_cross_table_lookups_circuit<
     builder: &mut CircuitBuilder<F, D>,
     cross_table_lookups: Vec<CrossTableLookup<F>>,
     ctl_zs_first: [Vec<Target>; N],
-    ctl_extra_looking_sums: Option<&[Vec<Target>]>,
+    ctl_extra_looking_sums: &HashMap<usize, Vec<Target>>,
```
Nashtare, line 979: Similar comment here.

starky/src/cross_table_lookup.rs
hratoanina: The issue is that we're not indexing by Table anymore, but by CTL.

Nashtare (approved):

PR #1592, opened 2024-05-21
change `set_stark_proof_target`'s witness to `WitnessWrite`

qope:
The witness used in `set_stark_proof_target` is sufficient with `WitnessWrite`. `Witness` (which inherits from `WitnessWrite`) cannot be used in cases like `run_once` of `SimpleGenerator`.

### DIFF
--- a/starky/src/recursive_verifier.rs
+++ b/starky/src/recursive_verifier.rs
@@ -14,7 +14,7 @@ use plonky2::hash::hash_types::RichField;
 use plonky2::iop::challenger::RecursiveChallenger;
 use plonky2::iop::ext_target::ExtensionTarget;
 use plonky2::iop::target::Target;
-use plonky2::iop::witness::Witness;
+use plonky2::iop::witness::WitnessWrite;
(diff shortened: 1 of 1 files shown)


Nashtare (commented):
Hi! Could you share a bit more context / a reproducible example of failure? `SimpleGenerator` implements both `WitnessWrite` and `Witness`, it is unclear to me why this change is necessary.

qope:
I am writing a generator that automatically generates and verifies a starky proof. I need to set the starky proof in the `out_buffer`, but `GeneratedValues<F>` only implements `WitnessWrite<F>` and does not implement `Witness<F>`. 
The link below shows the usage of Starky after applying this PR:
https://github.com/InternetMaximalism/plonky2_keccak/blob/5aecea14a6325f21b57e22e072fd2051bffc0607/src/generators/stark_proof_generator.rs#L155-L251

Even aside from the above issue, I believe it is desirable to change `set_stark_proof_target` to use `WitnessWrite<F>` instead of `Witness<F>` since it does not utilize the `Witness<F>` trait.

Nashtare (approved):

PR #1599, opened 2024-06-11
Add `Field::shifted_powers` and some iterator niceties

gio256:
Adds the `Field::shifted_powers` method from [Plonky3](https://github.com/Plonky3/Plonky3/blob/fde81db95b8eeb39e07890f4099a7d59daeec52f/field/src/field.rs#L122-L127) along with a few additions to the `Powers` iterator implementation.

Please pick and choose what's useful.

`<Powers<F> as Iterator>::nth` in particular might require some tweaking:
- It currently panics if `n` overflows `u64`. This seems at least as good as the current behavior of hanging trying to compute `u64::MAX` field multiplications. But, maybe it's worth the performance hit to use `Field::exp_biguint`.
- The current implementation incurs one extra multiplication in the case of `n = 1` (e.g. `generator.powers().skip(1)`). It's possible this is the most commonly expected case and `nth` should just be removed altogether.

### DIFF
--- a/field/src/types.rs
+++ b/field/src/types.rs
@@ -427,9 +427,13 @@ pub trait Field:
     }
 
     fn powers(&self) -> Powers<Self> {
+        self.shifted_powers(Self::ONE)
+    }
+
+    fn shifted_powers(&self, start: Self) -> Powers<Self> {
         Powers {
             base: *self,
-            current: Self::ONE,
+            current: start,
         }
     }
 
@@ -563,6 +567,11 @@ pub trait PrimeField64: PrimeField + Field64 {
 }
 
 /// An iterator over the powers of a certain base element `b`: `b^0, b^1, b^2, ...`.
+///
+/// This iterator panics on calls to [`nth`] if `n` overflows [`u64`].
+///
+/// [`nth`]: ::core::iter::Iterator::nth
+#[must_use = "iterators are lazy and do nothing unless consumed"]
 #[derive(Clone, Debug)]
 pub struct Powers<F: Field> {
     base: F,
@@ -577,6 +586,24 @@ impl<F: Field> Iterator for Powers<F> {
         self.current *= self.base;
         Some(result)
     }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        (usize::MAX, None)
+    }
+
+    fn nth(&mut self, n: usize) -> Option<F> {
+        let result = self.current * self.base.exp_u64(n.try_into().unwrap());
+        self.current = result * self.base;
+        Some(result)
+    }
+
+    fn last(self) -> Option<F> {
+        panic!("called `Iterator::last()` on an infinite sequence")
+    }
+
+    fn count(self) -> usize {
+        panic!("called `Iterator::count()` on an infinite sequence")
+    }
 }
 
 impl<F: Field> Powers<F> {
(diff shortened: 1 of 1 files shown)


field/src/types.rs
```diff
@@ -563,6 +567,11 @@ pub trait PrimeField64: PrimeField + Field64 {
 }
 
 /// An iterator over the powers of a certain base element `b`: `b^0, b^1, b^2, ...`.
+///
+/// This iterator panics on calls to [`nth`] if `n` overflows [`u64`].
```
wborgeaud, line 571: I'm not sure I understand this comment. The argument of `nth` is `usize`, so how can it overflow `u64`?

wborgeaud (commented):
> It's possible this is the most commonly expected case and nth should just be removed altogether

Why do you think this is the most commonly expected case? In general, I'm not sure where `nth` could be useful. Do you have use cases in mind?

gio256:
The reason for the PR as a whole was to hopefully make it easier to replace as many instances of `F::from_canonical_usize(1 << n)` as possible (for safety and readability reasons). Originally, I thought `nth` would be useful for bit shifts for example.

```rust
// n is the (hypothetical) number of bits to shift by.
// f_n is a flag indicating whether this is really the n to shift by.
// All bits in the output with index < n are 0.
// m in {n..32} is the LE index of a possibly nonzero bit in the output.
// We want to line up the nth bit of the output with the 0th bit of the input,
// so we take `output_bits[m] = input_bits[m - n]`.
let shift_left: P = lv
    .shift_by_indices
    .into_iter()
    .enumerate()
    .flat_map(|(n, f_n)| {
        (n..32)
            .zip(P::Scalar::TWO.powers().skip(n))
            .map(move |(m, base)| f_n * lv.input_bits[m - n] * base)
    })
    .sum();
```

But, this is more efficiently done with `Field::shifted_powers` anyway. `nth` is used under the hood by iterator adapters like `skip` and `step_by`, so if those seem useful in the context of `powers()` than maybe it's worth keeping. Otherwise, I'm happy to remove it.

field/src/types.rs
gio256: You're right that this is probably an unnecessary comment. If we decide to keep `nth` I'll get rid of the comment. If you want to get really into the pointer-sized weeds, [this](https://faultlore.com/blah/fix-rust-pointers/#distinguish-pointers-and-addresses) is a great read.

field/src/types.rs
```diff
@@ -592,3 +619,26 @@ impl<F: Field> Powers<F> {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::Field;
+    use crate::goldilocks_field::GoldilocksField;
+
+    #[test]
+    fn test_powers_nth() {
```
gio256, line 629: I've just realized that none of the unit tests in `plonky2_field` are run in the CI. Is there a better place for this test?

Nashtare: This is actually that _shouldn't_ be the case. Thanks for catching it, will update the CI jobs right away.

gio256: I might be misunderstanding the intent of all the config attributes here, but this line looks a bit suspicious to me.
https://github.com/0xPolygonZero/plonky2/blob/42a821dfde351e8579da6e8addbb440a30203f3a/field/src/lib.rs#L9

gio256:
Thinking about this more, I gave a misleading example as using `Field::shifted_powers` in the code above doesn't actually accomplish anything.

```rust
let shift_left: P = lv
    .shift_by_indices
    .into_iter()
    .enumerate()
    .flat_map(|(n, f_n)| {
        (n..32)
            .zip(P::Scalar::TWO.powers_shifted(P::Scalar::from_canonical_u32(1 << n)))
            .map(move |(m, base)| f_n * lv.input_bits[m - n] * base)
    })
    .sum();
```
It's a small thing, but this still requires verifying that at no point in the loop does `1 << n` exceed `u32::MAX` or the order of the field. For that reason, I think the version using `Iterator::skip` is slightly better if the overhead is negligible. So, I'll update my (soft) pitch for adding `nth` to the Iterator implementation:

- It makes the overhead of using `Field::powers` more intuitive and predictable, in that advancing the iterator by some large `n` costs one exponentiation rather than `n` muls. This enables the use of `Field::powers` to (efficiently) replace a mildly dangerous pattern.

- On the other hand, this means using exponentiation even if `n = 1` (e.g. `generator.powers().skip(1)`). It's also far from obvious that `P::Scalar::TWO.powers().skip(n)` translates to `(2^n, 2^n+1, ...)`, so you could argue that even this use case is contrived.

matthiasgoergens:
Nice, I miss `shifted_powers`!

gio256:
I merged upstream so the tests would run, but let me know if I should squash.

Otherwise, this should be ready to go. Thanks for the feedback.

Nashtare (approved):
LGTM too! 

> I merged upstream (https://github.com/0xPolygonZero/plonky2/pull/1601) so the tests would run, but let me know if I should squash.

No problem about this, PRs are squashed into `main` anyway!

PR #1604, opened 2024-06-26
ci: add PR check job

Nashtare:
cf internal discussion

cc @Freyskeyd, thank you for the template!

### DIFF
--- a/.github/pr_checking.yml
+++ b/.github/pr_checking.yml
@@ -1,0 +1,22 @@
+name: PR check
+
+on:
+  pull_request_target:
+    types:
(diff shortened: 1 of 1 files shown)


muursh (approved):

.github/pr_checking.yml
Freyskeyd, line 1: This file isn't at the right place, it should be in `.github/workflows/pr_checking.yml`

Nashtare: 🤦🏻‍♂️
My bad, will fix it.

PR #1614, opened 2024-08-14
Fix example and documentation rendering

simon-perriard:
List of changes:

- fixed the example in the documentation of `starky/lib.rs` and make it runnable
- bumped `katex` version to the latest available (`v0.16.11`)
- added `config.toml` in the different `.cargo` to automate KaTeX rendering
- updated README 
- tentatively added `--no-deps` flag for documentation rendering on https://docs.rs/, honestly don't know if it will work. Hopefully will render with KaTeX and fix the math rendering of pages like https://docs.rs/plonky2/latest/plonky2/gates/coset_interpolation/struct.CosetInterpolationGate.html


.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustdocflags = ["--html-in-header", ".cargo/katex-header.html"]
```
Nashtare, line 2: nit: could you add a whitespace here and in other similar files?

Nashtare (approved):
Thank you @simon-perriard for addressing this, and apologies for the delay. I've verified the display, and the rendering for the `CosetInterpolationGate` indeed looks much better, thanks!

muursh (approved):

plonky2/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustdocflags = ["--html-in-header", ".cargo/katex-header.html"]
```
muursh, line 2:
```suggestion
rustdocflags = ["--html-in-header", ".cargo/katex-header.html"]

```

starky/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+rustdocflags = ["--html-in-header", ".cargo/katex-header.html"]
```
muursh, line 2:
```suggestion
rustdocflags = ["--html-in-header", ".cargo/katex-header.html"]

```

PR #1620, opened 2024-08-27
Add `connect_array` convenience method in `CircuitBuilder`

Nashtare:
Mostly for reuse on zk_evm side where we heavily use arrays of targets.

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -524,7 +524,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
             .push(CopyConstraint::new((x, y), self.context_log.open_stack()));
     }
 
-    /// Enforces that two [`Target`] arrays are equal.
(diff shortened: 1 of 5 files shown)


plonky2/src/plonk/circuit_builder.rs
```diff
@@ -524,6 +524,13 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
             .push(CopyConstraint::new((x, y), self.context_log.open_stack()));
     }
 
+    /// Enforces that two [`Target`] arrays are equal.
```
muursh, line 527: I'm not convinced equal here is especially clear.

muursh (approved):

plonky2/src/plonk/circuit_builder.rs
Nashtare: should I precise `underlying values` perhaps?

muursh: Yeah that's more clear

PR #1628, opened 2024-09-25
add timing percentage

doutv:
Fix https://github.com/0xPolygonZero/plonky2/issues/1627


Nashtare:
Hi! Thanks for this and the related issue. We'd actually be interested in getting rid of the timing tree and rely on tracing, similarly to how plonky3 does. Would it be something you're interested in working on?
Ideally we would have it included in our next release incoming relatively soon.

doutv:
That would be great! I don't have enough time to replace timing tree with tracing haha.

PR #1629, opened 2024-09-30
Refactor CTL Handling

sai-deng:
This PR refactors the `get_ctl_vars_from_proofs` function to sync with the circuit version.


starky/src/proof.rs
```diff
@@ -186,31 +186,6 @@ where
...
-impl<F: RichField + Extendable<D>, C: GenericConfig<D, F = F>, const D: usize, const N: usize>
-    MultiProof<F, C, D, N>
-{
-    /// Returns the degree (i.e. the trace length) of each STARK proof,
-    /// from their common [`StarkConfig`].
-    pub fn recover_degree_bits(&self, config: &StarkConfig) -> [usize; N] {
-        core::array::from_fn(|i| self.stark_proofs[i].proof.recover_degree_bits(config))
-    }
-}
-
```
Nashtare, line 213: Are we sure we want to remove this from the API? It's a relatively convenient struct to group the whole system together. I'd rather suggest expanding it by implementing some additional helpers if needed (like extracting the aux_polys needed for the new signature above).

Nashtare (approved):
Non-blocking comments, though I'm not sure about the full removal of `MultiProof`

starky/src/proof.rs
sai-deng: I removed them simply because they are unused in the Plonky2 library. In our zkVM implementation, I changed:
`pub stark_proofs: [StarkProofWithMetadata<F, C, D>; N],`
to
`pub stark_proofs: [Option<StarkProofWithMetadata<F, C, D>>; N],`
However, it wouldn’t make sense to require every project that depends on Plonky2 to make the same changes.

sai-deng:
@Nashtare I also updated the `from_proof` function to sync with the circuit version. PTAL.

PR #1630, opened 2024-10-02
Added serialize and deserialize to starky proofs

armanthepythonguy:
Added serialize and deserialize to starky proofs. We needed this for our ZKML project where we needed to export the proofs into files and read them back to verify. It might be useful for other projects as well !!!


starky/Cargo.toml
```diff
@@ -23,6 +23,7 @@ anyhow = { workspace = true }
 hashbrown = { workspace = true }
 itertools = { workspace = true }
 log = { workspace = true }
+serde = { workspace = true, features = ["rc"] }
```
Nashtare, line 26: This is the same import in `plonky2`, it'd be best to import it at the workspace level and then

`serde = { workspace = true }`

Nashtare (commented):
Thanks, just a minor nit!

Nashtare (approved):

muursh (approved):

muursh:
@armanthepythonguy can you fix the clippy issue then we'll merge

Nashtare:
@armanthepythonguy you need latest nightly toolchain for this to be triggered

PR #1640, opened 2024-11-27
fix: changed conditioning for timing functionality in circuit_builder

### LINKED ISSUE #1639: Bug ? `circuit_builder` timing | features misconfiguration
Hey guys!

In `plonky2/src/plonk/circuit_builder.rs` there is enabled import then feature `timing` is activated:

```rust
#[cfg(feature = "timing")]
use web_time::Instant;
```

after that, there is something following:

```rust
#[cfg(feature = "std")]
let start = Instant::now();
```

This breaks compilation, if try to add plonky2 with only `std` feature:

```toml
[dependencies]
plonky2 = { version = "1.0.0", default-features = false, features = ["std"] }
``` 

since rust can't find `Instant` type (being unimported because I didn't enable `timing`)

I guess at the second code snippet shall be conditional for `timing` and not `std`, like:

```rust
#[cfg(feature = "timing")]
let start = Instant::now();
```

or maybe conditional import from std, if `timing` is disabled

NikitaMasych:
# What?

This PR resolves #1639

### DIFF
--- a/plonky2/src/plonk/circuit_builder.rs
+++ b/plonky2/src/plonk/circuit_builder.rs
@@ -1075,7 +1075,7 @@ impl<F: RichField + Extendable<D>, const D: usize> CircuitBuilder<F, D> {
     ) -> (CircuitData<F, C, D>, bool) {
         let mut timing = TimingTree::new("preprocess", Level::Trace);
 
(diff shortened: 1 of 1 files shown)


Nashtare (approved):
LGTM, thanks!

PR #1643, opened 2024-12-11
Relax visibility of `verify` function

ekovalev:
Render the `plonk/verifier`'s `verify()` function public to allow for proofs verification in external applications.


Nashtare:
External applications can already verify proofs by using the public `verify` method from `VerifierCircuitData`.

PR #1645, opened 2025-01-03
add serialization and deserialization for BytesHash

carlogf:
This PR adds the implementation of the `serialize` and `deserialize` traits for the `BytesHash` struct.

Since the `Keccak` hash function uses `BytesHash` as its hash type, it is now possible to serialize and deserialize proofs that use the `KeccakGoldilocksConfig`, like [in this example](https://gist.github.com/fndecroix/92b12b7a546113183314eace452ebf3e)

### DIFF
--- a/plonky2/src/hash/hash_types.rs
+++ b/plonky2/src/hash/hash_types.rs
@@ -1,7 +1,9 @@
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
+use core::fmt;
 
 use anyhow::ensure;
+use serde::de::{self, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::field::goldilocks_field::GoldilocksField;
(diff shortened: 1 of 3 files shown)


plonky2/src/hash/hash_types.rs
```diff
@@ -193,19 +195,48 @@ impl<F: RichField, const N: usize> GenericHashOut<F> for BytesHash<N> {
...
+    }
+
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: de::SeqAccess<'de>,
+    {
+        let mut bytes = [0u8; N];
+        for i in 0..N {
+            bytes[i] = seq.next_element().unwrap().unwrap();
+        }
```
Nashtare, line 222: It think it may be best to catch deserialization errors here instead of simply panicking.

Nashtare (commented):
Thank you for this, good catch! Just left a small comment inline.

plonky2/src/hash/hash_types.rs
fndecroix: Hi! Good call, we changed that piece of code to catch the errors. Could you look at it?

Also, there is an error when running cargo clippy, but it is not related to the code we modified. Should we fix that?

Nashtare: If you don't mind, you can include it in this PR, otherwise I'll open a separate one and merge yours after :)
Thanks!

fndecroix: Done! We included those changes in a separate commit

Nashtare (approved):
Thank you again! LGTM!

PR #1654, opened 2025-01-28
fix(field): added os_rng conditioning feature

### LINKED ISSUE #1653: `OsRng` in Plonky2 breaks no-std
Hello guys!

## Prerequisite 

I see in [field/src/types.rs](https://github.com/0xPolygonZero/plonky2/blob/main/field/src/types.rs) following usage:

```rust
use rand::rngs::OsRng;
```

For that, you add dependency as below:

```toml
rand = { workspace = true, features = ["getrandom"] }
```

## Problem

I am working on a project, where strict no-std compatibility is required. After adding `plonky2` with no default features, it came to my surprise that there was a compilation failure regarding `getrandom` being dependent on `std`. 

Upon investigating, I discovered, that `getrandom` does not support my target and, as a result, an error was thrown causing linking to std.

In my specific case, I don't actually need sampling functionality.

## Proposed Solution

I suggest adding a feature-flag, where default will remain as it was with `OsRng`, however, as an opt-in path for projects, as mine, that might not have a need or support for `OsRng`, seeded solution.

NikitaMasych:
# What?

Resolves #1653


NikitaMasych:
I see CI failing, because we need to add `--features os_rng` to `cargo test`

PR #1655, opened 2025-01-28
feat: configurable nightly optimizations

### LINKED ISSUE #1638: Nightly to Stable clarification
Hello guys!

I see that you rely on `nightly` rust, so I was wondering whether it is possible to patch the library with only degraded performance, but keeping security and all features working in order to switch to `stable`.

[This](https://github.com/distributed-lab/plonky2/commit/0464ed09d022ef3db3c356a259655eb36172d1d2) is a rough patch of how I guess it should look.

NikitaMasych:
# What?

Resolves #1638

## Details

`Nightly` is required due to reliance on `specialization` feature, which allows keeping default implementation but leverage more optimized algorithms for specific fields. However, considering this is more desirable than critical, this PR elevates the requirement for nightly toolchain, allowing usage of the library in `stable` mode with, likely, worsened performance.


NikitaMasych:
@Nashtare @LindaGuiga would appreciate your thoughts on this and #1654, as soon as you have time, guys!

Issue #1670, opened 2025-03-18
Documentation does not build on docs.rs


dgulotta:
The plonky2 documentation is not available on docs.rs because of a build error.  The most recent log is here:
https://docs.rs/crate/plonky2/1.0.2/builds/1691160

It seems that cargo rustdoc does not accept --no-deps as an argument.

PR #1671, opened 2025-03-18
Fix documentation rendering

Nashtare:
This should fix the documentation rendering issue in #1670.
Also `MultiProof` got removed so we probably could remove `MultiProofChallenges` at some point.


Nashtare:
@LindaGuiga

LindaGuiga (approved):
Thank you!

PR #1677, opened 2025-04-04
Aards bench

tideofwords:
Added some debug statements useful for benchmarking.  They are debug!() macros, so you'll have to use env_logger and call env_logger.init(), then at command line pass RUST_LOG=debug


tideofwords:
sorry meant to PR a fork

PR #1679, opened 2025-05-05
Add constraints binding

LindaGuiga:
This PR aims at binding the STARK constraints to the challenges used in the quotient polynomials.
@SyxtonPrime Does this look good to you?


Nashtare:
Clippy will be fixed with #1678

starky/src/cross_table_lookup.rs
```diff
@@ -451,13 +451,11 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
     ) -> Vec<Self> {
         // Get all cross-table lookup polynomial openings for the provided STARK proof.
```
hratoanina, line 452:
```suggestion
        // Get all cross-table lookup polynomial openings for the provided STARK opening set.
```

starky/src/cross_table_lookup.rs
```diff
@@ -648,9 +646,9 @@ pub struct CtlCheckVarsTarget<F: Field, const D: usize> {
 
 impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
     /// Circuit version of `from_proofs`, for a single STARK.
```
hratoanina, line 648:
```suggestion
    /// Circuit version of `from_opening_set`.
```

starky/src/cross_table_lookup.rs
```diff
@@ -659,14 +657,13 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
     ) -> Vec<Self> {
         // Get all cross-table lookup polynomial openings for each STARK proof.
```
hratoanina, line 658:
```suggestion
        // Get all cross-table lookup polynomial openings.
```

starky/src/cross_table_lookup.rs
```diff
@@ -659,14 +657,13 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
...
         let ctl_zs = {
-            let openings = &proof.openings;
-            let ctl_zs = openings
+            let ctl_zs = stark_opening_set
                 .auxiliary_polys
                 .as_ref()
                 .expect("We cannot have CTls without auxiliary polynomials.")
```
hratoanina, line 663:
```suggestion
                .expect("We cannot have CTLs without auxiliary polynomials.")
```
OCD nits.

starky/src/cross_table_lookup.rs
```diff
@@ -659,14 +657,13 @@ impl<'a, F: Field, const D: usize> CtlCheckVarsTarget<F, D> {
...
                 .iter()
                 .skip(num_lookup_columns);
-            let ctl_zs_next = openings
+            let ctl_zs_next = stark_opening_set
                 .auxiliary_polys_next
                 .as_ref()
                 .expect("We cannot have CTls without auxiliary polynomials.")
```
hratoanina, line 669:
```suggestion
                .expect("We cannot have CTLs without auxiliary polynomials.")
```

starky/src/get_challenges.rs
```diff
@@ -61,6 +71,40 @@ where
...
+        )
+    } else {
+        None
+    };
+
+    let stark_alphas_prime = challenger.get_n_challenges(num_challenges);
+    let zeta_prime = challenger.get_extension_challenge();
+
+    // Bind constraints.
+    let constraints = compute_eval_vanishing_poly::<F, C, S, D>(
```
hratoanina, line 93: Nit: maybe call it `constraint_evals` or something?

hratoanina (dismissed):
LGTM, only minor nits.

starky/src/get_challenges.rs
```diff
@@ -61,6 +71,40 @@ where
...
+        &poly_evals,
+        ctl_vars,
+        lookup_challenges.as_ref(),
+        &stark.lookups(),
+        public_inputs,
+        stark_alphas_prime.clone(),
+        zeta_prime,
+        degree_bits,
+        num_lookup_columns,
+    );
```
SyxtonPrime, line 104: It might make sense to add a comment here (and in the other places this turns up) explaining what the point of this step is.

SyxtonPrime (approved):
I think this looks reasonably good (Though I'm not really familiar with this codebase, I joined after we had switched to Plonky3).

Do you think it would be possible to add some comments to the Prover explaining the idea behind the commitment and to the verifier/recursive_verifier showing more clearly where this is checked? (I struggled a little to understand where the verifier was checking the provers commitment)

starky/src/cross_table_lookup.rs
```diff
@@ -440,24 +440,22 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
     CtlCheckVars<'a, F, F::Extension, F::Extension, D>
 {
     /// Extracts the `CtlCheckVars` from a single proof.
-    pub fn from_proof<C: GenericConfig<D, F = F>>(
+    pub fn from_opening_set<C: GenericConfig<D, F = F>>(
```
Nashtare, line 443: Method description outdated now.

Also I am not sure I understand the need for changing the API here, if we're using the proof opening set anyway I'd rather keep it unchanged to reduce the friction

starky/src/recursive_verifier.rs
```diff
@@ -51,10 +47,21 @@ pub fn verify_stark_proof_circuit<
     let max_degree_bits_to_support = proof_with_pis.proof.recover_degree_bits(inner_config);
 
     let mut challenger = RecursiveChallenger::<F, C::Hasher, D>::new(builder);
+    let degree_bits_target = proof_with_pis.proof.degree_bits;
```
Nashtare, line 50: minor: probably no need to extract here, as the `get_challenges` method below take a proof as `&self`

starky/src/vanishing_poly.rs
```diff
@@ -83,3 +93,222 @@ pub(crate) fn eval_vanishing_poly_circuit<F, S, const D: usize>(
...
+    ctl_vars: Option<&[CtlCheckVarsTarget<F, D>]>,
+    lookup_challenges: Option<&Vec<Target>>,
+    public_inputs: &[Target],
+    alphas: Vec<Target>,
+    zeta: ExtensionTarget<D>,
+    degree_bits: usize,
+    degree_bits_target: Target,
+    num_lookup_columns: usize,
+) -> Vec<ExtensionTarget<D>>
+where
```
Nashtare, line 242: nit: the first is just a wrapper of the second, so maybe best to keep only `compute_eval_vanishing_poly_circuit` so that we have `let vanishing_polys_zeta = compute_eval_vanishing_poly_circuit(` in the recursive verifier?

starky/src/cross_table_lookup.rs
```diff
@@ -442,22 +442,41 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
     /// Extracts the `CtlCheckVars` from a single proof.
     pub fn from_proof<C: GenericConfig<D, F = F>>(
         table_idx: TableIdx,
-        proof: &StarkProof<F, C, D>,
+        stark: &StarkProof<F, C, D>,
```
Nashtare, line 445: confusing change

starky/src/cross_table_lookup.rs
```diff
@@ -442,22 +442,41 @@ impl<'a, F: RichField + Extendable<D>, const D: usize>
...
+    /// Extracts the `CtlCheckVars` from a single opening set.
+    pub fn from_opening_set<C: GenericConfig<D, F = F>>(
+        table_idx: TableIdx,
+        stark_opening_set: &StarkOpeningSet<F, D>,
+        cross_table_lookups: &'a [CrossTableLookup<F>],
+        ctl_challenges: &'a GrandProductChallengeSet<F>,
+        num_lookup_columns: usize,
+        total_num_helper_columns: usize,
+        num_helper_ctl_columns: &[usize],
+    ) -> Vec<Self> {
```
Nashtare, line 472: These 2 are identical, can't we just keep `from_proof` as before?

starky/src/cross_table_lookup.rs
```diff
@@ -647,29 +666,49 @@ pub struct CtlCheckVarsTarget<F: Field, const D: usize> {
...
+    /// Circuit version of `from_opening_set`, for a single STARK.
+    pub fn from_opening_set(
+        table: TableIdx,
+        stark_opening_set: &StarkOpeningSetTarget<D>,
         cross_table_lookups: &'a [CrossTableLookup<F>],
         ctl_challenges: &'a GrandProductChallengeSet<Target>,
         num_lookup_columns: usize,
         total_num_helper_columns: usize,
         num_helper_ctl_columns: &[usize],
     ) -> Vec<Self> {
```
Nashtare, line 699: same here

starky/src/get_challenges.rs
```diff
@@ -25,18 +33,24 @@ use crate::proof::*;
...
+    ctl_vars: Option<&[CtlCheckVars<F, F::Extension, F::Extension, D>]>,
     openings: &StarkOpeningSet<F, D>,
     commit_phase_merkle_caps: &[MerkleCap<F, C::Hasher>],
     final_poly: &PolynomialCoeffs<F::Extension>,
     pow_witness: F,
     config: &StarkConfig,
     degree_bits: usize,
+    num_trace_polys: usize,
+    num_aux_polys: usize,
+    pow_degree: usize,
```
Nashtare, line 53: `num_trace_polys` and `pow_degree` can be extracted internally from the `stark` parameter no?

starky/src/get_challenges.rs
```diff
@@ -64,6 +78,89 @@ where
...
+                .map(|ch| ch.beta)
+                .collect::<Vec<_>>(),
+        )
+    } else {
+        None
+    };
+
+    // Before computing the quotient polynomial, we bind the constraints.
+    // To do so, the verifier gets all the necessary polynomials evaluated at the random extension challenge `zeta_prime`,
+    // and evaluates the constraints at that point. The constraints are combined usihg `stark_alphas_prime`.
```
Nashtare, line 98:
```suggestion
    // and evaluates the constraints at that point. The constraints are combined using `stark_alphas_prime`.
```

starky/src/get_challenges.rs
```diff
@@ -102,6 +199,75 @@ where
...
+{
+    let total_num_dummy_extension_evals = num_trace_polys * 2 + num_aux_polys * 2; // for auxiliary_polys and auxiliary_polys_next
+    let num_dummy_evals = num_ctl_zs_first; // for ctls_zs_first
+
+    // Get extension and base field challenges that will simulate the trace, ctl, and auxiliary polynomials.
+    // Since sampling challenges for all polynomials might be heavy, we sample enough challenges {c_i}_i and use:
+    // c_i, c_i^{pow_degree}, ..., c_i^{pow_degree * 50} as simulated values.
+    let simulating_zetas =
+        challenger.get_n_extension_challenges(total_num_dummy_extension_evals.div_ceil(50));
+    let simulating_betas = challenger.get_n_challenges(num_dummy_evals.div_ceil(50));
```
Nashtare, line 223: shouldn't `50` be actually a parameter of `pow_degree`?

starky/src/get_challenges.rs
```diff
@@ -337,7 +703,17 @@ impl<const D: usize> StarkProofWithPublicInputsTarget<D> {
...
-            .get_challenges::<F, C>(builder, challenger, challenges, ignore_trace_cap, config)
+        self.proof.get_challenges::<F, C, S>(
+            builder,
+            stark,
+            &self.public_inputs,
+            challenger,
+            challenges,
+            ctl_vars,
+            degree_bits,
+            self.proof.degree_bits,
```
Nashtare, line 714: We're calling `get_challenges` which is a method of `self.proof` so `self.proof.degree_bits` does not need to be an argument

starky/src/prover.rs
```diff
@@ -246,13 +249,137 @@ where
...
+    let pow_degree = stark.constraint_degree() + 1;
+
+    // Get extension and base field challenges that will simulate the trace, ctl, and auxiliary polynomials.
+    // Since sampling challenges for all polynomials might be heavy, we sample enough challenges {c_i}_i and use:
+    // c_i, c_i^{pow_degree}, ..., c_i^{pow_degree * 50} as simulated values.
+    let simulating_zetas =
+        challenger.get_n_extension_challenges(total_num_dummy_extension_evals.div_ceil(50));
+    let simulating_betas = challenger.get_n_challenges(total_num_ctl_polys.div_ceil(50)); // For ctl_zs_first
+
+    // For each zeta in zetas, we compute the powers z^{(constraint_degree + 1)^i} for i = 0..50.
```
Nashtare, line 286: same comment about 50

starky/src/cross_table_lookup.rs
LindaGuiga: Sorry, forgot to remove it indeed

starky/src/get_challenges.rs
```diff
@@ -64,6 +76,92 @@ where
...
+    // Then, the challenger observes the constraint evaluations, so that they are bound to `stark_alphas`
+    // (the challenges used in the quotient polynomials).
+    let stark_alphas_prime = challenger.get_n_challenges(num_challenges);
+
+    let total_num_ctl_polys = ctl_vars
+        .map(|ctls| ctls.iter().map(|ctl| ctl.helper_columns.len()).sum())
+        .unwrap_or_default();
+
+    // First power unreachable by the constraints.
+    let pow_degree = stark.constraint_degree() + 1;
```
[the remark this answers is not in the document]
LindaGuiga: Makes sense, thanks! But then instead of adding an assert (peopel might decide they want to allow higher constraint degrees), I'll make the 50 modular!

starky/src/get_challenges.rs
```diff
@@ -64,6 +76,92 @@ where
...
+        )
+    } else {
+        None
+    };
+
+    // Before computing the quotient polynomial, we bind the constraints.
+    // To do so, the verifier gets all the necessary polynomials evaluated at the random extension challenge `zeta_prime`,
+    // and evaluates the constraints at that point. The constraints are combined using `stark_alphas_prime`.
+    // Then, the challenger observes the constraint evaluations, so that they are bound to `stark_alphas`
+    // (the challenges used in the quotient polynomials).
```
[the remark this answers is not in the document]
Nashtare: I'm not sure I understand the claim `it's possible for a malicious prover to "break" this with enough work`. Wouldn't that basically be breaking the hash function within the challenger? (e.g. in your terms, be more of a ~100 bits grinding than anything else?) Or were you implying something else?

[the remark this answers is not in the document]
SyxtonPrime: So my mental model of what "breaking" this commitment is, would be to find `2` different AIR's which give the same commitment.

Assume we do the currently coded thing were for each random variable `alpha_i` we take the powers `alpha_i, alpha_i^4, ..., alpha_i^{4^49} =  alpha_i^{2^98}`. Then our "commitment polynomial" looks like:
`C_0(alpha_1, ..., alpha_1^{2^98}, ..., alpha_j^{2^98}) + zeta C_1(alpha_1, ..., alpha_1^{2^98}, ..., alpha_j^{2^98}) + ...`.

As all our constraints have degree `3`, the total degree of this thing as a polynomial in `alpha_1, ..., alpha_j, zeta` will be `3 * 2^{98} * # of constraints`. Say we have `2000` constraints, then the degree of this thing is roughly `2^{111}`.

So if we have `2` different AIRs, the Schwartz Zipple lemma says that the probability that the constraints are equal at a given randomly generated point is `< 2^{111}/|EF| ~ 2^{111}/2^{128} ~ 2^{-17}`. So it may be possible for the prover to find a pair of AIRs where this constraint agrees. (In actual practice I suspect this probability is much lower than this but that would require far more difficult analysis).

We could increase the number of random elements and only take powers up to `4^{24}` to mitigate this but I guess my sense is that it's not a massive disaster if this particular part of the proof gets broken and it doesn't allow for a false proof to be proven.

starky/src/get_challenges.rs
```diff
@@ -299,17 +617,24 @@ impl<const D: usize> StarkProofTarget<D> {
...
             quotient_polys_cap.as_ref(),
+            ctl_vars,
             openings,
             commit_phase_merkle_caps,
             final_poly,
             *pow_witness,
+            degree_bits,
+            self.degree_bits,
+            S::COLUMNS,
+            openings.auxiliary_polys.as_ref().map_or(0, |aux| aux.len()),
```
Nashtare, line 637: `openings` are arguments of the method already so can be extracted internally
`S::COLUMNS` can also be inferred internally

starky/src/prover.rs
```diff
@@ -246,13 +249,126 @@ where
...
+        .iter()
+        .flat_map(|&zeta| {
+            successors(Some(zeta), move |prev| {
+                Some(prev.exp_u64(pow_degree as u64))
+            })
+            .take(nb_dummy_per_zeta)
+        })
+        .collect::<Vec<_>>();
+
+    // Simulate the openning set.
```
Nashtare, line 299:
```suggestion
    // Simulate the opening set.
```

Nashtare (approved):
Thanks @LindaGuiga! Looking good! I hust have minor comments on some method arguments otherwise seems good to go

Nashtare (approved):

PR #1687, opened 2025-06-26
chore(ci): replace deprecated rust-toolchain@master with stable

eeemmmmmm:
Replaced `dtolnay/rust-toolchain@master` with the stable tag as the master branch is deprecated.

See: https://github.com/dtolnay/rust-toolchain


eeemmmmmm:
@Nashtare

Nashtare:
Hi!

Plonky2 is unmaintained as you can see in the README, and as such we do not accept any PRs anymore.
We suggest you to use Plonky3 instead, which info you'll also find in the README. Thank you for your understanding!

